PR #101, opened 2014-12-21
basic search for event, site, person, work in progress

barryrowlingson:
Adds haystack search for models.


gvwilson:
#115 uses simple-minded SQL search, which is probably good enough for now - shall I close this one?

PR #121, opened 2015-01-04
Supporting notes about events under negotiation

### LINKED ISSUE #2: Add list of provisional events to home page
Home page should show events whose dates and instructors still haven't been settled. This will probably require a change or addition to the data model, since `workshops_event` only lists things with dates.

gvwilson:
1.  Add a new model `Pending` to record minimal information about event being negotiated.
2.  Create migration for model change.
3.  Create `fake.sql` to inject synthesized information into development database.
4.  Add new target `fake` to `Makefile` to create such information.
5.  Add URLs for displaying, editing, and creating pending events.
6.  Add templates to display information.

**Warning:** editing and creating pending events does not work right now for reasons that I do not understand.
To repeat:

1.  make clean
2.  make migrations
3.  make import
4.  make fake
5.  make serve
6.  Go to http://127.0.0.1:8000/workshops/pending/1
7.  Click on 'edit this pending event'

or:

6.  Go to http://127.0.0.1:8000/workshops/pendings
7.  Click on 'create'

When finished, this will close #2.


wking:
On Sun, Jan 04, 2015 at 08:43:41AM -0800, Greg Wilson wrote:
> 1.  Add a new model `Pending` to record minimal information about
> event being negotiated.

Can't we just use Event for this?  Here are the current event fields:

    site       = models.ForeignKey(Site)
    project    = models.ForeignKey(Project)
    organizer  = models.ForeignKey(Site, related_name='organizer', null=True)
    start      = models.DateField()
    end        = models.DateField(null=True)
    slug       = models.CharField(max_length=STR_LONG, unique=True)
    url        = models.CharField(max_length=STR_LONG, unique=True, null=True)
    reg_key    = models.CharField(max_length=STR_REG_KEY, null=True)
    attendance = models.IntegerField(null=True)
    admin_fee  = models.DecimalField(max_digits=6, decimal_places=2)
    notes      = models.TextField(default="")

And here they are again with the null entries removed:

    site       = models.ForeignKey(Site)
    project    = models.ForeignKey(Project)
    start      = models.DateField()
    slug       = models.CharField(max_length=STR_LONG, unique=True)
    admin_fee  = models.DecimalField(max_digits=6, decimal_places=2)
    notes      = models.TextField(default="")

And here are your proposed Pending fields:

    site       = models.ForeignKey(Site)
    project    = models.ForeignKey(Project)
    organizer  = models.ForeignKey(Site, related_name='pending_organizer', null=True)
    notes      = models.TextField(default="")

Changes we'd need to use Event for pending events:

* Make ‘start’ nullable, and interpret a null start as “this event is
  still in negotiation, and may not actually happen” (i.e. it's
  pending).

* Require pending events to have a slug (this seems reasonable to me,
  and I'd recommend something like YYYY-MM-DD-toronto), filling in as
  much of the date information as has been settled on.  When you
  finalize an event, just update the slug to use the finalized start
  date.  As a knock-on for this, we probably want to refer to events
  in URLs and such with their immutable ID, instead of using the
  now-mutable slug.  Alternatively, we could have a null slug for
  pending events, and only set an immutable slug once we'd finalized
  the event.

* Make admin_fee nullable, and interpret a null admin fee as “we
  haven't decided yet”.  If you decide to waive the admin fee,
  explicitly set it to zero.

gvwilson:
1.  Many of the changes we've made in the last month have been to move 
from implicit to explicit semantics.
2.  We will eventually want to decorate Pending with much more complex 
times (e.g., "here are the five times at which the workshop might work").
3.  We will eventually want to decorate Pending with notes about 
required skills, which is *not* the same thing as recording what topics 
were actually taught at an actual workshop.

wking:
On Sun, Jan 04, 2015 at 11:31:59AM -0800, Greg Wilson wrote:
> 1.  Many of the changes we've made in the last month have been to
> move from implicit to explicit semantics.

I agree that explicit is better than implicit, but I think “a null
field value means "we haven't decided yet" or "we haven't told amy” is
explicit.  Mapping “we haven't decided (or told amy) about a start
date” to “the event is pending” is just my definition for pending,
although if you have another definition you'd prefer I imagine we can
work that in instead.

> 2.  We will eventually want to decorate Pending with much more complex 
> times (e.g., "here are the five times at which the workshop might work").
> 3.  We will eventually want to decorate Pending with notes about 
> required skills, which is *not* the same thing as recording what topics 
> were actually taught at an actual workshop.

So maybe have Event (with the changes I outlined above to match it to
your Pending model), and then Event.pending which is a nullable
one-to-one pointing to the more complex pending-specific information.
That avoids duplicating fields between Event and Pending.  You can
null Event.pending and delete the associated Pending entry when you
finalize an event.

gvwilson:
Your arguments are persuasive - in particular, I like being able to 
promote an event from "we don't yet have dates or instructors" to "we 
have instructors" *or* "we have dates" to "we're set to go" without 
copying information around.

The challenges I see are identifying and sorting events:

1.  We can't use a made-up slug like `yyyy-mm-dd-esu` for an event that 
doesn't yet have dates because we might be negotiating several at a 
particular site at once.  We could just use database record IDs, but 
'events/123' is a lot less mnemonic than 'events/2015-10-12-esu'.  We 
could create URL mappings that sent '/event/2015-10-12-esu' to the right 
place, and also sent '/pending/123' to the right place, while using the 
same model underneath them.

2. We'd also have to figure out how to create tasks that mapped to 
events that were still pending (i.e., that didn't yet have a valid 
slug), since we're going to want to assign organizers, instructors, 
etc., to events before the dates are set.

3. Sorting probably isn't as much of a problem as this - we can filter 
events into two sets (confirmed and pending) and render them separately 
whenever we need to.

Thoughts?

gvwilson:
@r-gaia-cs @pbanaszkiewicz @cudevmaxwell could you please join the conversation? The goal is to represent events whose instructors and dates haven't yet been finalized in a simple way.

wking:
On Sun, Jan 04, 2015 at 12:58:00PM -0800, Greg Wilson wrote:
> 1.  We can't use a made-up slug like `yyyy-mm-dd-esu` for an event
> that doesn't yet have dates because we might be negotiating several
> at a particular site at once.

Do the several have specific slugs?  Presumably, we'll either be able
to narrow down by month (e.g. 2015-01-dd-esu and 2015-02-dd-esu).  If
we have two in the works for the same month, we'll need some
additional information in the slug (e.g. 2015-01-dd-esu-internal and
2015-01-dd-esu-wise).

> We could just use database record IDs, but 'events/123' is a lot
> less mnemonic than 'events/2015-10-12-esu'.  We could create URL
> mappings that sent '/event/2015-10-12-esu' to the right place, and
> also sent '/pending/123' to the right place, while using the same
> model underneath them.

Right, but folks would probably prefer the slug-based links, and then
be dissapointed when changing the slug invalidated their old link.  We
*could* keep a list of previous slugs and automatically redirect folks
to the ID-based link, but that's not very elegant on the
implementation side.  We could also just track the current slug and
redirect folks from /event/2015-10-12-esu to the canonical /event/123,
so you would only need to remember the current slug but could easily
bookmark/copy-paste/… the permanent, ID-based URL.

> 2. We'd also have to figure out how to create tasks that mapped to
> events that were still pending (i.e., that didn't yet have a valid
> slug), since we're going to want to assign organizers, instructors,
> etc., to events before the dates are set.

I'd use something immutable (e.g. Event.id) for references to Events.

> 3. Sorting probably isn't as much of a problem as this - we can
> filter events into two sets (confirmed and pending) and render them
> separately whenever we need to.

Right, so long as we have a clear definition of what “pending” means.
As an alternative to the one-to-one relationship I proposed earlier
[1], I'd also support just adding the pending-specific fields to
Event.  If we had:

  class EventTime(models.Model):
      event = models.ForeignKey(Event)
      start = models.DateField()
      end   = models.DateField(null=True)

then we could have several EventTime instances associated with one
Event during the planning stage, but only a single EventTime instance
associated with the Event once it was finalized.  For required skills,

  class EventRequiredSkill(models.Model):
      event = models.ForeignKey(Event)
      skill = models.ForeignKey(Skill)

will works just as well pointing at Event as it would pointing at
a Pending model.

[1]: https://github.com/swcarpentry/amy/pull/121#issuecomment-68646812

gvwilson:
> Do the several have specific slugs? Presumably, we'll either be able
> to narrow down by month (e.g. 2015-01-dd-esu and 2015-02-dd-esu).
Unfortuately not - we can't even assume we know the month when we're 
negotiating.

> > We could just use database record IDs, but 'events/123' is a lot
> > less mnemonic than 'events/2015-10-12-esu'. We could create URL
> > mappings that sent '/event/2015-10-12-esu' to the right place, and
> > also sent '/pending/123' to the right place, while using the same
> > model underneath them.
>
> Right, but folks would probably prefer the slug-based links, and then
> be dissapointed when changing the slug invalidated their old link.
That's part of why I proposed the separation in the first place: we've 
had several workshops whose slugs have changed as dates were moved 
around, which messed up our records and other people's bookmarks.  A 
serial number is the "right" way to do this from every perspective 
except usability.

Another option is to add a 'confirmed' flag to the Event model, rather 
than using nullity of start date or anything else as an implicit 
indicator.  Doesn't solve the problem of identifying events as dates 
(and hence slugs) change, but makes other things simpler.

-- 

wking:
On Sun, Jan 04, 2015 at 01:33:40PM -0800, Greg Wilson wrote:
> Another option is to add a 'confirmed' flag to the Event model,
> rather than using nullity of start date or anything else as an
> implicit indicator.  Doesn't solve the problem of identifying events
> as dates (and hence slugs) change, but makes other things simpler.

What does ‘confirmed’ represent?  That the dates won't change?  That
instructors have committed themselves?  That the event won't be
canceled?  That the event is open to publication?  To student
registration?  If the meaning is directly tied to some other data,
e.g.:

  Event.task_set.filter(role__name='instructor').count() >= Event.required_instructors

I'd rather avoid the additional field, because then we don't have to
figure out what to do when the ‘confirmed’ flag value differs from the
criterion used to determine it.  If the ‘confirmed’ flag *does*
represent a new property that isn't otherwise encoded in Event, we'd
obviously need to add a new field (or fields), but I'd name the fields
after the properties themselves (e.g. Event.registration_open).

pbanaszkiewicz:
Keeping one event model (for both pending and past events) feels more natural to me.

As for the issue with slugs and links and usability: many sites use this SEO technique of preparing almost-user-friendly links: they make article ID and slug both part of the URL, for example: http://muzyka.onet.pl/pop/amputacja-trampolina-do-slawy/6f9f4.
Here you have an article ID at the very end (`6f9f4`). Going to slightly different URL (like http://muzyka.onet.pl/pop/amputacja-trampolina-do-slawa/6f9f4) returns HTTP 301 Moved Permanently.

In our case we could use this pattern for events: `/event/ID/SLUG`. I don't know if folks remember various workshops' slugs, but just in case we can handle these with URL `/event/STARTDATE-SLUGTXT` which would result in 301 as well.

@gvwilson @wking you people are fast typers…

gvwilson:
> What does ‘confirmed’ represent? That the dates won't change? That
> instructors have committed themselves? That the event won't be
> canceled? That the event is open to publication? To student
> registration?
In practice, it means "the instructors have bought their plane tickets".
> I'd rather avoid the additional field, because then we don't have to
> figure out what to do when the ‘confirmed’ flag value differs from the
> criterion used to determine it.
But all those criteria can change - dates can move around (which impacts 
the slug and the website URL), instructors - even the coordinating 
organization could change.  Are you more comfortable with calling the 
flag 'published' ?

gvwilson:
p.s. even if this one gets thrown away, I'd be grateful if someone could let me know why my attempt to make Pending createable and editable failed.

pbanaszkiewicz:
@gvwilson you're missing `templates/workshop/pending_form.html`:

```django
{% extends "workshops/_page.html" %}

{% load breadcrumbs %}
{% block breadcrumbs %}
    {% breadcrumb_url 'Amy' 'index'  %}
    {% breadcrumb_url 'All pending' 'all_pending' %}
    {% if object %}
        {% url 'pending_details' object as object_url %}
        {% breadcrumb object object_url %}
        {% breadcrumb_active 'Edit pending event'  %}
    {% else %}
        {% breadcrumb_active 'New pending event'  %}
    {% endif %}
{% endblock %}

{% block content %}
<form action="" method="post">
    {% csrf_token %}
    {{ form.as_p }}
    <input type="submit" value="Submit" />
</form>

{% endblock %}
```

wking:
On Sun, Jan 04, 2015 at 01:52:05PM -0800, Piotr Banaszkiewicz wrote:
> In our case we could use this pattern for events: `/event/ID/SLUG`

Once you require folks to know the ID, I don't see a point in using
the slug.  If we want to use mutable-but-usually-easy-to-remember
slugs (and I think we do), I'd rather have /event/SLUG 301 you to
/event/ID.  Then you can use the current slug (which you remember or
correctly guess) to get to the stable ID-based URL.

wking:
On Sun, Jan 04, 2015 at 02:11:57PM -0800, Greg Wilson wrote:
> > What does ‘confirmed’ represent? That the dates won't change? That
> > instructors have committed themselves? That the event won't be
> > canceled? That the event is open to publication? To student
> > registration?
>
> In practice, it means "the instructors have bought their plane
> tickets".

Then I'd prefer:

  class PersonEventTime(models.Model):
      person = models.ForeignKey(Person)
      event_time = models.ForeignKey(EventTime)
      travel_plans_finalized = models.BooleanField(default=False)

and I'd consider an Event confirmed when:

* event.eventtime_set.count() == 1
  (we've decided on a single EventTime),
* len(instructors) >= event.required_instructor_count
  (we have enough instructors committed)
* event.eventtime_set.get().personeventtime_set.filter(
      person__in=instructors, travel_plans_finalized=True).exists()
  (at least one of those instructors has finalized their travel plans)

> > I'd rather avoid the additional field, because then we don't have
> > to figure out what to do when the ‘confirmed’ flag value differs
> > from the criterion used to determine it.
>
> But all those criteria can change - dates can move around (which
> impacts the slug and the website URL), instructors - even the
> coordinating organization could change.  Are you more comfortable
> with calling the flag 'published' ?

I'd rather record measurable information in the model, and encode
associated policy in a separate model or manager method.  If we want
to make “an instructor has finalized their travel plans” the criterion
for publishing an event, that sounds good to me.  If we want a
separate ‘open_for_enrollment’ flag that the organizer can set once
they want to open enrollment, I'm fine with that too, and I'd have
*that* flag or model method trigger whether we publish the workshop
where students can find it.  Having a separate ‘published’ (or
‘confirmed’) flag that's independent of these more fundamental flags
seems redundant to me (and redundant information can go stale).  I
don't see why you'd want to distinguish between ‘published’ (visible
to students?) and ‘open_for_enrollment’.  Is there a reason to make a
distinction there?

gvwilson:
Thanks @pbanaszkiewicz - I clearly needed more coffee :-)

gvwilson:
We're using #122 instead.

PR #122, opened 2015-01-05
Handling provisional events

gvwilson:
This PR is an alternative solution to #2 - it allows events to have null fields for start date, end date, slug, and a few other things, then adds a 'published' field to show whether the event is ready to publish on the main web site.  That part is more or less working; what isn't is event editing and updating.

1.  `EventCreate` seems to work, but whenever *any* field is left NULL, Django reports an error.

2.  `EventUpdate` doesn't work at all: clicking on the 'edit event' link on the bottom of an event's detail page bring up an error:
    ~~~
    Generic detail view EventUpdate must be called with either an object pk or a slug.
    ~~~
    This occurs even though the event's PK is in the URL.  `slug_field` and `slug_url_kwarg` are *not* set in `EventUpdate`, but `pk_url_kwarg` is, so I'm mystified.


wking:
On Sun, Jan 04, 2015 at 08:48:43PM -0800, Greg Wilson wrote:
>   * Accepting either slug or ID to identify event

Instead of:

  +    if re.match(r'^\d+$', event_ident):
  +        return Event.objects.get(id=event_ident)
  +    elif re.match(r'^\d{4}-\d{2}-\d{2}-.+$', event_ident):
  +        return Event.objects.get(slug=event_ident)
  +    else:
  +        pass # FIXME: error (unrecognizable ident)

why not avoid specifying the slug format and just use:

  try:
      event_id = int(event_ident)
  except ValueError:
      return Event.objects.get(slug=event_ident)
  return Event.objects.get(id=event_id)

In other words, everything that can't be cast to an integer is a slug.
However, I think the support added in this commit isn't quite right,
and would prefer if the slug form redirected to the ID form [1,2].
That means using ‘redirect’ [3] with something like:

  def event_details(request, event_ident):
      '''List details of a particular event.'''
      try:
          event_id = int(event_ident)
      except ValueError:
          event = Event.objects.get(slug=event_ident)
          return redirect(event, permanent=True)
      event = Event.objects.get(id=event_id)
      …

I haven't had time to review the other commits in this pull request
yet.

[2]: https://github.com/swcarpentry/amy/pull/121#issuecomment-68664373
[3]: https://docs.djangoproject.com/en/1.7/topics/http/shortcuts/#redirect

gvwilson:
1.  I'd like to define the regexp that matches a valid slug in 
workshops/util.py and use it in workshops/url.py so that it's guaranteed 
to be consistent with what we're using workshops/models.py - is there an 
easy way to do this?

2.  I strongly prefer mnemonic slugs (yyyy-mm-dd-site) to numeric event 
IDs where possible - I think we should only use the latter when the 
former isn't available.  I've tested, and it appears that if an event is 
edited to change its slug, everything continues to work (but my testing 
hasn't been exhaustive).

3. If we can address point 1 above, I'd like to merge this, and then I 
will integrate your site/host split as well.  One thing I'd like to add 
(either to that, or separately) is to identify organizations that are 
organizers: when selecting "who is organizing this workshop?", only a 
handful of orgs (SSI, MSL, soon to be one or two others) should be 
choosable.  One possibility is a flag in the Site model; another is a 
separate model that contains nothing except a foreign key (essentially, 
an enumeration class).  Preferences?

Thanks,
Greg

gvwilson:
I'm going to go ahead and merge this, and open a PR to define regexp for things we use in multiple places (like event slugs).

wking:
On Mon, Jan 05, 2015 at 10:36:29PM -0800, Greg Wilson wrote:
> 1.  I'd like to define the regexp that matches a valid slug in
> workshops/util.py and use it in workshops/url.py so that it's
> guaranteed to be consistent with what we're using
> workshops/models.py - is there an easy way to do this?

Having a regular expression to validate slugs is fine, but you should
only need to use it in the Event-posting validation code.  That makes
sure you don't get invalid slugs in the database.  You don't need to
validate the slug used in a URL though, since if someone uses an
invalid slug there, we just won't find a matching Event, and they'll
get a 404 (or whatever we're handing out for “I don't have that
Event”).

PR #124, opened 2015-01-07
Tiny simplification to gender choices in Person model.

gvwilson:
Single-letter constants were defined for male, female, and other, but weren't used anywhere.


wking:
On Tue, Jan 06, 2015 at 08:28:32PM -0800, Greg Wilson wrote:
> Single-letter constants were defined for male, female, and other,
> but weren't used anywhere.

These are for eventual person filtering.  For example, if you only
want female instructors for a WiSE workshop, you'd use:

  Person.objects.filter(
      award__badge__name='instructor',
      gender=Person.FEMALE)

So I'm -1 on this PR.

gvwilson:
Ah - OK, we could use the letter in that case, but I understand the 
reason for the constant. I'll nuke the PR.

Issue #70, opened 2014-12-15
Replace Django's auth.User with our workshops.Person?


wking:
I expect everyone in the database will want to log in to see their dashboard (listing upcoming workshops that they're on, or workshops in their area that need instructors or helpers), manage their personal information (new email, etc.), or register a new event.  We probably don't *two* person tables with many duplicate entries, so I'd recommend dropping the stock auth.User and [substituting][1] workshops.Person.

[1]: https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#substituting-a-custom-user-model

gvwilson:
I don't understand the scope of changes or resulting fragility if we use `workshops.Person` instead of `auth.User`.  If the answer is "small" and "not much", I'm +1.  If third-party authentication modules like OAuth expect `auth.user`, we should use `auth.User` for logins (since there are only a few) and `workshops.Person` for recording information about people.  @wking @craigmbooth can you please comment?

wking:
On Wed, Jan 14, 2015 at 04:09:14AM -0800, Greg Wilson wrote:
> I don't understand the scope of changes or resulting fragility if we
> use `workshops.Person` instead of `auth.User`.

I haven't used AUTH_USER_MODEL in other projects yet, since the other
Django projects I'm involved were started in older versions of Django.
My understanding is that it will be a simple change, but (as mentioned
earlier in this issue) we can't migrate in or out of it, so we'll have
to ditch our existing migrations.

> If third-party authentication modules like OAuth expect `auth.user`,
> we should use `auth.User` for logins (since there are only a few)
> and `workshops.Person` for recording information about people.

As mentioned in the docs, this change should be transparent to any
libraries using get_user_model [1].  Craig suggested using
python-social-auth [2], which has an example using AUTH_USER_MODEL
[3], so I expect we can get it working for us too ;).

[1]: https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#django.contrib.auth.get_user_model
[2]: https://github.com/omab/python-social-auth#id4
[3]: https://github.com/omab/python-social-auth/blob/v0.2.1/examples/django_example/example/settings.py#L201

gvwilson:
1.  No problem ditching our current migrations.
2.  I'll trust you on everything else.

Thanks,
Greg

wking:
Sane values for REQUIRED_FIELDS [1] are blocked on the blank stuff
that's in a753408 (Allowing several fields in a Person record to be
empty, #125).

[...]

PR #125, opened 2015-01-07
Allowing several fields in a Person record to be empty.

gvwilson:
We may not know a person's middle name, email address, nearest airport, etc., so this PR modifies the Person model to allow those fields to be missing.


wking:
On Tue, Jan 06, 2015 at 08:32:43PM -0800, Greg Wilson wrote:
> We may not know a person's middle name, email address, nearest
> airport, etc., so this PR modifies the Person model to allow those
> fields to be missing.

Most of these make sense to me, but I think Person.slug should be
neither nullable nor blank.  It looks like we use Person.id in most
places as the identifier, but we use Person.slug for badge strings:

  result[badge.name] = [{"user" : p.slug, "name" : p.fullname()} for p in persons]

We may also use it in other places (I'm not sure).  In any case, if we
want the slug as an alternate ID field complementing Person.id, I'd
require it for all people.

gvwilson:
We only need a non-empty person.slug when we're issuing a badge (or 
adding someone to the website for some other reason), so in almost all 
cases it can safely be NULL.

wking:
On Wed, Jan 07, 2015 at 11:06:24PM -0800, Greg Wilson wrote:
> We only need a non-empty person.slug when we're issuing a badge (or
> adding someone to the website for some other reason), so in almost
> all cases it can safely be NULL.

Once we get student badges, I think a majority of Person instances
will have badges.  Event with the current database:

  $ python3.4 ./manage.py shell
  >>> from workshops.models import Person

Number of people:

  >>> Person.objects.all().count()
  7517

Badged people:

  >>> Person.objects.all().exclude(award=None).count()
  322

Learners (who may eventually be getting skill badges):

  >>> Person.objects.filter(task__role__name='learner').count()
  6559

So once we add learner badges, we'll have:

  >>> from django.db.models import Q
  >>> Person.objects.filter(
  ...     Q(task__role__name='learner') |
  ...     Q(award__isnull=False)
  ... ).distinct().count() / Person.objects.all().count()
  0.8903818012504988

almost 90% of our People badged.

I'm not sure how “added to the website for some other reason” is
represented in the database, but unless it's in there somewhere it
will be hard to ensure we've slugged everyone that falls into that
category.  I think it's easier to just create slugs for everyone.

wking:
On Thu, Jan 08, 2015 at 08:50:15AM -0800, W. Trevor King wrote:
> I think it's easier to just create slugs for everyone.

Alternatively with #70 we could use Person.username as a slug, and
have people slug themselves when they create an account (e.g. to
register as a trainee).

gvwilson:
Only a handful of people will need login accounts for the foreseeable future - this is *not* going to be a self-serve system, and we can add logic to check that Person.slug is non-empty when assigning badges.

wking:
On Sat, Jan 17, 2015 at 09:35:58PM -0800, Greg Wilson wrote:
> Only a handful of people will need login accounts for the
> foreseeable future - …

Ok.  I don't see anything in the docs that says a person needs to have
a non-null username [1], but the uniqueness constraint doesn't play
well with string fields if we have multiple blank-string entries, and
Django won't store NULLs for empty strings [2].  I guess that means
I'll need a separate username field for #147, and we'll just live with
the overlap between it and the slug field.

[1]: https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#django.contrib.auth.models.CustomUser.USERNAME_FIELD
[2]: https://code.djangoproject.com/ticket/4136

gvwilson:
Everyone who will ever be allowed to log in *will* have a non-null, 
non-empty slug slash username - will that get us through?

wking:
On Sun, Jan 18, 2015 at 12:46:36PM -0800, Greg Wilson wrote:
> Everyone who will ever be allowed to log in *will* have a non-null, 
> non-empty slug slash username - will that get us through?

It won't let us use the slug as the username, since the username's
unique requirement is about a database constraint, and Django doesn't
give you the “multiple NULL entries don't violate uniqueness” out with
string-based fields (as described in the ticket 4136 I linked
earlier).

gvwilson:
Gotcha.  OK, username field it is - or do you want to have a completely 
separate model for login accounts with username/password info and a 
possibly-null foreign key ref to a Person?

wking:
On Sun, Jan 18, 2015 at 02:50:41PM -0800, Greg Wilson wrote:
> … or do you want to have a completely separate model for login
> accounts with username/password info and a possibly-null foreign key
> ref to a Person?

No, since I see no reason to distinguish between “people that can log
in” and “people we track” beyond a null password (or false is_active,
or whatever).  They're still all people.

gvwilson:
This one has fallen out of date, and should not be a blocker.

PR #106, opened 2014-12-22
Add bulk person ingest from CSV

cudevmaxwell:
Fixes #27. 

Still requires testing and vetting.

PR #131, opened 2015-01-10
Bulk addition of people from CSV files

### LINKED ISSUE #27: Add form for bulk addition of new persons
The app should allow admins to upload a CSV file containing names, email addresses, and optionally roles as well.  The uploaded data should be validated and displayed for further editing before being added.

* If roles are not included in the data, everyone should be marked as a learner, but a warning should be raised.
* If roles are specified, an event must also be specified, and a task should be created for each person showing what role they had in that event.
* Instead of using a fixed column ordering, the upload form should allow users to specify which columns contain which fields on the fly.

gvwilson:
This incorporates @cudevmaxwell's work in #106 in order to fix #27:

1.  Form for uploading CSV files containing people (and tasks) from workshops.
2.  Some test data (but no unit tests as yet - those need to be added).
3.  Some lessons learned about Django's ORM.


gvwilson:
@cudevmaxwell can you please have a look at this one? And thank you again for getting the ball rolling...

workshops/views.py
```diff
@@ -136,6 +140,58 @@ def person_details(request, person_id):
...
+def person_bulk_add(request):
+    if request.method == 'POST':
+        form = PersonBulkAddForm(request.POST, request.FILES)
+        if form.is_valid():
+            try:
+                persons_tasks = _upload_person_task_csv(request, request.FILES['file'])
+            except csv.Error as e:
+                messages.add_message(request, messages.ERROR, "Error processing uploaded .CSV file: {}".format(e))
+            else:
+                for entry in persons_tasks:
```
pbanaszkiewicz: One thing I saw when testing your code a few days ago: we should have database transaction for this loop.

In case one entry in 1000 is wrong and the user has to fix it and then re-upload… they'll see database errors (unique constraint violation) due to re-adding previously added users / tasks.

Django transaction handling: https://docs.djangoproject.com/en/1.7/topics/db/transactions/#django.db.transaction.atomic

pbanaszkiewicz:
Thought: do we need to see `person_bulk_add_result` after bulk-adding people?

Wouldn't it be more user-friendly to have a "Confirm this is what you wanted to upload: Yes, proceed / No, cancel" with upload data in a table? Then, after confirming, data would end up inserted to DB.

gvwilson:
I believe that was @cudevmaxwell's original plan - but that display 
ought to be editable so that errors can be correct in the page (rather 
than people having to edit the CSV and re-upload).

workshops/views.py
gvwilson: There should definitely be a transaction here, and adding People and 
Task records to the DB should only happen after they've been validated 
and the user has pressed 'yes' on the confirmation form.

pbanaszkiewicz:
@gvwilson I'd make some adjustments and send you a PR, but I think changes would be big. Why don't you merge this instead and create a new issue to improve bulk addition?

gvwilson:
OK, I'll merge this one so that we can keep moving...

Issue #129, opened 2015-01-09
DataTables for quick in-table search


narrativecraig:
In the past I have had a lot of success using [DataTables](http://www.datatables.net) to enhance HTML tables with search, column sorting, etc.

![DataTable](http://imgur.com/hXlpvun.png)

It's a pretty quick, lightweight change and could complement the simple SQL search that currently exists in amy.  Would this be worth implementing?

wking:
On Thu, Jan 08, 2015 at 06:42:47PM -0800, Craig Booth wrote:
> In the past I have had a lot of success using DataTables…

Personally, I like either no JavaScript, or a full-blown framework
(like AngularJS).  Going the Django API + AngularJS route seemed a bit
heavy handed for amy at the moment, so I'm fine with using a jQuery
plugin if we want to assemble our JavaScript widgets in smaller
chunks (e.g. I floated bootstrap-datepicker as part of #110 [1,2]).

Tooling thoughts aside, I don't have strong preferences for whether we
have sortable tables or not.

[1]: https://github.com/wking/swc-amy/commit/ad5e3dd1ea68ad1d4ee7045d6babb11d28d876ed
[2]: https://github.com/swcarpentry/amy/pull/110#discussion-diff-22294959

gvwilson:
Sortable tables would be nice, particularly for search results, but I 
don't know enough to choose between DataTables, Angular.js, and what-not.

pbanaszkiewicz:
I really like the idea of having these tables in AMY. My only concern: is this really going to be lightweight? DataTables (or something very similar) in Google Summer of Code homepage (about 1.4k records, for example: http://www.google-melange.com/gsoc/projects/list/google/gsoc2014) are awefully slow.

wking:
On Sat, Jan 10, 2015 at 05:23:20AM -0800, Piotr Banaszkiewicz wrote:
> I really like the idea of having these tables in AMY.

Is this “I'd rather be sorting the results with the backing SQL
engine”?  That makes sense to me, but getting a nice UI in front of
that isn't as easy as dropping in DataTables.  I'm fine going either
way here and then re-opening the discussion if we start bumping into
problems.

pbanaszkiewicz:
@wking:
> Is this “I'd rather be sorting the results with the backing SQL engine”?

No, I'm +1 for DataTables and +1 on re-opening the discussion if any errors occur.

craigmbooth:
[...]

Issue #82, opened 2014-12-18
Display map of known airports


gvwilson:
The "all airports" page should display a map.

gvwilson:
Close? We seem to be managing OK without this...

PR #134, opened 2015-01-10
Make sortable DataTables tables for the all airports page

craigmbooth:
This pull request brings sortable [DataTables](http://www.datatables.net/) into amy.

airport datatable

airport datatable filtered

Given discussion in #129  I have implemented this initially only on the ``all_airports`` page, so that people can see how it works.  Assuming nobody has any big objections or issues, I'll go ahead and apply this change to all tables in a couple of days.  Until then, this pull request should be considered a work in progress.

### DIFF
--- a/workshops/templates/workshops/_page.html
+++ b/workshops/templates/workshops/_page.html
@@ -4,9 +4,16 @@
   <head>
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css">
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap-theme.min.css">
+    <link rel="stylesheet" href="https://cdn.datatables.net/1.10.4/css/jquery.dataTables.min.css">
     <link rel="stylesheet" href="{% static 'amy.css' %}">
     <script src="https://code.jquery.com/jquery-2.1.3.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/js/bootstrap.min.js"></script>
+    <script src="https://cdn.datatables.net/1.10.4/js/jquery.dataTables.min.js"></script>
+    <script type="text/javascript" class="init">
+      $(document).ready(function() {
+      $('#sortable_table').DataTable();
+      } );
+    </script>
     <title>Amy: {{ title }}</title>
   </head>
   <body>
--- a/workshops/templates/workshops/_page.html
+++ b/workshops/templates/workshops/_page.html
@@ -9,11 +9,6 @@
     <script src="https://code.jquery.com/jquery-2.1.3.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/js/bootstrap.min.js"></script>
     <script src="https://cdn.datatables.net/1.10.4/js/jquery.dataTables.min.js"></script>
-    <script type="text/javascript" class="init">
-      $(document).ready(function() {
-      $('#sortable_table').DataTable();
-      } );
-    </script>
     <title>Amy: {{ title }}</title>
   </head>
   <body>
--- a/workshops/templates/workshops/_single_table_page.html
+++ b/workshops/templates/workshops/_single_table_page.html
@@ -1,0 +1,28 @@
+{% extends "workshops/_page.html" %}
+
+{% load breadcrumbs %}
+{% block breadcrumbs %}
+    {% breadcrumb_url 'Amy' 'index'  %}
+    {% breadcrumb_active 'All events'  %}
+{% endblock %}
+
+{% block content %}
+  {% if data %}
+     <script type="text/javascript" class="init">
+       $(document).ready(function() {
+         $('#data_table').DataTable({
+           "data": {{data|safe}},
+           "columns": {{columns|safe}},
+           "deferRender": true,
+           {% block column_definitions %}{% endblock column_definitions %}
+         })
+       })
+     </script>
+     <table id="data_table"></table>
+  {% else %}
+    <p>No rows.</p>
+  {% endif %}
+
+{% block add_new_link %}{% endblock add_new_link %}
+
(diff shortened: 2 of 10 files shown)


wking:
On Sat, Jan 10, 2015 at 09:14:52AM -0800, Craig Booth wrote:
> Given discussion in #129 I have implemented this initially only on
> the ``all_airports`` page, so that people can see how it works…

This looks good to me.

gvwilson:
How does/should this interact with pagination?

wking:
On Sat, Jan 10, 2015 at 07:03:16PM -0800, Greg Wilson wrote:
> How does/should this interact with pagination?

As you can see from the previously-unpaginated airport listing,
DataTables adds its own in-JavaScript paging [1,2], so there's no need
for Django-side pagination.  As @craigmbooth pointed out in #129 [3],
we may want a threshold for high result counts that asks you to refine
your initial filter to keep DataTables performant.

[1]: https://datatables.net/examples/basic_init/alt_pagination.html
[2]: http://www.datatables.net/plug-ins/pagination/
[3]: https://github.com/swcarpentry/amy/issues/129#issuecomment-69462801

craigmbooth:
That's a good question.  It basically comes down to this.  There are three different ways to have DataTables accept data:

   1. Render data from the DOM (this is what I do here, just wrap the HTML table with the correct class).  This will scale up to ~5,000 rows before slowdown is noticeable.
   2. Pass in data as JSON, and have DataTables render only the data you can see.  This will scale up to ~50,000 rows before slowdown is noticeable.

Neither of these solutions use Django's pagination at all.  All data in the table is sent to the client.  At some point just sending the data will become a cause of latency (although we are not near that limit yet).  In this case we can do the processing server-side in the SQL database with e.g. [django-datatables-view](https://pypi.python.org/pypi/django-datatables-view)

craigmbooth:
Oh yes, and as @wking pointed out as I was writing that... It is also easy to have any table that is too big (say > 10,000 lines) fall back to Django's standard pagniation.

That said, the only table build I've seen so far that is slow is for the Tasks table, and the cause of slowness is not the rendering of the table, it is this line:

```
all_tasks = Task.objects.order_by('event', 'person', 'role')
```

wking:
On Sat, Jan 10, 2015 at 08:16:36PM -0800, Craig Booth wrote:
> all_tasks = Task.objects.order_by('event', 'person', 'role')

Which we can drop if we're using DataTables to do the ordering.  I
imagine we can configure DataTables for that view so we get a default
ordering on page-load?

narrativecraig:
I'm happy with this now.  In the last commit I made a few changes:

    1. All tables are now sortable
    2. I changed ``views.py`` to send json data into the template, which is then rendered into the table.  This makes the tables load far faster than parsing HTML, but means I had to make more changes.

   There is now some Javascript in the templates, I hope nobody minds this.  It is really just a dictionary giving configuration for what should be in each column and some simple functions with logic like "if data == 0: wrap the data in a warning div else: print the data", so my hope is that this is totally readable and maintainable by non-Javascript people.

gvwilson:
Thanks @craigmbooth - @wking and @cudevmaxwell, could you please review?

wking:
On Tue, Jan 13, 2015 at 08:44:28PM -0800, Craig Booth wrote:
>    There is now some Javascript in the templates, …

Looking this over, it looks like there's a reasonable amount of
duplicate code in these templates (and it looks like there was already
a reasonable amount of duplication before this PR).  Can we extract
the shared parts into a common list-view template?  That would make
reviewing easier.

narrativecraig:
So yeah, this is an interesting point.  I initially went 100% in this direction, had the views generate everything, and made it so that there was one template (``all.html``) that took care of every one of these pages.  I gave up on that when I realized that the pages do actually want to be different (e.g. #82), and broke everything back out.  That said, I did go too far with copying boilerplate code around.

I'll break this down with blocks tonight and should be able to cut down significantly on the line count.

wking:
On Wed, Jan 14, 2015 at 02:14:40PM -0800, Craig Booth wrote:
> I gave up on that when I realized that the pages do actually want to
> be different (e.g. #82), and broke everything back out.

I see a few approaches that could work.  In order of decreasing
preference:

a. Make the map it's own view, since finding an airport on a map
   seems only weakly associated with a DataTables listing.
b. Have a switch in the template to include additional per-model
   templates as needed.
c. Add more context variables for stuff like this, and have the map
   logic in the view.

craigmbooth:
I prefer (b), I could be convinced that a separate view is good for maps/airports, but I'm sure that we'll hit situations down the road where we do need to add custom content to each page, so let's plan around that.

Anyway, I just pushed again.  Following the gist of @wking's suggestion, I have moved the boilerplate into ``_single_table_page.html``, and the only things that need to be set on each page now are the HTML that comes below the table, and the column configuration options.  I don't want to try and programmatically generate the column definitions because there are far too many things we could need to do there...

This pull request now increases the codebase by a net of <30 lines (if you subtract newly added comments) and added a fair amount of functionality.

wking:
You currently have things like:

> +    all_sites = [[site.fullname, site.domain, site.notes,
> +                  reverse('site_details', args=[site.domain])] for
> +                  site in Site.objects.all()]

I don't like having to keep array ordering synchronized between the
data and column definitions.  Can we use objects instead [1]?  That
lets us re-order the table purely in the column definition (via
columns.data [2]) without having to touch tha all_sites construction.

I'd also consider adding .json methods to our classes so we could just
have:

  all_sites = [site.json() for site in Site.objects.all()]

That way it's easier to keep the rendered JSON in sync with the model
as we migrate fields and such.

[1]: http://www.datatables.net/manual/data#Objects
[2]: http://datatables.net/reference/option/columns.data

wking:
Since serving JSON from Django is so easy, I'd also consider using
DataTables' server-side processing [1].  A nice side effect would be
that other folks could use the JSON endpoints directly for their own
scripting or user interfaces.

[1]: http://datatables.net/manual/data#Server-side-processing

pbanaszkiewicz:
Closing in favor of #367.

Issue #132, opened 2015-01-10
Unifying workshops and training events


gvwilson:
This subsumes #29 (recording instructors for training events), #31 (recording training events as workshops), #32 (displaying training event completion rates), #33 (showing whether training counts toward an instructor badge), and #63 (do we need a separate table for training cohorts).

Amy currently represents workshops and instructor training events using completely separate models.  @wking has suggested unifying them; I agree that makes sense, but want to think through the implications before diving in.

Background:

*   A regular workshop is represented by an `Event` instance.
*   The things people can do at regular workshops (instructor, helper, learner, organizer) are enumerated by `Role`.
*   What a particular person did at a particular event is recorded by `Task`.
*   A specific offering of instructor training is recorded by `Cohort`.
*   Someone's instructor training status (in-progress, complete, incomplete) is enumerated by `TraineeStatus`.
*   Someone's participation in a particular round of instructor training is recorded by `Trainee`.

Translations:

*   `Event` and `Cohort` have start dates, venues, and slugs.
*   `Cohort` doesn't have most of the other things `Event` has, but they can be filled in:
    *   `published` is `True`.
    *   `project` (which enumerates SWC, DC, etc.) can be made `TTT` (train the trainers).
    *   `organizer` can be `software-carpentry.org`
    *   `url` can be NULL (or I can retroactively create pages for instructor training sessions).
    *   `reg_key` can be NULL.
    *   `attendance` can be the number who enrolled.
    *   `admin_fee` can be $0.0.
    *   `notes` can be empty.
*   One special case: `Cohort` doesn't have an end date right now; instead, it has an `active` field to show whether some people are still in training in that cohort.  As @wking pointed out, we could use an end date instead to show that a round of training has ended.  The only trick with this is that we don't know end dates for instructor training in advance (online courses can drag one).  However, `enddate` can already be NULL, so that seems straightforward.
*   `Cohort` *does* have a `qualifies` field to show whether this training qualifies someone to become an instructor.  We can get rid of that and simply mark people who took part in non-qualifying training rounds as done, then not issue them a badge.

The harder part is figuring out what to do about `TraineeStatus`.  The simplest answer is, don't bother, since we can infer status as follows:

*   registered = listed as a learner in a training event that hasn't started yet
*   in_progress = listed as a learner in a training event that has started but hasn't finished
*   complete = listed as a learner in a training event *and* have an instructor badge (but see below)
[...]

PR #136, opened 2015-01-11
Merging cohort with event

gvwilson:
Addresses #132 (I had time on the plane...)


wking:
The the workshops/ changes are minimal and they look good to me.  I
haven't checked the migrater.py changes, since I don't have access to
the input data it's using.

There are some things we discussed in #132 that aren't addressed in
this pull request:

* It doesn't address the auto-badging and assignment modeling (you'd
  still have to badge individuals by hand).
* It doesn't address the “attendance complete” modeling.

I feel like those are generic shortcomings of our Event model, and
shouldn't hold back this PR rolling Cohort into Event.

PR #137, opened 2015-01-14
Showing statistics graphically

gvwilson:
This PR is a proof of concept showing how we can generate statistical charts on the fly using matplotlib. To try it out, go to http://127.0.0.1:8000/chart/enrolment or .../chart/workshops.  Before merging it, we must:

1.  Replace the raw SQL with Django ORM (I don't know how to calculate a cumulative sum using Django).
2.  Create views like `stats/enrolment` and `stats/workshops` which produce HTML pages that contain `img` elements that in turn cause these images to be generated (rather than just returning the image directly).

### DIFF
--- a/workshops/urls.py
+++ b/workshops/urls.py
@@ -44,4 +44,6 @@ urlpatterns = [
     url(r'^search/?$', views.search, name='search'),
 
     url(r'^export/(?P<name>[\w\.-]+)/?$', views.export, name='export'),
+
+    url(r'^stats/(?P<name>[\w\.-]+)/?$', views.stats, name='stats'),
 ]
--- a/workshops/urls.py
+++ b/workshops/urls.py
@@ -45,5 +45,5 @@ urlpatterns = [
 
     url(r'^export/(?P<name>[\w\.-]+)/?$', views.export, name='export'),
 
-    url(r'^stats/(?P<name>[\w\.-]+)/?$', views.stats, name='stats'),
+    url(r'^chart/(?P<name>[\w\.-]+)/?$', views.chart, name='chart'),
 ]
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -487,6 +487,14 @@ def export(request, name):
 
 #------------------------------------------------------------
 
+def stats(request, name):
+    '''Display statistics.'''
+    context = {'title' : name,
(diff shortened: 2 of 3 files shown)


narrativecraig:
As an alternative solution that feels less heavyweight than matplotlib, how about pulling in [flot](http://www.flotcharts.org/).  Anything we can render into JSON, we can immediately make into a pretty interactive chart.

I can build a proof of concept off of this branch for people to check out if there's interest?

wking:
On Tue, Jan 13, 2015 at 07:16:09PM -0800, Craig Booth wrote:
> As an alternative solution that feels less heavyweight than
> matplotlib, how about pulling in [flot](http://www.flotcharts.org/).
> Anything we can render into JSON, we can immediately make into a
> pretty interactive chart.

Handling this in JavaScript sounds good to me.

gvwilson:
No objection to Javascript.

workshops/views.py
```diff
@@ -487,6 +487,52 @@ def export(request, name):
...
+    from matplotlib import pyplot
+    from matplotlib.backends.backend_agg import FigureCanvasAgg
+    from django.db import connection
+
+    if name == 'workshops':
+        query = '''select e1.start, count(e2.start)
+                   from workshops_event e1 join workshops_event e2
+                   where (e1.start||e1.slug)>=(e2.start||e2.slug)
+                   group by e1.start, e1.slug
+                   order by e1.start, e1.slug;'''
```
sburns: The Django ORM is good at many things, but IMO annotation is not one of them. I don't have a problem with a raw query like this. I would parameterize `datetime.date.today()` and add that as a WHERE clause in the statement so it's one less thing to do in python.

workshops/urls.py
```diff
@@ -44,4 +44,6 @@
     url(r'^search/?$', views.search, name='search'),
 
     url(r'^export/(?P<name>[\w\.-]+)/?$', views.export, name='export'),
+
+    url(r'^chart/(?P<name>[\w\.-]+)/?$', views.chart, name='chart'),
```
sburns: You might consider `r'^chart/(?P<name>{workshops,enrolment})/?$'` (my regex might be off) and then the URL dispatcher will handle 404s for unknown charts in both generating URLS in templates & dispatching to the view.

Issue #138, opened 2015-01-14
Bulk CSV upload: transaction, confirmation page


pbanaszkiewicz:
We need to handle CSV uploads better.

Current workflow:

1. Prepare CSV file
2. Upload
3. Pray for it to work

We should have something less stressful instead:

1. Prepare CSV file
2. Upload
3. Make changes to fields with wrong values
4. Confirm
5. Save & be happy

Required changes:

* pack whole thing in database transaction
* make confirmation page
* make editable confirmation page

wking:
On Wed, Jan 14, 2015 at 02:33:08AM -0800, Piotr Banaszkiewicz wrote:
> 1. Prepare CSV file
> 2. Upload

If we're preparing CSV by hand, we're doing it wrong (we should just
be using the Amy UI directly).  If CSV uploading doesn't work, we
should be fixing the CSV generator.  I doubt building a more
complicated CSV-upload-checker into amy is worth the trouble,
although…

> * pack whole thing in database transaction

… this makes sense to me (see my ATOMIC_REQUESTS suggestion [1]).

[1]: https://github.com/swcarpentry/amy/pull/106#issuecomment-68576325

pbanaszkiewicz:
@wking the bulk-adding people feature has landed. It was really needed. I just made it more user-friendly and (I hope!) less error-prone.
I don't know if CSVs are being generated or prepared by humans. @gvwilson ?

wking:
On Wed, Jan 14, 2015 at 01:16:29PM -0800, Piotr Banaszkiewicz wrote:
> the bulk-adding people feature has landed. It was really needed.

I'm still skeptical [1,2,3] ;).  But I'm not enough of a purist to
stand in the way of others who are willing to write, review, and
maintain things that I think could be done more easily ;).  I'll just
chime in with my point of view periodically to make sure folks aren't
confusing specific UI plans (like the bulk person upload) with the
usage requirements that those UIs are designed for (like adding people
to the database).

> I just made it more user-friendly and (I hope!) less error-prone.

Fair enough ;).  If you felt motivated to write the PR, I'm fine with
it landing.  Again, I just wanted to make sure we'd thought through
the “why do we need this, and could we accomplish that mission in a
simpler way” question.  If y'all feel satisfied that this is a
reasonably simple (or already working) solution to one of our usage
requirements, then I'm happy.

[1]: https://github.com/swcarpentry/amy/issues/27#issuecomment-67921209
[...]

PR #139, opened 2015-01-14
Bulk people CSV upload confirmation page

pbanaszkiewicz:
This PR is a solution to #138.

Work-in-progress.

### DIFF
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -267,6 +267,8 @@ def _verify_upload_person_task(data):
 
     for item in data:
         event, role = item.get('event', None), item.get('role', None)
+        email = item['person'].get('email', None)
+
         errors = []
         if event:
             try:
@@ -282,6 +284,15 @@ def _verify_upload_person_task(data):
             except Role.MultipleObjectsReturned:
                 errors.append('More than one role named {} exists.'
                               .format(role))
+
+        if email:
+            try:
+                Person.objects.get(email=email)
+                errors.append("User with email {} already exists."
+                              .format(email))
+            except Person.DoesNotExist:
+                # we want the email to be unique
+                pass
 
         if errors:
             # copy the errors just to be safe
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -6,6 +6,9 @@ import requests
 from django.contrib import messages
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 from django.core.urlresolvers import reverse
+from django.core.exceptions import ObjectDoesNotExist
+from django.http import Http404
+from django.db import IntegrityError, transaction
 from django.db.models import Count, Q
 from django.shortcuts import redirect, render, get_object_or_404
 from django.views.generic.edit import CreateView, UpdateView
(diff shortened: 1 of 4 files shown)


gvwilson:
Thanks @pbanaszkiewicz - @cudevmaxwell, would you like to review this, since you got the ball rolling?

pbanaszkiewicz:
@cudevmaxwell, @gvwilson: I'll have to use some JavaScript-Fu to make editable fields on the confirmation page.

I also plan to add "Verify" button.

Only these two features are missing.

pbanaszkiewicz:
@cudevmaxwell, @gvwilson okay, I think this PR is ready for review.

cudevmaxwell:
Great work, @pbanaszkiewicz! I didn't think to use sessions.

gvwilson:
I'm learning a lot from this too...

pbanaszkiewicz:
Thank you! :-)

workshops/views.py
```diff
@@ -145,51 +148,201 @@ def person_bulk_add(request):
...
+        # alters persons_tasks via reference
+        _verify_upload_person_task(persons_tasks)
+
+        context = {'title': 'Confirm uploaded data',
+                   'persons_tasks': persons_tasks}
+        return render(request, 'workshops/person_bulk_add_results.html',
+                      context)
+
+
 def _upload_person_task_csv(request, uploaded_file):
```
sburns: Mostly a style preference but I generally prefer to keep view functions (those returning HttpResponses) in `views.py` and non-view functions (those that don't) into a `<app_name>.utils.py` file to keep `views.py` from becoming a huge mess. Such a change would apply to  `_verify_upload_person_task` as well.

Then, when in testing, we can directly import the util functions and test them and then test the views with `TestCase.Client`.

workshops/views.py
```diff
@@ -145,51 +148,201 @@ def person_bulk_add(request):
...
 
+def person_bulk_add_confirmation(request):
+    """
+    This view allows for manipulating and saving session-stored upload data.
+    """
+    persons_tasks = request.session['bulk-add-people']
+
+    # if the session is empty, don't bother
+    if not persons_tasks:
+        raise Http404
```
sburns: I don't think the 404 would be helpful and it's not really semantically correct. I think it'd be better to add a warning message about not having the csv data from the previous operation and to try again, then redirect to the bulk upload page.

workshops/views.py
```diff
@@ -145,51 +148,201 @@ def person_bulk_add(request):
...
-                        entry['task'].save()
-                context = {'title' : 'Process CSV File',
-                           'form': form, 
-                           'persons_tasks': persons_tasks}
-                return render(request, 'workshops/person_bulk_add_results.html', context)
+                # instead of insta-saving, put everything into session
+                # then redirect to confirmation page which in turn saves the
+                # data
+
+                request.session['bulk-add-people'] = persons_tasks
```
sburns: Probably doesn't matter, but django's session serializer moved from pickle to json so we've got to make sure that if  in the future we add columns to the csv, the values will still be json-serializable. I'm probably just being paranoid though :)

workshops/views.py
```diff
@@ -145,51 +148,201 @@ def person_bulk_add(request):
...
-                messages.add_message(request, messages.ERROR, \
-                                     'More than one role named {} exists.'.format(row['role'])) 
-        persons_tasks.append(entry)
-    return persons_tasks
+                errors.append('More than one role named {} exists.'
+                              .format(role))
+
+        if email:
+            try:
+                Person.objects.get(email=email)
```
sburns: Email addresses aren't case-sensitive, so to ensure you don't have differently-cased but otherwise the same address, you should do `Person.objects.get(email__iexact=email)`.

sburns:
Other than the line notes, none of which are show-stoppers, I think this is good.

gvwilson:
Thanks Scott.  Do you have time to add a single end-to-end test in 
workshops/test/test_csv_upload.py (new file) to show us how you'd test 
this?  It's been over three years since I wrote tests for Django code, 
and I'm struggling to fit what I remember to how things are now done - 
guidance would be greatly appreciated.

sburns:
Yeah not a problem Greg. I'll try to minimize potential merge conflicts with #140.

@pbanaszkiewicz I'll make a PR your `pbanaszkiewicz:bulk-upload-enhancement` branch.

pbanaszkiewicz:
@sburns thanks a lot for the feedback!

I agree with your comments. Will your PR contain some fixes or only a new test? If the latter, I can work on changes not sooner than Friday. Cheers!

workshops/views.py
pbanaszkiewicz: Are CHAR UNIQUE fields in the DB case-sensitive?

If not ([1]) then it's fine the way it is.

[1] http://sqlite.1065341.n5.nabble.com/UNIQUE-Constraint-but-case-sensitive-td774.html

workshops/views.py
pbanaszkiewicz: That's a perfect test case right here :-)

workshops/views.py
sburns: This might not always be on SQLite so I'd prefer to let the orm handle this logic.

sburns:
@pbanaszkiewicz I'll roll these changes into my branch.

pbanaszkiewicz:
@sburns thanks a lot!

sburns:
thank you for the feature :)

PR #140, opened 2015-01-15
Adding tests

gvwilson:
This PR starts to build a more comprehensive test framework for Amy.  Please don't bother reviewing until more work has been done on utilities.

### DIFF
--- a/workshops/test/base.py
+++ b/workshops/test/base.py
@@ -4,6 +4,7 @@ import re
 import datetime
 import itertools
 import xml.etree.ElementTree as ET
+from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 from ..models import \
@@ -14,9 +15,6 @@ from ..models import \
     Qualification, \
     Site, \
     Skill
-
-
-TEMPLATE_STRING_IF_INVALID = 'XXX-unset-variable-XXX' # FIXME: get by importing settings
 
 
 class TestBase(TestCase):
@@ -129,12 +127,12 @@ class TestBase(TestCase):
                 writer.write(content)
 
         # Report unfilled tags.
-        if TEMPLATE_STRING_IF_INVALID in content:
+        if settings.TEMPLATE_STRING_IF_INVALID in content:
             self._save_html(content)
             lines = content.split('\n')
             hits = [x for x in enumerate(lines)
-                    if TEMPLATE_STRING_IF_INVALID in x[1]]
-            msg = '"{0}" found in HTML page:\n'.format(TEMPLATE_STRING_IF_INVALID)
+                    if settings.TEMPLATE_STRING_IF_INVALID in x[1]]
+            msg = '"{0}" found in HTML page:\n'.format(settings.TEMPLATE_STRING_IF_INVALID)
             assert not hits, msg + '\n'.join(['{0}: "{1}"'.format(h[0], h[1].rstrip())
                                               for h in hits])
 
--- a/workshops/test/base.py
+++ b/workshops/test/base.py
@@ -1,3 +1,4 @@
+import cgi
 import traceback
 import os
 import re
@@ -79,7 +80,7 @@ class TestBase(TestCase):
                                               twitter='herself', url='http://hermione.org', slug='granger.h')
         Award.objects.create(person=self.hermione,
                              badge=self.instructor,
-                             awarded=datetime.date(2014, 01, 01))
+                             awarded=datetime.date(2014, 1, 1))
         Qualification.objects.create(person=self.hermione, skill=self.git)
         Qualification.objects.create(person=self.hermione, skill=self.sql)
 
@@ -89,7 +90,7 @@ class TestBase(TestCase):
                                            twitter=None, url=None, slug='potter.h')
         Award.objects.create(person=self.harry,
                              badge=self.instructor,
-                             awarded=datetime.date(2014, 05, 05))
+                             awarded=datetime.date(2014, 5, 5))
         Qualification.objects.create(person=self.harry, skill=self.sql)
 
         self.ron = Person.objects.create(personal='Ron', middle=None, family='Weasley',
@@ -113,7 +114,7 @@ class TestBase(TestCase):
         self.blackwidow = Person.objects.create(personal='Natasha', middle=None, family='Romanova',
                                                 email=None, gender='F', active=False)
 
-    def _parse(self, content, save_to=None):
+    def _parse(self, response, save_to=None):
         """
         Parse the HTML page returned by the server.
         Must remove the DOCTYPE to avoid confusing Python's XML parser.
@@ -121,6 +122,10 @@ class TestBase(TestCase):
         If save_to is a path, save a copy of the content to that file
         for debugging.
         """
+        _, params = cgi.parse_header(response['content-type'])
+        charset = params['charset']
+        content = response.content.decode(charset)
+
         # Save the raw HTML if explicitly asked to (during debugging).
         if save_to:
             with open(save_to, 'w') as writer:
@@ -146,7 +151,7 @@ class TestBase(TestCase):
             doc = ET.XML(content)
             return doc
         # ...and save in a uniquely-named file if we can't.
-        except ET.ParseError, e:
+        except ET.ParseError as e:
             self._save_html(content)
             assert False, 'HTML parsing failed: {0}'.format(str(e))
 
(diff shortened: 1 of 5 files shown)


gvwilson:
Finally got this working - @sburns @pbanaszkiewicz @wking can you have a look?

wking:
The basic thrust looks good to me.  A few suggestions:

* You have:

    Person.GENDER_CHOICES_DICT.get(person.gender, None)),

  which should probably be:

    Person.get_gender_display()

  See [1].

* I'd also rather not unwrap the:

    active     = models.NullBooleanField(…

  line, to keep the source readable in 80-char-wide terminals.

* You've also still got a FIXME marker:

    TEMPLATE_STRING_IF_INVALID = 'XXX-unset-variable-XXX' # FIXME: get by importing settings

  Does importing settings not work or something?

* Setting up a test database is usualy done with fixtures [2,3], not
  your self._setUpSites() and similar.

* I'd rather drop the ‘assert’ statements and use the assert methods
  [4], which usually give good error messages out of the box (without
  us having to specify the error message).

* There are also some driveby changes that distract from the
  substantive changes.  For example:

    -            for person in persons:
    -                person.num_taught = \
    -                    person.task_set.filter(role__name='instructor').count()
    +            for p in persons:
    +                p.num_taught = \
    +                    p.task_set.filter(role__name='instructor').count()

[1]: https://docs.djangoproject.com/en/1.7/ref/models/instances/#django.db.models.Model.get_FOO_display
[2]: https://docs.djangoproject.com/en/1.7/topics/testing/tools/#topics-testing-fixtures
[3]: https://docs.djangoproject.com/en/1.7/howto/initial-data/#providing-initial-data-with-fixtures
[4]: https://docs.python.org/2/library/unittest.html#assert-methods

sburns:
I'll piggyback on the good comments from @wking. Typically, I don't test the actual HTML of responses because front-end changes that don't otherwise the site can stop the test-suite from passing.

I think it's better to [test the context](https://docs.djangoproject.com/en/1.7/topics/testing/tools/#django.test.Response.context) of the response to make sure the view is passing the correct objects to the template.

sburns:
Also, if you're expecting a redirect (e.g. the user isn't logged in and is trying to access a view that requires authentication), you can test the redirect chain with [`response.redirect_chain`](https://docs.djangoproject.com/en/1.7/topics/testing/tools/#django.test.Client.get)

gvwilson:
> You've also still got a FIXME marker:
> TEMPLATE_STRING_IF_INVALID = 'XXX-unset-variable-XXX' # FIXME: get by importing settings
> Does importing settings not work or something?

`import settings` and `import ..settings` and variations thereon don't work - the settings.py file is actually in the `amy` sub-directory.  I don't want to munge the path before import - would be grateful for directions on how to do this correctly.

gvwilson:
I've been bitten before by testing the context rather than the rendered HTML (graphic designer edited the latter, I was feeding it the right things, they weren't being displayed).

sburns:
> import settings and import ..settings and variations thereon don't work - the settings.py file is actually in the amy sub-directory.

The organization of apps in projects has undergone turbulent changes but for the most part django apps, e.g. workshops, should be _next_ to settings.py, urls.py, etc. and under `amy`. Then you can `from amy import settings` or some amount of dots up wrt your place in the particular application.

But...assuming the `DJANGO_SETTINGS_MODULE` envvar is set (probably `amy.settings`), then `from django.conf import settings` will be a module w/ read-only attributes coming from `amy/settings.py`. So you should be able to:

```python
from django.conf import settings

bad_template_string = settings.TEMPLATE_STRING_IF_INVALID
```

gvwilson:
I just tried:

 > from django.conf import settings

w/out luck.  How about we wait until #134, #139, and #140 land, then 
move 'workshops' to where it ought to be?

sburns:
> then move 'workshops' to where it ought to be?

agreed.

sburns:
> w/out luck.  

There isn't a setup.py file but i wonder if python setup.py develop would fix this...symlinks the package into your environment so it's both installed and editable. </curiousity>

wking:
On Tue, Jan 20, 2015 at 11:26:10AM -0800, Greg Wilson wrote:
> > You've also still got a FIXME marker:
> > TEMPLATE_STRING_IF_INVALID = 'XXX-unset-variable-XXX' # FIXME: get by importing settings
> > Does importing settings not work or something?
> 
> `import settings` and `import ..settings` and variations thereon
> don't work - the settings.py file is actually in the `amy`
> sub-directory.  I don't want to munge the path before import - would
> be grateful for directions on how to do this correctly.

Use [1]:

from django.conf import settings

[1]: https://docs.djangoproject.com/en/1.7/topics/settings/#using-settings-in-python-code

gvwilson:
Tried that - didn't work.

wking:
On Tue, Jan 20, 2015 at 12:52:49PM -0800, Scott Burns wrote:
> > w/out luck.  
> 
> There isn't a setup.py file…

You only need that if you're installing amy, and I don't think we need
that (running it out of a Git checkout should be fine).

Of course, I'd be fine if we wanted to package up the workshops app
for PyPI, and then have an amy repository with the minimal amy-app
boilerplate that we have now consuming the independent workshops app.
That's probably more work than we want for stage 1 though ;).

wking:
On Tue, Jan 20, 2015 at 01:28:34PM -0800, Greg Wilson wrote:
> Tried that - didn't work.

More information would make it easier to debug ;).  What did you try?
How did you launch the tests?  How did they fail?

wking:
On Tue, Jan 20, 2015 at 12:41:19PM -0800, Scott Burns wrote:
> The organization of apps in projects has undergone turbulent changes
> but for the most part django apps, e.g. workshops, should be _next_
> to settings.py, urls.py, etc. and under `amy`.

Hmm.  So long as workshops is in the PYTHONPATH (e.g. you're running
./manage.py from amy's root) it shouldn't matter where workshops is
located.  For example, it could be installed in your system's
site-packages (if we distributed the workshops package as a separate
project).  Importing the django project's settings file directly
(instead of using django.conf.settings) is breaks down the (useful)
wall between the app (workshops) and the site (amy).

gvwilson:
> > There isn't a setup.py file…
> That's probably more work than we want for stage 1 though ;).
Agreed - Version 2.

wking:
On Tue, Jan 20, 2015 at 01:39:13PM -0800, W. Trevor King wrote:
> Tue, Jan 20, 2015 at 01:28:34PM -0800, Greg Wilson:
> > Tried that - didn't work.
> 
> More information would make it easier to debug ;).  What did you try?
> How did you launch the tests?  How did they fail?

And after some Python 3 fixes, it's working for me (gvwilson/amy#12).

gvwilson:
import settings

import ..settings

import ..amy.settings

from ..amy import settings

from django.conf import settings

all failed for me when using `make test` to run tests

gvwilson:
OK, that works:

from django.config import settings

then using

settings.TEMPLATE_STRING_IF_INVALID

I'll add this to my PR and worry about "import .base" later.  (I presume 
the '.' to be sure you're importing from same directory rather than from 
elsewhere on path?)

gvwilson:
p.s. thanks

wking:
On Tue, Jan 20, 2015 at 02:34:28PM -0800, Greg Wilson wrote:
> from django.conf import settings
> 
> all failed for me when using `make test` to run tests

On my system, ‘python’ means ‘python3.3’, and I only have Django
installed for 3.4.  If I edit the Makefile to use ‘python3.4’ instead
of ‘python’, and use my adjusted branch (which uses the django.conf
line you mention, gvwilson/amy#12), I get:

  $ make test
  python3.4 manage.py test
  …
	Ran 32 tests in 2.390s

  OK

So can you give more information about what you're seeing than
“failed”?  Do you see an ImportError traceback?  What does it say?

wking:
On Tue, Jan 20, 2015 at 02:39:36PM -0800, Greg Wilson wrote:
> OK, that works:
> 
> from django.config import settings
> 
> then using
> 
> settings.TEMPLATE_STRING_IF_INVALID
> 
> I'll add this to my PR…

Or just merge the first commit in my gvwilson/amy#12.

> and worry about "import .base" later.  (I presume the '.' to be sure
> you're importing from same directory rather than from elsewhere on
> path?)

Yes.  For details on relative imports see [1].  Perhaps we could
discuss any issues you have with gvwilson/amy#12 there?  Putting it
off just leaves this branch broken for Python 3 and makes more work
for me later if I have to rebase the branch to apply to some future
amy version.

[1]: https://www.python.org/dev/peps/pep-0328

PR #141, opened 2015-01-15
Adding airport names to instructor search.

gvwilson:
Arliss requested that full airport names be displayed along with IATA codes when looking up instructors. This PR changes the way airport names are displayed so that the instructor search pulldown will do the right thing, then modifies the logic in the form's cleanup to match.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -33,21 +33,19 @@ class InstructorsForm(forms.Form):
 
     def clean(self):
         cleaned_data = super(InstructorsForm, self).clean()
-        iata = cleaned_data.get('airport')
+        airport = cleaned_data.get('airport')
         lat = cleaned_data.get('latitude')
         long = cleaned_data.get('longitude')
 
-        if iata is None:
+        if airport is None:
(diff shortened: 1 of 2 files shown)


wking:
> -                    'Must specify either an airport code or a '
> -                    'latitude/longitude')
> +                    'Must specify either an airport code or latitude/longitude')

This looks like a drive by unwrapping (and ‘ a’ dropping).  Now the
line is longer than 80 characters ;).  Not really a big deal, but I'd
have been happier if this change wasn't part of this PR.  ‘git add -p’
makes it easy to only stage the changes you want before committing.

The Airport.__str__ change is all you should need to add airport names.

The ‘iata’ → ‘airport’ changes just look like a bugfix because we
hadn't understood the reprecussions of to_field_name [1].  Did
cleaned_data.get('airport') ever return just an IATA string?

In any case, I'm ok with this change.

[1]: https://docs.djangoproject.com/en/1.7/ref/forms/fields/#django.forms.ModelChoiceField.to_field_name

PR #145, opened 2015-01-16
Context views breadcrumbs

### LINKED ISSUE #142: Title not set in template parameters when editing person
1.  Display all persons.
2.  Pick a person.
3.  Edit (e.g., go to http://127.0.0.1:8000/workshops/person/100/edit).
4.  View source.

If you have #140, you will see XXX-unset-variable-XXX as the page title, because `title` isn't included in the context used when the page is expanded.  How should `title` be set when editing?  The editing form is auto-generated from `PersonUpdate(UpdateView)`.

pbanaszkiewicz:
It may seem not connected, but thanks to new `CreateViewContext` and `UpdateViewContext` I was able to repair breadcrumbs.

1. These new views were created because Django basic `CreateView` and `UpdateView` do not carry template variables we needed. And because every update/create view uses variables `title` and (from now) `model`, I made `CreateViewContext` and `UpdateViewContext` that set these variables.

2. Breadcrumbs were neglected and inconsistent across pages. I removed unnecessary template tags code, took a look at what breadcrumbs are used most of the time and come up with some. More in commit message 2b164aaa482fdaeed8e9f835345f8968435954c7

This fixes #142.

EDIT: Forgot to mention that I fixed some URL regexp bug preventing sites editing.

### DIFF
--- a/workshops/templates/workshops/airport.html
+++ b/workshops/templates/workshops/airport.html
@@ -2,9 +2,9 @@
 
 {% load breadcrumbs %}
 {% block breadcrumbs %}
-    {% breadcrumb_url 'Amy' 'index'  %}
-    {% breadcrumb_url 'All airports' 'all_airports' %}
-    {% breadcrumb_active airport.fullname  %}
+    {% breadcrumb_main_page %}
+    {% breadcrumb_index_all_objects airport %}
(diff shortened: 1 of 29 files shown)


gvwilson:
Thanks very much for this - it's a big help.
G

Issue #127, opened 2015-01-08
What does Person.active mean?


wking:
It landed with cba8d659 (Adding more information about people, 2014-12-01), but we don't seem to use it for anything.

gvwilson:
People (including instructors) occasionally ask to have participation 
suspended (e.g., don't mail me for the next few months, I'm busy), and 
we need to respect that.

wking:
On Thu, Jan 08, 2015 at 04:56:09PM -0800, Greg Wilson wrote:
> People (including instructors) occasionally ask to have
> participation suspended (e.g., don't mail me for the next few
> months, I'm busy), and we need to respect that.

So it means “do not contact”?

gvwilson:
> So it means “do not contact”?
Yes - "do not contact for now"

PR #147, opened 2015-01-18
Replace Django's auth.User with our workshops.Person

wking:
Implementing #70.  Currently a work-in-progress [blocked][1] on #125.

[1]: https://github.com/swcarpentry/amy/issues/70#issuecomment-70391552


workshops/models.py
```diff
@@ -97,11 +98,13 @@ class Person(models.Model):
     github     = models.CharField(max_length=STR_MED, unique=True, null=True)
     twitter    = models.CharField(max_length=STR_MED, unique=True, null=True)
     url        = models.CharField(max_length=STR_LONG, null=True)
-    slug       = models.CharField(max_length=STR_LONG, null=True)
+    username   = models.CharField(max_length=STR_MED, unique=True)
 
     # Set the custom manager
     objects = PersonManager()
 
+    USERNAME_FIELD = 'username'
```
pbanaszkiewicz: I might have missed that in all these lengthy discussions about auth, but why add a new field? We can use `email`, as it's already a unique field.

wking: On Sun, Jan 18, 2015 at 08:37:45AM -0800, Piotr Banaszkiewicz wrote:
> > +    USERNAME_FIELD = 'username'
> 
> I might have missed that in all these lengthy discussions about
> auth, but why add a new field? We can use `email`, as it's already a
> unique field.

For the same reason we can't use the slug: the username field needs to
be unique, and Django doesn't do a good job with unique string fields
that aren't required [1].  If we required email addresses for every
Person, using email addresses would be fine, but I don't think we can
do that.

[1]: https://github.com/swcarpentry/amy/pull/125#issuecomment-70414088

pbanaszkiewicz: Alright, I missed that `Person.email` is now nullable.

gvwilson: Yup - if people ask us to remove their email address from our database, 
we have to comply.

sburns:
Collecting what's left to do related to authentication...

- [ ] Authentication flow (there's no reason not to use the standard `django.contrib.auth.views` and add our templates, correct?)
    - [ ] Login
    - [ ] Logout
    - [ ] Password change
    - [ ] Password reset request/email/reset confirmation view
    - [ ] Since there are many more `People` than will ever really interact w/ the site, we need a way for people to "activate" their account assuming they're already represented in the `People` table. This should set `People.is_active`. How will people know if they have an account?
    - [ ] Request an account (email to an admin?)
    - [ ] Invite a user/send email to them/account confirmation (redirects to password reset)
- [ ] Protect views with `@login_required`. @gvwilson how much is this site is meant to be public?
- [ ] Protect staff views with `@staff_member_required`
    - [ ] @gvwilson might need some help here determining what only staff should be able to do.

gvwilson:
HI Scott,

1. The "active" flag in the Person class indicates whether they're 
currently interested in being contacted, taking part in workshops, etc.  
It isn't related to whether or not their account is active - honestly, 
there are only going to be half a dozen people with login privileges 
this year (because the DB holds personal identifying information, etc.).

2.  By default, *none* of this stuff is public (again, PID).  For the 
first round, we can divide accounts into:

a) can't log in (99% of people)

b) can log in but can't modify things (nobody's in this category right 
now, but I'm sure someone will be some day)

c) can log in and modify anything they want (me, Arliss, Giacomo, maybe 
one more person)

Does that make sense?

Thanks,
Greg

sburns:
>  The "active" flag in the Person class indicates whether they're currently interested in being contacted, taking part in workshops, etc.

@wking the help text on this field is good, but we need to make sure future selves/others don't mix this up w/ the [`is_active`](https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#django.contrib.auth.models.CustomUser.is_active) attr that subclassing `AbstractUser` requires that the django auth machinery uses behind the scenes.

@gvwilson I amended the todo list, yay for simplicity!

> c) can log in and modify anything they want (me, Arliss, Giacomo, maybe one more person)

I'd say these accounts can either be staff or superusers.

pbanaszkiewicz:
Just a suggestion: maybe `contactable` will serve better than `active`?

gvwilson:
On 2015-01-21 9:35 AM, Scott Burns wrote:
>
>     The "active" flag in the Person class indicates whether they're
>     currently interested in being contacted, taking part in workshops,
>     etc.
>
> @wking <https://github.com/wking> the help text on this field is good, 
> but we need to make sure future selves/others don't mix this up w/ the 
> |is_active| 
> <https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#django.contrib.auth.models.CustomUser.is_active> 
> attr that subclassing |AbstractUser| requires that the django auth 
> machinery uses behind the scenes.
>

I'm +1 on changing the name to "may_contact" or something more explicit.

> @gvwilson <https://github.com/gvwilson> I amended the todo list, yay 
> for simplicity!
>
>     c) can log in and modify anything they want (me, Arliss, Giacomo,
>     maybe one more person)
>
> I'd say these accounts can either be staff or superusers.
>

+1.

wking:
On Wed, Jan 21, 2015 at 06:12:26AM -0800, Greg Wilson wrote:
> 1. The "active" flag in the Person class indicates whether they're
> currently interested in being contacted, taking part in workshops,
> etc.

For previous discussion, see #127 and #130.

> there are only going to be half a dozen people with login privileges
> this year (because the DB holds personal identifying information,
> etc.).

That's a permissions issue to me, not a “don't let anyone else log in”
issue.  But I'm happy to postpone that discussion until we have admin
logins.

wking:
On Wed, Jan 21, 2015 at 06:35:50AM -0800, Scott Burns wrote:
> > The "active" flag in the Person class indicates whether they're
> > currently interested in being contacted, taking part in workshops,
> > etc.
> 
> @wking the help text on this field is good, but we need to make sure
> future selves/others don't mix this up w/ the `is_active` attr that
> subclassing `AbstractUser` requires that the django auth machinery
> uses behind the scenes.

I wasn't going to use the is_active field.  The docs for
User.is_active say [1]:

  This doesn’t necessarily control whether or not the user can log
  in. Authentication backends aren’t required to check for the
  is_active flag, and the default backends do not. If you want to
  reject a login based on is_active being False, it’s up to you to
  check that in your own login view or a custom authentication
  backend. However, the AuthenticationForm used by the login() view
  (which is the default) does perform this check, as do the
  permission-checking methods such as has_perm() and the
  authentication in the Django admin. All of those functions/methods
  will return False for inactive users.

For folks who can't login, I was going to use set_unusable_password
[2].  Still, I'm fine renaming the field to something else, and I've
spun that off into #159.

[1]: https://docs.djangoproject.com/en/1.7/ref/contrib/auth/#django.contrib.auth.models.User.is_active
[2]: https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#django.contrib.auth.models.AbstractBaseUser.set_unusable_password

workshops/models.py
wking: I got a bit further this morning, and spun off a blocking issue into
#160.  I'll continue once we've resolved that.

pbanaszkiewicz:
Rebase done in https://github.com/pbanaszkiewicz/amy/tree/wking-use-person-for-auth. Now what can I do with it?

gvwilson:
Can this now be merged as-is, even though authentication isn't yet 
implemented?  Any reason *not* to?

wking:
On Fri, Jan 30, 2015 at 06:59:27PM -0800, Greg Wilson wrote:
> Can this now be merged as-is, even though authentication isn't yet
> implemented?  Any reason *not* to?

@pbanaszkiewicz's found docs showing that we *do* need a
PersonManager, although with different methods than we had before
[1,2].  Keeping this branch out of master until we've found and fixed
those implementation issues gives us the freedom to massage the
history until we have a logical series of commits that explains why
we're making the changes we're making.  Merging it now gets you
nothing that you can't get just by checking out the branch ;).  So I
see a downside and no upside to merging early.

[1]: https://github.com/wking/swc-amy/pull/3#issuecomment-72261035
[2]: https://github.com/wking/swc-amy/pull/3#issuecomment-72263592

pbanaszkiewicz:
Hey @gvwilson 

I believe that we can merge this as-is after implementing `PersonManager`. I'll do it today. Right now it's kind-of a bug that we don't have it.

Implementing `PersonManager` will let us create super users and it will be a bridge between auth-foundation and TODOs by @sburns (https://github.com/swcarpentry/amy/pull/147#issuecomment-70842356).

gvwilson:
Yes please and thank you - I'll have a couple of hours this afternoon 
(my time) to pick up from where you leave off.

pbanaszkiewicz:
@gvwilson `PersonManager` and some other fixes/additions are now in https://github.com/wking/swc-amy/issues/3

I'll hopefully add log in, log out now. I want to get as much as possible before @wking starts reviewing. Which, by the way, you are welcome to do, too (https://github.com/wking/swc-amy/issues/3).

PR #152, opened 2015-01-18
Fixing refs from people to tasks

gvwilson:


wking:
On Sun, Jan 18, 2015 at 11:31:58AM -0800, Greg Wilson wrote:
>     A schema.sql (43)

Why list the schema, since that information is already encoded in the
Django models, and the exact schema commands are already in db.sql.

gvwilson:
I'm finding it useful to be able to diff the actual schema against the 
last saved schema when migrating - even small changes to properties of 
models (e.g., nullability) changes the order of fields in the tables, 
and tracking that down is a constant pain in the ass.

wking:
On Sun, Jan 18, 2015 at 12:16:42PM -0800, Greg Wilson wrote:
> I'm finding it useful to be able to diff the actual schema against
> the last saved schema when migrating - even small changes to
> properties of models (e.g., nullability) changes the order of fields
> in the tables, …

Maybe you should explicitly name your columns instead of relying on
the ordering?  From the SQLite docs [1], you should be able to:

  INSERT INTO workshops_site (<column names>) VALUES (<values>);

Something like:

  table_name = 'workshops_site'
  data = {
    'site': …,
    'fullname': …,
    'country': …,
    }
  keys = list(data.keys())
  operation = 'INSERT INTO {name} ({columns}) VALUES ({values})'.format(
    name=table_name,
    columns=', '.join(keys),
    values=', '.join(':{}'.format(key) for key in keys))
  cursor.execute(operation, data)

using the named paramstyle [2].

[1]: https://www.sqlite.org/lang_insert.html
[2]: https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.execute

gvwilson:
Yeah, tried that, it didn't seem to work - I'll give it another whack.
Thanks,
Greg

On 2015-01-18 4:21 PM, W. Trevor King wrote:
> On Sun, Jan 18, 2015 at 12:16:42PM -0800, Greg Wilson wrote:
> > I'm finding it useful to be able to diff the actual schema against
> > the last saved schema when migrating - even small changes to
> > properties of models (e.g., nullability) changes the order of fields
> > in the tables, …
>
> Maybe you should explicitly name your columns instead of relying on
> the ordering? From the SQLite docs [1], you should be able to:
>
> INSERT INTO workshops_site (<column names>) VALUES (<values>);
>
> Something like:
>
> table_name = 'workshops_site'
> data = {
> 'site': …,
> 'fullname': …,
> 'country': …,
> }
> keys = list(data.keys())
> operation = 'INSERT INTO {name} ({columns}) VALUES ({values})'.format(
> name=table_name,
> columns=', '.join(keys),
> values=', '.join(':{}'.format(key) for key in keys))
> cursor.execute(operation, data)
>
> using the named paramstyle [2].
>
> [1]: https://www.sqlite.org/lang_insert.html
> [2]: 
> https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.execute
>
> —
> <https://github.com/swcarpentry/amy/pull/152#issuecomment-70426603>.
>

-- 

gvwilson:
OK, I've found something similar that works (and that I understand) - 
thanks very much for the tip.

PR #155, opened 2015-01-20
Add requirements.txt and update README

sburns:
The `requirements.txt` is a standard file to have in a python project
to declare dependencies. Also allows to pin dependencies to specific versions.
Updated README with instructions for using it to install.

If CI is ever added to the project, this file will also help with setup.


wking:
>   * Add requirements.txt

It looks like you're working off an old version of amy (19fb7d42, One
logo per page is enough, 2014-12-04).  The current version has a
requirements.txt file already [1], and has since #65 landed on
2014-12-15.

[1]: https://github.com/swcarpentry/amy/blob/master/requirements.txt

sburns:
Sorry should have fetched upstream :(

PR #156, opened 2015-01-20
Add default ordering to Event

### LINKED ISSUE #150: Events should be ordered by start date, not by insertion order
The default display of events should sort them by start date (or equivalently slug), not by table insertion order.  (Right now, training cohorts are converted to events after actual events have been added, so all the training cohorts show up out of time order.)

sburns:
Fixes #150.

### DIFF
--- a/workshops/migrations/0040_auto_20150120_1610.py
+++ b/workshops/migrations/0040_auto_20150120_1610.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('workshops', '0039_remove_tag_slug'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='event',
+            options={'ordering': ('-start',)},
+        ),
+    ]
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -225,6 +225,9 @@ class Event(models.Model):
     admin_fee  = models.DecimalField(max_digits=6, decimal_places=2, null=True, blank=True)
     notes      = models.TextField(default="", blank=True)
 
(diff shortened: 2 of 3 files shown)


wking:
On Tue, Jan 20, 2015 at 01:14:24PM -0800, Scott Burns wrote:
>   * Add default ordering to Event

Where else do we list events?  Do we want to add order_by() bits there
to avoid useless sorting when we don't need it?  From [1]:

  If you don’t want any ordering to be applied to a query, not even
  the default ordering, call order_by() with no parameters.

[1]: https://docs.djangoproject.com/en/1.7/ref/models/querysets/#django.db.models.query.QuerySet.order_by

sburns:
The admin page?

wking:
On Tue, Jan 20, 2015 at 01:55:24PM -0800, Scott Burns wrote:
> The admin page?

We'd want that sorted the same way, I think.  I don't see any
event_set or other calls that use Event queries for calculations
(instead of for displaying the query results to the user), but I
haven't looked all that hard.

sburns:
I went with Meta ordering so future devs get the ordering "for free". If you'd prefer just ordering in the view, let me know.

gvwilson:
Looks good to me.

wking:
On Tue, Jan 20, 2015 at 11:48:04PM -0800, Scott Burns wrote:
> I went with Meta ordering so future devs get the ordering "for
> free". If you'd prefer just ordering in the view, let me know.

I like the Meta ordering (clearing ordering where you don't need it is
DRYer than trying to manually setup the same ordering where you do
need it), but when adding Meta ordering I think it's good to check for
any existing calls that don't need newly-default ordering.  I haven't
turned any up (but as I said I didn't look very hard).  If noone else
is interested in looking, I'm fine with merging this as is.  If we
find unorderable calls we missed, we can always fix them later.

gvwilson:
+1.

PR #160, opened 2015-01-21
workshops/models.py: Drop PersonQuerySet and PersonManager

wking:
These don't actually save us any work, since the instructors() view is
simpler without them.

They also break superuser addition when using Person as the user
model.  In my use-person-for-auth branch (swcarpentry/amy#147) without
this change, I see:

    $ python3.4 ./manage.py syncdb
    ...
    You have installed Django's auth system, and don't have any superusers defined.
    Would you like to create one now? (yes/no): yes
    Username: ...
      File "./manage.py", line 10, in <module>
      File "/.../django/core/management/__init__.py", line 385, in execute_from_command_line
        utility.execute()
      File "/.../django/core/management/__init__.py", line 377, in execute
      File "/.../django/core/management/base.py", line 288, in run_from_argv
        self.execute(*args, **options.__dict__)
      File "/.../django/core/management/base.py", line 338, in execute
      File "/.../django/core/management/base.py", line 533, in handle
        return self.handle_noargs(**options)
      File "/.../django/core/management/commands/syncdb.py", line 46, in handle_noargs
        call_command("createsuperuser", interactive=True, database=options['database'])
      File "/.../django/core/management/__init__.py", line 115, in call_command
        return klass.execute(*args, **defaults)
      File "/.../django/contrib/auth/management/commands/createsuperuser.py", line 55, in execute
        return super(Command, self).execute(*args, **options)
      File "/.../django/core/management/base.py", line 338, in execute
      File "/.../django/contrib/auth/management/commands/createsuperuser.py", line 113, in handle
        self.UserModel._default_manager.db_manager(database).get_by_natural_key(username)
    AttributeError: 'PersonManager' object has no attribute 'get_by_natural_key'

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -48,33 +48,6 @@ class Airport(models.Model):
 
 #------------------------------------------------------------
 
-# Define custom queries on the QuerySet and the Manager
-class PersonQuerySet(models.query.QuerySet):
-    '''Handles finding past, ongoing and upcoming events'''
-
-    def have_skills(self, skills):
-        '''Returns persons who have all the skills listed
-        in skills, which must be a list of Skill objects.
-        '''
-
-        for s in skills:
-            self = self.filter(qualification__skill=s)
-
-        return self
-
-class PersonManager(models.Manager):
-    '''A custom manager which is essentially a proxy for PersonQuerySet'''
-
-    # Attach our custom query set to the manager
-    def get_queryset(self):
-        return PersonQuerySet(self.model, using=self._db)
-
-    # Proxy methods so we can call our custom filters from the manager
-    # without explicitly creating an PersonQuerySet first
-
-    def have_skills(self, skills):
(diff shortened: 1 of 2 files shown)


gvwilson:
OK - do we know why they were created in the first place? @rbeagrie?

wking:
On Wed, Jan 21, 2015 at 12:26:11PM -0800, Greg Wilson wrote:
> OK - do we know why they were created in the first place? @rbeagrie?

The initial commit is 8da67886 (Added have_qualifications method to
Person QuerySets, 2014-12-16), which has:

> This commit adds a new method to Person QuerySets called
> have_qualifications. It takes a list of Skill objects and returns
> only those Person objects which have ALL of the listed skills. I
> also added a usage note to the "match" view.

That's a good explanation of what it's doing, but not much help for
why it's using a PersonManager to do it.

PR #161, opened 2015-01-22
Renaming Person.active to Person.may_contact

### LINKED ISSUE #159: Rename Person.active to show that it's about allowing contact
Spun off from #147.  See also, #127 and #130.

On Wed, Jan 21, 2015 at 06:35:50AM -0800, Scott Burns wrote:
> > The "active" flag in the Person class indicates whether they're
> > currently interested in being contacted, taking part in workshops,
> > etc.
>
> @wking the help text on this field is good, but we need to make sure
> future selves/others don't mix this up w/ the `is_active` attr that
> subclassing `AbstractUser` requires that the django auth machinery
> uses behind the scenes.

I'm fine renaming the field to something else. `contactable` sounds like “is it possible to contact them?” to me. How about `away` (or `away_until`) as a DateField, with the following
semantics:

* `away == Null`, for “feel free to contact me”
* `away > today()`, for “I expect I'll be away until ‘away’.  Don't
    contact me”
* `away <= today()`, for “I thought I'd be back by now, but I'm not.
    Don't contact me”

If we want to stick with the boolean, maybe `open_to_communication`?

gvwilson:
1.  Renamed `Person` model field from `active` to `may_contact`.
2.  Updated migrations.
3.  Updated `migrater.py`.
4.  Updated dependent code and templates.

Closes #159.


pbanaszkiewicz:
This looks good to me.

wking:
On Thu, Jan 22, 2015 at 07:01:30AM -0800, Greg Wilson wrote:
>     A workshops/migrations/0045_auto_20150122_0957.py (24)

This is currently “remove & add” not “rename”.  We're blowing away
migrations in #147 anyway, so it doesn't really matter, but in the
future I'd recommend RenameField [1].

>     M workshops/models.py (23)

I'm not a big fan of aligning the equals signs, since it creates
unneeded churn in cases like this that trigger realignment.  That
makes it harder to blame through the history to figure out why they're
the way they are.  For example, blaming the current master shows that
Person.email is blankable because it helps with testing (4659c064).
After this pull request, the blamed commit for Person.email will be
dbdd9195, which has nothing to do with Person.email.

Other than that this looks good to me.

[1]: https://docs.djangoproject.com/en/1.7/ref/migration-operations/#renamefield

gvwilson:
On 2015-01-23 2:12 AM, W. Trevor King wrote:
> On Thu, Jan 22, 2015 at 07:01:30AM -0800, Greg Wilson wrote:
> > A workshops/migrations/0045_auto_20150122_0957.py (24)
>
> This is currently “remove & add” not “rename”. We're blowing away
> migrations in #147 anyway, so it doesn't really matter, but in the
> future I'd recommend RenameField [1].
>
I just ran:

~~~
python manage.py makemigrations
python manage.py migrate
~~~

I'm guessing it doesn't automatically use RenameField?
> > M workshops/models.py (23)
>
> I'm not a big fan of aligning the equals signs, since it creates
> unneeded churn in cases like this that trigger realignment.

I find it much easier to read, but my eyesight shouldn't be a 
constraint.  I'll avoid realignment in future.

wking:
On Fri, Jan 23, 2015 at 02:04:02AM -0800, Greg Wilson wrote:
> 2015-01-23 2:12 AM, W. Trevor King:
> > Thu, Jan 22, 2015 at 07:01:30AM -0800, Greg Wilson:
> > > A workshops/migrations/0045_auto_20150122_0957.py (24)
> >
> > This is currently “remove & add” not “rename”. We're blowing away
> > migrations in #147 anyway, so it doesn't really matter, but in the
> > future I'd recommend RenameField [1].
>
> I just ran:
> 
> ~~~
> python manage.py makemigrations
> python manage.py migrate
> ~~~
> 
> I'm guessing it doesn't automatically use RenameField?

Apparently not, although I have a fuzzy memory of makemigrations
prompting me once (“do you want to rename $X to $Y?”).  Maybe I just
dreamed about it ;).  Unless that prompt is real, I guess we'll just
peek into the migration after we create it, or catch it on review ;).

PR #163, opened 2015-01-23
Recording whether the admin fee has been paid or not.

gvwilson:
1.  Add a new field to the Event class.
2.  Migrate.
3.  Update views.


gvwilson:
@pbanaszkiewicz please review?

pbanaszkiewicz:
@gvwilson I guess we both resolved conflicts in `db.sql` in the pretty same time…

pbanaszkiewicz:
@gvwilson one question: is `admin_fee` field NULL-able by design? If yes then why?

The PR looks good to me.

gvwilson:
"NULL" means we don't know, it hasn't been decided, etc. - "0.0" means 
"we've decided, and there isn't one".

pbanaszkiewicz:
@gvwilson alright!

wking:
On Tue, Jan 27, 2015 at 10:12:17AM -0800, Piotr Banaszkiewicz wrote:
> @gvwilson one question: is `admin_fee` field NULL-able by design? If
> yes then why?

My guess would be: yes it's nullable, and a NULL-value means “we
haven't decided what the admin fee should be yet”.  A zero-value means
“we've decided to waive the admin fee”.

PR #164, opened 2015-01-23
Fix: '$' was missing in URL regex for task edit

pbanaszkiewicz:
Missing '$' (matches end-of-line or something like that) made
task edit view render task detail view.


wking:
On Fri, Jan 23, 2015 at 04:51:39AM -0800, Piotr Banaszkiewicz wrote:
>   * Fix: '$' was missing in URL regex for task edit

Looks good to me.

PR #165, opened 2015-01-25
Add various tweaks and improvements to CSV parsing

sburns:
* Redirect when no session data after bulk upload
* case-insensitive email look up
* Move csv parsing to util.py, add test
* Move verification to util.py, add tests
* Ensure data structure from csv parsing is json-serializable

### DIFF
--- a/workshops/test/test_util.py
+++ b/workshops/test/test_util.py
@@ -1,6 +1,7 @@
 # this probably needs refactored for py3
 from datetime import datetime
 from StringIO import StringIO
+from json import dumps
 
 from django.test import TestCase
 
@@ -40,6 +41,15 @@ class UploadPersonTaskCSVTestCase(TestCase):
         person = self.compute_from_string(csv)[0]
 
         self.assertEqual(person['person']['middle'], '')
+
+    def test_serializability_of_parsed(self):
+        csv = """personal,middle,family,email\njohn,a,doe,johndoe@email.com\n,jane,a,doe,janedoe@email.com"""
+        person_tasks = self.compute_from_string(csv)
+
+        try:
+            dumps(person_tasks)
+        except Exception:
+            self.fail('Dumping person_tasks to JSON unexpectedly failed!')
 
 
 class VerifyUploadPersonTask(TestBase):
--- a/workshops/test/test_util.py
+++ b/workshops/test/test_util.py
@@ -1,10 +1,14 @@
 # this probably needs refactored for py3
+from datetime import datetime
 from StringIO import StringIO
 
 from django.test import TestCase
 
-from .util import PERSON_UPLOAD_FIELDS, upload_person_task_csv
+from ..models import Site, Event, Role, Person
+from ..util import PERSON_UPLOAD_FIELDS, upload_person_task_csv,\
+    verify_upload_person_task
 
+from .base import TestBase
 
 class UploadPersonTaskCSVTestCase(TestCase):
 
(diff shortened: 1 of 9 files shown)


gvwilson:
Thanks @sburns - @pbanaszkiewicz can you please review and then merge if you think it's good? This and logins and we're ready to deploy.

pbanaszkiewicz:
Thanks @sburns!

@gvwilson I'll review as soon as learn some [phase portraits](https://en.wikipedia.org/wiki/Phase_portrait), which should be in about hour.

workshops/util.py
```diff
@@ -28,3 +34,75 @@ def earth_distance(pos1, pos2):
...
+    """
+    Read data from CSV and turn it into JSON-serializable list of dictionaries.
+    "Serializability" is required because we put this data into session.  See
+    https://docs.djangoproject.com/en/1.7/topics/http/sessions/ for details.
+    """
+    persons_tasks = []
+    reader = csv.DictReader(uploaded_file)
+    for row in reader:
+        person_fields = dict((col, row[col].strip())
+                             for col in PERSON_UPLOAD_FIELDS)
```
pbanaszkiewicz: Malformed files (try uploading, for example, a python script) will throw KeyError here: no key 'personal'.

I think the generator may look something like this:

```python
(col, row.get(col, "").strip() for col in PERSON_UPLOAD_FIELDS)
```

But I don't know if this makes any sense… A file without required columns?

Additionally, we need test for this case.

workshops/views.py
```diff
@@ -317,73 +314,6 @@ def person_bulk_add_confirmation(request):
                       context)
```
pbanaszkiewicz: In lines 288 and 309 (views.py) there's still `_verify_upload_person_task` (should be `verify_upload_person_task`).

Why aren't tests catching this?

workshops/util.py
```diff
@@ -28,3 +34,75 @@ def earth_distance(pos1, pos2):
...
+                Role.objects.get(name=role)
+            except Role.DoesNotExist:
+                errors.append('Role with name {} does not exist.'.format(role))
+            except Role.MultipleObjectsReturned:
+                errors.append('More than one role named {} exists.'
+                              .format(role))
+
+        if email:
+            try:
+                Person.objects.get(email__iexact=email)
```
pbanaszkiewicz: I'm fine with using case-insensitive match `email__iexact=email`, but I think we might have minor design issue: our DB constraints allow to insert two same emails with different-case…

workshops/test/test_util.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def test_serializability_of_parsed(self):
+        csv = """personal,middle,family,email\njohn,a,doe,johndoe@email.com\n,jane,a,doe,janedoe@email.com"""
+        person_tasks = self.compute_from_string(csv)
+
+        try:
+            dumps(person_tasks)
+        except Exception:
+            self.fail('Dumping person_tasks to JSON unexpectedly failed!')
+
```
pbanaszkiewicz: As I mentioned in https://github.com/swcarpentry/amy/pull/165/files#r23504978, we need to test against the regression with malformed files being uploaded.  It's enough to test for CSV files that have columns with wrong names.

workshops/test/test_util.py
```diff
@@ -0,0 +1,138 @@
...
+    def compute_from_string(self, csv_str):
+        ''' wrap up buffering the raw string & parsing '''
+        csv_buf = StringIO(csv_str)
+        # compute & return
+        return upload_person_task_csv(csv_buf)
+
+    def test_basic_parsing(self):
+        ''' See views.PERSON_UPLOAD_FIELDS for field ordering '''
+        csv = """personal,middle,family,email\njohn,a,doe,johndoe@email.com\n,jane,a,doe,janedoe@email.com"""
```
pbanaszkiewicz: (Personal preference) I'd rather have this split into multiple lines and not use `"\n"` explicitely. You have `"""` already.

workshops/test/test_util.py
```diff
@@ -0,0 +1,138 @@
...
+        # assert
+        self.assertEqual(len(person_tasks), 2)
+
+        person = person_tasks[0]
+        for key in ('person', 'role', 'event', 'errors'):
+            self.assertIn(key, person)
+
+        person_dict = person['person']
+        for key in PERSON_UPLOAD_FIELDS:
+            self.assertIn(key, person_dict)
```
pbanaszkiewicz: These two for-loops can be squizzed into:
```python
    self.assertSetEqual(set(('person', 'role', 'event', 'errors')), set(person.keys()))
    self.assertSetEqual(set(PERSON_UPLOAD_FIELDS), set(person_dict.keys()))
```
The advantage is that if `person` or `person_dict` have more keys than we anticipated, we'll get an error.  This ensures that `PERSON_UPLOAD_FIELDS` is up-to-date with `Person` model.

workshops/test/test_util.py
```diff
@@ -0,0 +1,138 @@
...
+
+        self.assertEqual(person['person']['middle'], '')
+
+    def test_serializability_of_parsed(self):
+        csv = """personal,middle,family,email\njohn,a,doe,johndoe@email.com\n,jane,a,doe,janedoe@email.com"""
+        person_tasks = self.compute_from_string(csv)
+
+        try:
+            dumps(person_tasks)
+        except Exception:
```
pbanaszkiewicz: We're using [Django wrapper around dumps](https://docs.djangoproject.com/en/1.7/_modules/django/core/signing/#dumps), not "pure" Python JSON serializer.

Please switch to Django `dumps` and check for proper exception (ie. TypeError).

pbanaszkiewicz:
Hey @sburns,

this PR looks okay. Mostly I need you to fix these small errors and ensure they don't pop out anymore. Tests don't catch them, so we need some tests to test this part of the code.

Other than that there are only some minor improvements to test suite.

Cheers,
Piotr

workshops/test/test_util.py
sburns: I was looking at `PERSON_UPLOAD_FIELDS` and `static/BulkPersonAddTemplate.csv`...this file don't match the fields as it is. I agree we need to test against malformed csv. But we should generate this template as runtime using this single source of column names.

sburns:
- [ ] switch to django's json dumps
- [ ] squeeze basic parse test into equal sets
- [ ] explicit newline in test csv string
- [ ] generate BulkPersonAddTemplate.csv dynamically using PERSON_UPLOAD_FIELDS as columns.
- [ ] show useful error message for KeyErrors in upload_person_task

workshops/test/test_util.py
pbanaszkiewicz: I didn't look at `BulkPersonAddTemplate.csv` but it's a great idea to keep it up to date with `PERSON_UPLOAD_FIELDS`.

Maybe we should throw `PERSON_UPLOAD_FIELDS` into `Person` model definition with a comment suggesting to keep that list up-to-date with `Person` fields?

workshops/util.py
```diff
@@ -1,5 +1,11 @@
+import csv
 from math import pi, sin, cos, acos
 
+from .models import Event, Role, Person
+
+PERSON_UPLOAD_FIELDS = ['personal', 'middle', 'family', 'email']
+PERSON_TASK_UPLOAD_FIELDS = PERSON_UPLOAD_FIELDS + ['event', 'role']
```
pbanaszkiewicz: Please make `PERSON_UPLOAD_FIELDS` and `PERSON_TASK_UPLOAD_FIELDS` immutable by turning them into tuples.

Additionally you can move them into `Person` model definition as stated here: https://github.com/swcarpentry/amy/pull/165#discussion_r23534853

sburns:
@pbanaszkiewicz I think I hit everything, let me know?

pbanaszkiewicz:
@sburns thanks a lot! I'll check this out tomorrow and let you know.

workshops/util.py
```diff
@@ -28,3 +66,84 @@ def earth_distance(pos1, pos2):
...
+            except Event.DoesNotExist:
+                errors.append('Event with slug {} does not exist.'
+                              .format(event))
+        if role:
+            try:
+                Role.objects.get(name=role)
+            except Role.DoesNotExist:
+                errors.append('Role with name {} does not exist.'.format(role))
+            except Role.MultipleObjectsReturned:
+                errors.append('More than one role named {} exists.'
```
pbanaszkiewicz: Strings in lines 122, 128 and 130 (`errors.append(…)`) should be unicode (ie. `u'Event with slug…'`). Otherwise when uploading non-ASCII characters there would be errors in these lines.

pbanaszkiewicz:
@sburns as part of the code review I found three bugs (including the bug with unicode that I commented in https://github.com/swcarpentry/amy/pull/165#discussion_r23617908).

All three are basically bugs in my original code, and I have fixed them to make sure the code's working, so I'll simply send you a PR so that you can update this PR with our code.

When it comes to your latest additions to this PR, I'm really happy with your code.

To sum up: thank you for your updates and tests. They're great. I'll now send you a PR, so you can merge it and then this PR (#165) should update automatically. Finally we'll be able to successfully merge this PR (#165) into the master branch.

pbanaszkiewicz:
@sburns: here you go https://github.com/sburns/amy/pull/1.
It took longer that I anticipated :/ but there should be less errors now.

sburns:
Merged, thanks @pbanaszkiewicz. @gvwilson I think this is ready to go.

pbanaszkiewicz:
Okay, on the permission given me by @gvwilson in https://github.com/swcarpentry/amy/pull/165#issuecomment-71377130 I shall merge.

PR #172, opened 2015-01-29
Fix Event identification regex to include YYYY-MM format

### LINKED ISSUE #171: Crash when viewing upcoming event
This is either a data import error or a bug in Amy; either way, Amy should give something better than a stack trace when something goes wrong.

1.  `make clean`
2.  `make database` to import faked data.
3.  `make serve` to run.
4.  go to http://127.0.0.1:8000/workshops/
5.  fifth event down under 'upcoming events' is labelled 2015-04-xx-ttt-nih - click on it.
6.  boom: "ObjectDoesNotExist at /workshops/event/2015-04-xx-ttt-nih"

Other upcoming events display properly, and occurs as of b2e0c6058.

pbanaszkiewicz:
When fetching Event by identification text, which could be either event
ID or slug, format `'YYYY-MM-text'` was not correctly matched.  Both
`'YYYY-MM-DD-text'` and `'ID'` worked correctly, though.

This is now fixed.  Fetching Event by ident now checks these 3 formats
(in this order):

* `'ID'`

* `'YYYY-MM-****'`

* `'YYYY-MM-DD-****'`

This fixes #171 .

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -232,12 +232,9 @@ class Event(models.Model):
         if re.match(r'^\d+$', ident):
             return Event.objects.get(id=ident)
 
-        # match event slug in format YYYY-MM-****
+        # match event slug in format YYYY-MM-**** (ie. this works for
+        # YYYY-MMMM-DD-**** too)
         if re.match(r'^\d{4}-\d{2}-.+$', ident):
-            return Event.objects.get(slug=ident)
-
-        # match event slug in format YYYY-MM-DD-****
-        if re.match(r'^\d{4}-\d{2}-\d{2}-.+$', ident):
             return Event.objects.get(slug=ident)
 
         raise ObjectDoesNotExist(ident)
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -228,9 +228,15 @@ class Event(models.Model):
         The real indicator is the 'published' flag.
         '''
 
+        # match event ID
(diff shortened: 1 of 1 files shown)


workshops/models.py
```diff
@@ -228,9 +228,15 @@ def get_by_ident(ident):
         The real indicator is the 'published' flag.
         '''
 
+        # match event ID
         if re.match(r'^\d+$', ident):
             return Event.objects.get(id=ident)
 
+        # match event slug in format YYYY-MM-****
+        if re.match(r'^\d{4}-\d{2}-.+$', ident):
```
gvwilson: Doesn't this capture all of the cases captured by the pattern below?

pbanaszkiewicz: Welp, you're right, it does capture these cases, but there's no error because all we do in both cases is `return Event.objects.get(slug=ident)`. I'll add fixing commit in a few minutes.

pbanaszkiewicz: @gvwilson Is this safe to assume that we know for sure event's year and month, but not necessarily event's day?

gvwilson: I think we can assume YYYY-MM-something, so there are just two cases: 
the numeric object ID, and something that looks like \d{4}-\d{2}-.+

pbanaszkiewicz: Okay, great. That's what I left in the code.

PR #174, opened 2015-01-31
More Python 3 fixes for the test suite

wking:
Fix Python 3 incompatibilities that have landed since #140.  After
this series,

  $ python3.4 manage.py test

passes again.


wking:
My last round of fixes landed in #140 via gvwilson/amy#12.

PR #175, opened 2015-02-02
Undoing Py3 compatibility for StringIO in order to get tests to pass

gvwilson:
0dae639 changed the import of StringIO to be compatible with Py3 as well as Py2.7.  Problem is, tests of CSV upload fail with it in place.  Since we're deploying on Py2.7, I'm submitting a PR to undo this change - while we need Unicode support for CSV upload, (a) the rest of our code hasn't been tested Unicode-clean yet and (b) this is blocking work on auth.

### DIFF
--- a/workshops/test/test_util.py
+++ b/workshops/test/test_util.py
@@ -1,10 +1,7 @@
 # coding: utf-8
 import cgi
 from datetime import datetime
-try:
-    from io import StringIO
-except ImportError:
-    from StringIO import StringIO
+from StringIO import StringIO
 from importlib import import_module
 
 from django.conf import settings
--- a/workshops/util.py
+++ b/workshops/util.py
@@ -1,9 +1,6 @@
 # coding: utf-8
 import codecs
-try:
-    from io import StringIO
-except ImportError:
-    from cStringIO import StringIO
+from cStringIO import StringIO
 import csv
 from math import pi, sin, cos, acos
 


pbanaszkiewicz:
Hi Greg, we can have Trevor's change that makes test suite pass on Py3 and on Py2.

All we have to do is change strings to unicode strings, for example:

```python
# bad
s = "Windows"

# good
s = u"Linux"
```

I'll send you a PR against your PR to fix this, okay?

pbanaszkiewicz:
@gvwilson https://github.com/gvwilson/amy/pull/13

wking:
On Mon, Feb 02, 2015 at 06:58:01AM -0800, Piotr Banaszkiewicz wrote:
> All we have to do is change strings to unicode strings, for example:
> 
> ```python
> # bad
> s = "Windows"
> 
> # good
> s = u"Linux"
> ```

Alternatively, we can embrace the fact that we're going to need
Unicode literals much more frequently than we'll need byte literals
(which we may never need).  Then just:

  from __future__ import unicode_literals

in every file that needs Unicode literals [1,2].

[1]: https://docs.python.org/2/library/__future__.html
[2]: https://www.python.org/dev/peps/pep-3112/

wking:
On Mon, Feb 02, 2015 at 06:40:34AM -0800, Greg Wilson wrote:
> Since we're deploying on Py2.7, I'm submitting a PR to undo this
> change - while we need Unicode support for CSV upload, (a) the rest
> of our code hasn't been tested Unicode-clean yet…

Is the Python 2.7 deployment a hard-requirement?  If we deploy with
Python 3, we're automatically Unicode clean.  That's a whole class of
issues that we wouldn't have to worry about.

gvwilson:
It's up to @drio - David, are you OK blowing away the Py2.7 and using Py3?

gvwilson:
Better solutions are emerging.

PR #176, opened 2015-02-04
Remove Unicode Py2 shenanigans

pbanaszkiewicz:
This commit removes conditional StringIO import in a few source files as
well as fixes uploading unicode CSV-files.

Conditional import
==================

To support both Py2 and Py3 we had conditional imports:

```python
try:
	from io import StringIO
except ImportError:
	from cStringIO import StringIO
```

By moving to Py3-only we can drop these imports in favor of nice and
simple `from io import StringIO`.

Unicode issues with uploaded CSV files
======================================

By moving to Py3, it appeared that Django serves uploaded files as byte
files, not text files.  This forces us to encode upcoming files into
UTF-8 text files (which `TextIOWrapper` does).

`UnicodeWriter` removal
=======================

Helper class `UnicodeWriter` was required to make unicode CSV files,
because `csv` module from Python 2.x standard library couldn't handle
unicode files.

With migration to Python 3.x we can get rid of it, because now `csv`
module fully supports unicode files.

### DIFF
--- a/workshops/test/test_util.py
+++ b/workshops/test/test_util.py
@@ -1,10 +1,7 @@
 # coding: utf-8
 import cgi
 from datetime import datetime
-try:
-    from io import StringIO
-except ImportError:
-    from StringIO import StringIO
+from io import StringIO
 from importlib import import_module
 
 from django.conf import settings
--- a/workshops/util.py
+++ b/workshops/util.py
@@ -2,6 +2,7 @@
 from math import pi, sin, cos, acos
 from io import TextIOWrapper, StringIO
 import csv
+
 from .models import Event, Role, Person
 
 
--- a/workshops/util.py
+++ b/workshops/util.py
@@ -3,6 +3,8 @@ from math import pi, sin, cos, acos
 from io import TextIOWrapper, StringIO
 import csv
 
+from django.conf import settings
+
 from .models import Event, Role, Person
 
 
@@ -36,7 +38,7 @@ def earth_distance(pos1, pos2):
     return arc * 6373
 
 
-def upload_person_task_csv(uploaded_file, encoding="utf-8"):
+def upload_person_task_csv(uploaded_file, encoding=None):
     """
     Read data from CSV and turn it into JSON-serializable list of dictionaries.
     "Serializability" is required because we put this data into session.  See
@@ -44,8 +46,14 @@ def upload_person_task_csv(uploaded_file, encoding="utf-8"):
 
     Also return a list of fields from Person.PERSON_UPLOAD_FIELDS for which
     no data was given.
+
+    :param string encoding: encoding used to encode incoming file. Defaults to
+                            ``django.conf.settings.DEFAULT_CHARSET``
     """
     persons_tasks = []
+
+    if not encoding:
+        encoding = settings.DEFAULT_CHARSET
 
     # we provide uploaded_file as StringIO in our tests (test_util.py)
     if not issubclass(StringIO, uploaded_file.__class__):
--- a/workshops/util.py
+++ b/workshops/util.py
@@ -1,45 +1,8 @@
 # coding: utf-8
-import codecs
-try:
-    from io import StringIO
-except ImportError:
-    from cStringIO import StringIO
-import csv
 from math import pi, sin, cos, acos
-
+from io import TextIOWrapper, StringIO
+import csv
 from .models import Event, Role, Person
-
-
-class UnicodeWriter:
-    """
-    A CSV writer which will write rows to CSV file "f",
-    which is encoded in the given encoding.
-
-    https://docs.python.org/2/library/csv.html#examples
-    """
-
-    def __init__(self, f, dialect=csv.excel, encoding="utf-8", **kwds):
-        # Redirect output to a queue
-        self.queue = StringIO()
-        self.writer = csv.writer(self.queue, dialect=dialect, **kwds)
-        self.stream = f
-        self.encoder = codecs.getincrementalencoder(encoding)()
-
-    def writerow(self, row):
-        self.writer.writerow([s.encode("utf-8") for s in row])
-        # Fetch UTF-8 output from the queue ...
-        data = self.queue.getvalue()
-        data = data.decode("utf-8")
-        # ... and reencode it into the target encoding
-        data = self.encoder.encode(data)
-        # write to the target stream
-        self.stream.write(data)
-        # empty queue
-        self.queue.truncate(0)
-
-    def writerows(self, rows):
-        for row in rows:
-            self.writerow(row)
 
 
 def earth_distance(pos1, pos2):
@@ -72,7 +35,7 @@ def earth_distance(pos1, pos2):
     return arc * 6373
 
 
-def upload_person_task_csv(uploaded_file):
+def upload_person_task_csv(uploaded_file, encoding="utf-8"):
     """
     Read data from CSV and turn it into JSON-serializable list of dictionaries.
     "Serializability" is required because we put this data into session.  See
@@ -82,10 +45,18 @@ def upload_person_task_csv(uploaded_file):
     no data was given.
     """
     persons_tasks = []
+
+    # we provide uploaded_file as StringIO in our tests (test_util.py)
+    if not issubclass(StringIO, uploaded_file.__class__):
+        # Django provides uploaded files as byte file objects.  We need to wrap
+        # `uploaded_file` and provide it as a text file with specific encoding.
+        # This way we force users to upload UTF-8 encoded files.
+        uploaded_file = TextIOWrapper(uploaded_file.file, encoding=encoding)
+
     reader = csv.DictReader(uploaded_file)
     empty_fields = []
     for row in reader:
-        person_fields= {}
+        person_fields = {}
         for col in Person.PERSON_UPLOAD_FIELDS:
             try:
                 person_fields[col] = row[col].strip()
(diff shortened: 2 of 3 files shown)


pbanaszkiewicz:
Please review, @gvwilson and @wking.

When this gets merged in, I'll rebase my https://github.com/wking/swc-amy/pull/3 PR. Then we should think about auth issue with empty passwords. Then (or simultaneously) we should fix tests for auth.

wking:
On Wed, Feb 04, 2015 at 09:34:54AM -0800, Piotr Banaszkiewicz wrote:
>     M workshops/util.py (53)

It looks like you're removing the blank line between stdlib and local
imports here [1].  I'd also keep it alphabetical by package, which
would look like:

 import csv
 from math import pi, sin, cos, acos
 from io import TextIOWrapper, StringIO

 from .models import Event, Role, Person

although I'm fine with alphabetizing by line:

 from io import TextIOWrapper, StringIO
 from math import pi, sin, cos, acos
 import csv

 from .models import Event, Role, Person

If request.FILES['file'] is in bytes, upload_person_task_csv should
take a codec-wrapped file, and the codec should come from the upload
headers.  Then you can drop the TextIOWrapper/StringIO business from
upload_person_task_csv.  Using getreader [2], that would be something
like:

  _, params = cgi.parse_header(request['content-type'])
  charset = params['charset']
  stream = codecs.getreader(charset)(request.FILES['file'])
  persons_tasks, empty_fields = upload_person_task_csv(stream)

Other than that, this looks good to me.

[1]: https://www.python.org/dev/peps/pep-0008/#imports
[2]: https://docs.python.org/3/library/codecs.html#codecs.getreader

pbanaszkiewicz:
I will introduce removed blank line between imports. I group imports by source (standard library vs our code) and then order by appearance in the file, so I won't change the order.

> If request.FILES['file'] is in bytes, upload_person_task_csv should take a codec-wrapped file, and the codec should come from the upload headers. Then you can drop the TextIOWrapper/StringIO business from upload_person_task_csv. Using getreader [2], that would be something like:
>   ```
>    _, params = cgi.parse_header(request['content-type'])
>    charset = params['charset']
>   ```

Is it the same as `request.encoding`? (See https://docs.djangoproject.com/en/1.7/ref/request-response/#django.http.HttpRequest.encoding and https://docs.djangoproject.com/en/1.7/ref/settings/#std:setting-DEFAULT_CHARSET).

wking:
On Wed, Feb 04, 2015 at 10:00:03AM -0800, Piotr Banaszkiewicz wrote:
> I will introduce removed blank line between imports. I group imports
> by source (standard library vs our code) and then order by
> appearance in the file, so I won't change the order.

I prefer a more stable ordering, since you likely don't want to
reorder the imports just because you've shifted things around inside
the module.  But this really doesn't matter.

> >    _, params = cgi.parse_header(request['content-type'])
> >    charset = params['charset']
> 
> Is it the same as `request.encoding`?

Yeah, looks like it is :).  Definitely go with request.encoding then.

The fallback to DEFAULT_CHARSET is pretty hacky, since DEFAULT_CHARSET
is really about the *response* encoding, and should have nothing to do
with the request encoding.  I guess it's just used because you need
some sort of fallback for folks that don't set Content-Type on their
uploads.

pbanaszkiewicz:
Okay, so from what I tested: `request.encoding` is `None` when I upload files (for both UTF-16 and UTF-8).

So we cannot encode file based on `request.encoding`. And I have bad feelings about guessing the encoding from file contents.

Therefore I suggest that we default to UTF-8 encoding for uploded files.

wking:
On Wed, Feb 04, 2015 at 11:03:12AM -0800, Piotr Banaszkiewicz wrote:
> Okay, so from what I tested: `request.encoding` is `None` when I
> upload files (for both UTF-16 and UTF-8).

Hmm.  Looking at the docs again [1], HttpRequest.encoding seems to be
a *write* field.  Can you share the headers set by your upload?  You
should be able to see them in your browser's dev-tools network access.

[1]: https://docs.djangoproject.com/en/1.7/ref/request-response/#django.http.HttpRequest.encoding

pbanaszkiewicz:
Here you go.

For 'unicode' (I think LibreOffice calls UTF-16 'unicode'):
```
Host: 127.0.0.1:8000
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:35.0) Gecko/20100101 Firefox/35.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
DNT: 1
Referer: http://127.0.0.1:8000/workshops/persons/bulkadd/
Cookie: …
Connection: keep-alive
```

For UTF-8:
```
Host: 127.0.0.1:8000
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:35.0) Gecko/20100101 Firefox/35.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
DNT: 1
Referer: http://127.0.0.1:8000/workshops/persons/bulkadd/
Cookie: …
Connection: keep-alive
```

wking:
Hmm, I don't see Content-Type there at all.  Maybe we need to set
enctype [1] and/or accept-charset [2] for the form itself?  There's
also some discussion of encodings and forms here [3,4].  It looks like
the best approach is to set enctype to multipart/form-data.  We might
need to do other things too, but getting away from
application/x-www-form-urlencoded is a good first step.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-enctype
[2]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-accept-charset
[3]: http://www.w3.org/TR/html5/forms.html#selecting-a-form-submission-encoding
[4]: http://www.w3.org/TR/html5/forms.html#url-encoded-form-data
[5]: http://www.w3.org/TR/html5/forms.html#multipart-form-data

pbanaszkiewicz:
> It looks like the best approach is to set enctype to multipart/form-data. We might need to do other things too, but getting away from application/x-www-form-urlencoded is a good first step.

In our template (`person_bulk_add_form.html`) we have it already:

```html
<form action="" method="post" enctype="multipart/form-data">
```

Do you have any other ideas? How does this form behave on your system?

wking:
On Wed, Feb 04, 2015 at 01:04:53PM -0800, Piotr Banaszkiewicz wrote:
> <form action="" method="post" enctype="multipart/form-data">

I would expect the browser to set that as Content-Type when POSTing
the results, but that's not based on any research I've done.

> Do you have any other ideas? How does this form behave on your system?

I've been staying away from the CVS-upload stuff ;).  I'll poke around
some tonight though.

pbanaszkiewicz:
> I've been staying away from the CVS-upload stuff ;). I'll poke around some tonight though.

It's not a bug that we force people to upload UTF-8 encoded data. It's just a limitation :)

wking:
On Wed, Feb 04, 2015 at 01:36:36PM -0800, Piotr Banaszkiewicz wrote:
> > I've been staying away from the CVS-upload stuff ;). I'll poke
> > around some tonight though.
> 
> It's not a bug that we force people to upload UTF-8 encoded
> data. It's just a limitation :)

It's a bug if they correctly declare the charset on their uploaded
file as something other than UTF-8 and we try to parse it as UTF-8
anyway.

pbanaszkiewicz:
> It's a bug if they correctly declare the charset on their uploaded file as something other than UTF-8 and we try to parse it as UTF-8 anyway.

Then we can use `request.encoding` and fall back to UTF-8 if it's empty, can we?

wking:
On Wed, Feb 04, 2015 at 02:11:10PM -0800, Piotr Banaszkiewicz wrote:
> > It's a bug if they correctly declare the charset on their uploaded
> > file as something other than UTF-8 and we try to parse it as UTF-8
> > anyway.
> 
> Then we can use `request.encoding` and fall back to UTF-8 if it's
> empty, can we?

We should be falling back to DEFAULT_CHARSET, as recommended by the
docs you linked.  And I'm still not clear if request.encoding is setup
by Django for us to read, or just something we set to tell Django how
to decode whatever it reads next.  A form POST could conceivably have
strings and files encoded in several different encodings, and I'm not
sure how Django tries to expose all of that information, or how the
browser is supposed to set it on POST.

gvwilson:
Please go ahead and merge as soon as you're both comfortable with this - and thanks for digging into it.

pbanaszkiewicz:
Now the encoding is set to `request.encoding` and if it's empty the default is `DEFAULT_CHARSET`.

gvwilson:
Is this good enough that we can merge, press on, and revisit later if 
need be?

pbanaszkiewicz:
@gvwilson in my opinion yes, we can merge and move on.

gvwilson:
OK - please push the button.

wking:
On Wed, Feb 04, 2015 at 01:17:28PM -0800, W. Trevor King wrote:
> Wed, Feb 04, 2015 at 01:04:53PM -0800, Piotr Banaszkiewicz:
> > Do you have any other ideas? How does this form behave on your
> > system?
> 
> I've been staying away from the CVS-upload stuff ;).  I'll poke
> around some tonight though.

It looks like I missed the merge window here, so I pushed the way I
think we should handle this in #177.

PR #177, opened 2015-02-05
workshops/util.py: Expect a Unicode stream in upload_person_task_csv

wking:
Instead of passing in an encoding and doing some issubclass stuff,
just require a Unicode stream in upload_person_task_csv.  In general,
decoding to Unicode as soon as possible and encoding to bytes as late
as possible makes for cleaner code (who better to know the proper
encoding than the first/last code that touches the content?).  We see
that here, with a simpler upload_person_task_csv implementation and
only a two more lines in person_bulk_add to wrap the incoming file in
a StreamReader.

The accept-charset business is a hack to work around shortcomings in
modern browsers.  For example, the data posted from my Firefox 24.8
looks like:

    Content-Type: multipart/form-data; boundary=---------------------------196554893016147250581211961313
    Content-Length: 477

    -----------------------------196554893016147250581211961313

    Content-Disposition: form-data; name="csrfmiddlewaretoken"

    D8gooO2bYBXwsVQndknbnHUkJbs9ZRHp

    -----------------------------196554893016147250581211961313

    Content-Disposition: form-data; name="file"; filename="BulkPersonAddTemplate.csv"
    Content-Type: text/csv

    personal,middle,family,email,event,role
    αβ,γδ,Ε,a@b.com,2015-02-05-esu,student

    -----------------------------196554893016147250581211961313--

It's setting a per-file Content-Type (yay!), but that Content-Type
doesn't have charset information (boo!).  It looks like [the Mozilla
folks tried to fix this][1], but [backed it out in 2007][2] because it
[broke][3] some [crummy parsers][4].

I see nothing in RFC 2388 that is file-specific, but [their only
Content-Type example does show a charset][5].  The only workable,
spec-compliant solution I can come up with is to explictly set
[accept-charset][6].  Otherwise, the user agent is supposed to use
some [wonky fallbacks][7] (although if `DEFAULT_CHARSET` was UTF-8,
those would end up as UTF-8 too).

The current approach is to respect the charset set by the user-agent
(which we can access via the [.charset property][8]).  If it doesn't
set one, we expect it to respect the form's accept-charset and use
`DEFAULT_CHARSET`.

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=116346
[2]: https://bugzilla.mozilla.org/show_bug.cgi?id=116346#c79
[3]: https://bugzilla.mozilla.org/show_bug.cgi?id=116346#c74
[4]: https://bugzilla.mozilla.org/show_bug.cgi?id=392982#c45
[5]: https://tools.ietf.org/html/rfc2388#section-4.5
[6]: http://www.w3.org/TR/html5/forms.html#attr-form-accept-charset
[7]: http://www.w3.org/TR/html5/forms.html#multipart/form-data-encoding-algorithm
[8]: https://docs.djangoproject.com/en/1.7/ref/files/uploads/#django.core.files.uploadedfile.UploadedFile.charset

### DIFF
--- a/workshops/util.py
+++ b/workshops/util.py
@@ -35,18 +35,22 @@ def earth_distance(pos1, pos2):
     return arc * 6373
 
 
-def upload_person_task_csv(uploaded_file):
-    """
-    Read data from CSV and turn it into JSON-serializable list of dictionaries.
+def upload_person_task_csv(stream):
+    """Read people from CSV and return a JSON-serializable list of dicts.
+
+    The input `stream` should be a file-like object that returns
+    Unicode data.
+
     "Serializability" is required because we put this data into session.  See
     https://docs.djangoproject.com/en/1.7/topics/http/sessions/ for details.
 
     Also return a list of fields from Person.PERSON_UPLOAD_FIELDS for which
     no data was given.
+
     """
     persons_tasks = []
 
-    reader = csv.DictReader(uploaded_file)
+    reader = csv.DictReader(stream)
     empty_fields = []
     for row in reader:
         person_fields = {}
--- a/workshops/util.py
+++ b/workshops/util.py
@@ -1,9 +1,6 @@
 # coding: utf-8
 from math import pi, sin, cos, acos
-from io import TextIOWrapper, StringIO
 import csv
-
-from django.conf import settings
 
 from .models import Event, Role, Person
 
@@ -38,7 +35,7 @@ def earth_distance(pos1, pos2):
     return arc * 6373
 
 
-def upload_person_task_csv(uploaded_file, encoding=None):
+def upload_person_task_csv(uploaded_file):
     """
     Read data from CSV and turn it into JSON-serializable list of dictionaries.
     "Serializability" is required because we put this data into session.  See
@@ -46,21 +43,8 @@ def upload_person_task_csv(uploaded_file, encoding=None):
 
     Also return a list of fields from Person.PERSON_UPLOAD_FIELDS for which
     no data was given.
-
-    :param string encoding: encoding used to encode incoming file. Defaults to
-                            ``django.conf.settings.DEFAULT_CHARSET``
     """
     persons_tasks = []
-
-    if not encoding:
-        encoding = settings.DEFAULT_CHARSET
-
-    # we provide uploaded_file as StringIO in our tests (test_util.py)
-    if not issubclass(StringIO, uploaded_file.__class__):
-        # Django provides uploaded files as byte file objects.  We need to wrap
-        # `uploaded_file` and provide it as a text file with specific encoding.
-        # This way we force users to upload UTF-8 encoded files.
-        uploaded_file = TextIOWrapper(uploaded_file.file, encoding=encoding)
 
     reader = csv.DictReader(uploaded_file)
     empty_fields = []
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -1,6 +1,8 @@
+import csv
+import io
 import re
 import yaml
-import csv
+
 import requests
 
 from django.contrib import messages
@@ -201,14 +203,10 @@ def person_bulk_add(request):
     if request.method == 'POST':
         form = PersonBulkAddForm(request.POST, request.FILES)
         if form.is_valid():
+            charset = request.FILES['file'].charset or settings.DEFAULT_CHARSET
+            stream = io.TextIOWrapper(request.FILES['file'].file, charset)
             try:
-                # Provide encoding that we can use when reading CSV file.
-                # It's probably set by the browser to indicate encoding of the
-                # file being uploaded.  However during (small) testing
-                # ``request.encoding`` turned to be ``None``.
-                # In this case we default to settings.DEFAULT_CHARSET.
-                persons_tasks, empty_fields = upload_person_task_csv(
-                    request.FILES['file'], encoding=request.encoding)
+                persons_tasks, empty_fields = upload_person_task_csv(stream)
             except csv.Error as e:
                 messages.add_message(
                     request, messages.ERROR,
@@ -217,7 +215,7 @@ def person_bulk_add(request):
                 messages.add_message(
                     request, messages.ERROR,
                     "Please provide a file in {} encoding."
-                    .format(request.encoding or settings.DEFAULT_CHARSET))
+                    .format(charset))
             else:
                 if empty_fields:
                     msg_template = ("The following required fields were not"
@@ -234,8 +232,11 @@ def person_bulk_add(request):
     else:
         form = PersonBulkAddForm()
 
-    context = {'title': 'Bulk Add People',
-               'form': form}
+    context = {
+        'title': 'Bulk Add People',
+        'form': form,
+        'charset': settings.DEFAULT_CHARSET,
+    }
     return render(request, 'workshops/person_bulk_add_form.html', context)
 
 
--- a/workshops/templates/workshops/person_bulk_add_form.html
+++ b/workshops/templates/workshops/person_bulk_add_form.html
@@ -10,7 +10,8 @@
 {% block content %}
 {% load staticfiles %}
 <p><a href="{% url 'person_bulk_add_template' %}">BulkPersonAddTemplate.csv</a> is a blank template of a well-formed .CSV file.</p>
-<form action="" method="post" enctype="multipart/form-data">
+<form action="" method="post" enctype="multipart/form-data"
+      accept-charset="{{charset}}">
     {% csrf_token %}
     {{ form.as_p }}
     <input type="submit" value="Submit" />


workshops/util.py
```diff
@@ -46,22 +43,9 @@ def upload_person_task_csv(uploaded_file, encoding=None):
...
-        encoding = settings.DEFAULT_CHARSET
-
-    # we provide uploaded_file as StringIO in our tests (test_util.py)
-    if not issubclass(StringIO, uploaded_file.__class__):
-        # Django provides uploaded files as byte file objects.  We need to wrap
-        # `uploaded_file` and provide it as a text file with specific encoding.
-        # This way we force users to upload UTF-8 encoded files.
-        uploaded_file = TextIOWrapper(uploaded_file.file, encoding=encoding)
-
     reader = csv.DictReader(uploaded_file)
```
pbanaszkiewicz: That conditional was required for tests (see for example test_util.py:104). Without this conditional, the test suite doesn't pass…

wking: On Thu, Feb 05, 2015 at 01:45:14PM -0800, Piotr Banaszkiewicz wrote:
> > -    # we provide uploaded_file as StringIO in our tests (test_util.py)
> > -    if not issubclass(StringIO, uploaded_file.__class__):
> > -        # Django provides uploaded files as byte file objects.  We need to wrap
> > -        # `uploaded_file` and provide it as a text file with specific encoding.
> > -        # This way we force users to upload UTF-8 encoded files.
> > -        uploaded_file = TextIOWrapper(uploaded_file.file, encoding=encoding)
> > -
> >      reader = csv.DictReader(uploaded_file)
> 
> That conditional was required for tests (see for example
> test_util.py:104). Without this conditional, the test suite doesn't
> pass…

They pass for me:

  $ git show HEAD | head -n1
  commit 2545157878bf93c9b5bddb02e838546702a94e10
  $ git status
  On branch drop-py27-changes
  Your branch is up-to-date with 'wking/drop-py27-changes'.
  Untracked files:
    (use "git add <file>..." to include in what will be committed)
  
          amy/whoosh_index/
  
  nothing added to commit but untracked files present (use "git add" to track)
  $ python3.4 manage.py test
  …
  Ran 43 tests in 2.613s

  OK

pbanaszkiewicz: Yeah, you're right, sorry. I didn't activate my Py3.4 environment when running tests.

wking: On Thu, Feb 05, 2015 at 01:52:04PM -0800, Piotr Banaszkiewicz wrote:
> Yeah, you're right, sorry. I didn't activate my Py3.4 environment
> when running tests.

No need to apologize ;).  Thanks for pushing back on anything that
isn't clear, since I certainly make stupid mistakes that I'd rather
have caught and fixed during the review phase.  Just because this one
thing seems ok, doesn't mean I'll be right the next time I do
something that you feel is suspicious ;).

pbanaszkiewicz: > No need to apologize ;). Thanks for pushing back on anything that isn't clear, since I certainly make stupid mistakes that I'd rather have caught and fixed during the review phase. Just because this one thing seems ok, doesn't mean I'll be right the next time I do something that you feel is suspicious ;).

Yup, I'm used to making mistakes :) (therefore I'm used to apologizing for my mistakes :) )

Anyway, that's what code review is, right? We take a critical look at each others code.

wking: On Thu, Feb 05, 2015 at 02:08:31PM -0800, Piotr Banaszkiewicz wrote:
> Anyway, that's what code review is, right? We take a critical look
> at each others code.

Yeah, I was just saying that “Yeah, you're right, sorry. I didn't
activate…” could have been “Oops, I didn't activate…”.  With all the
rest there, I just thought I'd make sure you weren't actually thinking
that the push-back bothered me.  Anyhow, it sounds like we were on the
same page from the start ;).

pbanaszkiewicz:
I'm +1 on making `upload_person_task_csv` work on unicode data only. However, I think this requirement should be mentioned in the docstring.

I'm not feeling good about using `codecs.StreamReader`. I don't have anything to back up my feelings but [this PEP 400](https://www.python.org/dev/peps/pep-0400/). Although this PEP was deferred, it still makes a better point in `codecs.StreamReader` vs `io.TextIOWrapper` than I do.

Please consider switching to `io.TextIOWrapper` and / or explaining with a comment what for and why we are using `charset` in the code.

workshops/util.py
pbanaszkiewicz: > Anyhow, it sounds like we were on the same page from the start ;)

Yes! And that's what matters :-)

workshops/views.py
```diff
@@ -201,14 +203,10 @@ def person_bulk_add(request):
     if request.method == 'POST':
         form = PersonBulkAddForm(request.POST, request.FILES)
         if form.is_valid():
+            charset = request.FILES['file'].charset or settings.DEFAULT_CHARSET
+            stream = codecs.getreader(charset)(request.FILES['file'])
```
pbanaszkiewicz: **HUUUGE nit-picking**: there's a space left at the end of this line.

wking: On Thu, Feb 05, 2015 at 02:19:29PM -0800, Piotr Banaszkiewicz wrote:
> > +            charset = request.FILES['file'].charset or settings.DEFAULT_CHARSET
> > +            stream = codecs.getreader(charset)(request.FILES['file']) 
> 
> **HUUUGE nit-picking**: there's a space left at the end of this line.

Thanks :).  Fixed and force-pushed.

wking:
On Thu, Feb 05, 2015 at 02:17:10PM -0800, Piotr Banaszkiewicz wrote:
> I'm +1 on making `upload_person_task_csv` work on unicode data
> only. However, I think this requirement should be mentioned in the
> docstring.

With Python 3, I expect the assumption to be that *everything*
requires Unicode data unless it explicitly says otherwise.  I'm fine
putting a note in the docstring if you don't agree though.

pbanaszkiewicz:
> With Python 3, I expect the assumption to be that *everything* requires Unicode data unless it explicitly says otherwise. I'm fine putting a note in the docstring if you don't agree though.

Just a note that this function doesn't expect a byte-file (default type in Django for uploaded files) but rather a unicode encoded text file.

wking:
On Thu, Feb 05, 2015 at 02:17:10PM -0800, Piotr Banaszkiewicz wrote:
> I'm not feeling good about using `codecs.StreamReader`. I don't have
> anything to back up my feelings but this PEP 400…

I guess I'm just showing my age ;).  Updated to use TextIOWrapper and
force-pushed.

wking:
On Thu, Feb 05, 2015 at 02:26:06PM -0800, Piotr Banaszkiewicz wrote:
> > With Python 3, I expect the assumption to be that *everything*
> > requires Unicode data unless it explicitly says otherwise. I'm
> > fine putting a note in the docstring if you don't agree though.
> 
> Just a note that this function doesn't expect a byte-file (default
> type in Django for uploaded files) but rather a unicode encoded text
> file.

Done in the follow-up 09a6c42 (workshops/util.py: Document Unicode
stream for upload_person_task_csv, 2015-02-05), although I'm still
skeptical.  Why would upload_person_task_csv care about Django upload
conventions?  It's just about parsing person info out of a stream
(regardless of where that stream comes from).  In Python 3, I'd expect
streams to always be Unicode, and to only deal with differences from
that assumption where they break down.  In this case, that's just in
person_bulk_add which uses TextIOWrapper internally to restore the
assumption.  By the time we get out to upload_person_task_csv we don't
have anything to worry about, because we're isolated from the broken
assumption.

wking:
On Thu, Feb 05, 2015 at 02:17:10PM -0800, Piotr Banaszkiewicz wrote:
> Please consider switching to `io.TextIOWrapper` and / or explaining
> with a comment what for and why we are using `charset` in the code.

I think I've addressed everything from this comment except explaining
charset.  I'd expect the fact that uploaded text/csv is encoded in a
charset is either clear or becomes clear if someone looks up
TextIOWrapper in the Python docs or blames the line to find my
e3b8cfad.  Personally, I'd rather put these sorts of comments in the
Git history to keep the code itself more condensed for future readers
who do have that background knowledge.  The Git history is always
accessible for newcomers to fill themselves in on why a line is the
way it is (which is part of the reason I think it's worth working to
make that history understandable and easy to review).

pbanaszkiewicz:
> Done in the follow-up 09a6c42 (workshops/util.py: Document Unicode stream for upload_person_task_csv, 2015-02-05), although I'm still skeptical. Why would upload_person_task_csv care about Django upload conventions?

Previously `upload_person_task` had a parameter `uploaded_file`. This was suggesting that we were passing an uploaded file to it. There's no more confusion because you changed the name of the parameter to `stream`.

> In Python 3, I'd expect streams to always be Unicode, and to only deal with differences from that assumption where they break down.

In Py3 you can have byte streams and text streams. That's why we need to say "this stream is supposed to be text stream".

> Personally, I'd rather put these sorts of comments in the Git history to keep the code itself more condensed for future readers who do have that background knowledge.

People are lazy - I'm lazy, too! - so having comments in the code definitely helps.


Anyway, I'm really happy with your changes so I'll test once again and then merge this PR in. Thanks!

wking:
On Thu, Feb 05, 2015 at 02:56:59PM -0800, Piotr Banaszkiewicz wrote:
> > Done in the follow-up 09a6c42 (workshops/util.py: Document Unicode
> > stream for upload_person_task_csv, 2015-02-05), although I'm still
> > skeptical. Why would upload_person_task_csv care about Django
> > upload conventions?
> 
> Previously `upload_person_task` had a parameter
> `uploaded_file`. This was suggesting that we were passing an
> uploaded file to it. There's no more confusion because you changed
> the name of the parameter to `stream`.

That suggestion is still there because I didn't rename the function
from upload_person_task_csv to parse_person_task_csv (which I think
would be a good idea, but is probably not worth the churn).

> > In Python 3, I'd expect streams to always be Unicode, and to only
> > deal with differences from that assumption where they break down.
> 
> In Py3 you can have byte streams and text streams. That's why we
> need to say "this stream is supposed to be text stream".

Listing all the assumptions everywhere gets tiresome ;).  I'd rather
assume all of our methods expect Unicode unless they explicitly say
otherwise.  I'd rather have “this stream is supposed to be text
stream” be the implicit case (like “Person.personal is Unicode” is
implicit, and not listed in the help_text), and then say “this stream
is supposed to be a byte stream” where I needed to break that
assumption.

> > Personally, I'd rather put these sorts of comments in the Git
> > history to keep the code itself more condensed for future readers
> > who do have that background knowledge.
> 
> People are lazy - I'm lazy, too! - so having comments in the code
> definitely helps.

When the issue is serious enough that I want to make sure someone sees
it, I'll put a comment in the file.  In this case, “uploaded content
is encoded using a charset and Django doesn't decode it by default so
we're doing it here” seemed pretty clear from the code itself, but I
agree that this can be a subjective call.

> Anyway, I'm really happy with your changes so I'll test once again
> and then merge this PR in.

Thanks for the review :).

PR #180, opened 2015-02-08
Remove old migrations

gvwilson:
1.  Remove old migrations.
2.  Update `migrater.py` migration script.
3.  Re-generate `db.sql` migration data.


pbanaszkiewicz:
Looks great, and by the way `2to3.py` shows only one reasonable change:

```
--- migrater.py	(original)
+++ migrater.py	(refactored)
@@ -61,7 +61,7 @@
 
 def insert(cursor, table, fields):
     INSERT_TEMPLATE = 'INSERT INTO {table} ({columns}) VALUES({values});'
-    keys = fields.keys()
+    keys = list(fields.keys())
     values = [fields[k] for k in keys]
     insert = INSERT_TEMPLATE.format(table=table,
                                     columns=','.join(keys),
```

I don't think it matters, so I'm merging this.

wking:
On Sun, Feb 08, 2015 at 10:10:31AM -0800, Piotr Banaszkiewicz wrote:
> Merged #180.

Do we really want to merge this?  I think everything it brings in was
already in #147:

* 709fa13 (Removing old migrations, 2015-02-08) was already in 4cd298d
  (workshops/migrations: Remove existing migrations, 2015-01-21).
* 610953d (Updating migration script and its output, 2015-02-08) was
  already in 23d5b50 (Updating migrater tool to Py3K and newest
  database schema, 2015-02-08).

PR #181, opened 2015-02-08
Remove old migrations

gvwilson:
1. Merge changes that use our `Person` for auth.
2. Modify `migrater.py`. Some choices here are dubious (see below).
3. Update `db.sql` for importing old data.
4. Modify tests to use 'yes'/'no' instead of True/False for "may contact".

Dubious changes in `migrater.py`:

1.  Set password to empty string - this is apparently not the same as saying "has no password", but password is not allowed to be NULL.
2.  Set `last_login` to now, because `last_login` isn't allowed to be NULL (even if someone has never actually logged in - again, this is an odd choice).
3.  Have a `username` field as well as a `slug` field - we should probably merge this.


gvwilson:
@wking @pbanaszkiewicz Please review and merge.

pbanaszkiewicz:
>  Set password to empty string - this is apparently not the same as saying "has no password", but password is not allowed to be NULL.

There's no "has no password" in Django auth. There's only "has unusable password" or "has usable password".

When there's an empty string in the database or if it starts with `!`, then the password is considered unusable and user cannot log in.

We can set an empty string to be user's password, but it's representation in the database will never be an empty string. User with usable password that is hashed empty string can log in.

Going to review it now :-)

pbanaszkiewicz:
@gvwilson some changes regarding login required views are missing. What about these?

gvwilson:
Yeah - it's not how I would have designed it...

gvwilson:
What branch are they sitting on?

pbanaszkiewicz:
@gvwilson https://github.com/pbanaszkiewicz/amy/tree/wking-use-person-for-auth here's the branch used for https://github.com/wking/swc-amy/pull/3

pbanaszkiewicz:
This is looking good to me. I'm merging

wking:
On Sun, Feb 08, 2015 at 10:18:12AM -0800, Greg Wilson wrote:
>   * Merging changes that user person for auth

This commit merges #180 and #147.  As I just pointed out, I'm not
clear on what #180 was adding to #147 [1], so I'm trying to figure out
what's happening in this branch.  It looks like there was more going
on in 24adc60 (Merging changes that user person for auth, 2015-02-08)
than a bare merge would suggest:

  $ git show 24adc60
  commit 24adc601f88806b7425cff127b8afe83500e9f6c
  Merge: 610953d 23d5b50
  Author: Greg Wilson <gvwilson@third-bit.com>
  Date:   Sun Feb 8 13:16:53 2015 -0500

    Merging changes that user person for auth

  diff --cc Makefile
  index 5fa9269,5fa9269..b5db0cd
  --- a/Makefile
  +++ b/Makefile
  @@@ -45,6 -45,6 +45,10 @@@ database
          rm -f ${APP_DB}
                  ${QUERY} < ${APP_SQL}
  
  ++## schema       : display the database schema
  ++schema :
  ++      ${QUERY} .schema
  ++
    ## notes        : load old notes
      notes :
              python notes-importer.py ${APP_DB} ${SRC_EVENTS} ${SRC_SITES}

Do we need this schema target?  That's just basic SQL right?  I don't
see a reason to encode it in the Makefile.

  diff --cc migrater.py
  index b39ee41,b39ee41..c496771
  --- a/migrater.py
  +++ b/migrater.py
  @@@ -145,6 -145,6 +145,7 @@@ new_crs.execute('delete from workshops_
    old_crs.execute('select person, personal, middle, family, email from person;')
    person_lookup = {}
    i = 1
  ++now = datetime.datetime.now()
    for (person, personal, middle, family, email) in old_crs.fetchall():
        person_lookup[person] = i
        if person in facts_lookup:
  @@@ -169,7 -169,7 +170,11 @@@
                      'github' : github,
                      'twitter' : twitter,
                      'url' : url,
  --                  'may_contact' : True if active else False}
  ++                  'may_contact' : True if active else False,
  ++                  'password' : '',
  ++                  'last_login' : now,
  ++                  'is_superuser' : False,
  ++                  'username' : person}
            insert(new_crs, 'workshops_person', fields)
        except Exception as e:
            fail('person', fields, e)

This seems useful, but I'd rather have it land in a content commit,
and not be a side-effect of a merge.

  diff --cc workshops/migrations/0001_initial.py
  index ae1a810,7ee17d3..36bb7e4
  --- a/workshops/migrations/0001_initial.py
  +++ b/workshops/migrations/0001_initial.py
  @@@ -11,9 -14,33 +14,33 @@@ class Migration(migrations.Migration)
  
        operations = [
            migrations.CreateModel(
  +             name='Person',
  +             fields=[
  -                ('id', models.AutoField(auto_created=True, verbose_name='ID', serialize=False, primary_key=True)),
  -                ('password', models.CharField(max_length=128, verbose_name='password')),
  -                ('last_login', models.DateTimeField(verbose_name='last login', default=django.utils.timezone.now)),
  -                ('is_superuser', models.BooleanField(verbose_name='superuser status', default=False, help_text='Designates that this user has all permissions without explicitly assigning them.')),
  ++                ('id', models.AutoField(serialize=False, auto_created=True, primary_key=True, verbose_name='ID')),
  ++                ('password', models.CharField(verbose_name='password', max_length=128)),
  ++                ('last_login', models.DateTimeField(default=django.utils.timezone.now, verbose_name='last login')),
  ++                ('is_superuser', models.BooleanField(default=False, verbose_name='superuser status', help_text='Designates that this user has all permissions without explicitly assigning them.')),
  +                 ('personal', models.CharField(max_length=100)),
   -                ('middle', models.CharField(max_length=100, blank=True, null=True)),
  ++                ('middle', models.CharField(blank=True, max_length=100, null=True)),
  …

I think the migration in #147 was fine.  Where did this one come from,
and what is it changing?  For example, the change to Person.middle
just looks like it shuffles the keyword argument order, and that
doesn't seem to be worth the churn.

  diff --cc workshops/test/test_person.py
  index 4e6a65d,4e6a65d..127b3cb
  --- a/workshops/test/test_person.py
  +++ b/workshops/test/test_person.py
  @@@ -64,7 -64,7 +64,7 @@@ class TestPerson(TestBase)
                      ('family', person.family),
                      ('email', person.email),
                      ('gender', person.get_gender_display()),
  --                  ('may_contact', person.may_contact),
  ++                  ('may_contact', 'yes' if person.may_contact else 'no'),
                      ('airport', person.airport),
                      ('github', person.github),
                      ('twitter', person.twitter),

This also looks useful, but I don't think it should be a side-effect
of a merge.

[1]: https://github.com/swcarpentry/amy/pull/180#issuecomment-73431407

wking:
On Sun, Feb 08, 2015 at 10:18:12AM -0800, Greg Wilson wrote:
> 3. Have a username field as well as a slug field - we should
> probably merge this.

The blocker here is the resistance to requiring slugs for every Person
(see [1] and later comments in #125), combined with Django's
(reasonable, I think) requirement that the username field be unique
and non-null [2]).

[1]: https://github.com/swcarpentry/amy/pull/125#issuecomment-69049496
[2]: https://github.com/wking/swc-amy/commit/c2638d139d3bb5e02a88188a3803c97b98acf531

PR #182, opened 2015-02-08
Additions to authentication and login.

gvwilson:


pbanaszkiewicz:
@gvwilson please rebase your branch `pb-auth-additions` against latest upstream `master` so that the commit history is cleaner.

wking:
On Sun, Feb 08, 2015 at 11:07:54AM -0800, Greg Wilson wrote:
>   * Person model: bring back is_staff()

This already landed via #147's 007fb48 (workshops/models.py: Add
Person.is_staff() (returns is_superuser), 2015-02-01) [1,2].

[1]: https://github.com/wking/swc-amy/pull/3#issuecomment-72332355
[2]: https://github.com/wking/swc-amy/pull/3#issuecomment-73420624

pbanaszkiewicz:
I'm pretty happy with these changes. Merging.

pbanaszkiewicz:
For future reference: to create a new super user (who can actually log in):

```
$ ./manage.py createsuperuser
```

gvwilson:
Should I add something like this as a Makefile target so that when we 
create/import faked data, a superuser is automatically set up?

pbanaszkiewicz:
Yes, that would be very handy. However I didn't see any option to automate `./manage.py createsuperuser` (have to enter passwords manually).

I guess we have to add some Python file that manually creates a super user upon `make database`:

```python

if __name__ == "__main__":
    Person.objects.create_superuser(username="admin", personal="admin", family="admin",
                                    email="admin@example.org", password="admin")
```

wking:
On Sun, Feb 08, 2015 at 02:44:51PM -0800, Greg Wilson wrote:
> Should I add something like this as a Makefile target so that when we 
> create/import faked data, a superuser is automatically set up?

I disagree with Piotr here.  I'd expect that with the release of amy
1.0 in the very near future, we can finally migrate completely into
amy and dump migrater.py and db.sql entirely.  Folks running amy
instances can create their own superuser by hand, but automating it
doesn't seem useful to me, since it's just a once-per-deployment issue
(which means it may be a once-ever issue ;).

PR #183, opened 2015-02-08
Creating an administrator when importing fake data.

gvwilson:
1.  Add management command `create_superuser`.
2.  Adding line to `Makefile` to invoke it.

See https://docs.djangoproject.com/en/1.7/howto/custom-management-commands/ for details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,7 @@ import :
 database :
 	rm -f ${APP_DB}
 	${QUERY} < ${APP_SQL}
+	python manage.py create_superuser
 
 ## schema       : display the database schema
 schema :
--- a/workshops/management/commands/create_superuser.py
+++ b/workshops/management/commands/create_superuser.py
@@ -1,0 +1,16 @@
+from django.core.management.base import BaseCommand, CommandError
+from workshops.models import Person
+
+class Command(BaseCommand):
+    args = 'no arguments'
+    help = 'Create a superuser called "admin" with password "admin".'
+
+    def handle(self, *args, **options):
+        try:
+            Person.objects.create_superuser(username='admin',
+                                            personal='admin',
+                                            family='admin',
+                                            email='admin@example.org',
+                                            password='admin')
+        except Exception as e:
+            raise CommandError('Failed to create admin: {0}'.format(str(e)))


wking:
On Sun, Feb 08, 2015 at 03:24:59PM -0800, Greg Wilson wrote:
> 1.  Add management command `create_superuser`.
> 2.  Adding line to `Makefile` to invoke it.

Ah, I hadn't gotten far enough down in my inbox ;).  As I just
mentioned in #182 [1], I think creating superusers, especially one
with hardcoded credentials, is not a particularly useful task, and
doesn't deserve automation.  Running:

  $ python manage.py createsuperuser

for each deploy isn't a difficult task, and is more flexible than the
create_superuser proposed here.

[1]: https://github.com/swcarpentry/amy/pull/182#issuecomment-73444958

pbanaszkiewicz:
> As I just mentioned in #182 [1], I think creating superusers, especially one with hardcoded credentials, is not a particularly useful task, and doesn't deserve automation.

While it not useful for production, for development it's really important. It would drive me nuts if I had to re-create super user every time I change branches to review some PRs (and I do that often).

My understanding is that `make database` is used only for development process, is that right? If so then I really like Greg's solution (except for the command's name, it should be something like "create default superuser").

wking:
On Mon, Feb 09, 2015 at 12:37:09AM -0800, Piotr Banaszkiewicz wrote:
> > As I just mentioned in #182 [1], I think creating superusers,
> > especially one with hardcoded credentials, is not a particularly
> > useful task, and doesn't deserve automation.
>
> While it not useful for production, for development it's really
> important. It would drive me nuts if I had to re-create super user
> every time I change branches to review some PRs (and I do that
> often).

You can create a local superuser once and then migrate into or out of
branches that add migrations (which I expect will be few and far
between).  When you don't have migrations in a branch, there's no need
to touch the database at all, just run the new code on the old
database.

> My understanding is that `make database` is used only for
> development process, is that right?

That's what I'd been using it for, but in the post-1.0 world, I don't
see a point to having anonymized seed data at all.  Folks interested
in local testing can just seed with:

  $ python manage.py migrate
  $ python manage.py createsuperuser --username admin --personal admin \
	>   --family admin --email admin@example.com

and away they go.  Typing out a password once per developer is not
going to be a great hardship.

That said, the new code is fairly isolated, so if my push-back isn't
convincing, don't worry about it ;).

pbanaszkiewicz:
Okay, I have a few thoughts on the subject here.

### Production

Creating superuser automatically on production is a bad idea. We all agree on that.

### Fixtures

> >  My understanding is that `make database` is used only for development process, is that right?
>
> That's what I'd been using it for, but in the post-1.0 world, I don't see a point to having anonymized seed data at all.

How do you see that working? Can we for example provide some fixtures that are used for development process?

### Forgotten command

I simply forgot that we can use `./manage.py syncdb` to create database, apply migrations and create superuser all at once:

```
(amy3k) ~/workspace/amy   master  ./manage.py syncdb
Operations to perform:
  Apply all migrations: workshops, auth, contenttypes, admin, sessions
Running migrations:
  Applying contenttypes.0001_initial... OK
  Applying auth.0001_initial... OK
  Applying workshops.0001_initial... OK
  Applying admin.0001_initial... OK
  Applying sessions.0001_initial... OK

You have installed Django's auth system, and don't have any superusers defined.
Would you like to create one now? (yes/no): yes
Username: piotr
Personal: Piotr
Family: Banaszkiewicz
Email: piotr@banaszkiewicz.org
Password: 
Password (again): 
Superuser created successfully.
```

### Summary

If we combine fixtures and `syncdb` command I think we can have a decent starting point for development process.

Therefore I'm closing this PR as unmerged. I'll open an issue regarding updates to `Makefile` in post-1.0 era where we can discuss changes to `make database` and development fixtures.

If we change our minds we can reopen this PR after all.

wking:
On Mon, Feb 09, 2015 at 12:06:34PM -0800, Piotr Banaszkiewicz wrote:
> I simply forgot that we can use `./manage.py syncdb` to create
> database, apply migrations and create superuser all at once:

With Django 1.7, ‘syncdb’ is deprecated in favor of just ‘migrate’ [1,2].

[1]: https://docs.djangoproject.com/en/1.7/ref/django-admin/#syncdb
[2]: https://docs.djangoproject.com/en/1.7/releases/1.7/#schema-migrations

pbanaszkiewicz:
> With Django 1.7, ‘syncdb’ is deprecated in favor of just ‘migrate’ [1,2].
> 
> [1]: https://docs.djangoproject.com/en/1.7/ref/django-admin/#syncdb
> [2]: https://docs.djangoproject.com/en/1.7/releases/1.7/#schema-migrations

Thanks Trevor, I did not know that. Unfortunately `migrate` doesn't run `createsuperuser` on it's own, so yeah…

wking:
On Mon, Feb 09, 2015 at 02:54:29PM -0800, Piotr Banaszkiewicz wrote:
> > With Django 1.7, ‘syncdb’ is deprecated in favor of just ‘migrate’ [1,2].
> > 
> > [1]: https://docs.djangoproject.com/en/1.7/ref/django-admin/#syncdb
> > [2]: https://docs.djangoproject.com/en/1.7/releases/1.7/#schema-migrations
> 
> Thanks Trevor, I did not know that. Unfortunately `migrate` doesn't
> run `createsuperuser` on it's own, so yeah…

… so if you need a superuser, you'll have to call ‘createsuperuser’ by
hand.  It's not hard ;).

I'd like to reopen #70 and move towards my initial proposal of
allowing *everyone* to log in and edit their own metadata (where
appropriate), have event leads manage their own events, etc., which
would make having a local superuser less important for local testing.

PR #186, opened 2015-02-15
Experimental. Do not merge. Example Deployment and Makefile changes

codersquid:
This is experimental. Do not merge.

Based on the deployment discussion in the mailing list, I put together a bash script that can set up a new VM to have all of the required system and python dependencies for amy. I also made some preliminary changes to the Makefile with the goal of someone being able to `sudo -i -u amy`, check out the desired amy branch, then run `make serve`. The `serve` target depends on a target that will set up a virtualenv if it does not exist. The targets that run amy scripts or django commands now will activate the virtualenv before running.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ htmlerror
 db.sqlite3
 *.sublime-project
 *.sublime-workspace
+venv
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,8 @@
+# Virtual environment
+AMY ?= .
+VIRTUALENV = $(AMY)/venv
+ACTIVATE=. $(VIRTUALENV)/bin/activate
+
 # Where original data lives.
 SRC_DIR = ~/s/admin
 
@@ -28,16 +33,16 @@ commands : Makefile
 
 ## test         : run all tests.
 test :
-	python manage.py test
+	$(ACTIVATE) && python manage.py test
 
 ## migrations   : create/apply migrations
 migrations :
-	python manage.py makemigrations
-	python manage.py migrate
+	$(ACTIVATE) && python manage.py makemigrations
+	$(ACTIVATE) && python manage.py migrate
 
 ## import       : import and save legacy data
 import :
-	python migrater.py ${SRC_DB} ${APP_DB}
+	$(ACTIVATE) && python migrater.py ${SRC_DB} ${APP_DB}
 	${QUERY} .dump > ${APP_SQL}
 
 ## database     : re-make database using saved data
@@ -51,11 +56,14 @@ schema :
 
 ## notes        : load old notes
 notes :
-	python notes-importer.py ${APP_DB} ${SRC_EVENTS} ${SRC_SITES}
+	$(ACTIVATE) && python notes-importer.py ${APP_DB} ${SRC_EVENTS} ${SRC_SITES}
 
 ## serve        : run a server
-serve :
-	python manage.py runserver
+serve : $(VIRTUALENV) migrations
+	$(ACTIVATE) && python manage.py runserver
+$(VIRTUALENV) :
+	virtualenv $(VIRTUALENV)
+	$(ACTIVATE) && pip install -r $(AMY)/requirements.txt
 
 ## clean        : clean up.
 clean :
--- a/deployment/.gitignore
+++ b/deployment/.gitignore
@@ -1,0 +1,1 @@
+.vagrant
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -1,0 +1,23 @@
+Deployment stuff, Bash edition
+
+This directory contains simple scripts that help prepare a brand new machine to run amy.
+
+## Manual prep
+
+### Rackspace example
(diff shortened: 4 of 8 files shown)


gvwilson:
This is great, @codersquid - thank you very much. @wking, can you please 
give it a test drive?

wking:
On Sun, Feb 15, 2015 at 01:33:49PM -0800, Greg Wilson wrote:
> @wking, can you please give it a test drive?

I don't have Vagrant installed locally, so I can't test that part of
this PR.

I also prefer the current Makefile which is Python-packaging-system
agnostic (e.g. it's currently working with my system Python and a few
packages installed in my ~/.local/ directory).  I'm fine suggesting
folks use virtualenv [1] or the standard library's venv [2], but I'm
fine if folks want to use their system packages too.  It doesn't seem
like the kind of think we should be making decisions about in
repository for developing the workshops module.  On the other hand, if
we split this into a swc-workshop-module repository, push that to
PyPI, and just *consume* that repository here [3], then this
repository would just be about the SWC deploy, and I'd be fine with it
being more opinionated about what that should look like.

[1]: https://pypi.python.org/pypi/virtualenv
[2]: https://docs.python.org/3/library/venv.html
[3]: https://github.com/swcarpentry/amy/pull/140#issuecomment-70738784

codersquid:
@wking I don't think the Vagrant part of this necessarily works. I stopped working on these changes based on feedback on the mailing list. It sounds like deploying to ASO and using the django server suffices for deployment, and it sounds like automation is not preferred by the group.

I went ahead and checked in my progress so far -- all the bash stuff is due to the earlier discussion where people asked for bash rather than some other toolchain. The only thing worth considering from this PR are Makefile changes. If they are helpful, I can make a new PR with only Makefile changes and we can pick up review from there.

With respect to enforcing use of virtual environments -- it is only a preference. My preferences leaked in to my PR. Since runserver is used in production, and since I like local environments to mimic production; I wanted the Makefile set up runserver in the exact way it is set up on ASO in production. _There should be one-- and preferably only one --obvious way to do it_, and it should be automated. :) :)

Based on your review, this PR should be closed since it reflects my customs and not the customs of the group.

wking:
On Thu, Feb 19, 2015 at 06:24:55AM -0800, Sheila Miguez wrote:
> _There should be one-- and preferably only one --obvious way to do
> it_, and it should be automated. :) :)

So much of deployment depends on the rest of the system.  If you're
using Docker, we'd want a Dockerfile here to help with deployment.
The Zen of Python is good stuff, but the set of potential deploy
environments is too broad for us to pick one here (I think).

> Based on your review, this PR should be closed since it reflects my
> customs and not the customs of the group.

Well, I'm neither the amy maintainer nor the amy.s-c.o sysadmin, so
I'm certainly not speaking for the group ;).  That said, I'd love
highly automated deploy tooling that's publically maintained under
github.com/swcarpentry/.  I just think it should be in a separate
repository from the generic model / UI decisions [1].  For example,
your Makefile could live in a swcarpentry/amy-deploy-virtualenv
repository and be of great use to all people that want to use that
deployment strategy.  My Dockerfile could go in
swcarpentry/amy-deploy-docker and be useful for folks running Docker
clusters.  But none of those deployment strategies or their tooling
should put any restrictions on the generic stuff in this repository.

That's my point of view anyway; rational people can disagree about
these things ;).

[1]: https://github.com/swcarpentry/amy/pull/186#issuecomment-74997039

codersquid:
On Thu, Feb 19, 2015 at 12:54 PM, W. Trevor King <notifications@github.com>
wrote:

> On Thu, Feb 19, 2015 at 06:24:55AM -0800, Sheila Miguez wrote:
> > _There should be one-- and preferably only one --obvious way to do
> > it_, and it should be automated. :) :)
>
> So much of deployment depends on the rest of the system. If you're
> using Docker, we'd want a Dockerfile here to help with deployment.
> The Zen of Python is good stuff, but the set of potential deploy
> environments is too broad for us to pick one here (I think).
>

While it is true that there are many ways of deploying things, for any one
project I'd suggest that the maintainers pick one way of deploying things
and have everyone follow suite.

(I don't like Vagrant and use lxc, but I was worried that it would confuse
people so I avoided it here)

highly automated deploy tooling that's publically maintained under
> github.com/swcarpentry/. I just think it should be in a separate
> repository from the generic model / UI decisions [1]. For example,
>

I agree and maintain deployment stuff in a separate repo for things I'm
responsible for.


> your Makefile could live in a swcarpentry/amy-deploy-virtualenv
> repository and be of great use to all people that want to use that
> deployment strategy. [...]
>

I do not like using bash scripts for deployment and don't prefer Makefiles.
I only did in this case because it seemed to fit the comfort level of the
group. Given that, it doesn't make sense for me to make a new repo to hold
this. doubly so since deployment automation is not on the wish list.

I'll look through the issue backlog to see what's open. I'm juggling
volunteer time, but will try to set aside blocks of time here.

-- 
shekay@pobox.com

Issue #190, opened 2015-02-18
Make sure new events can be added *without* a URL.


gvwilson:
We need to be able to add events (so that we can start collecting data) before we have a date (which means we won't yet have a URL or slug). Check that this works, and if it doesn't, fix it.

wking:
On Wed, Feb 18, 2015 at 09:19:38AM -0800, Greg Wilson wrote:
> We need to be able to add events (so that we can start collecting
> data) before we have a date (which means we won't yet have a URL or
> slug). Check that this works, and if it doesn't, fix it.

Can this be closed now that #195 has landed proving that we can do
this?

gvwilson:
Yes - thanks.

PR #195, opened 2015-02-18
workshops/test/test_event.py: Add TestEventViews.test_add_minimal_event

wking:
Only setting the non-blank fields (published, site, and tags).  This
ensures that the form respects our blanks (e.g. they don't need a URL,

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -295,6 +295,11 @@ class Event(models.Model):
 
         raise ObjectDoesNotExist(ident)
 
+    def save(self, *args, **kwargs):
+        self.url = self.url or None
+        super(Event, self).save(*args, **kwargs)
+
+
 #------------------------------------------------------------
 
 class Role(models.Model):
--- a/workshops/test/test_event.py
+++ b/workshops/test/test_event.py
@@ -219,6 +219,26 @@ class TestEventViews(TestBase):
             self._check_status_code_and_parse(response, 200)
             assert False, 'expected 302 redirect after post'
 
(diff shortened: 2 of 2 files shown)


gvwilson:
Looks good to me.

wking:
On Wed, Feb 18, 2015 at 11:13:52AM -0800, Greg Wilson wrote:
> Looks good to me.

Actually, it looks like the form currently struggles with two
successive empty entries, because both will have NULL URLs, and so the
form parser things the second isn't unique (SQLite doesn't mind
though).  We probably want another test checking that case before we
merge this PR.

wking:
On Wed, Feb 18, 2015 at 11:26:10AM -0800, W. Trevor King wrote:
> Actually, it looks like the form currently struggles with two
> successive empty entries, …

Ok, there's another test and a better description in 295dd1e8
(workshops/test/test_event.py: Test adding two minimal Events,
2015-02-18), and a fix following a Django core dev in 2171bc48
(workshops/models.py: Work around '' TextInput default for Event.url,
2015-02-18).  I think this is ready for merging now.  We can go
through, test-for, and fix other instances bumping into Django's 4136
[1] in follow-up PRs.

[1]: https://code.djangoproject.com/ticket/4136

wking:
On Wed, Feb 18, 2015 at 10:31:58PM -0800, W. Trevor King wrote:
> I think this is ready for merging now.

Bump :).  This fixes #190, which is tagged as “essential”.

PR #196, opened 2015-02-18
workshops/views.py: Drop 'groups' from form fields

### LINKED ISSUE #191: Remove the groups list.
The "new person" form shouldn't mention groups unless/until we start using Django's permission groups.

wking:
Until we actually have permission groups to choose from it just
clutters the form.

Fixes #191.


pbanaszkiewicz:
Looks good to me.

wking:
Bump.  This fixes #191, which is marked as “essential”.

PR #200, opened 2015-02-21
Keep secret_key secret.

drio:


wking:
On Sat, Feb 21, 2015 at 01:24:44PM -0800, David Rio Deiros wrote:
>   * Keep secret_key secret.

I'd rather have:

  import uuid

  SECRET_KEY = os.environ.get('SECRET_KEY', uuid.uuid4().hex)

so folks who aren't running a public deploy can skip setting the
environment variable.

pbanaszkiewicz:
+1 for defaulting to some value, but I'd rather have that value not rotate every time I reload settings [because it breaks many things](https://docs.djangoproject.com/en/1.7/ref/settings/#std:setting-SECRET_KEY).

wking:
On Mon, Feb 23, 2015 at 02:57:35AM -0800, Piotr Banaszkiewicz wrote:
> … but I'd rather have that value not rotate every time I reload
> settings [because it breaks many
> things](https://docs.djangoproject.com/en/1.7/ref/settings/#std:setting-SECRET_KEY).

The broken things don't look very important if you're just firing up a
local instance for testing.  Was there somthing in that list that
seemed more troublesome to you?

pbanaszkiewicz:
Hey @wking,

> The broken things don't look very important if you're just firing up a local instance for testing. Was there somthing in that list that seemed more troublesome to you?

yes I think generating `SECRET_KEY` every time settings are reloaded (and I think it happens every time dev server reloads - and therefore every time I make changes to the code) breaks: messages, sessions and crypto signing - all of these are used for bulk-upload.

I propose another approach:

```python
# in settings.py

SECRET_KEY = '3l$35+@a%g!(^y^98oi%ei+%+yvtl3y0k^_7-fmx2oj09-ac5@'
if not DEBUG:
    SECRET_KEY = os.environ.get('SECRET_KEY', None)
```

This way:

1) every developer uses the same `SECRET_KEY` and that key never changes
2) for production-ready deployment the application will refuse to start (https://docs.djangoproject.com/en/1.7/ref/settings/#std:setting-SECRET_KEY)

What do you think?

wking:
On Mon, Feb 23, 2015 at 10:37:58AM -0800, Piotr Banaszkiewicz wrote:
> SECRET_KEY = '3l$35+@a%g!(^y^98oi%ei+%+yvtl3y0k^_7-fmx2oj09-ac5@'
> if not DEBUG:
>     SECRET_KEY = os.environ.get('SECRET_KEY', None)

Ah, that looks good to me :).

pbanaszkiewicz:
Great! @drio can you update this PR, please?

drio:
Done.

wking:
On Mon, Feb 23, 2015 at 01:54:52PM -0800, David Rio Deiros wrote:
> Done.

It looks like we can drop the sys import now that we're not printing
to stderr.

drio:
Thanks @wking.

amy/settings.py
```diff
@@ -10,14 +10,16 @@
...
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/1.7/howto/deployment/checklist/
 
-# SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = '3l$35+@a%g!(^y^98oi%ei+%+yvtl3y0k^_7-fmx2oj09-ac5@'
+if not DEBUG:
+    SECRET_KEY = os.environ.get('SECRET_KEY', None)
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
```
pbanaszkiewicz: `DEBUG = True` should appear before `if not DEBUG: …`.

pbanaszkiewicz:
When we fix `NameError: name 'DEBUG' is not defined` (https://github.com/swcarpentry/amy/pull/200#discussion_r25237149) we're ready to merge.

drio:
How do you feel about this other alternative? It does not require changing the file
when deploying.

wking:
On Tue, Feb 24, 2015 at 06:27:41AM -0800, David Rio Deiros wrote:
> How do you feel about this other alternative? It does not require
> changing the file when deploying.

I don't want to hardcode deployment information in this repository.
How about:

  import json


  DEBUG = json.loads(os.environ.get('DEBUG', 'true'))

then folks who are deploying amy (wherever they happen to do that) can
use:

  DEBUG=true SECRET_KEY="..." ./manage.py runserver ...

Oh, and if a SECRET_KEY environment variable is set (regardless of
DEBUG), we should respect it.  I think DEBUG should just toggle the
default:

  if DEBUG:
      SECRET_KEY = '3l$35+@a%g!(^y^98oi%ei+%+yvtl3y0k^_7-fmx2oj09-ac5@'
  else:
      SECRET_KEY = None
  SECRET_KEY = os.environ.get('SECRET_KEY', SECRET_KEY)

drio:
I implemented @wking's latest suggestion.

We have been using the same key for production and development. I want to re-deploy (restart
the server really) with the new production key but I am afraid that will break some things. What do you
suggest?

wking:
On Wed, Feb 25, 2015 at 10:48:08AM -0800, David Rio Deiros wrote:
> We have been using the same key for production and development. I
> want to re-deploy (restart the server really) with the new
> production key but I am afraid that will break some things. What do
> you suggest?

I don't think running a production server with a compromised provate
key is salvageable.  I'd restart with DEBUG=false and a new
SECRET_KEY, and just accept the resulting breakage (which should be
small if you time the restart to avoid any running sessions).

wking:
On Wed, Feb 25, 2015 at 10:48:08AM -0800, David Rio Deiros wrote:
> I implemented @wking's latest suggestion.

Grr, and it looks like I suggested ‘DEBUG=true’ for production, it
should definitely be ‘DEBUG=false’.

pbanaszkiewicz:
It looks okay for me. @wking take one final look and merge if it's OK.

drio:
Ok. I think the PR looks good now. As soon as you merge this I will pull from
the vps and rerun using the production secret.

@gvwilson I have updated `$HOME/secret_key.sh` with the new production secret key. I will also update our passwords file in the sysadmin repo.

wking:
On Wed, Feb 25, 2015 at 11:37:14AM -0800, Piotr Banaszkiewicz wrote:
> It looks okay for me. @wking take one final look and merge if it's OK.

Sorry for the delay.  I responded via email yesterday, but GitHub
occasionally delays comments submitted by email :/.  Here's what I
said:

  Looks good to me :).  Maybe --squash the merge?  I doubt the details
  of our asymptotic approach to the current solution will be
  particularly useful for future developers.

And I don't think I should be merging this.  I'm just a contributor
here, not the maintainer, so I'm just giving advice, not making the
call about what lands when.  I'm not sure who the maintainers are [1],
but I'd guess it's currently @gvwilson.  I'm happy to do the squash
merge if he deputizes me to perform it.

[1]: http://lists.software-carpentry.org/pipermail/amy_lists.software-carpentry.org/2015-February/000024.html

gvwilson:
Merged.

PR #201, opened 2015-02-22
Creating command-line CSV upload tool

gvwilson:
This PR creates a new management command to upload user/task data from a CSV file from the command line (rather than through the web interface).

1.  `workshops/management/commands/upload.py` uses upload utilities to read and upload a CSV file.
2.  Some upload code moved from `workshops/views.py` to `workshops/util.py`.
3.  More test CSV files added to `workshops/test`.
4.  Small shell script added to `workshops/test` to exercise command-line upload.

Command-line upload from CSV currently fails because usernames aren't being generated.  We should not require every `Person` to have a username (since most will never log in).

Background: command-line upload may avoid some of the character set encoding problems we're encountering with web-based upload.  Being able to bulk upload from the command line is also generally useful.


wking:
On Sat, Feb 21, 2015 at 05:15:15PM -0800, Greg Wilson wrote:
>     A workshops/test/test-command-line-upload.sh (8)

This directory is for unittest tests in Python modules.  I don't think
'./manage.py test' is going to test your shell script.  We should be
able to call the upload command from Python, so I'd take that approach
in a regular TestCase subclass.

>     M workshops/util.py (30)

You can use [1]:

  Person.objects.create(**row['person'])

instead of:

  Person(**row['person'])
	p.save()

but I think this is a good place for get_or_create [2] to avoid
creating multiple people with the same email address.

>     M workshops/views.py (33)

PEP 8 prefers parenthesis to wrapping lines with backslashes [3].

Other than those minor issues, this looks reasonable to me.  I think
the create_uploaded_persons_tasks refactor can land now.  However, I'd
like to be clear on the need before we add the 'upload' command.  Will
a reasonable fraction of admins have SSH access to
amy.software-carpentry.org?  It might make more sense to write a
command-line client that POSTs to the existing HTTP endpoint, since
anyone with an admin account could use that.

[1]: https://docs.djangoproject.com/en/1.7/ref/models/querysets/#create
[2]: https://docs.djangoproject.com/en/1.7/ref/models/querysets/#get-or-create
[3]: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

     The preferred way of wrapping long lines is by using Python's
     implied line continuation inside parentheses, brackets and
     braces. Long lines can be broken over multiple lines by wrapping
     expressions in parentheses. These should be used in preference to
     using a backslash for line continuation.

pbanaszkiewicz:
Hey Greg and Trevor, sorry for the delay.

I'm not testing Greg's CLI command, but I'm using the web interface for testing. Here are some of my thoughts:

**Ideas**

* instead of heavy logic for setting user's default `username`, why not set `username` to user's email? Email has to be unique and provided anyway (see issue below)

**Issues**

* we cannot have two persons with empty email because of the unique constraint (we have to make email non-nullable to fix this; I don't imagine having non-unique emails in the DB)

* verification in `util.verify_upload_person_task` still checks if the person exists

* as @wking pointed out, we should be using `get_or_create` for creating a person in `util.create_uploaded_persons_tasks`

------

@gvwilson I think the issue with encoding is complex and consists of these factors:

* Excel (or, apparently, OpenOffice Calc) does not export to commonly used encoding formats like UTF-8

* we use encoding based on what browser tells us in the request, and we fall back to UTF-8 if browser doesn't say anything

* browsers don't say anything about encoding of the uploaded file (side note: why would they? Guessing an encoding is a hell of the problem)

I don't agree that we can fix the upload encoding issue by making a CLI command. We can fix the upload encoding issue by agreeing to the One True Encoding™ (for example UTF-8) :)

Last: I think the real issue we have is "how do we make Person model so that it works as an auth replacement model and also complies with our requirements".
We have:

* `username` field for user authentication - it must be unique and not-null
* `password` field for user authentication - it must be unique and not-null
* `email` field for ???? - it must be unique and not-null (see: issue above)
* `slug` field for ???? (my guess is that we can get rid of this one because we can use `username` instead)

Unless we actually need both `slug` and `username`, I'd drop the `slug` field.

wking:
On Tue, Feb 24, 2015 at 02:45:09PM -0800, Piotr Banaszkiewicz wrote:
> * we cannot have two persons with empty email because of the unique
>   constraint (we have to make email non-nullable to fix this; I
>   don't imagine having non-unique emails in the DB)

We can have unique, nullable strings.  It just requires a bit more
work.  See 2171bc48 (workshops/models.py: Work around '' TextInput
default for Event.url, 2015-02-18, #195).

> * `username` field for user authentication - it must be unique and not-null
> * `password` field for user authentication - it must be unique and not-null
> * `email` field for ???? - it must be unique and not-null (see:
>    issue above)
> * `slug` field for ???? (my guess is that we can get rid of this one
>    because we can use `username` instead)
> 
> Unless we actually need both `slug` and `username`, I'd drop the
> `slug` field.

The problem was that Greg didn't want to require either 'slug' or
'email' (see #125, especially [1,2]).  I agree with Greg that 'email'
should be optional.  I agree with you that I'd rather drop 'slug' in
favor of 'username', since once we have *a* required, unique
identifier, I don't see a point to having an optional, unique
identifier.

[1]: https://github.com/swcarpentry/amy/pull/125#issuecomment-69049496
[2]: https://github.com/swcarpentry/amy/pull/125#issuecomment-70430457

gvwilson:
> *Ideas*
>
>   * instead of heavy logic for setting user's default |username|, why
>     not set |username| to user's email? Email has to be unique and
>     provided anyway (see issue below)
>

1. I'd like to get rid of 'slug' and just have 'username'.

2. We use username as a stem for filenames (e.g., username.json is your 
badge file in the main website), so email addresses would be... odd.

3. But we can always default to email address for username on upload, 
then have the admin set a (manually generated) username when it becomes 
necessary.

> *Issues*
>
>  *
>
>     we cannot have two persons with empty email because of the unique
>     constraint (we have to make email non-nullable to fix this; I
>     don't imagine having non-unique emails in the DB)
>

We have to allow empty emails - sometimes we have emails that become 
invalid (so we want to clear them from the DB), and in some cases, we're 
not allowed to have people's emails (data privacy rules).

>  *
>
>     verification in |util.verify_upload_person_task| still checks if
>     the person exists
>
>  *
>
>     as @wking <https://github.com/wking> pointed out, we should be
>     using |get_or_create| for creating a person in
>     |util.create_uploaded_persons_tasks|
>

How will that interact with batching?  We don't want anyone created 
until all the new records are creatable, do we?

> I don't agree that we can fix the upload encoding issue by making a 
> CLI command. We can fix the upload encoding issue by agreeing to the 
> One True Encoding™ (for example UTF-8) :) 

Sure - now, how do we ensure that admins have done that, using the tools 
they have, with the skills they have?

> Last: I think the real issue we have is "how do we make Person model 
> so that it works as an auth replacement model and also complies with 
> our requirements".
> We have:
>
>   * |username| field for user authentication - it must be unique and
>     not-null
>   * |password| field for user authentication - it must be unique and
>     not-null
>   * |email| field for ???? - it must be unique and not-null (see:
>     issue above)
>   * |slug| field for ???? (my guess is that we can get rid of this one
>     because we can use |username| instead)
>
> Unless we actually need both |slug| and |username|, I'd drop the 
> |slug| field.
>
>
Agreed - my latest update to my branch does this.

Thanks,
Greg

-- 

gvwilson:
On 2015-02-24 6:17 PM, W. Trevor King wrote:
> The problem was that Greg didn't want to require either 'slug' or
> 'email' (see #125, especially [1,2]). I agree with Greg that 'email'
> should be optional. I agree with you that I'd rather drop 'slug' in
> favor of 'username', since once we have *a* required, unique
> identifier, I don't see a point to having an optional, unique
> identifier.
Agreed - #201 has eliminated 'slug' in favor of 'username'.

wking:
On Tue, Feb 24, 2015 at 06:34:05PM -0800, Greg Wilson wrote:
> >     as @wking pointed out, we should be using |get_or_create| for
> >     creating a person in |util.create_uploaded_persons_tasks|
> 
> How will that interact with batching?  We don't want anyone created
> until all the new records are creatable, do we?

1. create_uploaded_persons_tasks is already using:

     p = Person(**fields)
     p.save()

   and get_or_create will only be *less* agressive than that.

2. create_uploaded_persons_tasks wraps its database writes in:

     with transaction.atomic():`

   so any changes will… be atomic ;).

> > Unless we actually need both |slug| and |username|, I'd drop the
> > |slug| field.
>
> Agreed - my latest update to my branch does this.

Did you forget to push?  The most recent commit I see in this branch
is 2f431580, which still has the field [1].

[1]: https://github.com/gvwilson/amy/blob/2f431580764d86be8f40b389c9116ed8012b4130/workshops/models.py#L113

pbanaszkiewicz:
Hey everyone,

I'm glad we're removing `slug`.

> > I don't agree that we can fix the upload encoding issue by making a CLI command. We can fix the upload encoding issue by agreeing to the One True Encoding™ (for example UTF-8) :)
>
> Sure - now, how do we ensure that admins have done that, using the tools they have, with the skills they have?

Unfortunately I don't have a Mac to test what's working correctly and what's not (I'm assuming both Carliss and Giacomo are using Macs).

> > * we cannot have two persons with empty email because of the unique constraint (we have to make email non-nullable to fix this; I don't imagine having non-unique emails in the DB)
>
> We can have unique, nullable strings. It just requires a bit more work. See 2171bc4 (workshops/models.py: Work around '' TextInput default for Event.url, 2015-02-18, #195).

I don't see how this would be a solution for the problem with database.

It's SQLite that's saying "an nullable field with UNIQUE constraint can only have (at most) one record with NULL value". Can we fix this somehow? We'd have to drop UNIQUE constraint and check for uniqueness on our own.

pbanaszkiewicz:
Forgot to add that this PR breaks uploading CSV via website a little, so @gvwilson: do you want me to push some fixes against your PR branch?

gvwilson:
Yes please and thank you.

wking:
On Wed, Feb 25, 2015 at 12:25:10AM -0800, Piotr Banaszkiewicz wrote:
> > > * we cannot have two persons with empty email because of the
> > >   unique constraint (we have to make email non-nullable to fix
> > >   this; I don't imagine having non-unique emails in the DB)
> >
> > We can have unique, nullable strings. It just requires a bit more
> > work. See 2171bc4 (workshops/models.py: Work around '' TextInput
> > default for Event.url, 2015-02-18, #195).
> 
> I don't see how this would be a solution for the problem with database.
> 
> It's SQLite that's saying "an nullable field with UNIQUE constraint
> can only have (at most) one record with NULL value". Can we fix this
> somehow? We'd have to drop UNIQUE constraint and check for
> uniqueness on our own.

SQLite doesn't have a problem with it:

	$ sqlite3 /tmp/test.db
  SQLite version 3.8.7.4 2014-12-09 01:34:36
  Enter ".help" for usage hints.
  sqlite> CREATE TABLE test (
     ...>   "id" integer NOT NULL PRIMARY KEY AUTOINCREMENT,
     ...>   "email" varchar(100) NULL UNIQUE);
  sqlite> INSERT INTO test ( email ) VALUES (NULL);
  sqlite> INSERT INTO test ( email ) VALUES (NULL);
  sqlite> SELECT * FROM test;
  1|
  2|
  sqlite> INSERT INTO test ( email ) VALUES ("a@b.com");
  sqlite> INSERT INTO test ( email ) VALUES ("a@b.com");
  Error: UNIQUE constraint failed: test.email

And from the docs [1]:

  For the purposes of determining the uniqueness of primary key
  values, NULL values are considered distinct from all other values,
  including other NULLs.

[1]: https://www.sqlite.org/lang_createtable.html#constraints

pbanaszkiewicz:
@wking Django fails to add a `Person` object with `email=None` if there's one in the database already:

```
>>> Person.objects.create_user(username="pbanaszkiewicz", personal="Piotr", family="Banaszkiewicz", email=None, password=None)
<Person: pbanaszkiewicz:Piotr Banaszkiewicz <>>
>>> Person.objects.create_user(username="pbanaszkiewicz2", personal="Piotr", family="Banaszkiewicz", email=None, password=None)
  File "<console>", line 1, in <module>
  File "/home/piotr/workspace/SwC/amy/workshops/models.py", line 71, in create_user
    user.save(using=self._db)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/models/base.py", line 589, in save
    force_update=force_update, update_fields=update_fields)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/models/base.py", line 617, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/models/base.py", line 698, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/models/base.py", line 731, in _do_insert
    using=using, raw=raw)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/models/manager.py", line 92, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/models/query.py", line 921, in _insert
    return query.get_compiler(using=using).execute_sql(return_id)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/models/sql/compiler.py", line 920, in execute_sql
    cursor.execute(sql, params)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/backends/utils.py", line 81, in execute
    return super(CursorDebugWrapper, self).execute(sql, params)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/utils.py", line 94, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/utils/six.py", line 658, in reraise
    raise value.with_traceback(tb)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
  File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/db/backends/sqlite3/base.py", line 485, in execute
    return Database.Cursor.execute(self, query, params)
django.db.utils.IntegrityError: UNIQUE constraint failed: workshops_person.email
```

pbanaszkiewicz:
@wking I found out what's causing the problem. It's `normalize_email` that's returning string instead of `None`: https://github.com/gvwilson/amy/blob/cmdline-import/workshops/models.py#L68

wking:
On Wed, Feb 25, 2015 at 09:25:28AM -0800, Piotr Banaszkiewicz wrote:
> @wking I found out what's causing the problem. It's
> `normalize_email` that's returning string instead of `None`:
> https://github.com/gvwilson/amy/blob/cmdline-import/workshops/models.py#L68

Right.  For more details on the issue and the workaround I used
before, see the references I linked before ;).  2171bc4
(workshops/models.py: Work around '' TextInput > default for
Event.url, 2015-02-18, #195).

pbanaszkiewicz:
Hey @gvwilson @wking @neon-ninja,

please give this PR a test ride. Maybe we could deploy it for Giacomo to provide some feedback, too? Anyway, if you see something working unexpectedly or even not working at all, don't hesitate to blame me :)

Cheers,
Piotr

wking:
The following commits look good to me:
* 9c57b2f8da (models.py: make empty emails NULL, 2015-02-25)
* 9dedd1a4a4 (Fix: make generated usernames lowercase by default, 2015-02-25)

I'm fairly comfortable with 723ed9fdbe (Add tests that cover new
verification mechanism, 2015-02-25), but have some minor qualms.

On the style front, I'd prefer 723ed9fdbe use something like:

  bad_data = [
      {'email': 'harry@hogwarts.edu',
       'personal': 'Harry',
       …
      },
  ]

instead of your:

  bad_data = [dict(), ]
  bad_data[0]['email'] = 'harry@hogwarts.edu'
  bad_data[0]['personal'] = 'Harry'
  …

because I think it reads more clearly.

I also think it would be nice to test that we aren't creating
duplicate users in
VerifyUploadPersonTask.test_verify_email_caseinsensitive_matches.
Currently we only test that the almost-duplicate emails don't raise
errors.

I didn't review 710878bae (Fix: bulk upload allow existing Persons,
2015-02-25) in depth, because it's longish and I'm not familiar with
the code it's altering.  The get_or_create handling looked reasonable.
If the tests pass and the maintainer is satisfied with the code style,
I have no problem with this commit landing without further review.

pbanaszkiewicz:
I considered @wking comments and made another pull request against this pull request: https://github.com/gvwilson/amy/pull/2

Mostly I need your feedback on https://github.com/pbanaszkiewicz/amy/commit/8edea27fab3e934bf2901d644d4cfbfe8104fa23

Cheers.

pbanaszkiewicz:
Current status: we're waiting for #206 and #210 to be merged and closed. The plan is then to merge #200 (not sure about this one), #203, #208 and #209, #206, #210 into this PR.

Finally we'll merge #201 into the master and… probably test on production to get more feedback from Giacomo.

I'll spawn an issue for https://github.com/gvwilson/amy/pull/2#issuecomment-76458963 so I don't forget about it. It probably will be my next TODO after #201 gets merged.

wking:
On Mon, Mar 02, 2015 at 01:37:44AM -0800, Piotr Banaszkiewicz wrote:
> The plan is then to merge #200 (not sure about this one), #203, #208
> and #209, #206, …

These have been merged into master already, and seem mostly
independent of this PR.  Are you proposing additional merges for these
features?

> … #210 into this PR.

#210 looks good to me, and also seems largely independent of #201.

> Finally we'll merge #201 into the master and… probably test on
> production to get more feedback from Giacomo.

I don't mind if #210 lands before or after #201.

PR #203, opened 2015-02-28
Regenerating database to sync with schema changes

gvwilson:


pbanaszkiewicz:
Uh, it took me a while to dig through that change. Apart from thousands of new INSERTs, we have one schema change which is alters… order of the columns:

```
-CREATE TABLE "workshops_site" ("id" integer NOT NULL PRIMARY KEY AUTOINCREMENT, "domain" varchar(100) NOT NULL UNIQUE, "fullname" varchar(100) NOT NULL UNIQUE, "country" varchar(100) NULL, "notes" text NOT NULL);
+CREATE TABLE "workshops_site" ("id" integer NOT NULL PRIMARY KEY AUTOINCREMENT, "notes" text NOT NULL, "domain" varchar(100) NOT NULL UNIQUE, "fullname" varchar(100) NOT NULL UNIQUE, "country" varchar(100) NULL);
```

gvwilson:
Sorry about that - any schema change in models.py (even one that just 
changes a default) can trigger column reordering.

PR #204, opened 2015-02-28
Simplifying secret key logic.

gvwilson:
I was finding it difficult to follow the logic that set the secret key.

### DIFF
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -8,28 +8,26 @@ For the full list of settings and their values, see
 https://docs.djangoproject.com/en/1.7/ref/settings/
 """
 
-# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+import sys
 import os
-import json
 
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
 
 
-# Quick-start development settings - unsuitable for production
-# See https://docs.djangoproject.com/en/1.7/howto/deployment/checklist/
-
-
-# SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = json.loads(os.environ.get('DEBUG', 'true'))
-# For deployment in production: 
-# DEBUG=false SECRET_KEY="..." ./manage.py runserver ...
-
+# Manage the secret key.
+# 1. if AMY_DEBUG is set, use AMY_SECRET_KEY if available, or fall back to saved "secret" key
+# 2. if AMY_DEBUG is not set, must  have AMY_SECRET_KEY
+DEBUG = os.environ.get('AMY_DEBUG', None)
+AMY_SECRET_KEY = os.environ.get('AMY_SECRET_KEY', None)
 if DEBUG:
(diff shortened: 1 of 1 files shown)


wking:
I'm -1 on this PR.

On Sat, Feb 28, 2015 at 03:16:50AM -0800, Greg Wilson wrote:
> I was finding it difficult to follow the logic that set the secret
> key.

The logic from #200 was:

* Load DEBUG from a ‘DEBUG’ environment variable, defaulting to True.
  Can set True or False using DEBUG=true or DEBUG=false.
* Load SECRET_KEY from a ‘SECRET_KEY’ environment variable, defaulting
  to a hard-coded value if DEBUG (easier testing) and None if not
  DEBUG (safer production).

Your logic here looks like:

* Load DEBUG from an ‘AMY_DEBUG’ environment variable, defaulting to
  None.  Can set truthy with AMY_DEBUG=anything and falsy with
  AMY_DEBUG=''
* Load SECRET_KEY from an ‘AMY_SECRET_KEY‘ environment variable,
  defaulting to a hard-coded value if DEBUG (easier testing) and None
  if not DEBUG (safer production).

Modulo the AMY_* variable names, that sounds like pretty much the same
thing to me (although I prefer #200's =true and =false to this PR's
=anything and ='').

There's no need for the empty SECRET_KEY assertion you're adding here,
since Django checks that already.  For example, the current master
has:

  $ DEBUG=false python3.4 ./manage.py runserver
  ...
  django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.

pbanaszkiewicz:
I agree with @wking on this one.

@gvwilson: If you feel we need to have a better explanation of what's going on in the code, I can add more comments that clear things up.

For development purposes, it's enough to run `./manage.py runserver`. For production there has to be set:

```shell
DEBUG=false
SECRET_KEY=asdf...
```

By the way, I actually like "AMY_*" prefix to these env var names, but that's something minor we can think about another day.

gvwilson:
Unfortunately, DEBUG is already being used as an environment variable by 
another application I'm working on, which is why I strongly prefer all 
app-specific variables to have a prefix (just like PYTHONPATH).  Can you 
please PR a change to AMY_DEBUG and AMY_SECRET_KEY as the env vars?  The 
variable inside settings apparently *must* be DEBUG or Django complains.

wking:
On Sat, Feb 28, 2015 at 03:50:47PM -0800, Greg Wilson wrote:
> Unfortunately, DEBUG is already being used as an environment
> variable by another application I'm working on, which is why I
> strongly prefer all app-specific variables to have a prefix (just
> like PYTHONPATH).  Can you please PR a change to AMY_DEBUG and
> AMY_SECRET_KEY as the env vars?  The variable inside settings
> apparently *must* be DEBUG or Django complains.

I'd recommend setting the environment variables right before launching
the application, vs. globally in your ~/.bashrc or whatever.  That
said, I'd expect you'd be either running a production server (in which
case you'd want to disable DEBUG everywhere) or a testing server (in
which case you'd want to enable DEBUG everywhere).  And I'd rather
have applications share a single variable with consisten semantics
(e.g. HOME, TERM, ...) than each have their own variable with slightly
different syntax.  Does your other application require different
values for DEBUG than 'true' or 'false'?  If there's a pre-existing
standard for DEBUG, I'd be happy to have amy use it.

PR #205, opened 2015-02-28
Replacing person slug with username

gvwilson:
This isolates the replacement of Person.slug with Person.username.

### DIFF
--- a/workshops/migrations/0003_auto_20150223_1532.py
+++ b/workshops/migrations/0003_auto_20150223_1532.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('workshops', '0002_auto_20150219_1305'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='person',
+            name='slug',
+        ),
+        migrations.AlterField(
+            model_name='person',
+            name='username',
+            field=models.CharField(max_length=40, unique=True, blank=True),
+            preserve_default=True,
+        ),
+    ]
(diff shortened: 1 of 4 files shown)


wking:
I like the thrust, but have some issues with the current implementation.

On Sat, Feb 28, 2015 at 03:19:03AM -0800, Greg Wilson wrote:
>   * Unifying Person.slug and Person.username, since they're redundant.

Why make username blankable?  That will let you create a single person
with an empty-string username, but tying to create a second person
with a blank username will give:

  Person with this Username already exists.

If you want a blankable field, you could probably work around this by
autogenerating a random username in Person.save if the username field
was falsy.

>   * Merge branch 'master' into replacing-person-slug-with-username

Why this merge?  It looks like it's just pulling in #200, which seems
orthogonal to the slug/username unification.

You'll also likely need to update migrater.py to avoid setting
Person.slug.  It looks like workshops/static/BulkPersonAddTemplate.csv
also has a slug field, which I expect should become 'username' or be
dropped entirely (in favor of autogenerated random usernames, or
email-based usernames, or something) for bulk-uploaded people.

pbanaszkiewicz:
@wking `workshops/static/BulkPersonAddTemplate.csv` is actually deprecated - we're generating CSV template on the fly in `workshops.views.person_bulk_add_template`.

Usernames are also generated on the fly (with `util.create_username`).

@gvwilson why is username blankable?

wking:
On Sat, Feb 28, 2015 at 02:07:52PM -0800, Piotr Banaszkiewicz wrote:
> Usernames are also generated on the fly (with
> `util.create_username`).

Where?

  $ git grep create_username origin/pr/205
	(no results)

pbanaszkiewicz:
@wking in #201, in bulk upload (either from CLI or via website).

gvwilson:
> @gvwilson <https://github.com/gvwilson> why is username blankable?
That's probably a leftover from when we were using `slug` as the ident - 
since we're generating usernames as we load, let's make it 
non-blankable.  Can you please?
Thx,
Greg (who's in the middle of packing up the house to get it on the market)

wking:
On Sat, Feb 28, 2015 at 11:14:57PM -0800, Piotr Banaszkiewicz wrote:
> @wking in #201, in bulk upload (either from CLI or via website).

Ah, it was not clear to me that this PR was intended to block on #201.
In any case, not allowing blank usernames works too.

pbanaszkiewicz:
Closing because it's succeeded by #210.

PR #206, opened 2015-03-01
Add "AMY_" prefix to env vars DEBUG, SECRET_KEY

pbanaszkiewicz:
This is dictated by https://github.com/swcarpentry/amy/pull/204#issuecomment-76557668
and it's generally a good idea to use such variables with
per-application prefix.

If Greg's using specific "standard" for `DEBUG` values we can update our `settings.py` here to have: `AMY_DEBUG=$DEBUG AMY_SECRET_KEY=hjll... ./manage.py runserver`.

### DIFF
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -19,17 +19,16 @@ BASE_DIR = os.path.dirname(os.path.dirname(__file__))
 # See https://docs.djangoproject.com/en/1.7/howto/deployment/checklist/
 
 
-# SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = json.loads(os.environ.get('DEBUG', 'true'))
-# For deployment in production: 
-# DEBUG=false SECRET_KEY="..." ./manage.py runserver ...
+# SECURITY WARNING: don't run with DEBUG turned on in production!
+DEBUG = json.loads(os.environ.get('AMY_DEBUG', 'true'))
+# For deployment in production:
+# AMY_DEBUG=false AMY_SECRET_KEY="..." ./manage.py runserver ...
 
 if DEBUG:
-  SECRET_KEY = '3l$35+@a%g!(^y^98oi%ei+%+yvtl3y0k^_7-fmx2oj09-ac5@'
+    SECRET_KEY = '3l$35+@a%g!(^y^98oi%ei+%+yvtl3y0k^_7-fmx2oj09-ac5@'
 else:
-  SECRET_KEY = None
-SECRET_KEY = os.environ.get('SECRET_KEY', SECRET_KEY)
-
+    SECRET_KEY = None
+SECRET_KEY = os.environ.get('AMY_SECRET_KEY', SECRET_KEY)
 
 
 TEMPLATE_DEBUG = True


amy/settings.py
```diff
@@ -19,17 +19,16 @@
 # See https://docs.djangoproject.com/en/1.7/howto/deployment/checklist/
 
 
-# SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = json.loads(os.environ.get('DEBUG', 'true'))
-# For deployment in production: 
-# DEBUG=false SECRET_KEY="..." ./manage.py runserver ...
+# SECURITY WARNING: don't run with DEBUG turned on in production!
+DEBUG = json.loads(os.environ.get('AMY_DEBUG', 'true'))
```
gvwilson: Why are we doing a json.loads here?

gvwilson:
+1 to merge once someone explains the json.loads - thanks.

amy/settings.py
pbanaszkiewicz: It's there to handle 'true' / 'false' strings, ie. convert them to True/False objects in Python.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        From: Greg WilsonSent: Sunday, 1 March 2015 14:24To: swcarpentry/amyReply To: swcarpentry/amyCc: Piotr BanaszkiewiczSubject: Re: [amy] Add "AMY_" prefix to env vars DEBUG, SECRET_KEY (#206)In amy/settings.py:
> @@ -19,17 +19,16 @@
>  # See https://docs.djangoproject.com/en/1.7/howto/deployment/checklist/
>  
>  
> -# SECURITY WARNING: don't run with debug turned on in production!
> -DEBUG = json.loads(os.environ.get('DEBUG', 'true'))
> -# For deployment in production: 
> -# DEBUG=false SECRET_KEY="..." ./manage.py runserver ...
> +# SECURITY WARNING: don't run with DEBUG turned on in production!
> +DEBUG = json.loads(os.environ.get('AMY_DEBUG', 'true'))

Why are we doing a json.loads here?

—Reply to this email directly or view it on GitHub.

gvwilson: On 2015-03-01 8:28 AM, Piotr Banaszkiewicz wrote:
> > +DEBUG = json.loads(os.environ.get('AMY_DEBUG', 'true'))
> It's there to handle 'true' / 'false' strings, ie. convert them to 
> True/False objects in Python.

That seems like a really big hammer for a really small nail - it would 
be a lot simpler, and more conformant with Unix convention, to simply 
require AMY_DEBUG=1.
Thanks
Greg

wking: On Sun, Mar 01, 2015 at 05:47:54AM -0800, Greg Wilson wrote:
> 2015-03-01 8:28 AM, Piotr Banaszkiewicz:
> > > +DEBUG = json.loads(os.environ.get('AMY_DEBUG', 'true'))
> >
> > It's there to handle 'true' / 'false' strings, ie. convert them to
> > True/False objects in Python.
>
> That seems like a really big hammer for a really small nail - it
> would be a lot simpler, and more conformant with Unix convention, to
> simply require AMY_DEBUG=1.

It looks like the json-or-not issue has been spun off into #207.

wking:
On Sat, Feb 28, 2015 at 11:24:51PM -0800, Piotr Banaszkiewicz wrote:
> This is dictated by
> https://github.com/swcarpentry/amy/pull/204#issuecomment-76557668
> and it's generally a good idea to use such variables with
> per-application prefix.
> 
> If Greg's using specific "standard" for `DEBUG` values we can update
> our `settings.py` here to have: `AMY_DEBUG=$DEBUG…

As I said before [1], I prefer consolidating on standard variable
names and semantics, instead of having per-app variables with per-app
semantics.  But I don't care about it strongly enough to stand in the
way (as much as a third-party contributor can stand in the way ;) if
everyone else prefers the AMY_ prefix.

[1]: https://github.com/swcarpentry/amy/pull/204#issuecomment-76558570

pbanaszkiewicz:
@wking a few key points for "AMY_" prefix:

* readability (you can immediately tell that "AMY_DEBUG" is meant for Amy and not another application)
* flexibility (we can set DEBUG/SECRET independently from that other application and I can imagine at least a couple of scenarios when that's useful)

Please merge if this convinces you.

wking:
On Mon, Mar 02, 2015 at 01:17:18AM -0800, Piotr Banaszkiewicz wrote:
> * readability (you can immediately tell that "AMY_DEBUG" is meant
>   for Amy and not another application)
> * flexibility (we can set DEBUG/SECRET independently from that other
>   application and I can imagine at least a couple of scenarios when
>   that's useful)

At work, we launch services like this using runit's chpst [1], which
lets us load environment variables from a directory immediately before
launching the service (among other things you can tweak).  In any
case, I'd expect the environment setup to be part of the deploy
tooling (whichever tooling you want to use [2]), and unless you're
launching amy from *within* a process that has an incompatible
interpretation for DEBUG (or SECRET_KEY, or HOSTNAME, #209), you're
not going to run into problems with overlapping names.  And as I'd
said before, I *prefer* overlapping names when you can also get
overlapping semantics [3], since I expect a host to have the same name
for all services, and the same debug/production status, etc.
SECRET_KEY *is* amy-specific, but it's a secret, so you shouldn't be
setting it to amy's value for non-amy processes.

Anyhow, the AMY_ prefix seems like a solution in search of a problem.
Maybe you can elaborate on the situation in which the variable is
*not* clearly for amy?  My example:

  amy$ DEBUG=false SECRET_KEY=... ./manage.py runserver 0.0.0.0:80

seems pretty clear to me.  That example (setting the environment
variables when launching the process) also avoids overlap issues
(module the process nesting I mentioned above).  Is there a use-case
I'm missing where the readability or flexibility is compromised?  Can
you sketch it out in more detail for me?

[1]: http://smarden.org/runit/chpst.8.html
[2]: https://github.com/swcarpentry/amy/pull/186#issuecomment-75112545
[3]: https://github.com/swcarpentry/amy/pull/204#issuecomment-76558570

pbanaszkiewicz:
> Is there a use-case I'm missing where the readability or flexibility is compromised? Can you sketch it out in more detail for me?

Sure!

The use-cases (scenarios) I can think about are probably very rare and come down to "not maintaining good deployment practices".

1. A maintainer has two applications that use `DEBUG` envvar. He sets `DEBUG` globally for one application, then tries to test-run AMY via `./manage.py runserver` (which by default runs in debug mode).   (← flexibility issue)

2. A new maintainer comes in to maintain the server AMY is deployed at. He sees one `DEBUG` envvar. Can he guess this envvar is used to steer two applications?    (← readability issue)

As I said, these scenarios are rather unlikely, but on the other hand… we don't know anything about [that other application Greg is working on](https://github.com/swcarpentry/amy/pull/204#issuecomment-76557668).

Having per-app prefix ("AMY_" in this case) is not only common in Unix world, but also solves problems of having non-unique identifiers in global-only namespace (prefix introduces "scope" of some sort).

Also: having prefix solves overlapping issue (if there's ever any) without altering other applications.

wking:
On Mon, Mar 02, 2015 at 09:56:08AM -0800, Piotr Banaszkiewicz wrote:
> 1. A maintainer has two applications that use `DEBUG` envvar. He
> sets `DEBUG` globally for one application, then tries to test-run
> AMY via `./manage.py runserver` (which by default runs in debug
> mode).  (← flexibility issue)

And do we want amy to run in DEBUG mode?  Presumably you'd only set
DEBUG if you were working on a development box, so you *would* want to
enable debugging on both applications.

> 2. A new maintainer comes in to maintain the server AMY is deployed
> at. He sees one `DEBUG` envvar. Can he guess this envvar is used to
> steer two applications?  (← readability issue)

While it's good to have clarity and namespaces, it's also good to not
repeat yourself.  Also, having a single DEBUG variable is less
information for the new maintainer to figure out (“are you on a
production box, yes/no?”), vs. “do you want to debug amy?  $OTHER_APP?
…”).

> As I said, these scenarios are rather unlikely, but on the other
> hand… we don't know anything about that other application Greg is
> working on.

Maybe we should just match the values used by this application, since
I haven't been able to find a broad consensus for a DEBUG variable or
on values for boolean environment options more generally [1].

Or we just suggest that users set these at runtime, instead of setting
them globally without commenting on why ;).

[1]: https://github.com/swcarpentry/amy/pull/207#issuecomment-76632810

wking:
Anyhow, this discussion seems like one that could go around and around
for a while.  If you delegate me to make it, I'd decide against the
prefix (unless we get more data on problems it's causing).  If you
delegate someone else to make it, I'll accept their decision [1].  I
think we've spent enough time on it for the perspectives to become
clear, so the maintainer should just take them into consideration,
make the call, and move forward.

[1]: https://github.com/swcarpentry/amy/pull/206#issuecomment-76633431

PR #207, opened 2015-03-01
No need to load json. (follow up on PR #200)

drio:


wking:
On Sun, Mar 01, 2015 at 08:13:58AM -0800, David Rio Deiros wrote:
>   * No need to load json.

We use this to support:

  DEBUG=true

and:

  DEBUG=false

Without the ‘json.loads’ wrapper, both of those will enable DEBUG.

drio:
What do you mean? The output from `os.environ.get()` is not json so `json.loads()` fails.

```
$ python
Python 3.4.2 (default, Feb 12 2015, 14:15:15)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-11)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> import os
>>> os.environ.get('DEBUG', 'true')
'FALSE'
>>> json.loads('FALSE')
  File "/home/swc/drd_python/Python34/lib/python3.4/json/__init__.py", line 318, in loads
    return _default_decoder.decode(s)
  File "/home/swc/drd_python/Python34/lib/python3.4/json/decoder.py", line 343, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/home/swc/drd_python/Python34/lib/python3.4/json/decoder.py", line 361, in raw_decode
    raise ValueError(errmsg("Expecting value", s, err.value)) from None
ValueError: Expecting value: line 1 column 1 (char 0)
```

wking:
More discussion here [1], as part of #206.  In response to @gvwilson's
comments, the standard library is a useful hammer, but it doesn't cost
any more than the Python install we already need for amy ;).

I've looked online and through bash(1), python(1), and gcc(1) in
search of conventions for boolean environment variables.  Bash is all
over the place (EMACS is ‘t’ for true, OPTERR is ‘1’ for true, and
POSIXLY_CORRECT is set for true, possibly including being set to the
empty string).  Python is consistently “set to a non-empty string for
true”.  Personally, I think accepting ‘true’ and ‘false’ as values
makes for a more accessible interface than “non-empty” vs “empty or
unset”, but this is a Python application, so I'd be ok with:

  DEBUG = os.environ.get('DEBUG')

I don't think having a default-true value makes sense with the “empty
for false” interpretation, because:

  DEBUG='' ./manage.py …

isn't clearly turning off debugging (to me).  On the other hand, I
think:

  DEBUG=false ./manage.py …

is clearly turning off debugging.

[1]: https://github.com/swcarpentry/amy/pull/206/files#r25567257

wking:
On Sun, Mar 01, 2015 at 01:38:30PM -0800, David Rio Deiros wrote:
> What do you mean? The output from `os.environ.get()` is not json so
> `json.loads()` fails.

You can set the value of the environment variable however you like, so
obviously you can pick non-JSON values.  I'm fine crashing in those
cases, but I'm happy to set things up to give a more personal error
message if we want.

> >>> os.environ.get('DEBUG', 'true')
> 'FALSE'

That's not JSON, but ‘false’ works as expected:

  $ DEBUG=false python -c "import json, os; print(json.loads(os.environ.get('DEBUG', 'true')))"
	False

drio:
Ok. That makes sense.

PR #208, opened 2015-03-01
set ALLOWED_HOSTS properly in production.

drio:


wking:
On Sun, Mar 01, 2015 at 02:24:20PM -0800, David Rio Deiros wrote:
>   * set ALLOWED_HOSTS properly in production.

At work we use:

  ALLOWED_HOSTS = [os.environ.get('HOSTNAME', '*')]

to avoid hard-coding deploy-specific values in the deploy-agnostic
source repository.

PR #212, opened 2015-03-04
Improve README

r-gaia-cs:


wking:
On Tue, Mar 03, 2015 at 06:10:53PM -0800, Raniere Silva wrote:
>   * Improve README

Looks good to me.

Issue #213, opened 2015-03-04
Support debriefing sessions


r-gaia-cs:
What about add a button to generate a list of instructors that taught in the last n weeks so we could send a invitation for the debriefing session?

wking:
On Tue, Mar 03, 2015 at 06:15:21PM -0800, Raniere Silva wrote:
> What about add a button to generate a list of instructors that
> taught in the last n weeks so we could send a invitation for the
> debriefing session?

This (and #214) sound good to me, but should probably wait until after
v1.0.  That doesn't mean we can't start working on the specs now
though ;).  Do we want a special view for this particular feature, or
should we just add more flexibility to /workshops/instructors/?
Personally, I'd rather collapse all the person searching into a single
view, but I've only seen “flexible search with a single field” work
with some well-documented search strings (vs. checkboxes and other
non-text widgets for building queries).  That didn't get rave reviews
the last time I suggested it [1,2,3], but maybe there's a tipping
point that makes the approach simpler than remember the set of
special-purpose search pages ;).

[1]: https://github.com/swcarpentry/amy/issues/24#issuecomment-67971632
[2]: https://github.com/swcarpentry/amy/issues/24#issuecomment-67974983
[3]: https://github.com/swcarpentry/amy/issues/24#issuecomment-67983074

r-gaia-cs:
> > What about add a button to generate a list of instructors that
> > taught in the last n weeks so we could send a invitation for the
> > debriefing session?
> 
> This (and #214) sound good to me, but should probably wait until after
> v1.0.  That doesn't mean we can't start working on the specs now
> though ;).

=)

> Do we want a special view for this particular feature, or
> should we just add more flexibility to /workshops/instructors/?

In this case and #214 I don't see the need for a flexibility search.
In this case the input should be only a time range for workshops date.

wking:
On Wed, Mar 04, 2015 at 02:44:33AM -0800, Raniere Silva wrote:
> > Do we want a special view for this particular feature, or should
> > we just add more flexibility to /workshops/instructors/?
> 
> In this case and #214 I don't see the need for a flexibility search.
> In this case the input should be only a time range for workshops
> date.

Do you want to add that to the /workshops/instructors search?  That
page already has a lot of widgets for finding instructors for future
workshops, and I don't think those widgets would help with this case
or #214.  Would you just add more widgets, and trust users to not mix
and match?  Or do all the combinations actually mean something?  For
example:

[...]

PR #110, opened 2014-12-24
Add per-person award administration and per-event "Assign awards" button

wking:
Fixes #37, although see my uncertainty about the "Assign awards" logic
in the "POST /workshops/event/{slug}/award" commit message.

PR #216, opened 2015-03-05
Add selection to debrief session

r-gaia-cs:
Address #213.

**This doesn't have a link from the home page.**

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -71,3 +71,11 @@ class SearchForm(forms.Form):
     in_persons = forms.BooleanField(label='in persons',
                                     required=False,
                                     initial=True)
+
+class DebriefForm(forms.Form):
+    '''Represent general debrief form.'''
+
+    begin_date = forms.CharField(label='Begin date as YYYY-MM-DD',
+                           max_length=10)
+    end_date = forms.CharField(label='End date as YYYY-MD-DD',
+                           max_length=10)
--- a/workshops/templates/workshops/debrief.html
+++ b/workshops/templates/workshops/debrief.html
@@ -30,7 +30,7 @@
 	</tr>
     {% endfor %}
     </table>
-    <a href="mailto:{% for task in all_tasks %}{{ task.person.email }}{% if not last %},{% endif %}{% endfor %}">Send email</a>
+    <a href="mailto:{% for task in all_tasks %}{% if task.person.email %}{{ task.person.email }}{% if not last %},{% endif %}{% endif %}{% endfor %}">Send email</a>
 {% else %}
     <p>No instructors for debrief section.</p>
 {% endif %}
--- a/workshops/templates/workshops/debrief.html
+++ b/workshops/templates/workshops/debrief.html
@@ -30,7 +30,7 @@
 	</tr>
     {% endfor %}
     </table>
-    <a href="mailto:{% for task in all_tasks %}{% if task.person.email %}{{ task.person.email }}{% if not last %},{% endif %}{% endif %}{% endfor %}">Send email</a>
+    <a href="mailto:{% for task in all_tasks %}{% if task.person.email %}{{ task.person.email }}{% if not forloop.last %},{% endif %}{% endif %}{% endfor %}">Send email</a>
 {% else %}
     <p>No instructors for debrief section.</p>
 {% endif %}
--- a/workshops/templates/workshops/debrief.html
+++ b/workshops/templates/workshops/debrief.html
@@ -1,0 +1,41 @@
+{% extends "workshops/_page.html" %}
+
+{% load breadcrumbs %}
+{% block breadcrumbs %}
+    {% breadcrumb_main_page %}
+    {% breadcrumb_active title %}
+{% endblock %}
+
+{% block content %}
+<form action="{% url 'debrief' %}" method="post">
+    {% csrf_token %}
+    <table class="table table-striped">
+    {{ form.as_table }}
+    </table>
+    <input type="submit" value="Submit" />
+</form>
+
+{% if all_tasks %}
+    <table class="table table-striped">
+        <tr>
+	    <th>event</th>
+	    <th>person</th>
+	    <td></td>
+	</tr>
+    {% for task in all_tasks %}
+        <tr>
+            <td>{{ task.event }}</td>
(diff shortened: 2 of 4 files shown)


wking:
I'm still uncertain about per-task views [1], but other than that this
looks reasonable to me.  I don't think the trailing comma is legal
though [2].  You should probably use:

  <a href="mailto:{% for task in all_tasks %}{{ task.person.email }}{% if not last %},{% endif %}{% endfor %}">…</a>

[1]: https://github.com/swcarpentry/amy/issues/213#issuecomment-77088428
[2]: https://tools.ietf.org/html/rfc6068#section-2
[3]: https://docs.djangoproject.com/en/1.7/ref/templates/builtins/#for

wking:
Also, it's better to push as much of the filtering as possible down
into the database, to save the cost of converting everything into
Python objects.  Maybe:

  tasks = Task.objects.filter(
      event__end__gte=form.cleaned_data['begin_date'],
      event__begin__lte=form.cleaned_data['end_date'],
      role__name='instructor',
      ).order_by('event', 'person', 'role')

for your whole form.is_valid() block ;).

We don't want to mail instructors who have disabled access (see #161).
Do we want that filtering to happen in the view, or in the template
while constructing the mailto?  We might also have instructors with
NULL email (see #125, 4659c06, and #140).  Do we filter *them* out in
the view or the template?

r-gaia-cs:
@wking Thanks for the comments.

> Also, it's better to push as much of the filtering as possible down
> into the database, to save the cost of converting everything into
> Python objects

I got "Relation fields do not support nested lookups" when trying your suggestion.

> Do we want that filtering to happen in the view, or in the template
> while constructing the mailto? (...) Do we filter *them* out in
> the view or the template?

You can vote on this. =)

wking:
On Fri, Mar 06, 2015 at 06:42:02AM -0800, Raniere Silva wrote:
> > Also, it's better to push as much of the filtering as possible
> > down into the database, to save the cost of converting everything
> > into Python objects
>
> I got "Relation fields do not support nested lookups" when trying
> your suggestion.

Ah, looks like it's Event.start [1], so:

  tasks = Task.objects.filter(
      event__end__gte=form.cleaned_data['begin_date'],
      event__start__lte=form.cleaned_data['end_date'],
      role__name='instructor',
      ).order_by('event', 'person', 'role') 

> > Do we want that filtering to happen in the view, or in the
> > template while constructing the mailto? (...) Do we filter *them*
> > out in the view or the template?
> 
> You can vote on this. =)

I'm happy either way ;).  It's just a question of whether or not users
(SWC admins) will want to know about the matching instructors that
they can't contact (in which case we should filter in the mailto) or
if they only care about folks they can contact (in which case we
should filter in the view).  So if you're contacting instructors for
debriefing, do you want to see entries in the results for instructors
you can't contact?

[1]: https://github.com/swcarpentry/amy/blob/546004032019ecc57e8834280d5bbad6be233d1c/workshops/models.py#L256

r-gaia-cs:
> It's just a question of whether or not users
> (SWC admins) will want to know about the matching instructors that
> they can't contact (in which case we should filter in the mailto) or
> if they only care about folks they can contact (in which case we
> should filter in the view).  So if you're contacting instructors for
> debriefing, do you want to see entries in the results for instructors
> you can't contact?

I update this using @wking suggest to use `filter` and removing instructors that we can't contact.
Instructors without email or blank email are listed but not include in the "send email" link.

@wking Could you check this using as input `2015-07-01` and `2015-07-30`? Jinja's `last` variable isn't working.

wking:
On Fri, Mar 06, 2015 at 11:29:56AM -0800, Raniere Silva wrote:
> Jinja's `last` variable isn't working.

Ah, looks like you need the namespace mentioned in the docs I'd linked
[1].  Try:

  {% if not forloop.last %}

Django's template language is very similar to Jinja, but they're not
quite the same.

[1]: https://docs.djangoproject.com/en/1.7/ref/templates/builtins/#for

r-gaia-cs:
> Ah, looks like you need the namespace

Thanks. Fixed.

workshops/forms.py
```diff
@@ -71,3 +71,11 @@ class SearchForm(forms.Form):
     in_persons = forms.BooleanField(label='in persons',
                                     required=False,
                                     initial=True)
+
+class DebriefForm(forms.Form):
+    '''Represent general debrief form.'''
+
+    begin_date = forms.CharField(label='Begin date as YYYY-MM-DD',
+                           max_length=10)
+    end_date = forms.CharField(label='End date as YYYY-MD-DD',
```
pbanaszkiewicz: Little typo; should be "End date as YYYY-MM-DD". I'd also add a `min_length=10` parameter to help with validation. Or do we accept formats like "YYYY-MM"?

(We can go pro and use a custom validator https://docs.djangoproject.com/en/1.7/ref/validators/#writing-validators to make sure a valid date is provided. But I don't see it as a necessity.)

pbanaszkiewicz:
@r-gaia-cs: Looks good to me, although I think we need a link to `/debrief/` subpage somewhere.

workshops/forms.py
wking: On Sun, Mar 08, 2015 at 12:36:00PM -0700, Piotr Banaszkiewicz wrote:
> We can go pro and use a custom validator
> https://docs.djangoproject.com/en/1.7/ref/validators/#writing-validators
> to make sure a valid date is provided. But I don't see it as a
> necessity.

I've also proposed Bootstrap's datepicker (see #110) for handling date
input.  It's harder to break your input if you're just clicking on
calendar entries.

pbanaszkiewicz: > I've also proposed Bootstrap's datepicker (see #110) for handling date input. It's harder to break your input if you're just clicking on calendar entries.

It's a good UX idea, but if @r-gaia-cs doesn't want to include this widget we can create an issue for this.

r-gaia-cs:
@pbanaszkiewicz Can you add the link to `/debrief/` and change this to use Bootstrap's datepicker (see #110)?

pbanaszkiewicz:
Okay. I think both link to `/debrief/` and Bootstrap's datepicker should land after #221. I'll make an issue for that.

PR #217, opened 2015-03-05
Previous method was yielding a 400 in production.

drio:


wking:
On Thu, Mar 05, 2015 at 03:25:23PM -0800, David Rio Deiros wrote:
>   * Previous method was yielding a 400 in production.

Under what conditions?  And with what value of HOSTNAME?  I'm happy to
figure out a generic solution that fixes this issue, but I need enough
information to reproduce the problem locally.

drio:
This is when starting the app in the vps server via:

`$ DEBUG=false python ./manage.py runserver 0.0.0.0:8080`

```
(venv.amy.p3)swc@vps1:~$ python
Python 3.4.2 (default, Feb 12 2015, 14:15:15)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-11)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ.get('HOSTNAME', '*')
'vps1.sensibleadventures.com'
>>>
```

wking:
On Fri, Mar 06, 2015 at 07:36:13AM -0800, David Rio Deiros wrote:
> This is when starting the app in the vps server via:
> 
> `$ DEBUG=false python ./manage.py runserver 0.0.0.0:8080`

Since #206 landed with 51ebaf8 (2015-03-03) it should be ‘AMY_DEBUG’.

> >>> os.environ.get('HOSTNAME', '*')
> 'vps1.sensibleadventures.com'

Can you checkout the old code:

  $ git checkout b420810

And try:

  $ AMY_DEBUG=false AMY_SECRET_KEY='…' HOSTNAME='amy.software-carpentry.org' python ./manage.py runserver 0.0.0.0:8080

?

PR #219, opened 2015-03-07
Updated README.md to Python3

narayanaditya95:
Changing ```$ sudo python``` to  ```$ sudo python3``` at all places should suffice.
Please review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,16 +28,16 @@ commands : Makefile
 
 ## test         : run all tests.
 test :
-	python manage.py test
+	python3 manage.py test
 
 ## migrations   : create/apply migrations
 migrations :
-	python manage.py makemigrations
-	python manage.py migrate
+	python3 manage.py makemigrations
+	python3 manage.py migrate
 
 ## import       : import and save legacy data
 import :
-	python migrater.py ${SRC_DB} ${APP_DB}
+	python3 migrater.py ${SRC_DB} ${APP_DB}
 	${QUERY} .dump > ${APP_SQL}
 
 ## database     : re-make database using saved data
@@ -47,7 +47,7 @@ database :
 
 ## superuser    : make a super-user in the database
 superuser :
-	python manage.py create_superuser
+	python3 manage.py create_superuser
 
 ## schema       : display the database schema
 schema :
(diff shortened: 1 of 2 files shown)


r-gaia-cs:
@narayanaditya95 you should use `python3` instead of `python3.4`.

narayanaditya95:
Changes made. Please review.

wking:
On Sat, Mar 07, 2015 at 08:02:44AM -0800, Aditya Narayan wrote:
> Changes made. Please review.

I'd rather just mention “Python 3” in the text.  For example:

  To get started:

  1. Install Python 3
	2. Install amy's Python dependencies:

       $ python -m pip install --user -r requirements.txt

Because on my Gentoo box ‘python’ is the name for the Python 3 binary:

  $ python --version
  Python 3.4.2

If we really want, we could say something like:

	2. Install amy's Python dependencies:

       $ python -m pip install --user -r requirements.txt

     The name of your Python 3 executable may be `python3`, in which
     case run:
		 
       $ python3 -m pip install --user -r requirements.txt

     and make similar replacements in this file's other `python`
     commands as well as in the `Makefile`.

Because we're currently hard-coding `python` in the Makefile.

I also prefer a --user install here than suggesting folks sudo install
system packages, since system-wide installs are best left to your
system's package manager (in my opinion ;), and we shouldn't claim to
need root privileges if we don't need them.

narayanaditya95:
@wking 
Refering to a [wiki at Ubuntu](https://wiki.ubuntu.com/Python/3), it seems ```python``` always points to Python2.7, if installed. Although there are goals to shift everything to Python3 as mentioned in the wiki, it still hasn't materialised. I am aware that other distro may point ```python``` to Python3.
Using ```python``` instead of  ```python3``` does not solve the problem for me, unless I add an alias at my ```.bashrc```.
I suggest we explicitly mention ```python3``` in the Makefile for the same reason.

Please correct me if I am wrong.

r-gaia-cs:
@wking I prefer use `python3` since the build instructions from Python says

> This will install Python as python3.

wking:
On Sat, Mar 07, 2015 at 04:42:47PM -0800, Aditya Narayan wrote:
> Refering to a wiki at Ubuntu…

The wiki links to PEP 394 [1], which suggests ‘python3’ too, so I'll
reverse my position and support ‘python’ → ‘python3’.  This PR should
update the Makefile too.

[1]: http://legacy.python.org/dev/peps/pep-0394/

narayanaditya95:
Sorry for missing out the Makefile. Updated Makefile as instructed.

wking:
On Sun, Mar 08, 2015 at 05:23:14PM -0700, Aditya Narayan wrote:
> Updated Makefile as instructed.

Thanks :).  The PR looks good to me now.

pbanaszkiewicz:
These changes look good to me so I'll go ahead and merge them.

PR #220, opened 2015-03-08
Fix py2 depend

narayanaditya95:
Since basestring has been deprecated in Python3, I changed the use of ```basestring``` to ```str``` in the decorator ```look_for_fixme``` at ```workshops/checks.py```. Please review.

### DIFF
--- a/workshops/check.py
+++ b/workshops/check.py
@@ -138,7 +138,7 @@ def look_for_fixme(func):
     '''Decorator to fail test if text argument starts with "FIXME".'''
     def inner(arg):
         if (arg is not None) and \
-           isinstance(arg, basestring) and \
+           isinstance(arg, str) and \
            arg.lstrip().startswith('FIXME'):
             return False
         return func(arg)


wking:
On Sun, Mar 08, 2015 at 08:49:53AM -0700, Aditya Narayan wrote:
> Since basestring has been deprecated in Python3, I changed the use
> of ```basestring``` to ```str``` in the decorator
> ```look_for_fixme``` at ```workshops/checks.py```.

This check script is actually from swcarpentry/workshop-template (see
[1] and #111).  I'd rather keep this script in it's own repository and
distribute it via PyPI [2], but I really don't want to start sinking
time into maintaining a local fork.  In this case, the test script is
also used outside of amy, so I think it's best to have compatibility
with both Python 2 and 3.  If the current
swcarpentry/workshop-template version isn't compatible with Python 3,
maybe you can patch it there?

[1]: https://github.com/swcarpentry/workshop-template/blob/gh-pages/tools/check.py
[2]: https://github.com/swcarpentry/amy/pull/111#issuecomment-68059757

narayanaditya95:
@wking Closing this PR and opening one at [1]. The PR would be modified to support both Python 2 and 3.

[1][https://github.com/swcarpentry/workshop-template](https://github.com/swcarpentry/workshop-template)

narayanaditya95:
@wking 
The issue has been taken care at swcarpentry/worshop-template[1]. Should I reopen the PR with the changes incorporated in Amy.
[1][](https://github.com/swcarpentry/workshop-template/blob/gh-pages/tools/check.py#L16)

wking:
On Sun, Mar 08, 2015 at 06:16:32PM -0700, Aditya Narayan wrote:
> The issue has been taken care at swcarpentry/worshop-template[1].

Ah, it looks like swcarpentry/worshop-template#131 brought the fix.
Personally, I'd still prefer we just pulled that script out into the
external package, but if @gvwilson wants to kick that can down the
road again, a new PR copy-pasting the current
swcarpentry/worshop-template version of the script into this
repository is probably the best alternative.

Issue #194, opened 2015-02-18
Add a navigation bar.


gvwilson:
There should be a navigation bar. For now, go to the home page and select a link.

pbanaszkiewicz:
Can it be in the topbar that appears after you log in?

gvwilson:
Sure, but this one isn't required for launch - we can get through the 
first month with "click to go home / click on a link".

wking:
On Wed, Feb 18, 2015 at 12:00:36PM -0800, Greg Wilson wrote:
> Sure, but this one isn't required for launch…

I'm not sure how the “essential” label squares with “isn't required
for launch”.  See also my previous confusion with “essential” + “v1.1”
for #170 [1].  I'd recommend dropping the “essential” and “defer”
labels in favor of assigning issues to milestones, because that gives
you more granularity on priority without diluting the other labels.

[1]: https://github.com/swcarpentry/amy/issues/170#issuecomment-74915443

pbanaszkiewicz:
#221 was merged so I'm closing this issue.

PR #210, opened 2015-03-02
models.py: remove Person.slug

pbanaszkiewicz:
Person.slug was replaced with Person.username.

This PR is based on @gvwilson work done in #205, but instead of making `Person.username` blank, `Person.username` stays the way it was previously (unique, non-nullable, non-blankable).

PR #221, opened 2015-03-08
Extended navigation bar

pbanaszkiewicz:
This PR:

* adds `navbar_element` template tag
* adds 4 new positions in the navigation bar (on top of the page)
* adds a search form in the navbar - a shortcut to the official search
  page

Ref #194

### DIFF
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -12,6 +12,8 @@ https://docs.djangoproject.com/en/1.7/ref/settings/
 import os
 import json
 
+from django.conf import global_settings
+
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
 
 
@@ -65,16 +67,9 @@ MIDDLEWARE_CLASSES = (
 
 # WARNING: in Django 1.8 processors get moved from ``django.core`` to
 #          ``django.template``.
-TEMPLATE_CONTEXT_PROCESSORS = [
-    "django.contrib.auth.context_processors.auth",
-    "django.core.context_processors.debug",
-    "django.core.context_processors.i18n",
-    "django.core.context_processors.media",
-    "django.core.context_processors.static",
-    "django.core.context_processors.tz",
+TEMPLATE_CONTEXT_PROCESSORS = global_settings.TEMPLATE_CONTEXT_PROCESSORS + (
     "django.core.context_processors.request",
-    "django.contrib.messages.context_processors.messages"
-]
+)
 
 ROOT_URLCONF = 'amy.urls'
 
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -33,7 +33,7 @@ SECRET_KEY = os.environ.get('AMY_SECRET_KEY', SECRET_KEY)
 
 TEMPLATE_DEBUG = True
 
-ALLOWED_HOSTS = [ 
+ALLOWED_HOSTS = [
     'software-carpentry.org',
     'software-carpentry.org.'
 ]
@@ -62,6 +62,19 @@ MIDDLEWARE_CLASSES = (
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
 )
+
+# WARNING: in Django 1.8 processors get moved from ``django.core`` to
+#          ``django.template``.
+TEMPLATE_CONTEXT_PROCESSORS = [
+    "django.contrib.auth.context_processors.auth",
+    "django.core.context_processors.debug",
+    "django.core.context_processors.i18n",
+    "django.core.context_processors.media",
+    "django.core.context_processors.static",
+    "django.core.context_processors.tz",
+    "django.core.context_processors.request",
+    "django.contrib.messages.context_processors.messages"
+]
 
 ROOT_URLCONF = 'amy.urls'
 
--- a/workshops/templates/base.html
+++ b/workshops/templates/base.html
@@ -3,7 +3,7 @@
 {% block navbar %}
     {% load navigation %}
     <nav class="navbar navbar-default">
-      <div class="container-fluid">
+      <div class="container">
         <!-- Brand and toggle get grouped for better mobile display -->
         <div class="navbar-header">
           <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#bs-example-navbar-collapse-1">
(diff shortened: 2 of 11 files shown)


pbanaszkiewicz:
Preview:
screen shot 2015-03-08 at 22 12 52

wking:
On Sun, Mar 08, 2015 at 03:28:11PM -0700, Piotr Banaszkiewicz wrote:
> Preview: …

It looks a bit wierd to me to have a fixed-width for the page content
but a fluid width for the nav bar.  Personally, I'd prefer a fluid
layout everywhere.

wking:
Why do we need TEMPLATE_CONTEXT_PROCESSORS (it looks like you're only
adding django.core.context_processors.request to the defaults [1])?  I
guess it's for request.path in navbar_element?

The implementation looks reasonable to me, but I'd list everything
from the “Menu” column of the main page and then drop that column once
the navbar was populated.

[1]: https://docs.djangoproject.com/en/1.7/ref/settings/#std:setting-TEMPLATE_CONTEXT_PROCESSORS

pbanaszkiewicz:
Thanks @wking for the feedback, I changed things accordingly.

Here's little preview:
screen shot 2015-03-09 at 10 36 05

I had an idea to put a third column in Dashboard: "Events with no instructors", but it turned out too complicated… @gvwilson do you think it would be useful to have such events listed in the dashboard?

wking:
On Mon, Mar 09, 2015 at 02:38:27AM -0700, Piotr Banaszkiewicz wrote:
> I had an idea to put a third column in Dashboard: "Events with no
> instructors", but it turned out too complicated…

Can you spin that off into it's own issue?  I don't want to
overcomplicate this navbar PR.

Provided that others aren't concerned by the fluix/fixed styling issue
[1], I think this is ok to merge.

If the TEMPLATE_CONTEXT_PROCESSORS change isn't compatible with Django
1.8, we probably want to update the requirements.txt to have:

  Django>=1.7,<1.8

[1]: https://github.com/swcarpentry/amy/pull/221#issuecomment-77781622

pbanaszkiewicz:
> Can you spin that off into it's own issue? I don't want to overcomplicate this navbar PR.

Done in #224. (BTW: I did not add this feature because it is hard to do in Django ORM…)

> Provided that others aren't concerned by the fluix/fixed styling issue [1], I think this is ok to merge.

I have full HD resolution (1920x1080) - it makes most webpages look empty on the sides. Below is a preview of a fixed-width navbar:
screen shot 2015-03-09 at 18 22 13

I think I like it more. I'll change accordingly.

There's one issue with this many items in the navbar. It looks strange below 1200px width:
screen shot 2015-03-09 at 18 24 44

> If the TEMPLATE_CONTEXT_PROCESSORS change isn't compatible with Django 1.8, we probably want to update the requirements.txt to have: Django>=1.7,<1.8

There's more to templates: https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-TEMPLATES

I can add a conditional that discovers if we're in 1.8 yet and sets up `TEMPLATES` for >=1.8 or `TEMPLATE_CONTEXT_PROCESSORS`, `TEMPLATE_DEBUG`, `TEMPLATE_DIRS`, `TEMPLATE_LOADERS`, `TEMPLATE_STRING_IF_INVALID` for <1.8.

If you agree with my idea, I'll do this in separate PR.

wking:
On Mon, Mar 09, 2015 at 10:32:59AM -0700, Piotr Banaszkiewicz wrote:
> There's one issue with this many items in the navbar. It looks
> strange below 1200px width:

Maybe we should push tasks/badges/airports into the dropdown and
rename it from “Utilities” to “More”?  I think Bootstrap has a way to
trigger a full menu collapse at a give screen size, maybe we just want
to do that?

> I can add a conditional that discovers if we're in 1.8 yet…

Sounds good to me, and a separate PR for this would be great.

pbanaszkiewicz:
> I think Bootstrap has a way to trigger a full menu collapse at a give screen size, maybe we just want to do that?

I was unable to make Bootstrap collapse reasonably, so I moved Tasks, Badges, Airports to "More" menu - as you suggested. Now it looks fine and the odd visual glitch happens only between 768px and 991px. I think it's manageable.

If you give me a green light, I'll go ahead and merge.

wking:
On Mon, Mar 09, 2015 at 10:50:01AM -0700, Piotr Banaszkiewicz wrote:
> If you give me a green light, I'll go ahead and merge.

Looks good to me, but again [1,2], I'm not the maintainer ;).  I'd
recommend having a clear maintainer who performs (or delegates) review
to avoid things like the recent #200 (with discussion about some new
environment variables) being followed immediately by #204 and #206,
both of which (I think) should have been handled (or at least
discussed) before #200 landed.  Of course, @gvwilson merged #200, so
maybe it doesn't matter ;).

[1]: https://github.com/swcarpentry/amy/pull/200#issuecomment-76223021
[2]: https://github.com/swcarpentry/amy/pull/210#issuecomment-76758829

pbanaszkiewicz:
@wking I just updated this PR with 8d0a538 seconds before your email arrived. I agree we need to know who's in charge, but I guess @gvwilson is AFK at the moment.

wking:
On Mon, Mar 09, 2015 at 11:00:42AM -0700, Piotr Banaszkiewicz wrote:
> I just updated this PR with 8d0a538 seconds before your email
> arrived.

Heh, that looks like a good change ;).

> I agree we need to know who's in charge, but I guess @gvwilson is
> AFK at the moment.

The world won't explode if this PR waits for his return ;).  *I* can't
deploy anything I merge anyway, so it's quite likely that regardless
of when this gets merged, it won't get deployed until he's back.

pbanaszkiewicz:
Hey @wking 

this is blocking #223 (I certainly won't explode because of that :) ).  I think I'll just make a PR for #223 that's based on this PR (#221).

wking:
On Wed, Mar 11, 2015 at 11:44:35AM -0700, Piotr Banaszkiewicz wrote:
> I think I'll just make a PR for #223 that's based on this PR (#221).

For things like this that are likely to land without further edits, I
don't see a problem with stacking PRs like that.  On the other hand, I
don't see much point in getting *too* far out ahead of the maintainer
;).

pbanaszkiewicz:

PR #225, opened 2015-03-11
Updated test_util.py

darshan95:
New unit tests added :
1) Check number of columns in case of extra fields and empty fields
2) Check new person is not added if existing user(case sensitive/insensitive email) with a new task is uploaded

### DIFF
--- a/workshops/test/test_util.py
+++ b/workshops/test/test_util.py
@@ -70,6 +70,47 @@ jane,a,doe,janedoe@email.com"""
             self.fail('Dumping person_tasks to JSON unexpectedly failed!')
 
 
+    def test_no_new_column_added(self):
+        ''' Ensure more number of fields are not added by increasing the number of columns'''
+        csv = """personal,middle,family,email,event,role,phone_no
+john,a,doe,johndoe@email.com,2015-09-01-ttt-online,learner,9991239011
+jane,a,doe,janedoe@email.com,2015-09-01-ttt-online,learner,9991239021"""
+        person_tasks, _ = self.compute_from_string(csv)
+        person = person_tasks[0]
+        self.assertTrue('phone_no' not in person)
+
+    def test_number_of_columns(self):
+        ''' Ensure number of columns remains same in case of empty fields or extra fields '''
+
+        csv = """personal,middle,family,email,event,role,phone_no
+john,a,doe,johndoe@email.com,2015-09-01-ttt-online,learner,9991239011"""
+        person_tasks, _ = self.compute_from_string(csv)
+        person = person_tasks[0]
+        self.assertEqual(len(person),7)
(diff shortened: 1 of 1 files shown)


workshops/test/test_util.py
```diff
@@ -70,6 +70,47 @@ def test_serializability_of_parsed(self):
...
+        self.assertTrue('phone_no' not in person)
+
+    def test_number_of_columns(self):
+        ''' Ensure number of columns remains same in case of empty fields or extra fields '''
+
+        csv = """personal,middle,family,email,event,role,phone_no
+john,a,doe,johndoe@email.com,2015-09-01-ttt-online,learner,9991239011"""
+        person_tasks, _ = self.compute_from_string(csv)
+        person = person_tasks[0]
+        self.assertEqual(len(person),7)
```
pbanaszkiewicz: Instead of checking for number of items, I'd rather check if the `person.keys()` corresponds to `Person.PERSON_TASK_UPLOAD_FIELDS` - either by hardcoding field names, or by comparing directly with `Person.PERSON_TASK_UPLOAD_FIELDS`.

The same applies to other `self.assertEqual(len(person),7)` in this test.

workshops/test/test_util.py
```diff
@@ -70,6 +70,47 @@ def test_serializability_of_parsed(self):
...
+        self.assertEqual(person['event'],None)
+
+        csv = """personal,middle,family,email
+john,a,doe,johndoe@email.com"""
+        person_tasks, _ = self.compute_from_string(csv)
+        person = person_tasks[0]
+        self.assertEqual(len(person),7)
+        self.assertEqual(person['event'],None)
+        self.assertEqual(person['role'],None)
```
pbanaszkiewicz: Please follow PEP8 (https://www.python.org/dev/peps/pep-0008/) for styling your code. Specifically:

* remove excessive blank lines before class methods (lines 71-72, etc.) and before classes themselves (lines 111-113), and within method's code (lines 339-341, 371-372, etc.)
* add whitespace after commas

You'll get best results when you install a PEP8 plugin for your editor of choice. Or use `pep8` Python command line application (https://pypi.python.org/pypi/pep8/1.6.2).

darshan95:
@pbanaszkiewicz Thanks for the guidance I have updated the test_number_of_columns function and followed PEP8 for styling the code. Whitespace after commas cannot be intoduced in the case of manually entering csv values, so not updated.

gvwilson:
Is this one now out of date?  If so, can it be closed?

pbanaszkiewicz:
I think this is out of date. I didn't give any feedback before because I didn't know the rationale for this PR.

I think we need to rework tests in the future, but that's entirely else matter.

PR #227, opened 2015-03-11
Case-insensitive matching of email

### LINKED ISSUE #226: Bulk-upload: New User added even in case of existing user
According to #225 , If a existing user with case insensitive email Id is uploaded with a new task, a new user is being added with that case insensitive email and no task is being added to the existing user.

darshan95:
This commit resolves #226 

create_uploaded_persons_tasks function was required to match email case-insensitive before creating a new person

### DIFF
--- a/workshops/util.py
+++ b/workshops/util.py
@@ -88,6 +88,7 @@ def verify_upload_person_task(data):
         errors = []
 
         event = item.get('event', None)
+
         if event:
             try:
                 Event.objects.get(slug=event)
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
Good stuff. Thanks @darshan95 for the fix :)

wking:
GitHub lost another of my mails ;).

darshan95:
@wking updated the code

wking:
On Thu, Mar 12, 2015 at 10:52:30AM -0700, Darshan Agarwal wrote:
> @wking updated the code

Looks good to me, thanks.

pbanaszkiewicz:

Issue #211, opened 2015-03-02
Bulk-upload: no-op if Person and their Task exist


pbanaszkiewicz:
After #201 gets merged in, we'll need to alter it's behavior a little.

Currently if Person and their Task already exist in the DB, an error is raised ("can't have duplicated enries"). @wking suggested in https://github.com/gvwilson/amy/pull/2#issuecomment-76458963 that we should simply do nothing in such cases.

I'm considering an UI extension (new column titled "Notes") that would indicate what's going to happen with the uploaded user.

wking:
On Mon, Mar 02, 2015 at 01:42:01AM -0800, Piotr Banaszkiewicz wrote:
> I'm considering an UI extension (new column titled "Notes") that
> would indicate what's going to happen with the uploaded user.

As I mention in the comment you linked, I don't think we need this
extra UI.  My suggested fix is outlined here [1].

[1]: https://github.com/gvwilson/amy/pull/2#issuecomment-76465279

pbanaszkiewicz:
Fixed in #228.

PR #228, opened 2015-03-11
Bulk-upload: no-op when user and task exist

pbanaszkiewicz:
Do nothing if we try to upload user and assigned task that already exist.

This solves #211.

### DIFF
--- a/workshops/test/test_util.py
+++ b/workshops/test/test_util.py
@@ -69,6 +69,16 @@ jane,a,doe,janedoe@email.com"""
         except TypeError:
             self.fail('Dumping person_tasks to JSON unexpectedly failed!')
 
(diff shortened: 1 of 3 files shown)


pbanaszkiewicz:
I went ahead with @wking suggestion (https://github.com/swcarpentry/amy/issues/211#issuecomment-77415417): no UI changes to indicate a specific user ↔ task exist.

wking:
On Wed, Mar 11, 2015 at 11:39:26AM -0700, Piotr Banaszkiewicz wrote:
> Do nothing if we try to upload user and assigned task that already exist.

Looks good to me.

pbanaszkiewicz:
Additionally, a fix for #231 was included here.

pbanaszkiewicz:

Issue #193, opened 2015-02-18
Merge people


gvwilson:
People often change email addresses, or use multiple addresses, so we need a way to say, "These two Person records in the database are actually the same human being, so eliminate one and change all foreign key references to it to point to the other person."

PR #229, opened 2015-03-13
Capability to merge duplicate person entries

neon-ninja:
This PR solves #193.


wking:
On Thu, Mar 12, 2015 at 09:23:39PM -0700, Nick Young wrote:
>   * capability to merge duplicate person entries

I prefer:

  type(primary_object)

to:

  primary_object.__class__

since it's usually best to avoid accessing stuff with leading
underscores (PEP 8 has a single leading underscore for internal use,
and double leading underscores for name mangling [1].  Best to avoid
using both from code outside the class implementation unless you have
to).

You could also like use QuerySet.update(…) for some of the foreign key
manipulation.

Minor tweaks aside, I'm not sure how I feel about merge_model_objects.
On the one hand, it's nice to have a generic implementation.  On the
other, its a fairly complicated implementation and I'm not sure it
hits all the corner cases (e.g. hardcoding '' as a missing value makes
me a bit jumpy, but it might be fine).  I don't expect us to be
merging models besides Person frequently, so it might be simpler (and
generic enough for our purposes) to write a model-merging function
specific to Person.  Or maybe the generic merger could be split out
into a separate library with it's own unit tests, and we could just
add it to requirements.txt?  Then I wouldn't have to worry about the
details of the implementation ;).

[1]: https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

workshops/views.py
```diff
@@ -375,6 +375,61 @@ def person_bulk_add_confirmation(request):
         return render(request, 'workshops/person_bulk_add_results.html',
                       context)
 
+@login_required
+def person_find_duplicates(request):
```
pbanaszkiewicz: I have a couple of notes regarding this function.

1) Please follow [PEP8](https://www.python.org/dev/peps/pep-0008/) especially when it comes to line length
2) ...and to spaces around colons and commas and `>`
3) You duplicate about 9 lines of code in lines 397 and 432 - it's against Don't Repeat Yourself rule
4) There's no handling for methods other than `POST` and `GET` (I'm nitpicking here.)

pbanaszkiewicz:
I looked through your code and I have 3 general notes, although I might have missed something important.

1. The UI is not clear on what is being merged into what. What's happening with non-empty values?
2. There should be some extensive tests against `util.merge_model_objects`
3. There are many PEP8 style violations (see https://github.com/swcarpentry/amy/pull/229#discussion_r27514733)

Please let me know if you have time and will to fix these, otherwise I may be able to come up with a PR against your PR sometime during Easter.

workshops/views.py
wking: On Tue, Mar 31, 2015 at 12:36:28PM -0700, Piotr Banaszkiewicz wrote:
> 4) There's no handling for methods other than `POST` and `GET` (I'm
> nitpicking here.)

To save some lookup, the correct status for unsupported methods is 405
(Method Not Allowed) [1].

[1]: http://tools.ietf.org/html/rfc7231#section-6.5.5

neon-ninja: I found a Django decorator to limit HTTP methods, which returns 405 for other methods - see https://github.com/neon-ninja/amy/commit/53abcd592e6696fa54be0b6d9914ca4f9cdd3ece

neon-ninja:
Regarding 1: That's possibly because I wasn't clear what should be merged into what. Replacing empty values with non-empty values seems obvious, but what should happen with non-empty values? Should I provide toggles in the interface such that the user can pick which is the correct value?

I've pushed some fixes for 3, see the commits above ^

pbanaszkiewicz:
Hey @neon-ninja,

thanks for your commits!

Now we're only left with two issues:

1. better UI
2. more tests against `util.merge_model_objects`.

I personally feel that a good UI for this feature is the hardest part.

@neon-ninja:

> Replacing empty values with non-empty values seems obvious

Yes, that's definitely a good strategy - we lose no data.

> but what should happen with non-empty values?

I'd use a notion of "more important" vs "less important" model instance. Values from the "more important" object would always overwrite non-empty values in the "less important" object.

This solution, however, limits us to only 2 objects engaged in the merge.

> Should I provide toggles in the interface such that the user can pick which is the correct value?

I think it would be too much effort.

> If so that would also involve rewriting merge_model_objects

Yes, even with my proposed solution ("most/less important"). And I think  that we'd have to have a new view with radio buttons for user to select a more important person.

@neon-ninja, @wking can you perhaps think of any other solution?

wking:
On Wed, Apr 01, 2015 at 01:34:05PM -0700, Piotr Banaszkiewicz wrote:
> 1. better UI
> …
> This solution, however, limits us to only 2 objects engaged in the merge.
> …
> Yes, even with my proposed solution ("most/less important"). And I
> think that we'd have to have a new view with radio buttons for user
> to select a more important person.

I doubt we have cases where we really care all that much about which
is the most important.  Personally, I'm happy switching on GitHub ID
(since that will be the eventual user login mechanism, see #70).
Conflicting GitHub IDs would cause merge attempts to error out.  If no
entry defined a GitHub ID or only one entry did, then we just squash
everything down preferring the oldest entry (measured by lowest
Person.id), since that's the most likely to have received previous
editing.

> 2. more tests against `util.merge_model_objects`.

It's always good to have more tests ;).

I'm still skeptical that the auto-merge logic is worth the trouble
[1].  But it's not my maintenance burden, so if @pbanaszkiewicz is
comfortable I'm fine moving ahead with this.

[1]: https://github.com/swcarpentry/amy/pull/229#issuecomment-79168046

neon-ninja:
>I'd use a notion of "more important" vs "less important" model instance. Values from the "more important" object would always overwrite non-empty values in the "less important" object.

>preferring the oldest entry (measured by lowest
Person.id)

Note that is how merge_model_objects operates currently - it takes a more important primary_object and merges in a list of alias_objects. As this is currently, the lower id will be the primary_object as Person.objects.filter implicitly returns in order of id.

wking:
On Wed, Apr 01, 2015 at 03:39:23PM -0700, Nick Young wrote:
> Note that is how merge_model_objects operates currently - it takes a
> more important primary_object and merges in a list of
> alias_objects. As this is currently, the lower id will be the
> primary_object as Person.objects.filter implicitly returns in order
> of id.

But it's missing an explicit radio (to make Piotr happy) or a GitHub
ID check (to make me happy).  As far as I understand anyway, since I
haven't actually looked at the code recently.

neon-ninja:
Merges now fail if there's conflicting github ids and I've added a radio on the confirm screen to select which object is primary. Also updated my test to test for foreign key merge. To summarise the tests:

test_merge_duplicate_persons tests that a merge successfully overwrites empty values with non-empty values, foreign keys are moved across (using a Badge), and the primary's non-empty value persists in the case of 2 non-empty values.

test_merge_fails_when_fields_not_set checks that a merge will fail if the user doesn't select two matching entries

test_merge_fails_when_github_mismatch checks that a merge will fail when two matching entries have nonmatching non-empty github ids

Is there anything else I should test for?

pbanaszkiewicz:
Hi @neon-ninja and @wking,

you're both right, I forgot about this simple assumption: the newest data for the same person is probably the correct one. Based on this assumption we can make the newest `Person` object (ie. having biggest ID) "more important" (using my logic). However…

@wking:
> Conflicting GitHub IDs would cause merge attempts to error out.

Right now it's impossible to have more than one Person with the same GitHub ID (except for empty values) because `Person.github` is unique.

I think we might have chosen a bad direction here. I think we should handle duplicates in the bulk upload. Or loosen up `Person.github` restriction, but I'm not convinced about that solution.

wking:
On Thu, Apr 02, 2015 at 01:33:33PM -0700, Piotr Banaszkiewicz wrote:
> you're both right, I forgot about this simple assumption: the newest
> data for the same person is probably the correct one. Based on this
> assumption we can make the newest `Person` object (ie. having
> biggest ID) "more important" (using my logic). However…

Heh, I was assuming that the *oldest* information was probaby correct
;) [1].

> > Conflicting GitHub IDs would cause merge attempts to error out.
> 
> Right now it's impossible to have more than one Person with the same
> GitHub ID (except for empty values) because `Person.github` is
> unique.

Good point.  So we don't need special handling in a post-upload merge
handler.

> I think we might have chosen a bad direction here. I think we should
> handle duplicates in the bulk upload. Or loosen up `Person.github`
> restriction, but I'm not convinced about that solution.

Regardless of how well you clean bulk-upload data, you're still going
to get duplicate Person entries in the live database.  Maybe not so
many that you need a dedicated tool to merge them?  I don't know.
Anyhow, I'd rather we slept on this until we get per-Person logins
(#70).  Do we really care about duplicates enough to have admins spend
time de-duping?

I think we should it to users to de-dup themselves.  If the Person has
a GitHub ID, authorizing the user to edit the user-editable portions
of the entry is easy.  Where a GitHub ID is missing, they'll either
need to contact an admin to get one assigned, and in that case, the
admin might need trigger a merge (between the entry without a GitHub
ID and the entry that already uses the appropriate ID).  But in this
case, all the merge logic has to do is update any foreign keys and
delete the GitHub-ID-less entry.  We can leave it to the user to
(re)populate the other data as they see fit.

[1]: https://github.com/swcarpentry/amy/pull/229#issuecomment-88627651

pbanaszkiewicz:
> Anyhow, I'd rather we slept on this until we get per-Person logins (#70).

I'll hold on with my thoughts on this issue till tomorrow. Need to rethink some stuff.

pbanaszkiewicz:
In #193 Greg described a case of merging two duplicate records that exist in the database.

In that case the current solution would be exactly what we need (and I prefer this solution).

However, we might as well detect duplicates during bulk-upload. This would mean we'd have to make bulk upload even more complicated than it is right now. Additionally, we don't provide GitHub IDs during bulk-upload, which means 1) many people won't have it assigned, 2) we should toss out GH ID checking from the de-duplication code.

I'd stick to the current solution, but with some important changes:

1. toss out GH checking and rely only on names
2. separate tests for the view and for underlying function `merge_model_objects`
3. provide excellent coverage for `merge_model_objects`

These issues are yet to be fixed:

1. proper breadcrumbs
2. clear UI indicator for the primary merge object

neon-ninja:
@pbanaszkiewicz is https://github.com/neon-ninja/amy/commit/b812c8db8e48ebf4e365a90c97b8d28365e3f936 what you meant by >proper breadcrumbs ?

>clear UI indicator for the primary merge object

primary

^ note there is now a radio button to indicate the primary - is this not sufficiently clear?

neon-ninja:
Perhaps we should add some help text describing what the primary object is? Something like:

"The primary object is the main object during the merge - if the primary object and the duplicate objects have conflicting values for a field, the value for the primary object is the one that will persist in the newly merged object"

gvwilson:
Is this one going to be incorporated, or has it gone stale?

pbanaszkiewicz:
I'd leave it for discussion until we figure out what do we want to do and how do we want to do it. I'm not sold on current implementation and I'd like more feedback from admins regarding merging.

We can have discussion in #193.

PR #230, opened 2015-03-15
basestring does not exist in Python 3. use str instead.

neon-ninja:


wking:
On Sun, Mar 15, 2015 at 03:54:21PM -0700, Nick Young wrote:
>   * basestring does not exist in Python 3. use str instead.

I think this is a duplicate of #220.  See that issue for notes about
where check.py comes from and why I don't think we want amy-side
tweaks.

neon-ninja:
I'd agree this should be refactored out into an external package on pypi. Should I take care of that and then submit a pull request with the change to requirements.txt? For reference, the view that this is currently breaking is validate_event.

wking:
On Sun, Mar 15, 2015 at 06:34:53PM -0700, Nick Young wrote:
> I'd agree this should be refactored out into an external package on
> pypi. Should I take care of that and then submit a pull request with
> the change to requirements.txt? For reference, the view that this is
> currently breaking is validate_event.

I think the first step is fixing the script in it's upstream
swcarpentry/workshop-template location to work with both Python 2 and
3.  We should all agree that that's a good thing ;).

Personally, I'd also support anyone who wanted to split the script
into it's own package and push it to PyPI, but I'm not sure where the
swcarpentry/workshop-template maintainers stand on that issue.  It
can't hurt to ask them though :).

neon-ninja:
Upstream is already Python 2/3 compatible, it's just that amy's copy of check.py is outdated

wking:
On Sun, Mar 15, 2015 at 08:57:56PM -0700, Nick Young wrote:
> Upstream is already Python 2/3 compatible, it's just that amy's copy
> of check.py is outdated

So then go the PyPI route if the swcarpentry/workshop-template folks
sign off on it.  Otherwise, just copy the new version from
swcarpentry/workshop-template into this repository?  Maybe that's what
you are doing?  For bonus points on the copy, cherry-pick the commits
from swcarpentry/workshop-template to preserve authorship information
and the motivation behind the changes.

neon-ninja:
@gvwilson thoughts?

gvwilson:
Please just copy in the new version for now - we need to tackle our repository proliferation + packaging problem soon, but we're not there yet.

neon-ninja:
Closing and reopening at #232

PR #232, opened 2015-03-18
Pull latest version of check.py from https://github.com/swcarpentry/workshop-template/blob/gh-pages/tools/check.py

neon-ninja:
This fixes errors relating to the missing basestring in Python 3

### DIFF
--- a/workshops/check.py
+++ b/workshops/check.py
@@ -1,22 +1,8 @@
 #!/usr/bin/env python
 
 '''Check that index.html is valid and print out warnings and errors
-when the header is malformed.
-
-Checks for:
-1.  There should be the right number of categories
-2.  Categories are allowed to appear only once
-3.  Contact email should be valid (letters + @ + letters + . + letters)
-4.  Address and venue should be non-empty
-5.  Latitute/longitude should be 2 floating point numbers separated by comma
-6.  Start date should be a valid date; if end date is present, it should be valid as well
-7.  Human date should have three-letter month and four-letter year
-8.  Human time should have 'am' or 'pm' or both
-9.  Country should be a recognized hyphenated country name from the embedded list
-10. Instructor and helper lists should be valid lists
-11. Template header should not exist
-12. Layout should be 'workshop'
-13. Root must be '.'
+when the header is malformed.  See the docstrings on the checking
+functions for a summary of the checks.
 '''
 
 from __future__ import print_function
@@ -24,9 +10,13 @@ import sys
 import os
 import re
 import logging
-
 import yaml
 from collections import Counter
+
+try:  # Hack to make codebase compatible with python 2 and 3
+    basestring
+except NameError:
+    basestring = str
 
 __version__ = '0.6'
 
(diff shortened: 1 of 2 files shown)


workshops/check.py
```diff
@@ -1,22 +1,8 @@
...
-4.  Address and venue should be non-empty
-5.  Latitute/longitude should be 2 floating point numbers separated by comma
-6.  Start date should be a valid date; if end date is present, it should be valid as well
-7.  Human date should have three-letter month and four-letter year
-8.  Human time should have 'am' or 'pm' or both
-9.  Country should be a recognized hyphenated country name from the embedded list
-10. Instructor and helper lists should be valid lists
-11. Template header should not exist
-12. Layout should be 'workshop'
-13. Root must be '.'
```
gvwilson: Whoa - what happened with all the deletes?

workshops/check.py
```diff
@@ -24,10 +10,14 @@
...
 import re
 import logging
-
 import yaml
 from collections import Counter
 
+try:  # Hack to make codebase compatible with python 2 and 3
+    basestring
+except NameError:
+    basestring = str
```
gvwilson: What I'm used to doing is:

    try:
        our_basestring = basestring
    except NameError:
        our_basestring = str

and then using `our_basestring` everywhere, rather than giving readers the possibly-mistaken impression that `basestring` in our code is actually what they think of as `basestring`.  It's a minor (really minor) nit.

workshops/check.py
```diff
@@ -314,6 +357,16 @@ def check_validity(data, function, errors, error_msg):
     return valid
 
 
+def check_blank_lines(raw_data, errors, error_msg):
+    '''Check for blank line in category headers.'''
```
gvwilson: '''Blank lines are not allowed in category headers.''' <<== will be clearer to non-native English speakers.

workshops/check.py
```diff
@@ -324,82 +377,57 @@ def check_categories(left, right, errors, error_msg):
...
-
-    return True
+    # YAML header must be right at the start of the file.
+    if not text.startswith('---'):
+        return None, None
 
+    # YAML header must start and end with '---'
+    pieces = text.split('---')
+    if len(pieces) < 2:
+        return None, None
```
gvwilson: We've been tripped by this before:

    ---
    key: value
      ---
    main body

where the closing triple dash is indented - it looks valid according to the `text.split` test above, but fails YAML parsing.

gvwilson:
Nice work - thanks very much.

neon-ninja:
No problem. Regarding your comments on the diff - should those changes be made upstream?

gvwilson:
Yes please and thank you.

workshops/check.py
neon-ninja: The other alternative would be to use six

from six import string_types
isinstance(s, string_types)

workshops/check.py
neon-ninja: raw = pieces[1].strip()
return raw, yaml.load(raw)

means that "key: value\n   " will be converted to "key:value", which means that those leading spaces won't be seen by yaml.load.

neon-ninja: Although, shouldn't this be
if len(pieces) < 3
as we're expecting 3 elements - leading empty string, the YAML content, then following HTML

pbanaszkiewicz:
Summing up: this PR contains `check.py` pulled from https://github.com/swcarpentry/workshop-template/ and one-line change in `views.py` for compatibility with this new `check.py`.

This change looks good and works good.

However, I did not see any PR for upstream. I created an issue (https://github.com/swcarpentry/workshop-template/issues/190) so that we don't forget about Greg's comments here.

I think we can merge and close this PR.

wking:
On Tue, Mar 31, 2015 at 10:51:15AM -0700, Piotr Banaszkiewicz wrote:
> I think we can merge and close this PR.

Works for me.  Has @gvwilson authorized anyone to merge PRs (or just
this PR) on his behalf?

gvwilson:
Happy to have others merge PRs.

pbanaszkiewicz:
@gvwilson @wking @neon-ninja great!

wking:
On Tue, Mar 31, 2015 at 11:07:04AM -0700, Greg Wilson wrote:
> Happy to have others merge PRs.

Ok.  @pbanaszkiewicz, do you want to handle this?  I think there are a
number of PRs that were just blocking on a maintainer's approval.  If
you're not sure if a given PR is ready to go, feel free to ping me for
my opinion.  If you're too busy, I'm happy to lead the merging effort
too (but I get the feeling you have more time available ;).

pbanaszkiewicz:
Hey @wking,

> Ok. @pbanaszkiewicz, do you want to handle this? I think there are a number of PRs that were just blocking on a maintainer's approval. If you're not sure if a given PR is ready to go, feel free to ping me for my opinion. If you're too busy, I'm happy to lead the merging effort too (but I get the feeling you have more time available ;).

Yeah! There's Easter coming so I should have more free time to move on with our remaining PRs. I will make sure to let you know if I have doubts.

Issue #16, opened 2014-12-02
Allow hosts, instructors, and helpers to be selected on event details page


gvwilson:
The event details page should (only) allow instructors and helpers to be added by selection from a list of people who are in the database.  It should be possible to add hosts this way as well.  The wiki page for the event can be used for notes about people who aren't in the database.

pbanaszkiewicz:
Solved in #234.

PR #234, opened 2015-03-31
Ability to add and remove tasks from the event edit form

neon-ninja:
Solves #16

amy task edit delete


workshops/templates/base_without_navbar.html
```diff
@@ -4,9 +4,12 @@
   <head>
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css" />
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap-theme.min.css" />
+    <link rel="stylesheet" href="https://code.jquery.com/ui/1.11.4/themes/smoothness/jquery-ui.css">
```
pbanaszkiewicz: There's one char missing at the end: `/`. The line should look like `<link rel="stylesheet" href="..." />`, otherwise it makes a lot of noise in the tests.

workshops/views.py
```diff
@@ -441,11 +442,33 @@ class EventCreate(LoginRequiredMixin, CreateViewContext):
     model = Event
     fields = '__all__'
 
+@login_required
+def event_edit(request, event_ident):
+    event_form_factory = modelformset_factory(Event, fields='__all__', extra=0)
+    task_form_factory = inlineformset_factory(Event, Task, fields='__all__', extra=1, can_delete=False)
```
pbanaszkiewicz: I'm not certain about these two form factories. Wouldn't it be better to have ModelForms explicitely implemented in `workshops.forms`?

workshops/views.py
```diff
@@ -474,6 +497,13 @@ def task_details(request, task_id):
     return render(request, 'workshops/task.html', context)
 
 
+@login_required
+def task_delete(request, task_id):
```
pbanaszkiewicz: Please add a docstring explaining where this view is used.

wking: On Wed, Apr 01, 2015 at 02:00:24PM -0700, Piotr Banaszkiewicz wrote:
> > +@login_required
> > +def task_delete(request, task_id):
> 
> Please add a docstring explaining where this view is used.

I'm -1 on that.  Docstring's should explain what a function (or
whatever) does, not who's using it.  It's better to pick distinct
names and use:

  $ git grep task_delete origin/pr/234

or similar to figure out who's using the function.  It's also good to
motivate odd stuff in the commit message (but this example seems
obvious enough to not need that).  And it would be good if we had some
user-facing workflow documentation, but we don't have anything along
those lines yet.

pbanaszkiewicz:
Hey @neon-ninja,

thanks for your PR!

I found a few very minor issues in the code.

One note regarding UI/UX: when I tried this out without looking at the code, I thought it uses AJAX and would add a new task without reloading the page.

Now, after reading the code, I know that a new task is created and the event is saved, and then a redirect happens.

I think it would be better to have two submit buttons ("Submit" and "Add"). Upon clicking "Add", a new task would be created, but the event form data would be saved and restored after the redirect to the same page.
Upon clicking "Submit" only changes to the event would be saved (then redirect to `event_details` view).

What do you think about this idea, Nick?

workshops/views.py
pbanaszkiewicz: A usual CRUD would redirect to the list of objects of the same kind after deleting one object from that list.

> Docstring's should explain what a function (or whatever) does, not who's using it.

In case of deleting a task, that function redirects to the task's event. The redirect is there to help with inline edits from the event's page. I think this is unusual enough to at least deserve a comment. This redirect is also a part of function's behavior, so it may just as well be put in the docstring.

> It's better to pick distinct names

Personally I'd use a different name and URL for that view - or jump to "RESTful" fashion and return a different status code instead of the redirect.

wking: On Wed, Apr 01, 2015 at 02:27:23PM -0700, Piotr Banaszkiewicz wrote:
> > Docstring's should explain what a function (or whatever) does, not
> > who's using it.
> 
> In case of deleting a task, that function redirects to the task's
> event. The redirect is there to help with inline edits from the
> event's page. I think this is unusual enough to at least deserve a
> comment. This redirect is also a part of function's behavior, so it
> may just as well be put in the docstring.

Ah, I agree that it's good to explain any surprising redirects
returned by the function it that function's docstring.  When you said
“where this view is used”, I thought you meant the URL the view was
bound to and any views linking to that URL, not where you'd go after
visiting this view.

workshops/views.py
neon-ninja: Could do. The one line solution just seemed more elegant to me. Particularly inlineformset seemed to be the only way to generate a model form that uses foreign keys automatically referring to the event on the page.

neon-ninja:
Sounds good to me :)

pbanaszkiewicz:
Hi @neon-ninja

Thanks for your commits!

The latest flow in the event edit view is not exactly what I imagined in https://github.com/swcarpentry/amy/pull/234#issuecomment-88632603

I think there should be two buttons, for example "Save" and "Add". The former should be placed right below the event edit form, while the latter should be where it is right now - below the tasks table.

Upon clicking on "Save": Amy should attempt to save the event form, and remember the contents of the task form (without saving them to the database).

Upon clicking on "Add": Amy should attempt to add a new task, and remember the contents of the event form (without saving them to the database).

I hope my vision is now clearer :)

Cheers!

workshops/templates/workshops/event_form.html
```diff
@@ -3,21 +3,44 @@
...
 {% block breadcrumbs %}
     {% breadcrumb_main_page %}
-    {% breadcrumb_index_all_objects model %}
-    {% if object %}
-        {% breadcrumb_object object %}
-        {% breadcrumb_edit_object object %}
-    {% else %}
-        {% breadcrumb_new_object model %}
-    {% endif %}
+    {% breadcrumb_active title %}
```
pbanaszkiewicz: Why did you remove the breadcrumbs? Was there an error or something?

neon-ninja: yes, as I wasn't passing in model or object. I can, just for the sake of the breadcrumbs.

workshops/templates/workshops/event_form.html
```diff
@@ -20,4 +20,4 @@
     <input type="submit" value="Submit" />
 </form>
 
-{% endblock %}
+{% endblock %}
```
pbanaszkiewicz: Please bring back the last line of this file.

pbanaszkiewicz:
Okay, this looks nice.

Apart from [the missing last line](https://github.com/swcarpentry/amy/pull/234#discussion-diff-28274049) I have only one comment regarding redirects:

1. After clicking "Save": redirect to the event's detail view
2. After clicking "Add": leave as it is right now (ie. "refresh" the page)
3. After clicking "Delete": change to "refresh" (ie. redirect to the event edit view)

@wking do you have any comments regarding this PR? I might have missed something.

wking:
On Mon, Apr 13, 2015 at 01:17:52PM -0700, Piotr Banaszkiewicz wrote:
> … I have only one comment regarding redirects:

Staying on the event-edit page after adding/editing/deleting tasks
makes sense to me.

> @wking do you have any comments regarding this PR?

No, it looks good to me, and your comments seem reasonable.

neon-ninja:
@pbanaszkiewicz see above commits :)

pbanaszkiewicz:
Hey @neon-ninja,

I took your work and made some minor changes:

1. rebase against latest master
2. change formsets to "normal" forms
3. fix errors caused by 1 & 2

I'm going to merge everything now. Thanks a lot for your contributions! :-) Very appreciated. And sorry it took me so long to review and merge this…

pbanaszkiewicz:
Hmmm, this should've been resolved by GitHub automatically. I'm not sure why it didn't.

narayanaditya95:
@pbanaszkiewicz I am fixing it.

pbanaszkiewicz:
My guess is that because I rebased this PR against `master` and did not forcibly update the PR, it shows as unmerged, even though @neon-ninja's commits show in `master`.

So I'm closing this PR. Thanks again @neon-ninja for your contribution, and sorry everyone for the chaos I made due to silly rebasing.

PR #238, opened 2015-04-03
Fix User Edit Form bugs #235

### LINKED ISSUE #235: User edit form bugs
1. When editing user with improper data (for example a non-unique email) the title and header of the page change even though the person is not saved to the database.

2. The password field is required. We should probably make a separate form for editing user's password.

narayanaditya95:
This PR aims to fix bullet 1 of [#235](https://github.com/swcarpentry/amy/issues/235). Please review the PR. I am not sure if this fix is the most elegent way to achieve it.

### DIFF
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -80,7 +80,7 @@ class UpdateViewContext(UpdateView):
 
         # self.object is available in UpdateView as the object being currently
         # edited
-        context['title'] = str(self.object)
+        context['title'] = str(self.model.objects.get(pk=self.object.pk))
         return context
 
 


wking:
On Fri, Apr 03, 2015 at 03:24:24PM -0700, Aditya Narayan wrote:
> I am not sure if this fix is the most elegent way to achieve it.

Yeah, hitting the database to revert the unsaved changes doesn't seem
right, but I can't think of a better way either.  Personally, I don't
think setting the title for the edit-page is worth an extra database
query, so I'd just drop this and go back to generic, model-level
titles (or just have the title be ‘amy’).  If we want to keep the
entry-specific titles, this change is probably fine.

narayanaditya95:
@wking Yes, I agree with you. I am against generating database query for changing the title. I would go for generic titles, unless there is another way to achieve the same without hitting the database.

wking:
On Fri, Apr 03, 2015 at 03:50:50PM -0700, Aditya Narayan wrote:
> I would go for generic titles, unless there is another way to
> achieve the same without hitting the database.

I'm sure there is some way to cache the original values and then
restore the object if saving it fails, but I don't think per-object
titles matter enough to make it worth figuring out ;).

gvwilson:
Are we expecting to merge this, or has it gone stale?

pbanaszkiewicz:
I'll probably `wontfix` #235, but need some time to think about that issue. Should be this week.

Issue #237, opened 2015-04-03
Django 1.8 changes


pbanaszkiewicz:
Hi,

Django 1.8 is out and contains some backward-incompatible changes, for example new stuff for [templates](https://docs.djangoproject.com/en/1.8/ref/templates/upgrading/).

We should check what's working and what has stopped working in Amy with Django 1.8.

We should also decide if we keep Amy compatible with Django 1.7, or Django 1.8, or both.

wking:
On Fri, Apr 03, 2015 at 12:56:44PM -0700, Piotr Banaszkiewicz wrote:
> Django 1.8 is out and contains some backward-incompatible changes…

Previous Django 1.8 discussion in [1,2,3,4].

> We should check what's working and what has stopped working in Amy
> with Django 1.8.

We don't use all that much Django, so we should be able to handle this
by:

* Reading the 1.8 release notes [5] and updating amy accordingly.
* Running the test suite after bumping Django.

> We should also decide if we keep Amy compatible with Django 1.7, or
> Django 1.8, or both.

I'm fine bumping to 1.8 and dropping 1.7 support, but we may want
input from our sysadmin first.  @drio (I think your still the amy
sysadmin ;), do you have any concerns about bumping the stack?  If
you've installed Django using pip I don't foresee any problems
shifting to 1.8.

[1]: https://github.com/swcarpentry/amy/issues/158#issuecomment-71443616
[2]: https://github.com/swcarpentry/amy/pull/221#issuecomment-77890801
[3]: https://github.com/swcarpentry/amy/pull/221#issuecomment-77902272
[4]: https://github.com/swcarpentry/amy/pull/221#issuecomment-77903376
[5]: https://docs.djangoproject.com/en/1.8/releases/1.8/

pbanaszkiewicz:
Issues I found after running the test suite:

1. We need to run migrations (`makemigrations` and then `migrate`) because of https://docs.djangoproject.com/en/1.8/releases/1.8/#abstractuser-last-login-allows-null-values
2. We need to do something about our forms: #240 

After fixing 1. and running the test suite I only had some issues with the forms. I did not check them out so they may or may not be connected to #240.

PR #242, opened 2015-04-06
`None` values in forms + introduction of crispy-forms

### LINKED ISSUE #240: NULL values in forms displayed as text "None"
Hi,

as part of my testing for #237 I discovered that our forms display NULL values as a string "None":

screen shot 2015-04-04 at 17 24 24

(This is a screenshot taken from one of our tests' output - the test fails.)

I didn't find an easy solution to this issue. The only one I found was to display form fields as such:

```html
<input type="text" name="field" id="id_field" value="{{ form.field.value|default:"" }}" />
```

…which would require us to rewrite forms to HTML because we use `{{ form.as_p }}` in templates.

By the way, our forms could benefit from displaying them with Bootstrap markup, so we may hire [django-bootstrap-form](https://github.com/tzangms/django-bootstrap-form) or [django-crispy-forms](https://github.com/maraujop/django-crispy-forms/).

pbanaszkiewicz:
Resolves #240.

See discussion in #240 and #237.

### DIFF
--- a/workshops/templates/workshops/event_edit_form.html
+++ b/workshops/templates/workshops/event_edit_form.html
@@ -44,7 +44,6 @@
     {% else %}
     <p>No tasks.</p>
     {% endif %}
-
     {{ task_form.as_p }}
     <button type="submit" class="btn btn-primary" name="add">Add</button>
 </form>
--- a/workshops/templates/workshops/event_edit_form.html
+++ b/workshops/templates/workshops/event_edit_form.html
@@ -12,6 +12,8 @@
     {% endif %}
 {% endblock %}
 
+{% load crispy_forms_tags %}
+
 {% block content %}
 
 <script type="text/javascript">
@@ -20,10 +22,8 @@
     });
 </script>
 
-<form action="" method="post" id="event_edit_form" role="form">
-    {% csrf_token %}
-    {{ event_form.as_p }}
-    <button type="submit" class="btn btn-success" name="save">Save</button>
+<form action="" method="post" id="event_task_form" role="form" class="form-horizontal">
+    {% crispy event_form form_helper %}
 
     <h2 id="associated-tasks">Associated tasks</h2>
     {% if tasks %}
@@ -44,6 +44,7 @@
     {% else %}
     <p>No tasks.</p>
     {% endif %}
+
     {{ task_form.as_p }}
     <button type="submit" class="btn btn-primary" name="add">Add</button>
 </form>
--- a/workshops/templates/workshops/instructors.html
+++ b/workshops/templates/workshops/instructors.html
@@ -6,18 +6,13 @@
     {% breadcrumb_active title %}
 {% endblock %}
 
+{% load crispy_forms_tags %}
 {% block content %}
-<form action="{% url 'instructors' %}" method="post">
-    {% csrf_token %}
-    <table class="table table-striped">
-    {{ form.as_table }}
-    </table>
-    <input type="submit" value="Submit" />
-</form>
+    {% crispy form form_helper %}
 {% if persons == None %}
 {% elif persons %}
     <table class="table table-striped">
-        <tr>
+    <tr>
 	    <th>taught</th>
 	    <th>airport</th>
 	    <th>personal</th>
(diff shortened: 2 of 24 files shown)


pbanaszkiewicz:
@wking, @neon-ninja and @shapiromatron - this is ready for review. Can anyone of you provide some feedback? Thanks

workshops/templates/workshops/event_edit_form.html
```diff
@@ -44,6 +44,7 @@ <h2 id="associated-tasks">Associated tasks</h2>
     {% else %}
     <p>No tasks.</p>
     {% endif %}
+
```
wking: Drive-by whitespace change ;).

workshops/templates/workshops/instructors.html
```diff
@@ -6,18 +6,13 @@
...
-    {{ form.as_table }}
-    </table>
-    <input type="submit" value="Submit" />
-</form>
+    {% crispy form form_helper %}
 {% if persons == None %}
 {% elif persons %}
     <table class="table table-striped">
-        <tr>
+    <tr>
```
wking: Drive-by whitespace change.  You want the `<tr>` indented beyond it's wrapping `<table>`, no?

And personally, I'd use `<thead>`, `<tbody>`, etc., but that's beyond the scope of this PR.

workshops/templates/workshops/instructors.html
```diff
@@ -27,7 +22,7 @@
 	    <th></th>
 	</tr>
     {% for p in persons %}
-        <tr class="instructor_row" id="instructor_{{forloop.counter0}}">
+    <tr class="instructor_row" id="instructor_{{forloop.counter0}}">
```
wking: Drive-by whitespace change.  See previous comment.

workshops/views.py
```diff
@@ -388,7 +391,10 @@ class PersonCreate(LoginRequiredMixin, CreateViewContext):
 
 class PersonUpdate(LoginRequiredMixin, UpdateViewContext):
     model = Person
-    fields = PERSON_FIELDS
+    # don't display the 'password' field, reorder fields
+    fields = ["personal", "middle", "family", "username", "may_contact",
+              "email", "gender", "airport", "github", "twitter", "url",
+              "is_superuser", "user_permissions"]
```
wking: PEP 8 doesn't suggest single-vs.-double quotes, but[ does suggest consistency][1].  I personally prefer single quotes, since they have fewer distracting pixels, but other folks [seem][2] to [prefer][3] single quotes for internal strings (read by the software and developers) and double quotes for stuff that could be shown to users.  Obviously not a critical issue, but I think it's good to have some sort of policy to keep things stable and consistent.

[1]: https://www.python.org/dev/peps/pep-0008/#string-quotes
[2]: http://programmers.stackexchange.com/questions/155176/single-quotes-vs-double-quotes
[3]: http://stackoverflow.com/questions/56011/single-quotes-vs-double-quotes-in-python

wking:
On Sat, Apr 18, 2015 at 10:14:56AM -0700, Piotr Banaszkiewicz wrote:
> Can anyone of you provide some feedback?

I just posted some minor style suggestions as inline comments.  Other
than those, this looks good to me (and I'd be fine if this was merged
as it stands).  I ran through some brief local testing things seem to
mostly work.  I had some trouble with our null/blank/unique fields
though.  For example, clearing a person's email gave me:

  Person with this Email already exists.

so you may need some form-side handling for that case to None the
values besides the stuff we have in Person.save().  And I expect well
want similar stuff for other models with null/blank/unique fields.

Hmm.  I just checked with the current master branch (74447f21), and
I'm seeing the same errors, so maybe this issue is independent of the
PR.  I think the solution will require some work on form validation
though.

Anyway, the tests all pass and the forms look much prettier, so I'm
fine with merging this and working out the null/blank/unique business
later.

pbanaszkiewicz:
@wking 

thanks a lot for your feedback. I think this PR should fix behavior on saving empty values. I'll work on that as well as on things you mentioned in the inline comments. Thanks again.

pbanaszkiewicz:
Hey @wking,

> I had some trouble with our null/blank/unique fields
though.  For example, clearing a person's email gave me:
>
>>  Person with this Email already exists.

I can't reproduce this behavior. I cleared emails for two people and in both cases it saved `NULL` to the database (no errors occurred).

I was not able to reproduce this behavior with other unique, null and blank fields (like `Person.github`, `Person.twitter` or `Event.url`).

Can you check again if these errors occur?

wking:
On Sat, Apr 18, 2015 at 02:15:09PM -0700, Piotr Banaszkiewicz wrote:
> > I had some trouble with our null/blank/unique fields though.  For
> > example, clearing a person's email gave me:
> >
> >   Person with this Email already exists.
> 
> I can't reproduce this behavior

Ah, I bet my local database already has someone with the empty string
as their email value.  So I have:

1. Person submits form with empty-string values
2. Django validates the form.
3. Empty string collides with existing database value, violating
   uniqueness.

While folks with databases that don't have existing empty-strings in
those fields get:

1. Person submits form with empty-string values
2. Django validates the form.
3. No empty strings in the database, so form validator is happy.
4. Django tries to save the model.
5. Person.save() (or other *.save()) replaces empty strings with None.
6. Model saved to the database with NULLs.

So that all makes sense and means this is working as it should (I just
need to rebuild my local database).  Merge away!

PR #243, opened 2015-04-06
Upgrade to Django 1.8

### LINKED ISSUE #237: Django 1.8 changes

pbanaszkiewicz:
Resolves #237. (See also discussion in #240.)

This should be merged after #242, but will probably cause conflicts in `settings.py`.

### DIFF
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -12,8 +12,6 @@ https://docs.djangoproject.com/en/1.7/ref/settings/
 import os
 import json
 
-from django.conf import global_settings
-
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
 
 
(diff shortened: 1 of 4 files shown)


pbanaszkiewicz:
We can now upgrade to Django 1.8, so @wking can you please check this PR?

wking:
On Sun, Apr 19, 2015 at 12:05:44PM -0700, Piotr Banaszkiewicz wrote:
> We can now upgrade to Django 1.8, so @wking can you please check this PR?

Sorry for the delay.  This looks good to me, and works in my local UI
testing.  I'm getting some errors when running the test suite though:

  $ make test
  …
  ======================================================================
  FAIL: test_search_for_site_with_no_matches (workshops.test.test_search.TestSearchSite)
    File "/…/amy/workshops/test/test_search.py", line 17, in test_search_for_site_with_no_matches
      doc = self._check_status_code_and_parse(response, 200)
    File "/…/amy/workshops/test/base.py", line 208, in _check_status_code_and_parse
      doc = self._parse(response=response)
    File "/…/amy/workshops/test/base.py", line 188, in _parse
      for h in hits])
  AssertionError: "" found in HTML page:
  …
  Ran 49 tests in 7.099s

  FAILED (failures=17)
  …

All 17 errors were “AssertionError: "" found in HTML page”.  I haven't
bisected or done any further debugging on that error yet though.

The only thing I don't like in the commits is the hard-coded list of
context processors, where we used to use to just append to the
original TEMPLATE_CONTEXT_PROCESSORS value.  But that's what the
Django folks suggest we do [1], so I'll just live with it ;).

[1]: https://docs.djangoproject.com/en/1.8/ref/templates/upgrading/#the-templates-settings

pbanaszkiewicz:
@wking fixed - the issue was that we were using settings.TEMPLATE_STRING_IF_INVALID in tests, while we should use settings.TEMPLATES[0]["OPTIONS"]["string_if_invalid"].

wking:
On Fri, May 08, 2015 at 12:10:23PM -0700, Piotr Banaszkiewicz wrote:
> @wking fixed - the issue was that we were using
> settings.TEMPLATE_STRING_IF_INVALID in tests, while we should use
> settings.TEMPLATES[0]["OPTIONS"]["string_if_invalid"].

Good stuff.  Local tests pass for me now, so I think we're good to
merge.

I'm getting a handful of RemovedInDjango19Warning messages, but we can
worry about those later ;).

pbanaszkiewicz:
@wking yeah, me too. They're coming from our dependency, django-crispy-forms. I believe it's soon to be fixed (https://github.com/maraujop/django-crispy-forms/commit/2b808e955c74c87b31b8d2d66fb101f35cf15767#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR9).

PR #245, opened 2015-04-19
Introduce Bower

### LINKED ISSUE #241: Use Bower for JS, CSS dependencies
We need something to manage our JavaScript (and perhaps CSS) dependencies and this tool is called [Bower](http://bower.io).

@wking [wrote](https://github.com/swcarpentry/amy/pull/239#issuecomment-89585714) some handy instructions on getting Bower up and running.

Bower packages we'd use for now:

* `bootstrap`
* `bootstrap-datepicker`

(This should land after #239)

pbanaszkiewicz:
This resolves #241.

We have 4 dependencies managed by [Bower](http://bower.io) now:

* bootstrap
* bootstrap-datepicker
* jquery
* jquery-ui

What's not managed by Bower:

* amy's main CSS file (`workshops/static/amy.css`)
* combobox JS file (`workshops/static/combobox.min.js`) - this is because the `jquery-ui.combobox` hanged my browser

Introduction of Bower requires a new step in the setup pipeline, as well as a new `Makefile` command.

As we want to deploy Amy soon, this PR also includes some minor tweaks to the management of static files.


pbanaszkiewicz:
@wking, @neon-ninja and @shapiromatron - this is ready for review. Can you provide some feedback? Thanks

pbanaszkiewicz:
Bump.

wking:
On Fri, Apr 24, 2015 at 02:28:34PM -0700, Piotr Banaszkiewicz wrote:
> Bump.

Sorry :/, I should have time to get to this tomorrow.  Work, work,
work :p.

pbanaszkiewicz:
> Sorry :/, I should have time to get to this tomorrow.  Work, work, work :p.

That's alright, we are all volunteers here ;)

wking:
This looks pretty good to me.  I left suggestions for a few minor
tweaks inline.  My main concern is the combobox vendoring, but I'm ok
punting on that if we don't have time to get to the bottom of it now.

pbanaszkiewicz:
Hey @wking,

thanks for the feedback. I changed almost everything according to your suggestions, except for:

* I didn't move `amy.css` around after all
* I don't plan to check why `jquery-ui.combobox` hung my browser because we'll soon have #248 
* I didn't add `bower_components` as a prerequisite to `make test` because we don't test JS or CSS

Please give this PR a final glance and if everything's alright I'll merge.

Cheers,
Piotr

wking:
On Sat, Apr 25, 2015 at 04:49:20PM -0700, Piotr Banaszkiewicz wrote:
> * I didn't move `amy.css` around after all
> * I don't plan to check why `jquery-ui.combobox` hung my browser
>   because we'll soon have #248
> * I didn't add `bower_components` as a prerequisite to `make test`
>   because we don't test JS or CSS

These all sound good to me.  I looked through the new commits and
everything looks great, so +1 to merge.

shapiromatron:
Looks good to me; no comments, other than what's the point of adding bower? It's one more barrier to the dev/production stack (node.js), and instead of using CDN we'll be serving the files from the deployed server, so it'll be slower. I don't mean this in a rude way, just general curiousity as to why this is best-practice.

Code implementation looked fine however so plus +1 for merging.

wking:
On Sat, Apr 25, 2015 at 08:12:59PM -0700, Andy Shapiro wrote:
> Looks good to me; no comments, other than what's the point of adding
> bower? It's one more barrier to the dev/production stack (node.js),
> and instead of using CDN we'll be serving the files from the
> deployed server, so it'll be slower. I don't mean this in a rude
> way, just general curiousity as to why this is best-practice.

There's some discussion of this in #105.  The gist is that:

1. We want to be able to work on this while we're offline.
2. Auto-fallback is possible (see 80d15559), but in my opinion more
   complicated than it's worth.

Personally, I don't find a local Node install to be a particularly
large barrier.  There are pre-built installers for major platforms,
and I'd be surprised if there's a major Linux distribution that
doesn't carry it in the main package repository.

I also don't think the first-hit page load speedup you'd get from a
CDN is a big deal for amy, because we're unlikely to be serving *that*
many libraries for client-side processing, and we're serving them to
SWC admins who can probably wait the extra few milliseconds ;).

pbanaszkiewicz:
Hey @wking and @shapiromatron,

thanks a lot for your feedback. I'll merge now.

@shapiromatron additionally to what @wking said, I like these features of using Bower:

* we can manage our front-end dependencies' versions and their dependencies (so it's easier with many deps to run `bower update` than to go to every website of these deps and check if a newer version is available)
* we can hook up assets compression / compilation - something you can't do with public CDN (I plan to do so later) - having only 3 requests per page (1 for HTML, 1 for CSS, 1 for JS) would be desirable.

Regarding your concern about Node: thankfully Amy hosting has it installed already.

Cheers everyone,
Piotr

Issue #233, opened 2015-03-20
Autocomplete form fields


neon-ninja:
Would Amy benefit from installing ajax autocomplete for specific form fields - for example, when selecting a user for a new task.

https://pypi.python.org/pypi/django-ajax-selects
https://github.com/crucialfelix/django-ajax-selects/

wking:
On Thu, Mar 19, 2015 at 08:39:33PM -0700, Nick Young wrote:
> Would Amy benefit from installing ajax autocomplete for specific
> form fields - for example, when selecting a user for a new task.

That sounds reasonable.  How does task-selection work now?  There are
few enough roles that selecting the role with Django's stock select
should work fine, but picking a from an event page or an event from a
person page would probably benefit from this approach.  I'd wait for
more folks to chime in before diving into the implementation though.

gvwilson:
+1 to doing this now for fields with a small number of fixed values 
(tasks, badges) - we can extend to larger things (users, events) later.

shapiromatron:
I've worked a good-bit with [django-selectable](https://github.com/mlavin/django-selectable), and find it works really well and is simple to integrate. Used it extensively with `django-crispy-forms` as well so I don't think there are any conflicts. Very few changes to the html templates; otherwise it's as simple as customizing the widgets used in the forms, and creating "lookups" which define an AJAX request and a queryset. 

Docs: http://django-selectable.readthedocs.org/en/v0.9.X/quick-start.html

I would generally implement in any cases where there are FK or M2M fields with greater than 30 or so options; these are the fields I identified from just scanning the source

- InstructorsForm.airport 
- Person.airport 
- Event.tags, Event.site in admin
- Task in admin (event, person, likely not role as there aren't too many options)

Want me to give it a go?

pbanaszkiewicz:
Hey @shapiromatron!

django-selectable looks really nice and I'd love to see it in Amy. However, I don't want you to start working on it yet, because we have #234 and #242 blocking (and I'd rather you avoided unnecessary rebasing).

By the way, do you perhaps have any links on how to use django-selectable and django-crispy-forms and class based views altogether? I hope we won't have to write a ModelForm for every Create and UpdateView we have.

Cheers,
Piotr

shapiromatron:
Great, I'll wait for you to merge before starting. 

Here's an example from my own project of how to integrate, using the `EffectTag` as an example. My project is overly complicated as I was learning as I went, but you can see the general steps with this example: 

[...]

PR #248, opened 2015-04-19
Autocomplete

shapiromatron:
Implemented #233.

Added autocomplete for foreign-key fields:
- event create [site, organizer]
- event update [site, organizer] 
- person update [airport]
- instructor search [airport]
- task create from event update [person]
- task create [event, person]
- task update [event, person]

I did not add autocomplete here:
- The person create view; I'm having issues with the creation view using django 1.8; it may be good to revisit the single person-create view in a separate issue; there are potential security issues with creating unsalted passwords and setting group-permissions outside an admin view
- The search-box on the main dashboard; this is more difficult to implement since it comes from 3 different model fields. It could probably be done by creating a custom search field, but perhaps that could be an additional issue to be opened since it is more involved than these changes

Since I was changing forms, I made a few extra form edits:
- Removed `workshops/static/combobox.js` since the functionality has now been replaced by this library. 
- Added input-validation in the EventForm to Event.slug to ensure that it has the proper form of "YYYY-MM-*" which is required for Event.get_by_ident method
- Restored the datepickers for the create and update event form


pbanaszkiewicz:
Hi @shapiromatron,

thanks a lot for this contribution! Highly appreciated :)

A quick note:
> I did not add autocomplete here:
>
> * The person create view; I'm having issues with the creation view using django 1.8; it may be good to revisit the single person-create view in a separate issue; there are potential security issues with creating unsalted passwords and setting group-permissions outside an admin view

It's unfortunately not stated in the `requirements.txt`, but Amy is not supposed to work with Django 1.8 *yet*. The PR for upgrade to 1.8 (#243) is in progress, so for now you should downgrade to Django 1.7.

> * The search-box on the main dashboard; this is more difficult to implement since it comes from 3 different model fields. It could probably be done by creating a custom search field, but perhaps that could be an additional issue to be opened since it is more involved than these changes

I don't think we need the autocompletion on the search form.

> * Added input-validation in the EventForm to Event.slug to ensure that it has the proper form of "YYYY-MM-*" which is required for Event.get_by_ident method

I honestly lost track of the requirements for the event slug. Indeed, *published* events have slugs in `YYYY-MM-DD-something` format, but - I think - in case we don't know a formal starting date for an event, the slug looks different.
@wking, @gvwilson: what should the event's slug look like in case we don't know the starting date?

I'll add some inline comments now, but I want to once again thank you for your work.

Cheers,
Piotr

workshops/templates/base_without_navbar.html
```diff
@@ -7,12 +7,12 @@
     <link rel="stylesheet" href="{% static 'css/amy.css' %}" />
     <link rel="stylesheet" href="{% static 'css/bootstrap-datepicker3.min.css' %}" />
     <link rel="stylesheet" href="https://code.jquery.com/ui/1.11.4/themes/smoothness/jquery-ui.css" />
+    {% block extrastyle %}{% endblock %}
```
pbanaszkiewicz: There's an issue. `django-selectable` has some inclusion template tags that don't use valid XHTML, but our tests rely on XML parsing library so… we have to add `workshops/templates/selectable/jquery-css.html` with this content:
```html
<link rel="stylesheet" href="//ajax.googleapis.com/ajax/libs/jqueryui/{{ version }}/themes/{{ theme }}/jquery-ui.css" type="text/css" />
```

The original file is located here: https://github.com/mlavin/django-selectable/blob/master/selectable/templates/selectable/jquery-css.html

workshops/forms.py
```diff
@@ -1,11 +1,14 @@
...
 from django import forms
-from django.forms import HiddenInput
-from django.forms.models import modelform_factory
 
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import Submit
+from selectable import forms as selectable
+
+from workshops.models import Skill, Airport, Event, Task, Person
+from . import lookups
```
pbanaszkiewicz: Please use `from workshops import lookups` for consistency.

workshops/forms.py
```diff
@@ -32,21 +35,24 @@ class BootstrapHelperWithoutForm(BootstrapHelper):
...
+    latitude = forms.FloatField(
+        label='Latitude',
+        min_value=-90.0,
+        max_value=90.0,
+        required=False)
+    longitude = forms.FloatField(
+        label='Longitude',
+        min_value=-180.0,
+        max_value=180.0,
+        required=False)
```
pbanaszkiewicz: Lines 38-51 should not be "repaired" this way: we loose information about the author of the original code in the `git blame`. Please revert these changes. (Possibly via `git rebase --interactive HEAD~5` and 'edit' in proper commit e60174d9.)

workshops/forms.py
```diff
@@ -109,8 +115,77 @@ class DebriefForm(forms.Form):
...
+    def clean_slug(self):
+        # required for Event.get_by_ident
+        data = self.cleaned_data['slug']
+        if data and not re.match(r'^\d{4}-\d{2}-.+$', data):
+            raise forms.ValidationError("Slug must begin with YYYY-MM- syntax")
+        return data
+
+    class Meta:
+        model = Event
+        exclude = tuple()
```
pbanaszkiewicz: Does this have to be explicitly empty tuple? `fields = '__all__'` looks nicer.

workshops/forms.py
```diff
@@ -109,8 +115,77 @@ class DebriefForm(forms.Form):
...
+
+    person = selectable.AutoCompleteSelectField(
+        lookup_class=lookups.PersonLookup,
+        label='Person',
+        required=True,
+    )
+
+    def __init__(self, *args, **kwargs):
+        event = kwargs.pop('event', None)
+        super(TaskForm, self).__init__(*args, **kwargs)
```
pbanaszkiewicz: In Py3 we can use simply: `super().__init__(*args, **kwargs)`.

workshops/forms.py
```diff
@@ -109,8 +115,77 @@ class DebriefForm(forms.Form):
...
+
+    event = selectable.AutoCompleteSelectField(
+        lookup_class=lookups.EventLookup,
+        label='Event',
+        required=True,
+    )
+
+    class Meta:
+        model = Task
+        exclude = tuple()
```
pbanaszkiewicz: https://github.com/swcarpentry/amy/pull/248/files#r28721114

workshops/lookups.py
```diff
@@ -0,0 +1,41 @@
+from selectable.base import ModelLookup
+from selectable.registry import registry
+
+from . import models
```
pbanaszkiewicz: https://github.com/swcarpentry/amy/pull/248/files#r28720279

workshops/models.py
```diff
@@ -296,17 +296,20 @@ def get_by_ident(ident):
...
+        except ValueError:
+            # match event slug in format YYYY-MM-****
+            #(ie. this works for YYYY-MM-DD-**** too)
+            if re.match(r'^\d{4}-\d{2}-.+$', ident):
+                kwargs["slug"] = ident
+            else:
+                raise ObjectDoesNotExist(ident)
+
+        # may throw DoesNotExist exception
+        return Event.objects.get(**kwargs)
```
pbanaszkiewicz: Hmmm, I have mixed feelings about this method.

I don't like that this method may throw two different exceptions (but it's not fault of your code but rather of the original code). But I like the that we have (partial) duck typing here.

What I think we should do here is something like:
```python
def get_by_ident(ident):
    try:
        return Event.objects.get(pk=int(ident))
    except ValueError:
        return Event.objects.get(slug=ident)
```

It's simpler, it's more Pythonic (I hope), it doesn't include slug validation where that validation should not happen.

If we really need to validate the slug format, we should do that in `save()`.

@wking, any thoughts?

workshops/models.py
```diff
@@ -296,17 +296,20 @@ def get_by_ident(ident):
...
-            return Event.objects.get(slug=ident)
-
-        raise ObjectDoesNotExist(ident)
+        kwargs = {}
+        try:
+            # first try to match using ID
+            kwargs["id"] = int(ident)
+        except ValueError:
+            # match event slug in format YYYY-MM-****
+            #(ie. this works for YYYY-MM-DD-**** too)
```
pbanaszkiewicz: Nit-picking: if we decide to go with this method, there's a space missing at the beginning of this comment. It should look like:
```python
            # (ie. this works for …
```

workshops/templates/workshops/event_create_form.html
```diff
@@ -1,5 +1,12 @@
 {% extends "workshops/_page.html" %}
 
+{% load crispy_forms_tags %}
+{% load selectable_tags %}
+
+{% block extrastyle %}
+  {% include_ui_theme %}
+{% endblock %}
+
```
pbanaszkiewicz: If we move loading tags to the top of the template, we can move `{% load breadcrumbs %}`, too.

workshops/templates/workshops/event_edit_form.html
```diff
@@ -1,5 +1,12 @@
 {% extends "workshops/_page.html" %}
 
+{% load crispy_forms_tags %}
+{% load selectable_tags %}
```
pbanaszkiewicz: https://github.com/swcarpentry/amy/pull/248/files#r28722582

workshops/templates/workshops/event_edit_form.html
```diff
@@ -49,3 +47,14 @@ <h2 id="associated-tasks">Associated tasks</h2>
 </form>
 
 {% endblock %}
+
+
```
pbanaszkiewicz: One newline too much.

workshops/templates/workshops/instructors.html
```diff
@@ -1,12 +1,19 @@
 {% extends "workshops/_page.html" %}
 
+{% load crispy_forms_tags %}
+{% load selectable_tags %}
+
```
pbanaszkiewicz: https://github.com/swcarpentry/amy/pull/248/files#r28722582

workshops/templates/workshops/person_form.html
```diff
@@ -1,5 +1,11 @@
 {% extends "workshops/_page.html" %}
 
+{% load selectable_tags %}
+
```
pbanaszkiewicz: https://github.com/swcarpentry/amy/pull/248/files#r28722582 + lack of consistency (you previously moved `{% load crispy_forms_tags %}`, but not in this template).

workshops/views.py
```diff
@@ -33,7 +33,8 @@
 from workshops.check import check_file
 from workshops.forms import (
     SearchForm, DebriefForm, InstructorsForm, PersonBulkAddForm, EventForm,
-    TaskForm, bootstrap_helper, bootstrap_helper_without_form)
+    TaskForm, bootstrap_helper, bootstrap_helper_without_form,
+    PersonForm, TaskFullForm)
```
pbanaszkiewicz: Very minor thing: you can put these imports to order: first forms, then helpers.

workshops/models.py
pbanaszkiewicz: @shapiromatron you actually thought about this and you check for slug validness [in another place](https://github.com/swcarpentry/amy/pull/248/files#diff-7dcb558eec23b9d2fe5e2f83d414e275R136). So I think `get_by_ident()` should be simplified.

shapiromatron: I also prefer your proposed method, if we don't need the `YYYY-MM-*` slug-format. If we go with this method, I can remove the `EventForm.clean_slug` check as well.

workshops/forms.py
shapiromatron: @pbanaszkiewicz will do, wish me luck!

pbanaszkiewicz:
Hey @shapiromatron,

I left some inline comments. Most of the code looks good, my comments are really nitpicks.

screenshot from 2015-04-20 22 05 15

Do you perhaps know if we could somehow indicate that:

1. Site and Organizer are autocompletable? (Previously these were `<select>`s, if someone forgot the options they could simply look for them). I'm thinking about something like a combobox.

2. Start and End pop up a calendar?

workshops/forms.py
pbanaszkiewicz: Good luck! If you have spare 30mins go watch this excellent talk https://www.youtube.com/watch?v=4EOZvow1mk4 it covers interactive rebase, too (I think).

workshops/models.py
pbanaszkiewicz: No, please don't get rid of `EventForm.clean_slug()`. And let's wait for @wking insight - I'm not sure if we should add another validation to `Event.save()` or not.

wking:
On Mon, Apr 20, 2015 at 11:35:27AM -0700, Piotr Banaszkiewicz wrote:
> > * Added input-validation in the EventForm to Event.slug to ensure
> >   that it has the proper form of "YYYY-MM-*" which is required for
> >   Event.get_by_ident method
> 
> I honestly lost track of the requirements for the event
> slug. Indeed, *published* events have slugs in
> `YYYY-MM-DD-something` format, but - I think - in case we don't know
> a formal starting date for an event, the slug looks different.
> @wking, @gvwilson: what should the event's slug look like in case we
> don't know the starting date?

As far as I'm concerned you're free to use any slug you like, but I'd
prefer if mutable slugs redirected to an immutable ID [1,2].  I
haven't checked if that's how we currently handle this or not.

[2]: https://github.com/swcarpentry/amy/pull/122#issuecomment-68676399

workshops/models.py
wking: On Mon, Apr 20, 2015 at 12:50:08PM -0700, Piotr Banaszkiewicz wrote:
> def get_by_ident(ident):
>     try:
>         return Event.objects.get(pk=int(ident))
>     except ValueError:
>         return Event.objects.get(slug=ident)

Works for me, and looks similar to my earlier suggestion [1].

[1]: https://github.com/swcarpentry/amy/pull/122#issuecomment-68676399

wking: On Mon, Apr 20, 2015 at 01:20:57PM -0700, Piotr Banaszkiewicz wrote:
> No, please don't get rid of `EventForm.clean_slug()`.

I think it's good to encourage users to use YYYY-MM-DD-SITE slugs, but
I see no reason to *require* it.  So I'd just drop a note to that
effect in the help_text for Event.slug, drop the clean_slug addition,
and use the int-fallback-to-slug approach when interpreting URLs.

I think folks who use integer slugs that collide with event IDs
deserve any suffering they get ;), but if you want to check to make
sure that the slug isn't an integer, I'm ok with that.

pbanaszkiewicz: Thanks, @wking. Let's then stick to simpler `get_by_ident()`.

Does `check.py` check event URL? I think it doesn't, so we can drop that check in `EventForm.clean_slug`, and use `help_text` with a format suggestion. Or we can allow any slug but integer by matching `\d+` in `EventForm.clean_slug`.

Personally, the combo "suggestive help_text" + "not allowing integer slugs" appeals the most to me. But I'll be fine even if we leave it as it is right now.

workshops/templates/base_without_navbar.html
shapiromatron: Sorry, I'm confused as to what the solution is. How can we resolve this?

workshops/models.py
wking: On Mon, Apr 20, 2015 at 10:22:14PM -0700, Piotr Banaszkiewicz wrote:
> Or we can allow any slug but integer by matching `\d+` in
> `EventForm.clean_slug`.
>
> Personally, the combo "suggestive help_text" + "not allowing integer
> slugs" appeals the most to me. But I'll be fine even if we leave it
> as it is right now.

So raise a ValidationError if we can match the slug with an
integer-only regexp [1]?

[1]: https://docs.djangoproject.com/en/1.8/ref/forms/validation/

workshops/templates/base_without_navbar.html
pbanaszkiewicz: Hey @shapiromatron,

Sorry, I wasn't clear.

You're using `block extrastyle` to add one tag call: `include_ui_theme` (for example here: https://github.com/swcarpentry/amy/pull/248/files#diff-780224fa654fedfe332f47ca01cca44fR7). This tag, `include_ui_theme`, in an [inclusion tag](https://docs.djangoproject.com/en/1.8/howto/custom-template-tags/#inclusion-tags). It simply includes `selectable/jquery-css.html` template.

That template is here: https://github.com/mlavin/django-selectable/blob/master/selectable/templates/selectable/jquery-css.html

It breaks some of our tests, because it doesn't follow XHTML syntax, and we rely upon XML parser in the tests.

So we want to override that template. We have to put it in our application (overriding it in `django-selectable` sources is not a good option).

To override that template, add `workshops/templates/selectable/jquery-css.html` with following content:
```html
<link rel="stylesheet" href="//ajax.googleapis.com/ajax/libs/jqueryui/{{ version }}/themes/{{ theme }}/jquery-ui.css" type="text/css" />
```

I hope I'm explained well this time, but ask if something isn't clear. Cheers!

workshops/models.py
pbanaszkiewicz: @wking yes. Thanks for pointing out to the validators.

@shapiromatron please replace `EventForm.clean_slug` with a validator that throws ValidationError when `slug` matches `\d+`. Thanks

workshops/forms.py
shapiromatron: @pbanaszkiewicz Great suggestion; https://youtu.be/4EOZvow1mk4?t=1136 is exactly what I needed to do; let's see what happens when I force-push to github...

workshops/templates/base_without_navbar.html
shapiromatron: @pbanaszkiewicz It turns out we don't need to `include_ui_theme` at all since a jQuery UI theme is included in our base template; I just removed the tag from all templates, so this should be resolved.

shapiromatron:
@pbanaszkiewicz 

I added help-text for the selectable fields, as well as the date-picker fields. I think the latest round of changes addresses your comments; let me know if I missed something or should make any other tweaks.

workshops/forms.py
```diff
@@ -43,10 +49,11 @@ class InstructorsForm(forms.Form):
...
                                  required=False)
-    airport = forms.ModelChoiceField(label='airport',
-                                     queryset=Airport.objects.all(),
-                                     to_field_name='iata',
-                                     required=False)
+    airport = selectable.AutoCompleteSelectField(
+                lookup_class=lookups.AirportLookup,
+                label='Airport',
+                required=False,
+                help_text=AUTOCOMPLETE_HELP_TEXT)
```
shapiromatron: @pbanaszkiewicz We lost the conversation here because of the interactive rebase session, but you can see that the only-line changed is the `airport` line, so the blame should be intact as you originally requested.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 Django>=1.7
-PyYAML
+PyYAML==3.11
 requests>=2.0
 django-crispy-forms>=1.4.0
+django-selectable==0.9.0
```
shapiromatron: specified an exact library version rather than >= to prevent unintended upgrades which may break functionality without testing

workshops/forms.py
```diff
@@ -109,8 +116,92 @@ class DebriefForm(forms.Form):
...
+        required=True,
+        help_text=AUTOCOMPLETE_HELP_TEXT
+    )
+
+    organizer = selectable.AutoCompleteSelectField(
+        lookup_class=lookups.SiteLookup,
+        label='Organizer',
+        required=True,
+        help_text=AUTOCOMPLETE_HELP_TEXT
+    )
```
pbanaszkiewicz: I checked and for fields that don't produce many options (like site, organizer or airport) we should be using [AutoComboboxSelectWidget](http://django-selectable.readthedocs.org/en/latest/widgets.html#autocomboboxselectwidget) - even if it looks quirky on the website.

Then you don't need AUTOCOMPLETE_HELP_TEXT.

workshops/forms.py
```diff
@@ -109,8 +116,92 @@ class DebriefForm(forms.Form):
...
+
+
+class TaskForm(forms.ModelForm):
+
+    person = selectable.AutoCompleteSelectField(
+        lookup_class=lookups.PersonLookup,
+        label='Person',
+        required=True,
+        help_text=AUTOCOMPLETE_HELP_TEXT
+    )
```
pbanaszkiewicz: There's a lot of people in our database so I would leave this field as it is right now.

workshops/forms.py
```diff
@@ -109,8 +116,92 @@ class DebriefForm(forms.Form):
...
+
+
+class PersonForm(forms.ModelForm):
+
+    airport = selectable.AutoCompleteSelectField(
+        lookup_class=lookups.AirportLookup,
+        label='Airport',
+        required=False,
+        help_text=AUTOCOMPLETE_HELP_TEXT
+    )
```
pbanaszkiewicz: [AutoComboboxSelectWidget](http://django-selectable.readthedocs.org/en/latest/widgets.html#autocomboboxselectwidget) for `airport`.

workshops/forms.py
```diff
@@ -109,8 +116,92 @@ class DebriefForm(forms.Form):
...
+
+
+class TaskFullForm(TaskForm):
+
+    event = selectable.AutoCompleteSelectField(
+        lookup_class=lookups.EventLookup,
+        label='Event',
+        required=True,
+        help_text=AUTOCOMPLETE_HELP_TEXT
+    )
```
pbanaszkiewicz: There's quite a few events in our database, so I'd not use [AutoComboboxSelectWidget](http://django-selectable.readthedocs.org/en/latest/widgets.html#autocomboboxselectwidget) here.

workshops/templates/workshops/event_edit_form.html
```diff
@@ -47,5 +41,14 @@ <h2 id="associated-tasks">Associated tasks</h2>
...
 </form>
-
 {% endblock %}
+
+{% block extrajs %}
+{{event_form.media}}
+{{task_form.media}}
+<script type="text/javascript">
+$(document).ready(function(){
+    $('#id_event-start,#id_event-end').datepicker();
```
pbanaszkiewicz: Please add `{format: "yyyy-mm-dd"}` here - otherwise this plugin refuses to work properly.

workshops/templates/workshops/event_create_form.html
```diff
@@ -12,8 +15,15 @@
...
-
 {% block content %}
 {% crispy form form_helper %}
 {% endblock %}
+
+{% block extrajs %}
+{{ form.media }}
+<script type="text/javascript">
+$(document).ready(function(){
+    $('#id_start,#id_end').datepicker();
```
pbanaszkiewicz: Please add `{format: "yyyy-mm-dd"}` here - otherwise this plugin refuses to work properly.

workshops/forms.py
```diff
@@ -1,14 +1,20 @@
...
+from workshops import lookups
 
-from workshops.models import Skill, Airport, Event, Task
 
 INSTRUCTOR_SEARCH_LEN = 10   # how many instrutors to return from a search by default
 
+AUTOCOMPLETE_HELP_TEXT = """Autocomplete field; type characters to view available options,
+                            then select desired item from list."""
+
+DATE_HELP_TEXT = "Select date using widget, or enter in MM/DD/YYYY format."
```
pbanaszkiewicz: Please change to "YYYY-MM-DD".

workshops/forms.py
```diff
@@ -109,8 +116,92 @@ class DebriefForm(forms.Form):
...
+    )
+
+    class Meta:
+        model = Person
+        # don't display the 'password' field, reorder fields
+        fields = ['personal', 'middle', 'family', 'username', 'may_contact',
+                  'email', 'gender', 'airport', 'github', 'twitter', 'url',
+                  'is_superuser', 'user_permissions']
+
+
```
pbanaszkiewicz: Too many blank lines at the end of `forms.py`.

pbanaszkiewicz:
Hi @shapiromatron,

your PR looks solid, there's very few changes to make:
* change widget type for some fields
* fix datepicker by changing its format
* rebase against master (there should not be many conflicts)

Thanks a lot for your work. Cheers,
Piotr

workshops/templates/workshops/event_edit_form.html
shapiromatron: It's working fine for me with the default format; what errors are you getting?

pbanaszkiewicz: The `event-start` field had `yyyy-mm-dd` date. After the widget had popped out, it did not render properly. It only had weekdays (Monday, Tuesday, …), but no actual calendar.

When I changed to `"yyyy-mm-dd"` format, the widget worked alright.

pbanaszkiewicz:
Hey @shapiromatron,

if you need any help with that PR, let me know, I'll gladly submit a PR against your PR.

Cheers,
Piotr

shapiromatron:
Thanks; I'm trying to fix the formatting of the auto select combo box which you requested and haven't had time to tweak it; I hope to submit a new PR soon!

pbanaszkiewicz:
@shapiromatron great! I think rebasing this PR against master could be easier than sending a new PR, but if the result is the same, I'm happy :) Cheers.

pbanaszkiewicz:
Superseded by #285.

PR #250, opened 2015-04-29
Adding the Amy logo to the login page

gvwilson:
This should probably be done using an 'include' directive of some kind, since it's exactly the same HTML as is used in index.html.

### DIFF
--- a/workshops/templates/account/logged_out.html
+++ b/workshops/templates/account/logged_out.html
@@ -7,6 +7,7 @@
 {% endblock %}
 
 {% block content %}
+{% include "amy-logo.html" %}
 
 <p>Thanks for spending some quality time with Amy today.</p>
 
--- a/workshops/templates/account/login.html
+++ b/workshops/templates/account/login.html
@@ -1,5 +1,4 @@
 {% extends "account/_page.html" %}
-{% load static %}
 {% load breadcrumbs %}
 {% block breadcrumbs %}
     {% breadcrumb_main_page %}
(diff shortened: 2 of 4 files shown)


pbanaszkiewicz:
> This should probably be done using an 'include' directive of some kind, since it's exactly the same HTML as is used in index.html.

Yes, can you please add `amy-logo.html` template with `<table>…</table>`, and then replace it in both `index.html` and `login.html` with inclusion tag: `{% include 'amy-logo.html' %}`?

gvwilson:
Sure - where does amy-logo.html go in order for both 
workshops/templates/workshops and workshops/templates/accounts to find it?
G

On 2015-04-29 2:16 AM, Piotr Banaszkiewicz wrote:
>
>     This should probably be done using an 'include' directive of some
>     kind, since it's exactly the same HTML as is used in index.html.
>
> Yes, can you please add |amy-logo.html| template with 
> |<table>…</table>|, and then replace it in both |index.html| and 
> |login.html| with inclusion tag: |{% include 'amy-logo.html' %}|?
>
> —
> <https://github.com/swcarpentry/amy/pull/250#issuecomment-97319083>.
>

-- 

pbanaszkiewicz:
Hi,

if you put the template into `workshops/templates/amy-logo.html`, it will work.

The inclusion tag should address `amy-logo.html`, ie. for both `index.html` and `login.html`: `{% include "amy-logo.html" %}`.

Remember to put `{% load static %}` on top of `amy-logo.html`.

Cheers

pbanaszkiewicz:
Sorry, I did notice you updated this PR. It looks good.

Issue #168, opened 2015-01-26
Implement backups of data on server


gvwilson:
Very frequent, since new code.

pbanaszkiewicz:
@gvwilson what is required for backups? Does future AMY server have some backups configured already?

gvwilson:
Nothing in place yet; we'll need backups very, very frequently during 
deployment (like, every 15 minutes if there's been a change to the DB).

wking:
On Tue, Feb 10, 2015 at 02:18:01PM -0800, Greg Wilson wrote:
> Nothing in place yet; we'll need backups very, very frequently
> during deployment (like, every 15 minutes if there's been a change
> to the DB).

I'd dump the “if there's been a change to the DB” requirement, and
just setup a cron job to copy db.sqlite3 to a date-stamped file every
15 minutes (e.g. /var/lib/amy/backup/2015-02-10T14:15-db.sqlite3).
I'd use a COW filesystem like btrfs (with ‘cp --reflink …’) to avoid
using extra space (besides the metadata entry) when the content didn't
change.

wking:
On Tue, Feb 10, 2015 at 02:28:13PM -0800, W. Trevor King wrote:
> I'd use a COW filesystem like btrfs…

And you could fake this on filesystems without copy-on-write by using
Git to store the backups with a single filename and commits for each
15-minute interval.  It would be harder to remove old versions (you'd
need to make the repository more shallow, and a quick search didn't
turn up an obvious command to do that.  With the COW filesystem, you
could just use ‘rm’).  In any case, I think it's best to handle this
sort of per-installation issue outside of amy, using an
amy-independent approach.

pbanaszkiewicz:
@wking is btrfs safe enough?

I'm unable to set up such configuration on a file system. Personally I'm using [duplicity](http://duplicity.nongnu.org/) with software RAID - kind of an easy solution.

wking:
On Tue, Feb 10, 2015 at 02:55:05PM -0800, Piotr Banaszkiewicz wrote:
> @wking is btrfs safe enough?

I've been using it since January 2011 without any data loss.

> I'm unable to set up such configuration on a file system.

Why not?  Btrfs is fairly widely adopted, and I've never had a problem
setting it up.

> Personally I'm using duplicity with software RAID - kind of an easy
> solution.

How does that handle de-duplication?  With snapshots every fifteen
minutes, the current 2 MB db.sqlite3 will use about 1.3 GB per week of
[...]

PR #251, opened 2015-05-06
Mark events and tasks as deleted

pbanaszkiewicz:
This PR changes the way events and tasks are deleted: they're marked as deleted (`object.deleted=True`) instead of being removed from the database for real.

It was achieved by adding a filter to default query sets for both models.

A downside of this solution is that we can still delete an event/task by:

```python
task.delete()
event.delete()
Task.objects.filter(…).delete()
Event.objects.filter(…).delete()
```

These commands will result in objects being removed from the database.

### DIFF
--- a/workshops/migrations/0005_auto_20150506_0848.py
+++ b/workshops/migrations/0005_auto_20150506_0848.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('workshops', '0004_merge'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='event',
+            name='deleted',
+            field=models.BooleanField(default=False),
+            preserve_default=True,
+        ),
+        migrations.AddField(
+            model_name='task',
+            name='deleted',
+            field=models.BooleanField(default=False),
+            preserve_default=True,
+        ),
+    ]
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -233,7 +233,8 @@ class EventManager(models.Manager):
 
     # Attach our custom query set to the manager
     def get_queryset(self):
-        return EventQuerySet(self.model, using=self._db)
+        """Fetch only existing events (ie. not deleted)."""
+        return EventQuerySet(self.model, using=self._db).filter(deleted=False)
 
     # Proxy methods so we can call our custom filters from the manager
     # without explicitly creating an EventQuerySet first - see
@@ -267,6 +268,7 @@ class Event(models.Model):
     admin_fee  = models.DecimalField(max_digits=6, decimal_places=2, null=True, blank=True)
     fee_paid   = models.NullBooleanField(default=False, blank=True)
     notes      = models.TextField(default="", blank=True)
+    deleted = models.BooleanField(default=False)
 
     class Meta:
         ordering = ('-start', )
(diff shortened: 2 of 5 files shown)


wking:
On Wed, May 06, 2015 at 07:10:50AM -0700, Piotr Banaszkiewicz wrote:
> This PR changes the way events and tasks are deleted: they're marked
> as deleted (`object.deleted=True`) instead of being removed from the
> database for real.

What's the use-case for this?  I'd expect that when you wanted to
delete something, you should actually be deleting it.

gvwilson:
The use case is "Greg screwed up and deleted the wrong event so we want 
an undo feature".

wking:
On Wed, May 06, 2015 at 10:19:50AM -0700, Greg Wilson wrote:
> The use case is "Greg screwed up and deleted the wrong event so we
> want an undo feature".

I'd address that with:

1. A “Delete is forever, are you sure you want to remove $THING?”
   confirmation dialog.
2. Frequent backups (#168) from which we can get the deleted
   information for folks who accidentally delete the wrong thing and
   then say “yes, I want to confirm that deletion” and *then* realize
   that it was the wrong thing ;).

Out of curiousity, what was your motivation for deleting an event?
The only cause I can think of is “oops, I accidentally clicked ‘create
an event’ and then ‘submit’ but didn't mean to do either”, which seems
like a low-probability situation ;).

gvwilson:
On 2015-05-06 1:31 PM, W. Trevor King wrote:
>
> 1. A “Delete is forever, are you sure you want to remove $THING?”
> confirmation dialog.

They've been shown to be ineffective - people get so used to clicking 
"yes" that they muscle-memory do it when they shouldn't.

> 2. Frequent backups (#168) from which we can get the deleted
> information for folks who accidentally delete the wrong thing and
> then say “yes, I want to confirm that deletion” and *then* realize
> that it was the wrong thing ;).
>
Agreed; Piotr's trying to figure out why his cron job isn't backing 
things up - can you lend a hand?

> Out of curiousity, what was your motivation for deleting an event?
> The only cause I can think of is “oops, I accidentally clicked ‘create
> an event’ and then ‘submit’ but didn't mean to do either”, which seems
> like a low-probability situation ;).
Already happened (partly because of case sensitivity, but partly because 
of sleepy operator)

-- 

wking:
On Wed, May 06, 2015 at 10:37:09AM -0700, Greg Wilson wrote:
> On 2015-05-06 1:31 PM, W. Trevor King wrote:
> > 1. A “Delete is forever, are you sure you want to remove $THING?”
> > confirmation dialog.
> 
> They've been shown to be ineffective - people get so used to
> clicking "yes" that they muscle-memory do it when they shouldn't.

I have ‘rm’ aliased to ‘rm -i’.  It's not 100% effective, but I've
certainly had cases where it helped protect me from blowing away the
wrong thing.

> > 2. Frequent backups (#168) from which we can get the deleted
> > information for folks who accidentally delete the wrong thing and
> > then say “yes, I want to confirm that deletion” and *then* realize
> > that it was the wrong thing ;).
>
> Agreed; Piotr's trying to figure out why his cron job isn't backing
> things up - can you lend a hand?

Sure.  Just have hime comment in #168 explaining what he's doing and
the symptoms he sees.

> > Out of curiousity, what was your motivation for deleting an event?
> > The only cause I can think of is “oops, I accidentally clicked
> > ‘create an event’ and then ‘submit’ but didn't mean to do either”,
> > which seems like a low-probability situation ;).
>
> Already happened (partly because of case sensitivity, but partly
> because of sleepy operator)

I think we should limit the height of the “operator error” barriers,
because they make the system more complicated (e.g. with this PR there
are still ways you can delete objects, and we need a new interface for
interacting with “deleted” objects and restoring them to active
status, …).  If operator errors are common, we likely need to redesign
the UI to avoid the error-generating workflow.  If operator errors are
uncommon, restoring from backups should be acceptable.

pbanaszkiewicz:
Okay, I'm merging because we need the functionality right now.

PR #252, opened 2015-05-06
Putting search results in columns for easier reading

gvwilson:
It's a lot easier to find things this way.


pbanaszkiewicz:
This may need to be rebased on #258 (`event.slug` → `event.get_ident`).

pbanaszkiewicz:
I'm marging this one and will rebase #258 instead.

PR #253, opened 2015-05-06
Trying to get a list of workshops that owe us money

gvwilson:
The aim of this PR is to show a third column on the dashboard page listing all the workshops that owe us an admin fee. The test for these workshops is:

* admin_fee > 0
* fee_paid != True (i.e., is either False or unknown (NULL))

EventQuerySet.unpaid_events is supposed to return this set, but it isn't working.  The test I conducted is to modify the standard (fake) test data like this:

~~~
update workshops_event set admin_fee = 1500 where id=344;
update workshops_event set admin_fee = 1500 where id=343;
update workshops_event set fee_paid = 0 where id=343;
~~~

and then check with:

~~~
select * from workshops_event where admin_fee > 0.0;

343|1|2015-07-02|2015-07-03|2015-07-02-UQ|https://github.com/bio-swc-bne/2015-07-02-UQ|||1500|0|||241
344|1|2015-07-13|2015-07-14|2015-07-13-amos|https://github.com/damienirving/2015-07-13-amos|||1500||||241
~~~

so there are two workshops that should show up, but none do.  The most likely explanation is that the custom query set is wrong - help!

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -227,6 +227,16 @@ class EventQuerySet(models.query.QuerySet):
 
         return self.filter(published=False)
 
+    def unpaid_events(self):
+        '''Return a queryset for events that owe money.'''
+
+        # All events that have a non-zero admin fee.
+        queryset = self.filter(admin_fee__ne=0)
+
+        # Of those, all those whose paid status is not 'Yes' (i.e., is 'No' or NULL).
+        queryset = queryset.filter(fee_paid__ne=True)
+
+        return queryset
 
 class EventManager(models.Manager):
     '''A custom manager which is essentially a proxy for EventQuerySet'''
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -230,13 +230,8 @@ class EventQuerySet(models.query.QuerySet):
     def unpaid_events(self):
         '''Return a queryset for events that owe money.'''
 
-        # All events that have a non-zero admin fee.
-        queryset = self.filter(admin_fee__ne=0)
-
-        # Of those, all those whose paid status is not 'Yes' (i.e., is 'No' or NULL).
-        queryset = queryset.filter(fee_paid__ne=True)
-
-        return queryset
+        # All events that have an admin fee but are not marked as paid.
+        return self.filter(admin_fee__gt=0).exclude(fee_paid=True)
 
 class EventManager(models.Manager):
     '''A custom manager which is essentially a proxy for EventQuerySet'''
@@ -260,6 +255,9 @@ class EventManager(models.Manager):
 
     def unpublished_events(self):
         return self.get_queryset().unpublished_events()
+
+    def unpaid_events(self):
+        return self.get_queryset().unpaid_events()
 
 class Event(models.Model):
     '''Represent a single event.'''
--- a/workshops/templates/workshops/index.html
+++ b/workshops/templates/workshops/index.html
@@ -34,7 +34,7 @@
     {% endfor %}
     </table>
   </div>
-  <div class="col-sx-4">
+  <div class="col-xs-4">
     <h3>Unpaid events</h3>
     <table class="table table-striped">
     {% for event in unpaid_events %}
--- a/workshops/templates/workshops/index.html
+++ b/workshops/templates/workshops/index.html
@@ -8,7 +8,7 @@
 {% include "amy-logo.html" %}
 
 <div class="row">
-  <div class="col-xs-6">
+  <div class="col-xs-4">
     <h3>Upcoming events</h3>
     <table class="table table-striped">
     {% for event in upcoming_events %}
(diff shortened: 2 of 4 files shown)


gvwilson:
Should add a fourth view: those workshops that don't yet have dates, instructors, or other important fields (i.e., things that are half-baked).

workshops/models.py
```diff
@@ -227,6 +227,16 @@ def unpublished_events(self):
 
         return self.filter(published=False)
 
+    def unpaid_events(self):
```
pbanaszkiewicz: You need an `unpaid_events` method on `EventManager` below, too:
```python
class EventManager:
    …
    def unpaid_events(self):
        return self.get_queryset().unpaid_events()
```

workshops/templates/workshops/index.html
```diff
@@ -34,5 +34,15 @@
     {% endfor %}
     </table>
   </div>
+  <div class="col-sx-4">
+    <h3>Unpaid events</h3>
+    <table class="table table-striped">
+    {% for event in unpaid_events %}
```
pbanaszkiewicz: You need to send `unpaid_events` in `workshops.views.index`.

workshops/models.py
```diff
@@ -227,6 +227,16 @@ def unpublished_events(self):
...
+        '''Return a queryset for events that owe money.'''
+
+        # All events that have a non-zero admin fee.
+        queryset = self.filter(admin_fee__ne=0)
+
+        # Of those, all those whose paid status is not 'Yes' (i.e., is 'No' or NULL).
+        queryset = queryset.filter(fee_paid__ne=True)
```
pbanaszkiewicz: `__ne` didn't work for me. Here's corrected code:
```python
        # All events that have a non-zero admin fee.
        queryset = self.filter(Q(admin_fee__gt=0) | Q(admin_fee__lt=0))

        # Of those, all those whose paid status is not 'Yes' (i.e., is 'No' or NULL).
        queryset = queryset.filter(~Q(fee_paid=True))
```

workshops/templates/workshops/index.html
```diff
@@ -34,5 +34,15 @@
     {% endfor %}
     </table>
   </div>
+  <div class="col-sx-4">
```
pbanaszkiewicz: A small typo: should be `col-xs-4`.

pbanaszkiewicz:
Hi @gvwilson,

I posted some inline comments, should help you with that queryset and also the PR itself.

> Should add a fourth view: those workshops that don't yet have dates, instructors, or other important fields (i.e., things that are half-baked).

I was thinking about this and it's very tricky to do. Especially hard is calculating number of instructors per Event - Django ORM does not support this kind of queries. You might want to spin this off to a new issue.

workshops/models.py
wking: On Wed, May 06, 2015 at 12:44:45PM -0700, Piotr Banaszkiewicz wrote:
>         # All events that have a non-zero admin fee.
>         queryset = self.filter(Q(admin_fee__gt=0) | Q(admin_fee__lt=0))

What is the < 0 case for?

>         # Of those, all those whose paid status is not 'Yes' (i.e., is 'No' or NULL).
>         queryset = queryset.filter(~Q(fee_paid=True))

No need to split this into multiple filter calls.  How about:

  self.filter(admin_fee__gt=0).exclude(fee_paid=True)

?

pbanaszkiewicz: Not sure if it's possible to store negative numbers in DecimalField, but I thought a little and it makes no sense to list overpaid events under "unpaid" label.

And I like the chained calls.

wking: On Wed, May 06, 2015 at 01:14:18PM -0700, Piotr Banaszkiewicz wrote:
> Not sure if it's possible to store negative numbers in DecimalField,
> but I thought a little and it makes no sense to list overpaid events
> under "unpaid" label.

Hrm, “overpaid” makes it sound like “we're adjusting admin_fee to be
the outstanding balance for this event”.  I'd expected it to be “how
much we charged (or will charge) for this event”, with payment status
only showing up via the fee_paid boolean.  If we *do* want to track
received money, I'd rather make fee_paid a DecimalField (or add a
separate Payment table tracking per-host payment instances).  But
adjusting admin_fee after the fact is going to make it hard to answer
questions like “how much did we charge for their last workshop?” or
even “how much money should we have earned last month?”.

pbanaszkiewicz: @wking you're totally right. I doubt Amy would be used for accounting, so no need to track payments, unless @gvwilson has some plans for that.

gvwilson:
Thanks for all the feedback - now passes local tests.

pbanaszkiewicz:
This PR only needs some tests in `workshops/tests/test_event.py` and we can merge.

gvwilson:
tests pushed

PR #255, opened 2015-05-06
Removing user slug from display of user name

gvwilson:
Instead of `wilson.g:Greg Wilson <gvwilson@software-carpentry.org>`, we now display just `Greg Wilson <gvwilson@software-carpentry.org>`.

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -131,10 +131,10 @@ class Person(AbstractBaseUser, PermissionsMixin):
         return self.personal
 
     def __str__(self):
-        email = ''
+        result = self.get_full_name()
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
Looks benign, but why do you want to get rid of usernames?

gvwilson:
We don't need them most of the time, and when we want to copy/paste 
names (e.g., from the list of people who should be in the next 
debriefing) they're in the way.  I think it's cleaner to just show `name 
<email>` most of the time.

pbanaszkiewicz:
Okay, it sounds reasonable.

Issue #257, opened 2015-05-06
Searching for 'ssi' causes a server error


gvwilson:
1. Go to http://amy.software-carpentry.org:8080/workshops/search
2. Enter 'ssi' (without quotes) in the search box.
3. Search.
4. Boom.

Search works for 'toronto'.  Maybe 'ssi' is returning too many results?

pbanaszkiewicz:
Reproduced with production database. The error is:

```
Environment:


Request Method: POST
Request URL: http://127.0.0.1:8000/workshops/search

Django Version: 1.7.3
Python Version: 3.4.3
Installed Applications:
('django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'workshops',
 'django.contrib.admin',
 'crispy_forms')
Installed Middleware:
('django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware')


Template error:
In template /home/piotr/workspace/SwC/amy/workshops/templates/workshops/search.html, error at line 26
   Reverse for 'event_details' with arguments '('',)' and keyword arguments '{}' not found. 1 pattern(s) tried: ['workshops/event/(?P<event_ident>[\\w-]+)/?$']
   16 :   {% for s in sites %}


   17 :   <li><a class="searchresult" href="{% url 'site_details' s.domain %}">{{ s.domain }}</a></li>


   18 :   {% endfor %}


   19 : </ul>


   20 : {% endif %}


   21 : 


   22 : <h2>Events</h2>


   23 : {% if events %}


   24 : <ul>


   25 :   {% for e in events %}


   26 :   <li><a class="searchresult" href=" {% url 'event_details' e.slug %} ">{{ e.slug }}</a></li>


   27 :   {% endfor %}


   28 : </ul>


   29 : {% endif %}


   30 : 


   31 : <h2>Persons</h2>


   32 : {% if persons %}


   33 : <ul>


   34 :   {% for p in persons %}


[...]

PR #258, opened 2015-05-06
Fix: template rev. URL error for events w/o slug

pbanaszkiewicz:
In rare occasion when events don't have any deatils about them,
especially the slug, we should display event details using
`event.get_ident` as it returns either slug or ID.

Ref #257

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -114,3 +114,4 @@ class DebriefForm(forms.Form):
 EventForm = modelform_factory(Event, fields="__all__")
 TaskForm = modelform_factory(Task, fields="__all__",
                              widgets={'event': HiddenInput})
+
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -297,24 +297,13 @@ class Event(models.Model):
     def get_by_ident(ident):
         '''
         Select event that matches given identifier.
-
-        If the event identifier is purely numeric, it's an ID (and
-        probably indicates a pending event whose details are still being
-        negotiated).  If it contains dashes, it's probably a
-        YYYY-MM-DD-site slug, and indicates an event whose dates are firm.
-        The real indicator is the 'published' flag.
+        If ident is an int, search for matching primary-key;
+        otherwise get matching slug. May throw DoesNotExist error.
         '''
-
-        # match event ID
-        if re.match(r'^\d+$', ident):
-            return Event.objects.get(id=ident)
-
-        # match event slug in format YYYY-MM-**** (ie. this works for
-        # YYYY-MM-DD-**** too)
-        if re.match(r'^\d{4}-\d{2}-.+$', ident):
+        try:
+            return Event.objects.get(pk=int(ident))
+        except ValueError:
             return Event.objects.get(slug=ident)
-
-        raise ObjectDoesNotExist(ident)
 
     def save(self, *args, **kwargs):
         self.url = self.url or None
--- a/workshops/templates/workshops/all_events.html
+++ b/workshops/templates/workshops/all_events.html
@@ -36,8 +36,8 @@
 	      {% if not event.slug %}
 	        —
 	      {% else %}
-	        <a href="{% url 'event_details' event.slug %}">
-		  {{ event.slug }}
+	        <a href="{% url 'event_details' event.get_ident %}">
+		  {{ event.get_ident }}
 	        </a>
 	      {% endif %}
 	    </td>
--- a/workshops/templates/workshops/search.html
+++ b/workshops/templates/workshops/search.html
@@ -29,7 +29,7 @@
       {% if events %}
       <ul>
 	{% for e in events %}
-	<li><a class="searchresult" href="{% url 'event_details' e.slug %}">{{ e.slug }}</a></li>
+	<li><a class="searchresult" href="{% url 'event_details' e.get_ident %}">{{ e.get_ident }}</a></li>
 	{% endfor %}
       </ul>
       {% endif %}
--- a/workshops/templates/workshops/site.html
+++ b/workshops/templates/workshops/site.html
@@ -38,7 +38,7 @@
     <td>{{ e.tags.all | join:", " }}</td>
     <td>{{ e.start }}</td>
     <td>{{ e.end }}</td>
-    <td><a href="{% url 'event_details' e.slug %}">{{ e.slug }}</a></td>
+    <td><a href="{% url 'event_details' e.get_ident %}">{{ e.get_ident }}</a></td>
     <td><a href="{{ e.url }}">{{ e.url }}</a></td>
   </tr>
   {% endfor %}
--- a/workshops/templates/workshops/validate_event.html
+++ b/workshops/templates/workshops/validate_event.html
@@ -20,7 +20,7 @@
     <p>No errors</p>
   {% endif %}
 {% endif %}
-<p>... <a href="{% url 'event_details' event.slug %}">return to event</a></p>
+<p>... <a href="{% url 'event_details' event.get_ident %}">return to event</a></p>
 <p>... <a href="{% url 'all_events' %}">all events</a></p>
 <p>... <a href="{% url 'index' %}">index</a></p>
 {% endblock %}


wking:
On Wed, May 06, 2015 at 01:11:48PM -0700, Piotr Banaszkiewicz wrote:
> In rare occasion when events don't have any deatils about them,
> especially the slug, we should display event details using
> `event.get_ident` as it returns either slug or ID.

So:

1. I'd rather just use Event.id in all of these places.  The
   slug-based form should just be for folks who want to type out the
   URL directly (where the slug is more memorable if it exists).  For
   internal links (typed by the templating engine), the ID form is
   always going to better.  See the mutable-vs.-permanent discussion
   in #121, especially [1].
2. There are a number of event_details reverse lookups, and any
   changes to their handling should update all occurrences (‘git grep
   event_details’).

gvwilson:
So it looks like I caused this by creating an event without a slug as a 
placeholder.  Maybe the simplest solution is to insist on a slug (since 
we can edit it later if e.g. dates change)?

pbanaszkiewicz:
@wking:

1. Implementing 301's everywhere would take more time, I'd rather do it in another issue (and another time).

2. I changed the rest of those lookups.

@gvwilson:
I don't think that solution would work fine from usability point of view without 301 redirects in place.

wking:
On Wed, May 06, 2015 at 01:37:17PM -0700, Greg Wilson wrote:
> So it looks like I caused this by creating an event without a slug
> as a placeholder.  Maybe the simplest solution is to insist on a
> slug (since we can edit it later if e.g. dates change)?

Requiring event slugs works for me [1].  In the past, I think this had
blocked on some cases where we had no idea what to use for the initial
slug [2].

But really, I think the simplest solution is just to use Event.id for
template-written links.  Why require a human-generated ID when we
already have an auto-generated ID?

[1]: https://github.com/swcarpentry/amy/pull/121#issuecomment-68642791
[2]: https://github.com/swcarpentry/amy/pull/121#issuecomment-68650003

gvwilson:
On 2015-05-06 4:48 PM, W. Trevor King wrote:
> Why require a human-generated ID when we already have an 
> auto-generated ID?
Because the YYYY-MM-DD-site slug is what we use everywhere else to 
identify the workshop (e.g., in invoicing, in GitHub, etc.), so it 
should be the most commonly used thing in Amy as well.

wking:
On Wed, May 06, 2015 at 01:44:09PM -0700, Piotr Banaszkiewicz wrote:
> 1. Implementing 301's everywhere would take more time, I'd rather do
> it in another issue (and another time).

Where do 301s come in?  I thought we were already using
Event.get_by_ident to transparently handle both IDs and slugs?

pbanaszkiewicz:
@wking:

> Where do 301s come in?  I thought we were already using Event.get_by_ident to transparently handle both IDs and slugs?

Idea of 301 redirect:
* https://github.com/swcarpentry/amy/pull/121#issuecomment-68650894
* https://github.com/swcarpentry/amy/pull/121#issuecomment-68664373

These URLs would have to redirect to IDs:
```python
    url(r'^event/(?P<event_ident>[\w-]+)/?$', views.event_details, name='event_details'),
    url(r'^event/(?P<event_ident>[\w-]+)/edit$', views.event_edit, name='event_edit'),
    url(r'^event/(?P<event_ident>[\w-]+)/validate/?$', views.validate_event, name='validate_event'),
```

Additionally, I believe some tests would have to be rewritten.

wking:
On Wed, May 06, 2015 at 01:49:17PM -0700, Greg Wilson wrote:
> On 2015-05-06 4:48 PM, W. Trevor King wrote:
> > Why require a human-generated ID when we already have an
> > auto-generated ID?
>
> Because the YYYY-MM-DD-site slug is what we use everywhere else to
> identify the workshop (e.g., in invoicing, in GitHub, etc.), so it
> should be the most commonly used thing in Amy as well.

I don't understand the connection there.  Why should the externally
used slug effect amy's internal linking?  I agree that we want to
support folks using the slug to get an event's details, create tasks,
… from *outside* amy, but once you're in the amy UI, I think users
should be treating URLs as opaque identifiers, and Event.id is a
better choice for that.  For folks copy/pasting links from amy into
the outside world, Event.id's are *still* better, because we know
they're stable.  The only place where slugs are useful are for folks
who are typing out an event identifier by hand.

wking:
On Wed, May 06, 2015 at 01:53:45PM -0700, Piotr Banaszkiewicz wrote:
> > Where do 301s come in?  I thought we were already using
> > Event.get_by_ident to transparently handle both IDs and slugs?
> 
> Idea of 301 redirect:
> * https://github.com/swcarpentry/amy/pull/121#issuecomment-68650894
> * https://github.com/swcarpentry/amy/pull/121#issuecomment-68664373
> 
> These URLs would have to redirect to IDs:
> ```python
>     url(r'^event/(?P<event_ident>[\w-]+)/?$', views.event_details, name='event_details'),
>     url(r'^event/(?P<event_ident>[\w-]+)/edit$', views.event_edit, name='event_edit'),
>     url(r'^event/(?P<event_ident>[\w-]+)/validate/?$', views.validate_event, name='validate_event'),
> ```

Ah right.  I agree that we should put that off.  The current “accept
either an ID or slug” is fine for now, and this pull can focus on what
we should use in template-generated, internal links.

gvwilson:
As more data: I just tried to edit the three workshops in the production 
version that had empty slugs.  The one that had start and end dates, but 
no slug, let me set 2015-xx-xx-gatech.  The other two, which *didn't* 
have dates, gave me a 500 error when I tried to save - but then 
apparently did the save anyway - see 384 @ uconn.edu : 2015-xx-xx-uconn 
<http://amy.software-carpentry.org:8080/workshops/event/2015-xx-xx-uconn> and 
395 @ vt.edu : 2015-xx-xx-vt 
<http://amy.software-carpentry.org:8080/workshops/event/2015-xx-xx-vt> .

wking:
On Wed, May 06, 2015 at 01:58:26PM -0700, Greg Wilson wrote:
> The other two, which *didn't* have dates, gave me a 500 error when I
> tried to save…

Do we have a traceback for that error?

pbanaszkiewicz:
@wking I do:

```
Environment:


Request Method: GET
Request URL: http://127.0.0.1:8000/workshops/event/2015-xx-xx-uconn

Django Version: 1.7.3
Python Version: 3.4.3
Installed Applications:
('django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'workshops',
 'django.contrib.admin',
 'crispy_forms')
Installed Middleware:
('django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware')


Traceback:
File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/core/handlers/base.py" in get_response
  111.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/home/piotr/workspace/anaconda/envs/amy3k/lib/python3.4/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
  21.                 return view_func(request, *args, **kwargs)
File "/home/piotr/workspace/SwC/amy/workshops/views.py" in event_details
  420.     event = Event.get_by_ident(event_ident)
File "/home/piotr/workspace/SwC/amy/workshops/models.py" in get_by_ident
  309.         raise ObjectDoesNotExist(ident)

Exception Type: ObjectDoesNotExist at /workshops/event/2015-xx-xx-uconn
Exception Value: 2015-xx-xx-uconn
```

The error is because we expect the slug in the form YYYY-MM-DD, where Y, M, D are numbers, not "x". This is fixed in #248 (not yet merged…).

pbanaszkiewicz:
@wking, @gvwilson do you want me to backport the fix for that, too?

wking:
On Wed, May 06, 2015 at 02:06:23PM -0700, Piotr Banaszkiewicz wrote:
> @wking, @gvwilson do you want me to backport the fix for that, too?

Backport as in “cherry-pick into this branch”?  No preference really,
but I don't see a pressing need to fix that issue if we're using IDs
for internal links instead of slugs ;).  The only issue would be if
folks would rather use the xx-style links to reference these workshops
from outside of amy.  If that's the case, then cherry-picking that
commit into this PR (or an independent PR) is fine with me.

pbanaszkiewicz:
Fix cherry-picked.

gvwilson:
+1

pbanaszkiewicz:
I'm merging; if we decide in #259 that we want IDs everywhere, it will require some more changes (ie. not only reverting this PR).

PR #265, opened 2015-05-08
Fix datepicker widget

pbanaszkiewicz:
Due to some changes in recent PRs, the datepicker widget broke. Here are my fixes to get it working.


wking:
On Fri, May 08, 2015 at 02:28:27PM -0700, Piotr Banaszkiewicz wrote:
> Assigned #265 to @wking.

These look reasonable to me.  I haven't tested them locally, but if
they work for you I'd just merge them ;).

pbanaszkiewicz:
Thanks, @wking.

PR #268, opened 2015-05-10
TaskManager methods for roles lookup

### LINKED ISSUE #214: Support first workshop session
I'm thinking in run once a month a session for instructors that will teach their first workshop. What about generate this list?

pbanaszkiewicz:
Hi,

I've added some methods to `TaskManager`, so that we can easily filter by the role name.

This for example makes easier for us to calculate the number of instructors per event - we don't have to run a loop for that. One eyesore less.

Additionally I updated `/debrief/` so that it provides a more verbose output: links to event/person detail pages, number of workshops a person has taught. This should fix #214.

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -330,6 +330,19 @@ class TaskManager(models.Manager):
         """Fetch only existing tasks (ie. not deleted)."""
         return super().get_queryset().filter(deleted=False)
 
+    def instructors(self):
+        """Fetch tasks with role 'instructor'."""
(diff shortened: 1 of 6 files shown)


gvwilson:
+1 to merge.

pbanaszkiewicz:
I'll rebase against master after I'm done merging #267.

pbanaszkiewicz:
Still looking good, @gvwilson, @wking?

PR #270, opened 2015-05-12
README.md: Fix grammar, new *Upgrading* section

### LINKED ISSUE #261: Create superuser step fails with NOT NULL constraint failed: workshops_person.last_login
Trying to install Amy locally using the instructions in the README. At the step to create a superuser, I get a failure - details below. Have replicated trying different superuser details in case spaces in the entry fields were the issue. I'm using the test dataset created using the previous step, no errors reported.

(py3k)MBANCH:amy nhong$ python3 manage.py createsuperuser
Username: neil
Personal: Neil
Family: Chue Hong
Email: n.chuehong@software.ac.uk
Password: 
Password (again): 
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/backends/sqlite3/base.py", line 318, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.IntegrityError: NOT NULL constraint failed: workshops_person.last_login

The above exception was the direct cause of the following exception:

  File "manage.py", line 10, in <module>
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/core/management/__init__.py", line 330, in execute
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/core/management/base.py", line 390, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/contrib/auth/management/commands/createsuperuser.py", line 50, in execute
    return super(Command, self).execute(*args, **options)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/core/management/base.py", line 441, in execute
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/contrib/auth/management/commands/createsuperuser.py", line 149, in handle
    self.UserModel._default_manager.db_manager(database).create_superuser(**user_data)
  File "/Users/nhong/Documents/src/npch/amy/workshops/models.py", line 83, in create_superuser
    user.save(using=self._db)
  File "/Users/nhong/Documents/src/npch/amy/workshops/models.py", line 159, in save
    super().save(*args, **kwargs)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/models/base.py", line 710, in save
    force_update=force_update, update_fields=update_fields)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/models/base.py", line 738, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/models/base.py", line 822, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/models/base.py", line 861, in _do_insert
    using=using, raw=raw)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/models/query.py", line 920, in _insert
    return query.get_compiler(using=using).execute_sql(return_id)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/models/sql/compiler.py", line 971, in execute_sql
    cursor.execute(sql, params)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/backends/utils.py", line 79, in execute
    return super(CursorDebugWrapper, self).execute(sql, params)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/utils.py", line 97, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/utils/six.py", line 658, in reraise
    raise value.with_traceback(tb)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/Users/nhong/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/backends/sqlite3/base.py", line 318, in execute
    return Database.Cursor.execute(self, query, params)
django.db.utils.IntegrityError: NOT NULL constraint failed: workshops_person.last_login

pbanaszkiewicz:
This commit should resolve #261 where @wking suggested to update
`README.md`.

A few grammatical errors were fixed.

Sections were introduced (`Getting started`, `Upgrading`, `Contact`).

A new step to *Getting started* section was introduced (cloning the
repository).  Some notes were added.

A new section, *Upgrading*, was added (this directly relates to the
suggestion in #261).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,76 +1,130 @@
 ![](workshops/static/amy-logo.png)
 
-[Software Carpentry](http://software-carpentry.org) is now running three events a week.
-That number could double by the end of 2015
-but only if we can
+[Software Carpentry](http://software-carpentry.org) is running three events
+a week.  The number could double by the end of 2015, but only if we can
 [streamline setup and organization](http://software-carpentry.org/blog/2014/12/plans-for-2015-workshop-organization.html).
-The goal of this project is to build
-a web-based workshop administration application using Django to do that.
-It is not intended to be visible to the general public,
-or even to instructors (at least initially --- we may add that later).
-Instead,
-the target audience is administrators,
-most of whom are non-programmers,
-who need to keep track of
-what workshops are being arranged,
-when they're supposed to occur,
-who's teaching what,
-and so on.
+The goal of this project is to build a web-based workshop administration
+application using Django to do that.  It is not intended to be visible to the
+general public, or even to instructors (at least initially — we may add that
+later).
 
-To get started:
+Instead, the target audience is administrators, most of whon are
+non-programmers, who need to keep track of what workshops are being arranged,
+when they're supposed to occur, who's teaching what, and so on.
 
-1.  Install Django and its dependencies.
+# Getting started
+
+1. Clone the repository:
+
+    ~~~
+    $ git clone https://github.com/swcarpentry/amy.git
+    $ cd amy
+    ~~~
+
+2. Install Django and its dependencies:
 
     ~~~
     $ sudo python3 -m pip install -r requirements.txt
     ~~~
 
-2. Install [Bower](http://bower.io/) - the tool that manages Amy's JavaScript and CSS dependencies:
+    If you're experienced Python programmer, feel free to create a Python3-compatible virtualenv for Amy and install dependencies from `requirements.txt`.
+
+3. Install [Bower](http://bower.io/) — the tool that manages Amy's JavaScript and CSS dependencies:
 
     ~~~
-    $ npm install -g bower
+    $ sudo npm install -g bower
     ~~~
 
-3.  Setup your local database.  There are two ways you can do this.
+    You may need some additional dependencies to install [Bower](http://bower.io/), such as [NodeJS](https://nodejs.org/) and [npm](https://www.npmjs.com/).
+
+4. Setup your local database.  There are two ways you can do this:
 
     1. If you have access to the legacy data:
 
         1. Create an empty database by running:
 
-           ~~~
-           $ make migrations
-           ~~~
+            ~~~
+            $ make migrations
+            ~~~
 
         2. Fill that database by running:
 
-           ~~~
-           $ make import
-           ~~~
+            ~~~
+            $ make import
+            ~~~
 
-    2. Otherwise set up the redacted data with:
+    2. Otherwise set up the redacted (development-ready) data with:
 
-       ~~~
-       $ make database
-       ~~~
+        ~~~
+        $ make database
+        ~~~
 
-4.  Create a administrator account.
+5. Create an administrator account:
 
     ~~~
     $ python3 manage.py createsuperuser
     ~~~
 
-5.  Start a local Django development server by running (this also installs front-end dependencies for Amy):
+6. Start a local Django development server by running:
 
     ~~~
     $ make serve
     ~~~
 
-6.  Open [http://localhost:8000/workshops/](http://localhost:8000/workshops/) in your browser and start clicking.
+    **Note**:  this also installs front-end dependencies for Amy, such as jQuery or Bootstrap.
+
+7. Open [http://localhost:8000/workshops/](http://localhost:8000/workshops/) in your browser and start clicking.
 
     Use the administrator account that you created.
 
-**Note**: please [check with us](mailto:gvwilson@software-carpentry.org)
-or open a [discussion issue](https://github.com/swcarpentry/amy/labels/discussion)
+# Upgrading
+
+1. Update the code:
+
+    1. Get the list of changes:
+
+        ~~~
+        $ git fetch
+        ~~~
+
+    2. Look for the newest tag:
+
+        ~~~~
+        $ git tag -n
+        ~~~~
+
+    3. Get the code from the newest tag:
+
+        ~~~~
+        $ git checkout tags/<tag_name>
+        ~~~~
+
+2. Update dependencies:
+
+    ~~~
+    $ sudo python3 -m pip install --upgrade -r requirements.txt
+    $ bower update
+    ~~~
+
+3. (Optional) use newer development-ready database:
+
+    ~~~
+    $ make database
+    ~~~
+
+3. Run database migrations:
+
+    ~~~~
+    $ python3 manage.py migrate
+    ~~~~
+
+4. Enjoy your new version of Amy.
+
+# Contact
+
+Please [check with us](mailto:gvwilson@software-carpentry.org) or open
+a [discussion issue](https://github.com/swcarpentry/amy/labels/discussion)
 before adding any new features.
-A few things have to get built in order to meet present demand,
-and they should take precedence over everything else.
+
+A few things have to get built in order to meet present demand, and they
+should take precedence over everything else.
--- a/README.md
+++ b/README.md
@@ -4,11 +4,9 @@
 a week.  The number could double by the end of 2015, but only if we can
 [streamline setup and organization](http://software-carpentry.org/blog/2014/12/plans-for-2015-workshop-organization.html).
 The goal of this project is to build a web-based workshop administration
-application using Django to do that.  It is not intended to be visible to the
+application using Django.  It is not intended to be visible to the
 general public, or even to instructors (at least initially — we may add that
-later).
-
-Instead, the target audience is administrators, most of whon are
+later).  Instead, the target audience is administrators, most of whon are
 non-programmers, who need to keep track of what workshops are being arranged,
 when they're supposed to occur, who's teaching what, and so on.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,76 +1,130 @@
...
-what workshops are being arranged,
-when they're supposed to occur,
-who's teaching what,
-and so on.
-
-To get started:
-
-1.  Install Django and its dependencies.
+The goal of this project is to build a web-based workshop administration
+application using Django to do that.  It is not intended to be visible to the
```
wking: Maybe just “using Django.".

README.md
```diff
@@ -1,76 +1,130 @@
...
+general public, or even to instructors (at least initially — we may add that
+later).
+
```
wking: I don't think you need the paragraph-break here.

README.md
```diff
@@ -1,76 +1,130 @@
...
+# Getting started
+
+1. Clone the repository:
+
+    ~~~
+    $ git clone https://github.com/swcarpentry/amy.git
+    $ cd amy
+    ~~~
+
+2. Install Django and its dependencies:
```
wking: Maybe update this to “Install Django and our other dependencies” now that there's more in `requirements.txt`.

README.md
```diff
@@ -1,76 +1,130 @@
...
 
     ~~~
     $ sudo python3 -m pip install -r requirements.txt
     ~~~
 
-2. Install [Bower](http://bower.io/) - the tool that manages Amy's JavaScript and CSS dependencies:
+    If you're experienced Python programmer, feel free to create a Python3-compatible virtualenv for Amy and install dependencies from `requirements.txt`.
+
+3. Install [Bower](http://bower.io/) — the tool that manages Amy's JavaScript and CSS dependencies:
```
wking: Is the canonical capitalization “Amy”?  I'd been using “amy”, but I'll switch if we want to use the capitalized form.  I'd recommend we *don't* use the capitalized form to avoid confusion with people named Amy, but I'm not the maintainer ;).

README.md
```diff
@@ -1,76 +1,130 @@
...
 
     ~~~
-    $ npm install -g bower
+    $ sudo npm install -g bower
```
wking: I prefer installing to a per-user location to avoid involving root.  See for example [my `config set prefix` example][1].

[1]: http://blog.tremily.us/posts/Node/

README.md
```diff
@@ -1,76 +1,130 @@
...
+    3. Get the code from the newest tag:
+
+        ~~~~
+        $ git checkout tags/<tag_name>
+        ~~~~
+
+2. Update dependencies:
+
+    ~~~
+    $ sudo python3 -m pip install --upgrade -r requirements.txt
```
wking: Again, I prefer `python3 -m pip install --user …` to avoid involving root.  I'd rather leave managing system packages to the system's package manager.

README.md
```diff
@@ -1,76 +1,130 @@
...
+    $ bower update
```
wking: This should happen automatically (if `bower.json` was updated) the next time you run `make serve`.

pbanaszkiewicz: We're using `bower install` in `Makefile:77`. Will it update packages if they're outdated?

wking: On Tue, May 12, 2015 at 12:50:16PM -0700, Piotr Banaszkiewicz wrote:
> > +    $ bower update
> 
> We're using `bower install` in `Makefile:77`. Will it update
> packages if they're outdated?

It will if you've bumped a dependency in your bower.json.  The only
difference will be if:

a. You have an existing install that already satisfies the bower.json
   dependency, and
b. There have been subsequent releases of that package that would also
   satisfy the bower.json dependency.

In that case, I see no need to update the local package.  If you do
have a reason to update the local package (e.g. there was a bug-fix in
a dependency), you should be bumping the bower.json listing.

If you periodically want to upgrade to current releases (e.g. because
you just assume upstream is fixing bugs that you don't know about),
then an update call is fine, but you can do that by hand.  And a move
like that would be part of amy development (you're forward-porting amy
to new dependencies), not part of a standard amy upgrade.

README.md
pbanaszkiewicz: While I agree with you that we should not use root account for security, I also know that to access `bin` under either local npm installation dir, or local pip installation dir, it would require changing $PATH, adding something to either .bashrc/.zshrc, etc, etc…

I'd leave `sudo` where it is and provide a link to some resources for people who choose not to use it.

What do you say?

README.md
pbanaszkiewicz: Okay, in this case I'll change that paragraph. Thanks!

README.md
wking: On Tue, May 12, 2015 at 01:01:14PM -0700, Piotr Banaszkiewicz wrote:
> >  
> >      ~~~
> > -    $ npm install -g bower
> > +    $ sudo npm install -g bower
> 

> While I agree with you that we should not use root account for
> security, I also know that to access `bin` under either local npm
> installation dir, or local pip installation dir, it would require
> changing $PATH, adding something to either .bashrc/.zshrc, etc, etc…

Right [1].  I guess that maybe makes it too complicated for our docs.

> I'd leave `sudo` where it is and provide a link to some resources
> for people who choose not to use it.

Fair enough.  Or just use sudo, and trust folks who don't like that to
lookup alternatives on their own.

[1]: https://github.com/wking/dotfiles-public/blob/980cb1cbaf67b9f060acb588d8984ce35347b42f/src/.bash_profile#L13

pbanaszkiewicz: Yeah, mine `$PATH` grew a little bit too:

```shell
$ echo $PATH
/home/piotr/.local/bin:/home/piotr/workspace/anaconda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/home/piotr/.rvm/bin
```

I'm thinking about adding a note about `npm` settings:

```markdown
**Note**: if you don't want to use `sudo`, you can install `bower` locally. You'll need to set up your `$PATH` correctly, though. Look [here](https://docs.npmjs.com/getting-started/fixing-npm-permissions#option-2-change-npm-s-default-directory-to-another-directory) for details.
```

And a note about virtualenvs (this one would replace https://github.com/swcarpentry/amy/pull/270/files#diff-04c6e90faac2675aa89e2176d2eec7d8R30):

```markdown
**Note**: if you don't want to use `sudo`, you can install `amy` into Python virtualenv. Look [here](https://virtualenv.pypa.io/en/latest/userguide.html) for details. Make sure you're using Python3-compatible virtualenv!
```

Looks okay?

README.md
pbanaszkiewicz: Git is using "Git" for project name, "git" for the tool name. Could we use "AMY" or would that look awkward? Maybe we could make it an abbreviation :)

I think I've seen Greg using capitalized "Amy". And why is this project called "Amy"?

Also, I think that our log-out text might need to be changed.

README.md
wking: On Tue, May 12, 2015 at 01:20:23PM -0700, Piotr Banaszkiewicz wrote:
> **Note**: if you don't want to use `sudo`, you can install `bower`
> locally. You'll need to set up your `$PATH` correctly, though. Look
> [here](https://docs.npmjs.com/getting-started/fixing-npm-permissions#option-2-change-npm-s-default-directory-to-another-directory)
> for details.

Looks good to me.

> **Note**: if you don't want to use `sudo`, you can install `amy`
> into Python virtualenv. Look
> [here](https://virtualenv.pypa.io/en/latest/userguide.html) for
> details. Make sure you're using Python3-compatible virtualenv!

No need for a virtualenv here, since Python loads stuff from your
~/.local by default.  Just use ‘pip install --user …’ with the
appropriate pip *or* use a virtualenv.  I'm not sure where Anaconda's
pip puts things if you use the virtualenv's pip and --user, but I
expect they have some sane default.

README.md
pbanaszkiewicz: I think turning "Amy" into "AMY" (abbrv) would be neat. But I'm out of ideas. For now I leave it as in original document ("Amy").

gvwilson: Please leave it as "Amy" - upper-casing "AMY" suggests that it's an 
acronym, and I'm too tired to try to reverse engineer a meaning for it.

Issue #269, opened 2015-05-12
Update development-ready database data with production database


pbanaszkiewicz:
The anonymizing script (`migrater.py`) @gvwilson wrote is not compatible with our production database. We're at the point when it's easier to gather development data from the Amy production database.

We can either update `migrated.py` or use something like [django-anonymizer](http://pythonhosted.org/django-anonymizer/) to blur our production data.

I'd like for this update to be the last big item on the TODO list for v0.2.

I'm looking for insight on either of two options I mentioned above.

wking:
On Tue, May 12, 2015 at 01:44:52AM -0700, Piotr Banaszkiewicz wrote:
> We can either update `migrated.py` or use something like
> django-anonymizer to blur our production data.

Or we can explicitly code up fixtures for testing (#184).  It looks
like I volunteered to write those fixtures once #134 lands, but that's
still open…

pbanaszkiewicz:
@wking:
> Or we can explicitly code up fixtures for testing (#184).

Yes. I liked the idea of anonymizer (since that's something similar to what Greg wrote), but was a little unsatisfied that we can unintentionally [erase our production database](http://pythonhosted.org/django-anonymizer/commands.html#anonymize-data) when using django-anonymizer.

> It looks like I volunteered to write those fixtures once #134 lands, but that's still open…

Which way do you like more?

wking:
On Tue, May 12, 2015 at 09:10:08AM -0700, Piotr Banaszkiewicz wrote:
> > It looks like I volunteered to write those fixtures once #134
> > lands, but that's still open…
> 
> Which way do you like more?

I'd like having fixtures that were just big enough to support our test
cases.  It looks like I cited #134 from #184 because I didn't feel
like excercising pagination was particulary important.  In that case,
I'm happy to start working on fixture-based tests this weekend or so,
if that sounds good to you.

pbanaszkiewicz:
@wking:
>I'd like having fixtures that were just big enough to support our test
cases.  It looks like I cited #134 from #184 because I didn't feel
like excercising pagination was particulary important.  In that case,
I'm happy to start working on fixture-based tests this weekend or so,
if that sounds good to you.

That actually sounds great!

Having fixtures:
1) would enable us to get rid of `migrater.py` and `db.sql`
2) would make each version release easier to handle (no need to update huge `db.sql`, just fixtures)
3) would be no risk (compared to having django-anonymizer)

I like it a lot. Thanks!

[...]

PR #273, opened 2015-05-14
Anonymizing production data for testing purposes.

gvwilson:
1.  Getting rid of old migration tools (no longer up to date).
2.  Building new tool to anonymize production database in `scripts/anonymizer.py`.
3.  Creating new `db.sql` by anonymizing current production database.

This addresses #269.


pbanaszkiewicz:
This anonymizer is a really nice piece of code.

If I were to make some changes in it, I'd not generate notes for every site/event (maybe a 5% chance for them to get notes? This way `db.sql` is ~20x smaller) and I'd rewrite `randword` to generate more legible output.

I don't consider these as urgent issues at all.

By the way, the use of randomization everywhere has some specific impact:

1. all slugs become invalid as soon as new development database is generated
2. it's way harder to track people this way
3. data seems illegible

To sum up, this looks solid.

PR #275, opened 2015-05-15
Show version (and possibly Git commit details) in footer

wking:
Make it easy for users to answer “what version of Amy is this?”
without admin action.  Details, such as they are, in the commit
messages.


pbanaszkiewicz:
Looks good. I'll merge with a slight change to styling and one DRY change.

Issue #271, opened 2015-05-13
Add all the other airports to the database


gvwilson:
We should add another few thousand airports to the database in one go so that we don't have to keep adding them one by one as we add new instructors.

gvwilson:
Close? We seem to be managing OK.

pbanaszkiewicz:
@maneesha do you feel like we're missing some airports from the database?

gvwilson:
We can add more airports as we need them.

Issue #62, opened 2014-12-14
Separate country table


wking:
Instead of trusting `airport.country` and `site.country` to avoid duplicates, its easier to just have a country table that can be referenced from other tables.  This also makes it easier to enforce canonical spelling (e.g. by requiring admin credentials to add new countries).

gvwilson:
@wking was right: we should add a country table so that when we're editing a site, we can select its country from a pulldown instead of typing it in.  These are the country names that match the flags we have:

~~~
Abkhazia
Afghanistan
Aland
Albania
Algeria
American-Samoa
Andorra
Angola
Anguilla
Antarctica
Antigua-and-Barbuda
Argentina
Armenia
Aruba
Australia
Austria
Azerbaijan
Bahamas
Bahrain
Bangladesh
Barbados
Basque-Country
Belarus
Belgium
Belize
Benin
Bermuda
Bhutan
Bolivia
Bosnia-and-Herzegovina
Botswana
Brazil
British-Antarctic-Territory
British-Virgin-Islands
Brunei
Bulgaria
Burkina-Faso
Burundi
Cambodia
Cameroon
Canada
Canary-Islands
Cape-Verde
Cayman-Islands
Central-African-Republic
Chad
Chile
China
Christmas-Island
Cocos-Keeling-Islands
Colombia
Commonwealth
Comoros
Cook-Islands
Costa-Rica
Cote-dIvoire
Croatia
Cuba
Curacao
Cyprus
Czech-Republic
Democratic-Republic-of-the-Congo
Denmark
Djibouti
Dominica
Dominican-Republic
East-Timor
Ecuador
Egypt
El-Salvador
England
Equatorial-Guinea
Eritrea
Estonia
Ethiopia
European-Union
Falkland-Islands
Faroes
Fiji
Finland
France
French-Polynesia
French-Southern-Territories
Gabon
Gambia
Georgia
Germany
Ghana
Gibraltar
GoSquared
Greece
Greenland
Grenada
Guam
Guatemala
Guernsey
Guinea-Bissau
Guinea
Guyana
Haiti
Honduras
Hong-Kong
Hungary
Iceland
India
Indonesia
Iran
Iraq
Ireland
Isle-of-Man
Israel
Italy
Jamaica
Japan
Jersey
Jordan
Kazakhstan
Kenya
Kiribati
Kosovo
Kuwait
Kyrgyzstan
Laos
Latvia
Lebanon
Lesotho
Liberia
Libya
[...]

PR #276, opened 2015-05-16
Script to show location of airports for addition to database

gvwilson:
Adding all airports (#271) is impractical without a better database, so this script allows easy lookup of airports when/as they need to be added.


pbanaszkiewicz:
Why do you think it's impractical?

gvwilson:
The data we have would take a lot of work to tidy up.  For example, I've already fixed up countries listed as "UK", "United Kingdom", and "England" to all be "United-Kingdom", and "US" and "USA" to be "United-States"; I'm sure there will be more of those before we can start to think sensibly about translating country names to refs to the country table, and then we have to sort out airport naming vs. city naming, etc.  With over 9000 entries, I'm figuring 9 hours of work (?).

pbanaszkiewicz:
Do you think we should rework the way we handle airports to use this list only?

gvwilson:
I think every DB ref to an airport is already a foreign key reference to the `Airport` table, no?

pbanaszkiewicz:
@gvwilson what do you say if we migrate to [`django-airports`](https://github.com/bashu/django-airports)? Similarily as we migrate to `django-countries` in #62.

@wking any thoughts?

gvwilson:
If they have better data than ours, sure.

wking:
On Thu, May 21, 2015 at 10:39:11PM -0700, Piotr Banaszkiewicz wrote:
> @gvwilson what do you say if we migrate to django-airports?
> Similarily as we migrate to `django-countries` in #62.

That looks like a solid package, code-wise.  It's a bit young (just a
few months old), but it has a few forks and a number of folks who've
starred it.  I'd be happier if there were more contributors or issue
activity, but a bunch of stars and few issues isn't a bad combination
;).  So I wouldn't make a serious bet on its long-term stability, but
I think it's safe enough for our use-case.

And the more functionality we can offload to stable, generic, external
packages, the easier it will be to maintain the Amy-specific code, so
+1 from me to migrate.

gvwilson:
OK, go for it.

pbanaszkiewicz:
What I'm a little worried is that django-airport requires spatial-enabled database, and it's a little bit tricky to set up in case of SQLite.

I need some time to think about this issue. It's likely that django-airports can be used only after we completely migrate Amy to Rackspace.

wking:
On Wed, May 27, 2015 at 11:11:43AM -0700, Piotr Banaszkiewicz wrote:
> What I'm a little worried is that django-airport requires
> spatial-enabled database, and it's a little bit tricky to set up in
> case of SQLite.

Are you having problems with SpatiaLite [1,2]?

[1]: https://www.gaia-gis.it/fossil/libspatialite/index
[2]: https://docs.djangoproject.com/en/dev/ref/contrib/gis/db-api/#module-django.contrib.gis.db.backends

pbanaszkiewicz:
I didn't have problems installing in on my own machine but:

1. I use Ubuntu (spatialite comes in as a package), and some devs certainly use Macs
2. I checked briefly for spatialite and it's not installed on our current VPS. From @drio's experience with Small Orange: they'd request money for installing spatialite or deny installing it at all.

gvwilson:
Last time I tried to install SpatiaLite on my Mac it took a while - 
making that a must-have for people doing dev worries me a little. Let's 
defer this one until more pressing things are fixed up?

pbanaszkiewicz:
So the resolution is:

1. don't switch to django-airports
2. don't use the script proposed by @gvwilson in this PR

am I correct?

gvwilson:
Yes to both - we'll add airports manually for now, then come back after 
0.4 and have another look.

wking:
On Wed, May 27, 2015 at 10:17:55PM -0700, Piotr Banaszkiewicz wrote:
> I didn't have problems installing in on my own machine but:
> 
> 1. I use Ubuntu (spatialite comes in as a package), and some devs
> certainly use Macs

Docker FTW ;).  I don't think requiring Docker is too high a bar,
especially with our limited dev pool.

> 2. I checked briefly for spatialite and it's not installed on our
> current VPS. From @drio's experience with Small Orange: they'd
> request money for installing spatialite or deny installing it at
> all.

I'm fine punting on this until we move to a more serious hosting
company :p.

pbanaszkiewicz:
@wking:
> Docker FTW ;)

Good point. Especially now as they added more tools to the set. I can even see [an easy way to set up Docker on OS X](https://docs.docker.com/kitematic/).

> I'm fine punting on this until we move to a more serious hosting company :p.

I've actually migrated to Rackspace, but apparently no-one wants to update the DNS records…

gvwilson:
I'll ask Jon P. to update DNS next week - just want to be on the ground 
in Toronto with him in case anything goes sour.

pbanaszkiewicz:
I talked with an admin and he suggested this workflow:

1. Change TTL of the domain record (`amy.software-carpentry.org`) to something small, like 60
2. Wait 1 day
3. Change the record
4. Wait 1 day to see if everything's okay (if it's not you can quickly rollback thanks to the small TTL )
5. Change the TTL back to previous value

wking:
On Thu, May 28, 2015 at 10:40:13PM -0700, Piotr Banaszkiewicz wrote:
> I talked with an admin and he suggested this workflow…

Sounds good to me.  And the current pool of users is small enough that
in the unlikely event of a catastrophic DNS mixup, you can have folks
add /etc/hosts entries (or their equivalent on other OSes) to resolve
amy.software-carpentry.org while the DNS gets straigtened out.

gvwilson:
This weekend would be a good time to make the change (Amy-the-person 
isn't going to dig in until Monday).  If you're able to do it then, 
please mail her, Giacomo, Aleksandra, and Jon P to confirm.

wking:
On Fri, May 29, 2015 at 03:24:54AM -0700, Greg Wilson wrote:
> Amy-the-person isn't going to dig in until Monday.

This is the main reason I prefer the lowercase ‘amy’ for both the
project and the resulting code ;).

gvwilson:
You reverse engineer a meaning such that 'amy' is an acronym, I'll 
switch to lower case :-)

wking:
On Fri, May 29, 2015 at 11:21:07AM -0700, Greg Wilson wrote:
> You reverse engineer a meaning such that 'amy' is an acronym, I'll
> switch to lower case :-)

If it was an acronym I'd be suggesting uppercase ;).

pbanaszkiewicz:
'AMY' - AdMinister Yourself?

gvwilson:
I like it.

PR #284, opened 2015-05-19
Stripping old stuff out of Makefile

gvwilson:
We no longer need to import data from the old database, and we're generating reports programmatically now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,8 @@
 # How to run Python.
 PYTHON = python3
 
-# Where original data lives.
-SRC_DIR = ~/s/admin
-
-# Where original database lives (only on Software Carpentry administrator's machine).
-SRC_DB = ~/s/admin/roster.db
-
 # Database used by this application.
 APP_DB = db.sqlite3
-
-# Saved SQL for regenerating the database.
-APP_SQL = db.sql
-
-# Where event notes live.
-SRC_EVENTS = ${SRC_DIR}/bootcamp-notes/archive
-
-# Where site notes live.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -40,15 +25,6 @@ migrations :
 	${PYTHON} manage.py makemigrations
 	${PYTHON} manage.py migrate
 
-## import       : import and save legacy data
-import :
```
pbanaszkiewicz: If this is removed, I believe we have to rework `README.md` a little, too.

Makefile
```diff
@@ -1,24 +1,9 @@
...
-SRC_DIR = ~/s/admin
-
-# Where original database lives (only on Software Carpentry administrator's machine).
-SRC_DB = ~/s/admin/roster.db
-
 # Database used by this application.
 APP_DB = db.sqlite3
 
-# Saved SQL for regenerating the database.
-APP_SQL = db.sql
```
pbanaszkiewicz: You should not remove this.

README.md
```diff
@@ -39,23 +39,7 @@ when they're supposed to occur, who's teaching what, and so on.
...
-
-            ~~~
-            $ make import
-            ~~~
-
-    2. Otherwise set up the redacted (development-ready) data with:
+4. Setup your local database wit redacted (development-ready) data with:
 
         ~~~
         $ make database
```
pbanaszkiewicz: Can you deindent this block (`~~~ $make database ~~~`)? Otherwise it's not rendered properly. Thanks

PR #288, opened 2015-05-22
Use django-countries

pbanaszkiewicz:
This solves #62.

Features:

* Data migration for old country fields (Site.country, Airport.country: shorten country names to [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1)). Not all countries can be migrated this way, but I used production database to get list of the countries and provided a mapping for them myself.
* "Normal" migration.
* Display of country flags in site/airport detail pages.
* Minor style fixes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ PyYAML
 requests>=2.0
 django-crispy-forms>=1.4.0
 django-selectable>=0.9.0,<1.0
+django-countries==3.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ PyYAML
 requests>=2.0
 django-crispy-forms>=1.4.0
 django-selectable>=0.9.0,<1.0
-django-countries==3.3
+django-countries>=3.3,<4.0
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -80,6 +80,7 @@ INSTALLED_APPS = (
     'django.contrib.admin',
     'crispy_forms',
     'selectable',
+    'django_countries',
 )
 
 CRISPY_TEMPLATE_PACK = 'bootstrap3'
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -12,6 +12,8 @@ https://docs.djangoproject.com/en/1.7/ref/settings/
 import os
 import json
 
+from django.utils.translation import ugettext_lazy as _
+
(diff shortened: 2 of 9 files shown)


pbanaszkiewicz:
I need to figure out what do we do with flags before this gets merged.

requirements.txt
```diff
@@ -2,3 +2,4 @@ Django>=1.8,<1.9
 PyYAML
 requests>=2.0
 django-crispy-forms>=1.4.0
+django-countries==3.3
```
wking: Can we do `>=3.3,<4.0`?  I don't see a need to hard-code a specific version, since all we need is an implementation that supports their 3.x API.

pbanaszkiewicz: Yes we can!

wking:
On Fri, May 22, 2015 at 11:21:24AM -0700, Piotr Banaszkiewicz wrote:
> This solves #62.

Modulo the backwards-compatibility issues we're discussing in #62,
this PR looks good to me.

pbanaszkiewicz:
This is a little bit more complicated than I originally thought, because to switch to ISO 3166-1 we'd have to:

* change https://github.com/swcarpentry/amy/blob/master/workshops/check.py
* change https://github.com/swcarpentry/site/blob/master/bin/workshops.py
* convert https://github.com/swcarpentry/site/blob/master/config/archived.yml

There's one thing I still don't understand.

Do we import or export events to/from Amy? I think not. Then there's no need to change swcarpentry/site files because event's country representation is 100% internal to Amy.

I can provide mappings to/from ISO 3166-1 so that we can use them if we get an importer.

What are your opinions, @wking and @gvwilson?

gvwilson:
You are correct, we don't currently export from Amy for use in site, 
though we might in future.  Consistency in country identifiers between 
Amy and site is mostly for human benefit (i.e., simpler for me if 
there's One Way to Do It).  Given that, work on translating between the 
two tools can wait.

pbanaszkiewicz:
@gvwilson:
>You are correct, we don't currently export from Amy for use in site, though we might in future. Consistency in country identifiers between Amy and site is mostly for human benefit (i.e., simpler for me if there's One Way to Do It). Given that, work on translating between the two tools can wait.

So is it +1 on merging this PR?

gvwilson:
Go for it.

PR #290, opened 2015-05-23
Regularizing use of names 'badge' and 'award'.

gvwilson:
The "all badges" page displays all the badges we can hand out.  However, the "badge details" page was displaying particular awards, not the details of the kind of badge (such as the criteria).  This inconsistent naming became a stumbling block when I wanted to add a new form to let us award a badge (e.g., give someone an instructor badge).  This PR changes things so that the "all badges" page links to pages showing who has been awarded particular badges.  Once this is merged, we can add another form that will allow for creation of new awards (i.e., assignment of badges to persons).

### DIFF
--- a/workshops/templates/workshops/all_awards.html
+++ b/workshops/templates/workshops/all_awards.html
@@ -8,14 +8,14 @@
 {% endblock %}
 
 {% block content %}
-{% if all_awards %}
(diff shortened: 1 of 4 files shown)


pbanaszkiewicz:
Looks & works OK.

PR #295, opened 2015-05-23
Trying to award badges

gvwilson:
* The URL for the page that displays instructor badges is `awards/instructor`
* This is very similar in structure to `awards/add`, which identifies the page for adding a new badge, but we can avoid ambiguity by putting `awards/add` ahead of the regexp-based URL for displaying particular awards.  (This problem doesn't come up with things like tasks because we display them all mixed together rather than trying to segregate by type.)
* However, `awards/add` invokes the `AwardCreate` class...
* ...which uses `breadcrumb_index_all_objects` from line 45 of `workshops/templatetags/breadcrumbs.py`, which expects to be able to look up the "display all" page as `all_whatever`...
* ...which doesn't work because `all_awards` needs a badge name.

We *could* modify `all_awards` to allow an empty badge name, and just display all awards, but that looks pretty messy - and we're going to want to modify `all_tasks` to display particular kinds of tasks (e.g., only show me instructors, only show me learners), so we should come up with a generic solution.

@wking @pbanaszkiewicz any suggestions?

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -406,3 +406,6 @@ class Award(models.Model):
 
     def __str__(self):
         return '{0}/{1}/{2}/{3}'.format(self.person, self.badge, self.awarded, self.event)
+
+    def get_absolute_url(self):
+        return reverse('award_details', kwargs={'award_id': self.id})
--- a/workshops/templates/base.html
+++ b/workshops/templates/base.html
@@ -26,7 +26,7 @@
               <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false">More <span class="caret"></span></a>
               <ul class="dropdown-menu" role="menu">
                 {% navbar_element "Tasks" "all_tasks" %}
-                {% navbar_element "Badges" "all_badges" %}
+                {% navbar_element "Awards" "all_awards" %}
                 {% navbar_element "Airports" "all_airports" %}
                 <li class="divider"></li>
                 {% navbar_element "Bulk add persons" "person_bulk_add" %}
--- a/workshops/templates/base.html
+++ b/workshops/templates/base.html
@@ -59,6 +59,7 @@
                 {% navbar_element "Add event" "event_add" %}
                 {% navbar_element "Add site" "site_add" %}
                 {% navbar_element "Add person" "person_add" %}
+                {% navbar_element "Add award" "award_add" %}
               </ul>
             </li>
             <li class="dropdown">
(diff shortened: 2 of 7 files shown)


pbanaszkiewicz:
# First way

1. `badges` for all badges
2. `award/add` (singular instead of plural) with not generic form template (because we need to link to index with `badge_name` parameter)
3. `awards/badge_name` to list all awards of the badge `badge_name`

# Second way

1. `badges` for all badges
2. `badge/badge_name/` for details about the badge `badge_name`
3. `badge/badge_name/edit` to edit the badge `badge_name`
4. `badge/badge_name/awards` to list all awards of the badge `badge_name`
5. `badges/add` to add a new badge
5. `awards/add` to add a new award (not generic form template)

I'm thinking that the latter option is better, but maybe @wking has a nicer solution?

gvwilson:
PR updated to:

1.  Display properties of all badges on a single page.
2.  Display all awards on a single page (without allowing us to select by type).
3.  Allow addition of a single award (which is the most important thing right now).
4.  Display details of a single award (needed by 3 above).

This does *not* allow us to filter awards by type, but right now, we can't filter tasks by type either.  I recommend that we merge this, then figure out how to handle filtering of tasks, awards, and events (by tags) more generally.

wking:
On Sat, May 23, 2015 at 02:21:04PM -0700, Piotr Banaszkiewicz wrote:
> I'm thinking that the latter option is better, but maybe @wking has
> a nicer solution?

I haven't looked over the code for this PR, but the way I like to
handle this in REST APIs is:

 * /{plural} to get multiple results, with query parameters for
   filtering (/{plural}?{param}={value}).  In this case, that would
   look like /awards?badge=badge_name to show all awards of the
   ‘badge_namde’ badge.  I also POST here for creating new entries.
* /{singular}/{id} to manage a single result, e.g. the details about a
  given badge.  With GET to see the details, PUT to update the values,
  and DELETE to remove an entry.

In a form-driven application, you may want separate GET views for
“display the details for this object” and “display an edit form for
this object”, so you can split it out into:

* /{singular}/{id} to see the detail view.
* /{singular}/edit/{id} to see (with GET) and submit (with POST) the
  form for editing the object.

To get more compact namespacing, some folks prefer to use only the
plural form or the singular form.  For example, if you want to add
logic to an Nginx reverse-proxy to put basic auth in front of your
badge endpoints, the split plurality form requires logic to match both
/badge and /badges.  That makes sense to me, but I think the “plural
for queries and singular for manipulating a single object” approach is
useful enough to API consumers (of which there are hopefully many) to
be worth the cost to your Nginx admin (which is probably just you ;).

Re, @gvwilson's “I recommend that we merge this, then figure out how
to handle filtering of tasks, awards, and events (by tags) more
generally”, that seems fine with me (but again, I haven't looked at
the PR itself).

gvwilson:
Should we close this in favor of #303?

pbanaszkiewicz:
I think so.

PR #296, opened 2015-05-24
Cleaning up the interface a bit.

gvwilson:
1.  Buttons to add things are redundant now that we have a pull-down 'add' menu.
2.  Breadcrumbs show page titles, so separate page titles are redundant.

### DIFF
--- a/workshops/templates/base.html
+++ b/workshops/templates/base.html
@@ -59,6 +59,7 @@
                 {% navbar_element "Add event" "event_add" %}
                 {% navbar_element "Add site" "site_add" %}
                 {% navbar_element "Add person" "person_add" %}
+		{% navbar_element "Add airport" "airport_add" %}
               </ul>
             </li>
             <li class="dropdown">
--- a/workshops/templates/base_without_navbar.html
+++ b/workshops/templates/base_without_navbar.html
@@ -19,7 +19,6 @@
   <body>
     {% block navbar %}{% endblock %}
     <div class="container">
-      {% if title %}<h1>{{ title }}</h1>{% endif %}
       <ol class="breadcrumb">{% block breadcrumbs %}{% endblock %}</ol>
       {% if messages %}
       <ul class="messages">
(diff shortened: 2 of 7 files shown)


pbanaszkiewicz:
I'm -1 to these changes. I'm pretty sure we shouldn't get rid of header titles, and probably we shouldn't get rid of the buttons, too.

We could consider removing the buttons if they're somehow misleading or confusing the users.

gvwilson:
1.  I should have submitted these in separate PRs.

2. Maybe get the breadcrumbs to show everything *except* the page you're 
now on?  I just find it redundant to have the page title show up in both.

3. At a minimum, we should use buttons everywhere, or links everywhere 
(for add, edit, delete).

pbanaszkiewicz:
Re: 2. I think that's why breadcrumbs have the latest element inactive.

Re: 3. I agree (buttons), may I provide a UI enhancement PR that deals with links/buttons?

gvwilson:
On 2015-05-24 7:47 AM, Piotr Banaszkiewicz wrote:
>
> Re: 2. I think that's why breadcrumbs have the latest element inactive.
>
Still find it annoying.

> Re: 3. I agree (buttons), may I provide a UI enhancement PR that deals 
> with links/buttons?
>
Yes please and thank you.

gvwilson:
Rejected.

pbanaszkiewicz:
I took a look at other big websites and how they do breadcrumbs and you're right, @gvwilson.

They don't repeat current site at the end of breadcrumbs.

This means we'd only have breadcrumbs at most 3 links long: `Amy / All * / * [ / action ]`, for example:

Was:
* Amy / All events / Event 2015-05-25-something / Edit

Will be:
* Amy / All events / Event 2015-05-25-something

But this does not bug me. In case of `All *` pages, we can just as well drop breadcrumbs (because they'd look like "Amy / ").

So I don't really know what to do:

1. Display breadcrumbs on the same pages as now, but hide the last item.
2. Display breadcrumbs on the pages that would have more than 1 breadcrumbs.
3. Drop breadcrumbs completely.

I propose to revisit this later. I'll open an issue so that we don't forget.

Issue #321, opened 2015-05-28
Add a "checkpoint this" button


gvwilson:
It would be reassuring to have a "make a backup of the database" button that we could press before doing big chunks of work (like uploading attendance figures from multiple workshops).

pbanaszkiewicz:
I'm not sure about this. Wouldn't you rather run manually the backup scipt (#168)?

gvwilson:
Sure, but the admins need to be able to launch it, and they're not going 
to have shell access to the server.  If the 'checkpoint' button just 
kicks off the script, that should be fine (?).

wking:
On Thu, May 28, 2015 at 04:19:26AM -0700, Greg Wilson wrote:
> It would be reassuring to have a "make a backup of the database"
> button that we could press before doing big chunks of work (like
> uploading attendance figures from multiple workshops).

“Every 15 minutes” from #168 seems frequent enough for me, so I'm not
sure we need a way to fire off additional backups manually.

gvwilson:
If it actually is every 15 minutes, yes - is it now?

wking:
On Thu, May 28, 2015 at 12:07:18PM -0700, Greg Wilson wrote:
> If it actually is every 15 minutes, yes - is it now?

In the example @pbanaszkiewicz posted it looks like it's “every day at
midnight” [1].  You can bump that up to “every 15 minutes” with:

  */15 * * * * …

but as I mention in #168, I think it's better to use Btrfs snapshots
or some other de-duping approach at that frequency.  Still, time spent
in that direction seems more useful to me than time spent adding and
using a manual trigger.

[1]: https://github.com/swcarpentry/amy/issues/168#issuecomment-99579041

pbanaszkiewicz:
Greg wants to have 96 backups per day for next 2 weeks.

I proposed this solution:

1. Set up 2 Rackspace VPS (in different locations) to gather backups every 15 minutes (I don't have enough space on my laptop for that)
2. Keep getting full backups every day to my and Greg's computer (assuming he's still running the cronjob for that).

I'm thinking about setting `rsync --backup` for point 1 (faster solution for me than getting to know Btrfs).

wking:
On Mon, Jun 01, 2015 at 03:56:28PM -0700, Piotr Banaszkiewicz wrote:
> 1. Set up 2 Rackspace VPS (in different locations) to gather backups
>    every 15 minutes (I don't have enough space on my laptop for that)
> …
[...]

PR #314, opened 2015-05-28
Starting on contributor guidelines

gvwilson:

PR #279, opened 2015-05-18
Creating reports

gvwilson:
Starting to work on #260.

PR #322, opened 2015-05-28
Producing YAML report of all instructor activity

gvwilson:
We need this to generate email to instructors to check database contents - we'll pipe the YAML to a separate Python script that sends the emails.

### DIFF
--- a/scripts/mail-all-activity.py
+++ b/scripts/mail-all-activity.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python
+
+'''Mail people to check their activity based on YAML dump of Amy database.'''
+
+import sys
+import yaml
+
+
+USAGE = 'Usage: mail-all-activity [-r|--real] [filename]'
+
+BODY = '''We would like to bring our records up to date, and would be grateful
+if you could check the information below and send us corrections or additions.
+
+name: {name}
+preferred email: {email}
+became instructor: {became_instructor}
+{optional}
+{skills}
+{tasks}'''
+
+
+def main(prog_name, args):
+    '''Main driver.'''
+
+    # Default is dummy run - only actually send mail if told to.
+    sender = display
+    if args[0] in ('-r', '--real'):
+        sender = send
+        args = args[1:]
+
+    # If no filename provided, use stdin.
+    if len(args) == 0:
+        process(sender, sys.stdin)
+    elif len(args) == 1:
+        with open(args[0], 'r') as reader:
+            process(sender, reader)
+    else:
+        fail(USAGE)
+
+
+def process(sender, reader):
+    '''Process the YAML data loaded through reader.'''
+    info = yaml.load(reader)
+    for record in info:
+        address, subject, body = make_message(record)
+        sender(address, subject, body)
+
+
+def make_message(record):
+    '''Construct address, subject, and body of message.'''
+
+    assert 'optional' not in record, \
+           'Keyword "optional" should not be a key in records'
+    assert 'skills' not in record, \
+           'Keyword "skills" should not be a key in records'
+
+    address = record['email']
+
+    subject = 'Checking your Software Carpentry information'
+
+    record['optional'] = ''
+    for key in ('airport', 'twitter', 'github'):
+        record['optional'] += format_optional(record, key)
+
+    record['skills'] = format_skills(record['can_teach'])
+
+    record['tasks'] = format_tasks(record['tasks'])
+
+    body = BODY.format(**record)
+    return address, subject, body
+
+
+def format_optional(record, key):
+    '''Format an optional entry for inclusion in the message body.'''
+
+    if record.get(key, None):
+        return 'Your {0} is "{1}"\n'.format(key, record[key])
+    else:
+        return 'Your {0} is unknown\n'.format(key)
+
+
+def format_skills(skills):
+    '''Format zero or more skills for inclusion in the message body.'''
+
+    if skills:
+        return 'You can teach: {0}\n'.format(', '.join(skills))
+    else:
+        return ''
+
+
+def format_tasks(tasks):
+    '''Format tasks and collaborators for inclusion in the message body.'''
+
+    result = ''
+    for t in tasks:
+        entry = 'You were a {0} at {1} with {2}\n'.format(
+            t['role'],
+            t['slug'],
+            ', '.join(t['others']))
+        result += entry
+    return result
+
+
+def display(address, subject, body):
+    '''Display a message that is not being sent.'''
+    print('To:', address)
+    print('Subject:', subject)
+    print(body)
+
+
+def send(address, subject, body):
+    '''Send email.'''
+    fail('Not yet able to send')
+
+
+def fail(msg):
+    '''Halt and catch fire.'''
+    print(msg, file=sys.stderr)
+    sys.exit(1)
+
+
+if __name__ == '__main__':
+    main(sys.argv[0], sys.argv[1:])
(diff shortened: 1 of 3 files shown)


wking:
On Thu, May 28, 2015 at 06:21:23AM -0700, Greg Wilson wrote:
> We need this to generate email to instructors to check database
> contents…

Is this one-off?  Do we want folks doing periodic reviews of database
contents via email.  Personally I'd just live with the current,
unaudited content until we get the per-instructor logins discussed in
#70 (e.g. [1]).

[1]: https://github.com/swcarpentry/amy/issues/321

gvwilson:
Asking people to remember to check is less likely to work than poking 
them annually to say, "Is this right?"

wking:
On Thu, May 28, 2015 at 12:08:00PM -0700, Greg Wilson wrote:
> Asking people to remember to check is less likely to work than
> poking them annually to say, "Is this right?"

Hmm.  Maybe have Amy mail them with a link to their page when we
record activity on their behalf (once they can login to that page)?
That way we're pushing notifications (which they can opt-out of via
Person.may_contact), we have more detail about what changed, and we
get faster feedback on bad data than waiting until the next yearly
cycle.  And we offload some of the fixup data-entry from admins to the
instructors themselves.

gvwilson:
I'd like to merge this one - we may change the details of what we produce, but a semi-annual ping asking people to update information is likely to recur.

scripts/mail-all-activity.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def fail(msg):
+    '''Halt and catch fire.'''
+    print(msg, file=sys.stderr)
+    sys.exit(1)
+
+
+if __name__ == '__main__':
+    main(sys.argv[0], sys.argv[1:])
```
pbanaszkiewicz: I'd rather pass whole `sys.argv` into `main`, and handle `[0]`, `[1]`, etc. in there, because that's where you check if `argv` is long enough.

scripts/mail-all-activity.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def main(prog_name, args):
+    '''Main driver.'''
+
+    # Default is dummy run - only actually send mail if told to.
+    sender = display
+    if args[0] in ('-r', '--real'):
+        sender = send
+        args = args[1:]
```
pbanaszkiewicz: I'd move this `if` into the chunk of code that checks length of `args` (lines 34-36).

scripts/mail-all-activity.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # If no filename provided, use stdin.
+    if len(args) == 0:
+        process(sender, sys.stdin)
+    elif len(args) == 1:
+        with open(args[0], 'r') as reader:
+            process(sender, reader)
```
pbanaszkiewicz: Check if `len(args)` is at least 1, and then check `args[0] in ('-r', '--real')`.

scripts/mail-all-activity.py
```diff
@@ -0,0 +1,123 @@
...
+'''Mail people to check their activity based on YAML dump of Amy database.'''
+
+import sys
+import yaml
+
+
+USAGE = 'Usage: mail-all-activity [-r|--real] [filename]'
+
+BODY = '''We would like to bring our records up to date, and would be grateful
+if you could check the information below and send us corrections or additions.
```
pbanaszkiewicz: I'd change the message a little. I'd add "what to do if my skills, email, or tasks don't match", and "what if everything's okay".

scripts/mail-all-activity.py
```diff
@@ -0,0 +1,123 @@
...
+    '''Construct address, subject, and body of message.'''
+
+    assert 'optional' not in record, \
+           'Keyword "optional" should not be a key in records'
+    assert 'skills' not in record, \
+           'Keyword "skills" should not be a key in records'
+
+    address = record['email']
+
+    subject = 'Checking your Software Carpentry information'
```
pbanaszkiewicz: I'd move this next to `BODY` on top of the file, so it serves as a constant, too.

scripts/mail-all-activity.py
```diff
@@ -0,0 +1,123 @@
...
+
+def format_tasks(tasks):
+    '''Format tasks and collaborators for inclusion in the message body.'''
+
+    result = ''
+    for t in tasks:
+        entry = 'You were a {0} at {1} with {2}\n'.format(
+            t['role'],
+            t['slug'],
+            ', '.join(t['others']))
```
pbanaszkiewicz: Not sure how `others` is processed on the backend side, but I've seen events with dozens of tasks. I'd take a lot of space to include `others` in the message.

scripts/mail-all-activity.py
```diff
@@ -0,0 +1,123 @@
...
+def display(address, subject, body):
+    '''Display a message that is not being sent.'''
+    print('To:', address)
+    print('Subject:', subject)
+    print(body)
+
+
+def send(address, subject, body):
+    '''Send email.'''
+    fail('Not yet able to send')
```
pbanaszkiewicz: Just a suggestion: you can use `NotImplementedError`. `fail` here is not bad, though.

workshops/management/commands/report_all_instructor_activity.py
```diff
@@ -0,0 +1,52 @@
...
+        if len(args) != 0:
+            raise CommandError('Usage: report_all_instructor_activity')
+
+        instructor_badge = Badge.objects.get(name='instructor')
+
+        result = []
+        all_awards = Award.objects.filter(badge=instructor_badge)
+        for award in all_awards:
+            person = award.person
+            if not person.email:
```
pbanaszkiewicz: What about `can_contact`?

pbanaszkiewicz: I think we can send this periodic email and add "Do you still don't want to hear from us?", but that's somewhat contrary to what `can_contact` says.

workshops/management/commands/report_all_instructor_activity.py
```diff
@@ -0,0 +1,52 @@
...
+                continue
+            tasks = Task.objects.filter(person=person)\
+                                .filter(Q(role=self.instructor_role) | Q(role=self.helper_role))
+            tasks = [{'slug' : t.event.slug,
+                      'start' : t.event.start,
+                      'role' : t.role.name,
+                      'others' : self.others(t.event, person)}
```
pbanaszkiewicz: See my comment about others: https://github.com/swcarpentry/amy/pull/322/files#r31453621

workshops/management/commands/report_all_instructor_activity.py
```diff
@@ -0,0 +1,52 @@
...
+                     for t in tasks]
+            can_teach = self.qualifications(person)
+            record = {'name' : person.get_full_name(),
+                      'email' : person.email,
+                      'airport' : person.airport.iata if person.airport else None,
+                      'github' : person.github,
+                      'twitter' : person.twitter,
+                      'can_teach' : can_teach,
```
pbanaszkiewicz: I think `mail-all-activity.py` uses `skills`, not `can_teach`.

pbanaszkiewicz:
I added a few comments and suggestions.

All in all I understand the code and I'm fine with it, except:

1. it violates some PEP8 rules, and we're just about to merge [contributor guidelines](https://github.com/swcarpentry/amy/pull/314)
2. it's incomplete (missing implementation of `send` method – what do you want to do about it, @gvwilson?)

gvwilson:
Closed in favor of #279 (after integrating feedback from @pbanaszkiewicz).

PR #303, opened 2015-05-26
Awarding badges

pbanaszkiewicz:
This is my take on awarding badges.

Only point 2 from https://github.com/swcarpentry/amy/issues/37#issuecomment-99558308
as I find point 1 more tricky.

This is WIP: I'm waiting for #285 to be merged so that I can rework this.

I have Greg's blessing to merge it as soon as I rebase against #285.

PR #324, opened 2015-05-30
Renaming Event.fee_paid to Event.invoiced.

gvwilson:
It isn't Amy's job to track whether money has arrived - that's the job of NumFOCUS's accounting system.  Instead, we should track whether an invoice has been sent, i.e., have we handed responsibility off to NumFOCUS.  This change of name makes those semantics clearer.


pbanaszkiewicz:
We'll have a migration conflict in #303, but that's not a big deal.

From what I can see, this PR looks good.

My suggestion is to split future PRs into more commits.

Issue #323, opened 2015-05-28
Add a 'notes' field to persons


gvwilson:
We may want to record extra information about instructors - @tracykteal has requested.

wking:
On Thu, May 28, 2015 at 08:17:57AM -0700, Greg Wilson wrote:
> We may want to record extra information about instructors -
> @tracykteal has requested.

This sounds reasonable to me, but I find it useful to at least have a
guess at the sort of information you want to store before adding a
field like this.  For example, see [1], where I argue for a structured
approach to storing waitlist-y information.

[1]: http://lists.software-carpentry.org/pipermail/amy_lists.software-carpentry.org/2015-February/000064.html

gvwilson:
The problem is that we don't know what we'll want to store - is someone 
only able to teach outside the academic year in Sweden?  Do they have a 
nut allergy?  Etc.

tracykteal:
Right, the idea for this field is that it's the miscellaneous information that we want to keep track of, so basically a free text field. Is this reasonable?

wking:
On Thu, May 28, 2015 at 12:06:55PM -0700, Greg Wilson wrote:
> The problem is that we don't know what we'll want to store - is
> someone only able to teach outside the academic year in Sweden?  Do
> they have a nut allergy?  Etc.

So before adding any data like that to an unstructured field, I want
to understand what the workflow around it is going to look like.  If
we store nut allergies in Person.notes, how does that get used?  Is it
someone's job to go through and check the notes fields before each
event to look for possible allergy conditions?  Maybe we want an
'Accessability' table that's many-to-many for people and sites, so we
can automatically check that we aren't assigning elevator-requiring
instructors to stair-only sites.  That's a bit trickier for one-off
things like menus (I don't expect nut-factory sites), but it would at
least make it easy for a host to see the union of all instructor (and
student?) accessability requirements automatically.

For “can only teach outside the academic year in Sweden”, I'd consider
modeling free/busy time following RFC 3.6.4 [1] and then feed that
into an iCalendar library to figure out which instructors were free
for a particular time slot.

[1]: https://tools.ietf.org/html/rfc5545#section-3.6.4

wking:
On Thu, May 28, 2015 at 12:19:46PM -0700, Tracy Teal wrote:
> Right, the idea for this field is that it's the miscellaneous
> information that we want to keep track of, so basically a free text
> field. Is this reasonable?

It's easy to add, but in almost every scenario I play through in my
[...]

PR #327, opened 2015-05-31
Add `Person.notes`

pbanaszkiewicz:
Solves #323.

One bug was discovered and fixed, see commits in this PR.

### DIFF
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -404,10 +404,12 @@ class PersonCreate(LoginRequiredMixin, CreateViewContext):
 
 class PersonUpdate(LoginRequiredMixin, UpdateViewContext):
     model = Person
-    # don't display the 'password' field, reorder fields
+    # don't display the 'password', 'user_permissions', 'group_permissions'
+    # fields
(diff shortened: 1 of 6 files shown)


workshops/views.py
```diff
@@ -407,7 +407,7 @@ class PersonUpdate(LoginRequiredMixin, UpdateViewContext):
     # don't display the 'password' field, reorder fields
     fields = ['personal', 'middle', 'family', 'username', 'may_contact',
               'email', 'gender', 'airport', 'github', 'twitter', 'url',
-              'is_superuser', 'user_permissions']
```
gvwilson: No longer showing `user_permissions` ?  I'm +1 on that in this PR (though it's separate from adding notes), but the comment at the top about not displaying `password` should be updated, and `reorder fields` can probably be removed. Otherwise +1 to merge.

pbanaszkiewicz:
@gvwilson amended that commit. Green light to merge?

gvwilson:
:+1: to merge and deploy.

Issue #315, opened 2015-05-28
Have separate columns for person and email in display of badges


gvwilson:
The person field should link to the person's individual record; the email address should be a mailto.

wking:
On Wed, May 27, 2015 at 07:10:14PM -0700, Greg Wilson wrote:
> The person field should link to the person's individual record; the
> email address should be a mailto.

Duplicate of #312?  Or is this just building on #312 by splitting the
name and email into separate columns?

gvwilson:
Made the same note for tasks and badges; I'll probably find other places 
where we should make the same change as well.  Or create a snippet that 
takes a `person` and produces `<td>{{ person.get_full_name}}</td><td>{{ 
person.email }}</td>` with appropriate wrappers to create the hyperlink 
and mailto.

Issue #312, opened 2015-05-27
Make person's full name a link to the person in the 'all tasks' display


gvwilson:
We should always display people as `name <email>` with `name` hyperlinking to the person and `<email>` a mailto: the person.

Issue #311, opened 2015-05-27
Make 'event' entry a hyperlink in 'all tasks' display


gvwilson:
The event name should link to the event in the display of all tasks.

Issue #310, opened 2015-05-27
Remove 'Add Many People' button from all_persons


gvwilson:
For consistency, just show the 'add single' - keep 'add many' under the 'More' menu.

PR #329, opened 2015-06-02
Graphical updates

### LINKED ISSUE #307: Remove Eventbrite ID from all_events display
We just need a checkmark to show whether there's an Eventbrite (which can be linked to the Eventbrite page) - we don't need the numeric ID (which takes up space).

pbanaszkiewicz:
This resolves #307, #310, #311, #312 (and #315, and fixes the same issue on other pages), plus this enhances person details page.


gvwilson:
:+1:

PR #336, opened 2015-06-03
Changing passwords and permissions

### LINKED ISSUE #326: User password and permissions
We should allow to set up user password by admins correctly.

I'm thinking about such scenario:

1. Admins creates a user. We don't provide "password", "repeat password" fields, but in the backend we set up `person.set_unusable_password()`.
2. On person details page we provide two buttons: "Change password", "Change permissions".
3. Change password has "password", "repeat password" fields
4. Change permissions has "is superuser", "user permissions", "user groups" fields

pbanaszkiewicz:
This resolves #326.

Not sure if I need to test password changing feature, because it relies upon Django forms.

### DIFF
--- a/workshops/test/test_person.py
+++ b/workshops/test/test_person.py
@@ -1,6 +1,8 @@
 # coding: utf-8
 
 from django.core.urlresolvers import reverse
+from django.contrib.auth.models import Permission, Group
+
 from ..models import Person
 from .base import TestBase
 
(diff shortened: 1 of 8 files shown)


pbanaszkiewicz:
I decided to add tests for `person_password` view. After all, we have some additional logic in there.

pbanaszkiewicz:
@wking and @gvwilson, can you comment on this PR?

workshops/test/test_person.py
```diff
@@ -147,3 +149,155 @@ def test_edit_person_notes(self):
         else:
             self._check_status_code_and_parse(response, 200)
             assert False, 'expected 302 redirect after post'
+
+    def test_edit_person_permissions(self):
+        "Make sure we can set up user permissions correctly."
+
+        # make sure Hermione does not have any perms, nor groups
+        assert self.hermione.is_superuser is False
```
gvwilson: assert not self.hermione.is_superuser

workshops/test/test_person.py
```diff
@@ -147,3 +149,155 @@ def test_edit_person_notes(self):
...
+        }
+
+        response = self.client.post(
+            reverse('person_permissions', args=[self.hermione.id]),
+            data,
+        )
+
+        assert response.status_code == 302
+
+        self.hermione.refresh_from_db()  # updates self.hermione
```
gvwilson: Comment unnecessary.

workshops/test/test_person.py
```diff
@@ -147,3 +149,155 @@ def test_edit_person_notes(self):
...
+        assert self.hermione.is_superuser is True
```
gvwilson: assert self.hermione.is_superuser

workshops/test/test_person.py
```diff
@@ -147,3 +149,155 @@ def test_edit_person_notes(self):
...
+        assert (
```
gvwilson: Why the outermost parens?

gvwilson:
Minor comments, but otherwise :+1: (warning: I don't know enough about 
Django password management to spot subtle glitches)

pbanaszkiewicz:
@gvwilson we actually use Django default password management system. I changed code according to your suggestions, please feel free to merge if you don't find any other obstructions in it.

PR #337, opened 2015-06-03
'Organizer' field of Event is optional.

### LINKED ISSUE #330: Some workshops at the database are missing the Organizer field

gvwilson:
The `Event` model allows the `organizer` field to be NULL, but the form wasn't allowing it to be empty.  This PR fixes that by allowing `organizer` to be left empty in the event creation form.

This fixes #330.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -133,7 +133,7 @@ class EventForm(forms.ModelForm):
     organizer = selectable.AutoCompleteSelectField(
         lookup_class=lookups.SiteLookup,
         label='Organizer',
-        required=True,
+        required=False,
         help_text=AUTOCOMPLETE_HELP_TEXT,
         widget=selectable.AutoComboboxSelectWidget,
     )


wking:
On Wed, Jun 03, 2015 at 12:31:42PM -0700, Greg Wilson wrote:
> The `Event` model allows the `organizer` field to be NULL, but the
> form wasn't allowing it to be empty.  This PR fixes that by allowing
> `organizer` to be left empty in the event creation form.

From [1]:

> …in the case of self-organized workshops, we can guess that the
> organizer is the same as the host institution, but it isn't always,
> and it doesn't really matter.

I'm on the fence here.  Possible organizer situations (I think
“organizers” are “coordinators” in our membership docs [2]):

a. SCF admins organize an event.  We usually charge an admin fee for
   these, although we might wave the fee for special cases.  I see no
   reason to not record these (maybe with an SCF Site/Host?).

b. Member organizations organize an event.  This is part of their
   contract (“Coordinates at least … workshops annually” [2]), so we
   definitely want to record this.

c. Someone else organizes the event.  We don't have a need to track
   these, although it would be nice to be able to query the
   information for organizing further events in a given area.

So I think this PR is about allowing us to skip the organizer field
for (c), and maybe (a).  But we *must* track organizer information for
(b).  It doesn't seem to difficult to specify this information at
event-creation time, and it's hard to go back and add it to old events
after the fact, so I'm personally in favor of requiring the field (so
it's neither null nor empty).  So the admin cost of managing this
field is a price I think is worth paying to get the database to ensure
we don't miss a (b) case.

For migrating to a non-null field, we could explicitly create a dummy
“Unknown organizer” site/host for migrating existing workshops.  Then
we'd disallow that organizer for future event saves (i.e. if you want
to touch an event, you need to figure out who organized it).

[1]: https://github.com/swcarpentry/amy/issues/330#issuecomment-108588037
[2]: https://github.com/swcarpentry/board/blob/fe53ef85313ae5cdbaeb03d9c38fbf195f8effa5/membership.md

pbanaszkiewicz:
@gvwilson are you for "self-organized"/"Unknown organizer" option in `event.organizer` (+ `null=False`) or are you for leaving the field both `null` and `blank`?

Personally, I'm for option °1.

gvwilson:
The admins who are using Amy daily find it annoying to have to specify an organizer when, in a growing number of cases, the workshop is self-organized (i.e., we don't care who the organizer was).  If you can put a "self-organized" button beside the "organizer" pulldown, such that clicking the button automatically sets the pulldown to the same value as the "site" field, that would help - can that easily be implemented, and then we can see how the workflow improves?

pbanaszkiewicz:
> can that easily be implemented, and then we can see how the workflow improves?

Unfortunately it's way harder (but not impossible) than just letting the organizer field be blank.

pbanaszkiewicz:
I think I'm going to merge this, and set a meaning to `Event.organizer` to something like:

> An event organizing institution, if we care about it.

PR #340, opened 2015-06-05
Make the 'email' field in display of a single person a mailto:

gvwilson:


pbanaszkiewicz:
Well, there was `{{ person.email | urlize }}` there before, which should turn email into `<a href="mailto:{{person.email}}">{{ person.email }}</a>`.

Was it not producing proper `<a href="mailto:…">`?

I think I had similar issue with person.url, and I solved it with this code: `{{ person.url|default:""|urlize|default:"—" }}` instead of using if-else.

gvwilson:
Nope, wasn't a mailto, but is now.

On 2015-06-05 11:54 AM, Piotr Banaszkiewicz wrote:
>
> Well, there was |{{ person.email | urlize }}| there before, which 
> should turn email into |<a href="mailto:{{person.email}}">{{ 
> person.email }}</a>|.
>
> Was it not producing proper |<a href="mailto:…">|?
>
> I think I had similar issue with person.url, and I solved it with this 
> code: |{{ person.url|default:""|urlize|default:"—" }}| instead of 
> using if-else.
>
> —
> <https://github.com/swcarpentry/amy/pull/340#issuecomment-109339543>.
>

-- 

pbanaszkiewicz:
Okay, so I checked and for people without email addresses it was simply "None", but should be "—".

Issue #37, opened 2014-12-02
Allow award of new badges


gvwilson:
There should be a way to award someone a badge, either through a stand-alone form or by adding a row to the page showing everyone who has that badge.

wking:
On Mon, Dec 29, 2014 at 10:33:11AM -0800, W. Trevor King wrote:
> On Sat, Dec 27, 2014 at 02:39:06PM -0800, Greg Wilson wrote:
> > 1. I think the simplest thing to start with is to show checkboxes
> >    for all as-yet-unawarded badges on the "person details" page,
> >    so that when editing the person, the admin can give that person
> >    badges.
> 
> I'd rather stick with an autogenerated many-to-many UI (which
> django-extra-views gives us)…

This may not be the most stable dependency.  We should wait to see how
the maintainer transition [1] works out before moving forward with it.
I'm still in favor of anything that keeps *us* from re-implementing
this sort of logic locally though.

[1]: https://github.com/AndrewIngram/django-extra-views/issues/96

gvwilson:
Further discussion:

1. The "edit this person" page should allow us to multi-select badges that the person *doesn't* yet have and award them.  The date of the award will be the date that the selection was done.

2. The page displaying badges (http://amy.software-carpentry.org:8080/workshops/badges) should have an "award badge" button that lets me select a person and a badge (via pulldowns).  When submitted, the page displays the newly-awarded badge and offers me the same pair of pulldowns so that I can add another.

pbanaszkiewicz:
Point 2. from the comment above is now resolved via #303.

Point 1. from the comment above will be resolved via #110.

r-gaia-cs:
@pbanaszkiewicz Can we close this? #303 was merge.

pbanaszkiewicz:
@r-gaia-cs we'll close as soon as we implement awarding badges from user edit page. I have something for it, and I'll submit a PR soon.

pbanaszkiewicz:
#345 fulfills all requirements now.

PR #345, opened 2015-06-06
Award badges from person-edit view

pbanaszkiewicz:
This is a little bit alternative to #110, and somewhat more limited in options, as it only allows us to add one badge per request from the person-edit view.

However, in here we use two crispy forms that integrate easily and visually appeal, + we have tests in place.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -1,6 +1,6 @@
 from django import forms
 from django.forms import HiddenInput
-from django.forms.models import modelform_factory
+from django.forms import formset_factory
 
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import Submit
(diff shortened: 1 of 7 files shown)


pbanaszkiewicz:
This may require a little rebase if we merge #336 first.

pbanaszkiewicz:
Rebased.

@gvwilson are you happy with this PR as a solution to #37 °1?

gvwilson:
I'll test it tonight - about to head out for a bit.

gvwilson:
1.  Since we can now set superuser in the "edit person" view, we should show superuser status on the "person details" page.
2.  When I clicked "Change password" for a person who didn't already have a password, the "New Password" field was auto-filled with one of my browser's saved passwords.
3.  Badge award works!

So I'm :+1: to merge as-is, and file separate issues for 1 and 2 above.

pbanaszkiewicz:
Thanks!!!

PR #347, opened 2015-06-07
Make AMY show list of skills horizontally

narayanaditya95:
Preview (UPDATE):  
screenshot from 2015-06-07 23 48 56
UPDATE: As discussed, I wrapped the three fields in a single row. I utilised panels to show lat/long.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -3,7 +3,8 @@ from django.forms import HiddenInput
 from django.forms.models import modelform_factory
 
 from crispy_forms.helper import FormHelper
-from crispy_forms.layout import Submit
+from crispy_forms.layout import Layout, Div, HTML, Submit
+from crispy_forms.bootstrap import FormActions
 from selectable import forms as selectable
 
 from workshops.models import Skill, Airport, Event, Task, Award, Person
@@ -56,15 +57,43 @@ class InstructorsForm(forms.Form):
         lookup_class=lookups.AirportLookup,
         label='Airport',
         required=False,
-        help_text=AUTOCOMPLETE_HELP_TEXT,
+        widget=selectable.AutoComboboxSelectWidget(
+            lookup_class=lookups.AirportLookup,
+        ),
     )
 
     def __init__(self, *args, **kwargs):
         '''Build checkboxes for skills dynamically.'''
         super(InstructorsForm, self).__init__(*args, **kwargs)
+        self.helper = FormHelper(self)
+        self.helper.form_class = 'form-inline'
+        self.helper.layout = Layout(
+            'wanted',
+            Div(
+                Div(
+                    'latitude',
+                    'longitude',
+                    css_class='col-sm-6'
+                ),
+                Div(
+                    HTML('<br><strong>OR</strong>'),
+                    css_class='col-sm-2',
+                ),
+                Div(
+                    'airport',
+                    css_class='col-sm-4'
+                ),
+                css_class='row panel panel-default panel-body',
+            ),
+            HTML('<label class="control-label">Skills</label>'),
+            FormActions(
+                Submit('submit', 'Submit'),
+            ),
+        )
         skills = Skill.objects.all()
         for s in skills:
             self.fields[s.name] = forms.BooleanField(label=s.name, required=False)
+            self.helper.layout.insert(3, s.name)
 
     def clean(self):
         cleaned_data = super(InstructorsForm, self).clean()
--- a/workshops/templates/workshops/instructors.html
+++ b/workshops/templates/workshops/instructors.html
@@ -11,8 +11,7 @@
 
 {% block content %}
     {% crispy form %}
-{% if persons == None %}
-{% elif persons %}
+{% if persons %}
     <table class="table table-striped">
         <tr>
     	    <th>taught</th>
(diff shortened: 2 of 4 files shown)


gvwilson:
Great start - can you move the 'submit' button down a bit so it doesn't 
crowd the list?

And if you'd like to tackle another thing on this form (as a separate 
PR), can you find a way to make it clear that:

1.  Latitude and longitude are needed together, and

2. lat+long is XOR with Airport (i.e., you either give us a location 
*or* you specify an airport)

Thanks,
Greg

narayanaditya95:
@gvwilson 
Layout will allow us:
1. Add simple "OR" text between lat/long and Airport fields. 
2. Wrap lat/long in one tab and airport in another so that only one is visible to the user.
Cannot think of another way. See [1](http://django-crispy-forms.readthedocs.org/en/latest/layouts.html#bootstrap-layout-objects)

To make lat/long appear required together, we can wrap them together in one line itself. (Optional: Add a background for both so they appear as a unit).

@gvwilson

gvwilson:
How about:

     ( [lat] + [long] ) OR [airport]

all on one line, where `[xyz]` is a pulldown and the parentheses and the 
word `OR` are actually on the page.

narayanaditya95:
Yes, that is easily possible.I will open a new PR for the same.

narayanaditya95:
@gvwilson 
I made the required changes. I did not create a new PR as the new commit nullifies the previous commits.  
Please review.

gvwilson:
It's looking good - a few suggestions:

1.  Removed the box around Latitude/Longitude.
2.  Remove the text under Airport about autocomplete (people will figure that out).
3.  Line up the titles and pulldowns vertically.
4.  Put the word 'OR' on the same line as the pulldown boxes.

Thanks!

pbanaszkiewicz:
>  2. Remove the text under Airport about autocomplete (people will figure that out).

I'd also change the widget on that box to `selectable.AutoComboboxSelectWidget`, so it shows a little drop-down button.

I also suggest a label or header "Skills" before all the checkboxes.

narayanaditya95:
UPDATE:
screenshot from 2015-06-08 11 48 20

>  1. Removed the box around Latitude/Longitude.
>  2. Remove the text under Airport about autocomplete (people will figure that out).
>  4. Put the word 'OR' on the same line as the pulldown boxes.

They have been implemented.

> 3. Line up the titles and pulldowns vertically.

Do you mean that I should revert to using ```form-horizontal``` as the form class for all fields except the skills? Bit confused here. Something similar to this?
screenshot from 2015-06-08 11 50 07


> I'd also change the widget on that box to selectable.AutoComboboxSelectWidget, so it shows a little drop-down button. I also suggest a label or header "Skills" before all the checkboxes.

Thanks that is done.

gvwilson:
This looks great - I'm :+1: to merge.  @pbanaszkiewicz ?

workshops/forms.py
```diff
@@ -56,15 +57,43 @@ class InstructorsForm(forms.Form):
         lookup_class=lookups.AirportLookup,
         label='Airport',
         required=False,
-        help_text=AUTOCOMPLETE_HELP_TEXT,
+        widget=selectable.AutoComboboxSelectWidget(
+            lookup_class=lookups.AirportLookup,
+        )
```
pbanaszkiewicz: Please add a trailing comma in this line. To do that, start an interactive rebase, edit your commit which changes this line (bdc7aae) and then amend that commit.

More info here: https://www.youtube.com/watch?v=4EOZvow1mk4

workshops/templates/workshops/instructors.html
```diff
@@ -10,7 +10,7 @@
 {% endblock %}
 
 {% block content %}
-    {% crispy form form_helper %}
+    {% crispy form %}
 {% if persons == None %}
```
pbanaszkiewicz: Could you change that conditional? We don't need to check `if persons == None`, and it's not Pythonic.

Make that change in a separate commit, and explain in the message why you're making that change.

pbanaszkiewicz:
The code looks good to me. I'll merge after @narayanaditya95 implements two minor edits.

workshops/forms.py
narayanaditya95: I have been following the issue tracker and have watched the video already. It was a useful git tutorial. Thanks :)

workshops/templates/workshops/instructors.html
narayanaditya95: Can you please explain to me why is this conditional unnecessary? As far as I can understand, it works exclusively for GET requests and renders nothing below the form (no table headers).  
Am I mising something?

narayanaditya95: Should I squash the four commits into one since it is an atomic update. Is that right?

narayanaditya95:

workshops/templates/workshops/instructors.html
pbanaszkiewicz: @narayanaditya95:
> ```
{% if persons %}
    {% for p in persons %}
        ## Tables
    {% empty %}
        <p>No matches.</p>
{% endfor %}
```

Yeah, that's what I wanted.

> Should I squash the first four commits into one since it is an atomic update. Is that right?

I'll answer in the comment thread to that.

workshops/forms.py
pbanaszkiewicz: That's wonderful :-)

pbanaszkiewicz:
@narayanaditya95:

> Should I squash the first four commits into one since it is an atomic update. Is that right?

Yes, I'd go and squash commits from 45bb6252aa0e3984fc622a22062157723186c599 to 49127d46e2e959f9ccb814ce3be64c45b4c83a5c. Please remember to provide a useful commit message.

narayanaditya95:
Finished with squashing the commits.

pbanaszkiewicz:
Good job, @narayanaditya95. Merging now.

PR #349, opened 2015-06-08
Switch from 'Amy' to 'AMY'

### LINKED ISSUE #328: Change name of app
Having the same name for a person and the app is confusing, especially since we both work in the same areas. Suggest "amybot" for the app.

pbanaszkiewicz:
We're switching from 'Amy' to 'AMY' to distinguish between real people and this project.

I think we have a consensus to leave 'amy' as the name for the eventual packages (PyPI / NPM), and 'AMY' for the project.

This resolves #328.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-Amy is an open source project, and we welcome contributions of all
+AMY is an open source project, and we welcome contributions of all
 kinds.  By contributing, you are agreeing that Software Carpentry may
 redistribute your work under [this license][license].  You also agree
 to abide by our [contributor code of conduct][conduct].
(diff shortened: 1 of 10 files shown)


gvwilson:
Mmm.. I like aMY, just to be conTRARY, but I can lIVE wITH tHIS.

pbanaszkiewicz:
Honestly I had no idea that https://github.com/swcarpentry/amy/commit/25e67c0ebf3afd9cb6d60f1b2d0592ad12c7e854 was "aMY". I can rebase, if you're more decided on aMY than AMY.

gvwilson:
I'm mostly trolling - I'll sleep on it and we'll figure it out 
tomorrow.  bUT I tHINK iT wOULD bE kIND oF cOOL.

pbanaszkiewicz:
:-)

gvwilson:
If you can update this, we can merge it - @amyrbrown would be very grateful if we switched to AMY in all caps.  I will then update the logo...

PR #352, opened 2015-06-08
Verbose messages

### LINKED ISSUE #339: Indicate successful operation on forms
Right now (2504e4aee6c, 2015-06-04) when we give someone a badge, the form is redisplayed with all the fields still filled in and no indication that the badge has been issued - so if you blink, you're not aware the form has been submitted, and you're all too likely to click 'submit' again.  We should indicate successful submission somehow (an acknowledgment message on the form).  This should also be done on other forms.

pbanaszkiewicz:
This PR:

1. adds messages to edit/create views
2. makes messages visually nicer (with http://getbootstrap.com/components/#alerts)
3. cleans up event-edit view (so it works very similar to person-edit)
4. shifts second form up on event/person-edit pages

Resolves #339.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -200,15 +200,10 @@ class TaskForm(forms.ModelForm):
         widget=selectable.AutoComboboxSelectWidget,
     )
 
-    def __init__(self, *args, **kwargs):
-        event = kwargs.pop('event', None)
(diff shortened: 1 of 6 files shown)


gvwilson:
Looks good to me - merging manually because of conflict with breadcrumb removal.

Issue #64, opened 2014-12-14
Additional event attendance information (slots and applicants)?


wking:
As [proposed][1] in swcarpentry/site#701, I think it would be good to have an `event.slots` or similar to count how many seats were available. And maybe `event.applicants` if anyone was waitlisted. Comparing with `event.attendance` will let us see how full the event was, which gives a hint at how soon we'll want to run a follow-up workshop. 

[1]: https://github.com/swcarpentry/site/pull/701#issuecomment-66926188

pbanaszkiewicz:
Do we have historical data for these?

gvwilson:
Mostly not.

pbanaszkiewicz:
Both Giacomo and Amy think this is not a good idea.

PR #354, opened 2015-06-10
Add Event.slots, Event.applicants

pbanaszkiewicz:
Additionally, help_text was added to Event.attendance, Event.slots,
Event.applicants to help admins understand role of each of these fields:

* Event.attendance: number of people who turned out to the workshop
* Event.slots: number of slots available for people
* Event.applicants: number of people who applied for the workshop.

For example, we can have a workshops for maximum of 40 people.
42 people registers, and only 36 comes to the workshop.  So
attendance=36, slots=40, and applicants=42.

This solves #64.

### DIFF
--- a/workshops/migrations/0011_auto_20150610_0609.py
+++ b/workshops/migrations/0011_auto_20150610_0609.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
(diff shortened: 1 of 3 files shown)


gvwilson:
I'm -0.5 on this - we're asking admins and others to do a lot already for record-keeping. What do Amy and Giacomo think?

pbanaszkiewicz:
Both Giacomo and Amy don't think this is a good idea. Closing then.

PR #356, opened 2015-06-11
Searching in airports as well

gvwilson:


wking:
Looks good to me.

pbanaszkiewicz:
Searching from topbar does not search airports by default. You need to add:
```html
            <input type="hidden" name="in_airports" value="on" />
```
In line 57 in `workshops/templates/base.html`.

gvwilson:
Nice catch.

On 2015-06-11 9:34 AM, Piotr Banaszkiewicz wrote:
>
> Searching from topbar does not search airports by default. You need to 
> add:
>
>              <input  type="hidden"  name="in_airports"  value="on"  />
>
> In line 57 in |workshops/templates/base.html|.
>
> —
> <https://github.com/swcarpentry/amy/pull/356#issuecomment-111136790>.
>

-- 

Issue #289, opened 2015-05-23
Need a way to delete things


gvwilson:
We need to be able to delete airports, badges, tasks, etc.

pbanaszkiewicz:
Would that be actual deletion or just "mark as deleted" (similarly to events, tasks)?

gvwilson:
In this case, I'm happy with pure deletion - the cost of lost data is pretty small.  But we should start logging changes to the DB in a recovery file somewhere.

pbanaszkiewicz:
Closed via #357.

PR #120, opened 2015-01-03
WIP: workshops/models.py: Add Host and EventHost models

wking:
To distinguish between the Site (a physical location, e.g. TGAG, the
Genome Analysis Centre, Norwich Research Park, Colney, Norwich,
Norfolk, United Kingdom, 52.621699 N, 1.21888 E) and the Host
(possibly a diffuse entity, e.g. ELIXIR, the European life-sciences
Infrastructure for biological Information).  A diffuse host can have
many sites, and a particular workshop can have many collaborating
hosts.  The many-to-many relationship (one host can support several
events, and one event may have several hosts) is not layed out
explicitly in our charter, but [we have][1]:

> *Hosting* a workshop means providing space for it, supporting travel
> costs for instructors (if applicable), etc.

and its conceivable that one host could provide space while others
could fund instructor travel, etc.  I've added the free-form
EventHost.notes to keep track of this, and I expect it will have
values like "providing the site", "funding travel for instructor Alice
Bobsdaughter", "providing coffee and snacks for the first day of the
event", etc.  If a single host provides for the whole event, I'd just
leave this field blank.

This pull request is currently a work-in-progress while we work out
the models, after which I'll go through and fill in the migrations,
view changes, etc.  I'll remove the WIP note in the summary once I
think it's ready for merging.

Fixes #5 and #6.

[1]: https://github.com/swcarpentry/board/blob/master/membership.md#terminology

PR #357, opened 2015-06-11
A way to delete things

pbanaszkiewicz:
This solves #289.


pbanaszkiewicz:
@gvwilson I'm thinking about reverting "mark as deleted" for events and tasks, what do you think?

gvwilson:
Yeah, I think it causes more trouble than it's worth.  But we're 
definitely going to need a "merge these people" tool - I just discovered 
that Mike/Michael Croucher has entries under both names...

pbanaszkiewicz:
@gvwilson yeah, since I'm kind of blocked on Eventbrite stuff, I'll move to fixing bulk-upload. But let me finish this one first.

gvwilson:
Thanks.

workshops/models.py
```diff
@@ -108,7 +108,7 @@ class Person(AbstractBaseUser, PermissionsMixin):
     email       = models.CharField(max_length=STR_LONG, unique=True, null=True, blank=True)
     gender      = models.CharField(max_length=1, choices=GENDER_CHOICES, null=True, blank=True)
     may_contact = models.BooleanField(default=True)
-    airport     = models.ForeignKey(Airport, null=True, blank=True)
+    airport     = models.ForeignKey(Airport, null=True, blank=True, on_delete=models.SET_NULL)
```
wking: Hmm.  If a `Person` references an `Airport` I don't think we should be deleting the airport.  Can we have a check in the airport-delete view that says "You can't delete this airport, because it's being used by: {list-of-linkers}”?  Then someone interested in deleting the airport can use information about the to-be-deleted airport to figure out what to do about the existing linkers.  For duplicate airport entries (should be impossible with unique `iata` and `fullname` fields?) that may be “shift them all over to {alternative-iata}”;

pbanaszkiewicz: I figured that we didn't need to forbid deletion of Airport if some Person links to it, because we don't manipulate airports at all.

I can change `on_delete` to `on_delete=models.PROTECT`, but this would raise a question: what would we do with Event.site and Event.organizer fields, that have `on_delete=models.SET_NULL`, too?

pbanaszkiewicz:
Reverted that change, but I'd very much like @wking to take a look at 831d7e2.

workshops/models.py
wking: So in my view this should be [`on_delete=models.PROTECT`][1].

[1]: https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.PROTECT

workshops/models.py
```diff
@@ -272,9 +272,10 @@ class Event(models.Model):
     '''Represent a single event.'''
 
     published  = models.BooleanField(default=False)
-    site       = models.ForeignKey(Site)
+    site       = models.ForeignKey(Site, null=True, on_delete=models.SET_NULL)
```
wking: Same comments as for `Person.airport`.  I don't want site-less `Event`s in the database, and I don't want to blow away associated `Event`s when someone tries to remove a `Site`.

workshops/models.py
```diff
@@ -272,9 +272,10 @@ class Event(models.Model):
...
     tags       = models.ManyToManyField(Tag)
-    organizer  = models.ForeignKey(Site, related_name='organizer', null=True, blank=True)
+    organizer  = models.ForeignKey(Site, related_name='organizer', null=True,
+                                   blank=True, on_delete=models.SET_NULL)
```
wking: Same comments as for `Person.airport` and `Event.site`.

workshops/templates/workshops/airport.html
```diff
@@ -10,5 +10,8 @@
   <tr><td>longitude:</td><td>{{ airport.longitude }}</td></tr>
 </table>
 
-<p class="edit-object"><a href="{% url 'airport_edit' airport.iata %}" class="btn btn-primary">Edit</a></p>
+<div class="clearfix">
+  <p class="edit-object pull-left"><a href="{% url 'airport_edit' airport.iata %}" class=" btn btn-primary">Edit</a></p>
+  <p class="delete-object pull-right"><a href="{% url 'airport_delete' airport.iata %}" onclick='return confirm("Are you sure you wish to remove \"{{ airport }}\"?")' class="btn btn-danger">Delete airport</a></p>
```
wking: I like these lightweight confirmation boxes :).

workshops/views.py
```diff
@@ -178,6 +178,16 @@ class SiteUpdate(LoginRequiredMixin, UpdateViewContext):
     slug_url_kwarg = 'site_domain'
     template_name = 'workshops/generic_form.html'
 
+
+@login_required
+def site_delete(request, site_domain):
+    """Delete specific site."""
+    site = get_object_or_404(Site, domain=site_domain)
+    site.delete()
```
wking: I tend to prefer idempotent deletes, because it seems odd to have one of a pair of racy deletes return a 404 because the object has already been deleted.  The deleter just wants to remove the item from play, and if it's already gone (even if it never existed) that seems like success to me ;).  In Django, that's:

    Site.objects.filter(domain=site_domain).delete()

workshops/views.py
```diff
@@ -178,6 +178,16 @@ class SiteUpdate(LoginRequiredMixin, UpdateViewContext):
...
+
+    messages.success(request, 'Site was deleted successfully.')
```
wking: Hmm.  I hadn't noticed out [`django.contrib.messages`][1] use before (looks like it came in with b05d4bd2 (First pass, not processed yet, 2014-12-22, #131).  But “[The messages framework allows you to temporarily store messages in one request and retrieve them for display in a subsequent request (usually the next one)][1]” sounds like a lot of state to store in the expectation that you're going to get a particular future request.  Do we really need this here?

[1]: https://docs.djangoproject.com/en/1.8/ref/contrib/messages/

workshops/views.py
```diff
@@ -178,6 +178,16 @@ class SiteUpdate(LoginRequiredMixin, UpdateViewContext):
...
+    return redirect(reverse('all_sites'))
```
wking: Maybe we can set that message in a query parameter here, so we're not racing with any other requests that session may have in flight.

workshops/views.py
pbanaszkiewicz: I don't understand your concerns about messages framework. It's a pretty standard way of displaying results of backend operations to the enduser. It came with #352.

> sounds like a lot of state to store in the expectation that you're going to get a particular future request

With `return redirect()` we will get it.

messages are invalidated as soon as they're traversed through, which happens in every view that uses our original template.

workshops/models.py
wking: On Thu, Jun 11, 2015 at 06:28:17AM -0700, Piotr Banaszkiewicz wrote:
> > -    airport     = models.ForeignKey(Airport, null=True, blank=True)
> > +    airport     = models.ForeignKey(Airport, null=True, blank=True, on_delete=models.SET_NULL)
> 
> I figured that we didn't need to forbid deletion of Airport if some
> Person links to it, because we don't manipulate airports at all.

If you don't delete airports, then leave this line alone and remove
the airport-deletion endpoint ;).

> I can change `on_delete` to `on_delete=models.PROTECT`, but this
> would raise a question: what would we do with Event.site and
> Event.organizer fields, that have `on_delete=models.SET_NULL`, too?

I'd use PROTECT there too, or stay with the default CASCADE if you're
feeling adventurous.  SET_NULL is for things like “Oops, not only does
that {link-target} not exist, but that {linker} doesn't even have a
{link-target}”.  That may be true, but I want it to be an explicit
decision for each linker, not something that happens automatically
after you've decided that a particular Site is a duplicate (in which
case Event.site should be updated to reference the deduplicated Site).
Why else would you be deleting Site/Airport/… that are referenced from
other fields besides deduping?

wking:
On Thu, Jun 11, 2015 at 07:11:15AM -0700, Piotr Banaszkiewicz wrote:
> Reverted that change…

Good, I never liked #251 anyway ;).

> … but I'd very much like @wking to take a look at 831d7e2.

I left a few comments.  Deletion is complicated though (whether it's
real or not), since you need to decide what to do with related models
downstream of the one you're trying to remove.  For example, what's
downstream of Airport?

  class Person(AbstractBaseUser, PermissionsMixin):
      airport     = models.ForeignKey(Airport, null=True, blank=True)

What's downstream of Person?

  $ grep '^class\|ForeignKey.*Person' workshops/models.py | grep -B1 'ForeignKey.*Person'
  class Task(models.Model):
      person     = models.ForeignKey(Person)
  --
  class Qualification(models.Model):
      person     = models.ForeignKey(Person)
  --
  class Award(models.Model):
      person     = models.ForeignKey(Person)

What's downstream of those?

  $ grep '^class\|ForeignKey.*Task' workshops/models.py | grep -B1 'ForeignKey.*Task'
  … no results…
  $ grep '^class\|ForeignKey.*Qualification' workshops/models.py | grep -B1 'ForeignKey.*Qualification'
  … no results…
  $ grep '^class\|ForeignKey.*Award' workshops/models.py | grep -B1 'ForeignKey.*Award'
  … no results…

Personally, I consider Person entries more important than Airport
entries.  A goofed/duplicate airport doesn't mean that all the
downstream objects are also goofed/duplicate, so I'd PROTECT
Person.airport to break the cascade.  On the other hand, I consider
Person more important than Task/Qualification/Award.  If you want to
delete a goofed/duplicate Person, I'll happily cascade that down to
blow away any associated Tasks/Qualififications/Awards, because it's
very unlikely that you care about a Task with a NULL Task.person.

workshops/views.py
wking: On Thu, Jun 11, 2015 at 07:35:16AM -0700, Piotr Banaszkiewicz wrote:

> > sounds like a lot of state to store in the expectation that you're
> > going to get a particular future request
> 
> With `return redirect()` we will get it.

Assuming that the client bothers to follow the redirect.  And what
happens if the client has simultaneous requests to two view that both
consume these messages?  You might have an event listing that shows
the “person created” message intended for the parallel person listing.
That's likely a low-probability event, but I don't like racy state ;).
And passing the message(s) as query parameters means you cover both
the racy-request case and the “client never bothered to follow the
redirect” case.

pbanaszkiewicz:
@wking 

> Personally, I consider Person entries more important than Airport entries. A goofed/duplicate airport doesn't mean that all the downstream objects are also goofed/duplicate, so I'd PROTECT Person.airport to break the cascade. On the other hand, I consider Person more important than Task/Qualification/Award. If you want to delete a goofed/duplicate Person, I'll happily cascade that down to blow away any associated Tasks/Qualififications/Awards, because it's very unlikely that you care about a Task with a NULL Task.person.

I agree && I'm actually making fixes accordingly. Additionally I want to preserve Awards when related Events are deleted, do you agree?

wking:
On Thu, Jun 11, 2015 at 08:35:48AM -0700, Piotr Banaszkiewicz wrote:
> Additionally I want to preserve Awards when related Events are
> deleted, do you agree?

No.  If the backing event didn't happen, that award should not have
been granted.  Compare with “if the backing airport didn't happen,
that person should not have been created”.

But again, this should really be accompanied by a list of situations
for which you expect to be deleting Events.  For “Oops, I accidentally
made this five seconds ago” [1] it doesn't matter either way.  Do we
have other workflows that would lead to Event deletion?

[1]: https://github.com/swcarpentry/amy/pull/251#issuecomment-99546446

pbanaszkiewicz:
> No. If the backing event didn't happen, that award should not have been granted. Compare with “if the backing airport didn't happen, that person should not have been created”. 

Okay.

> Do we have other workflows that would lead to Event deletion?

I can't think of any other, except a double misclick.

pbanaszkiewicz:
@wking Okay, I rebased.

1. didn't change `object = get_object_or_404(); object.delete()`
2. added a new view for failed deletion, it should display related objects that may prevent deletion

Could you take a look again? Thanks.

workshops/views.py
```diff
@@ -215,6 +228,18 @@ class AirportUpdate(LoginRequiredMixin, UpdateViewContext):
...
+
+@login_required
+def airport_delete(request, airport_iata):
+    """Delete specific airport."""
+    try:
+        airport = get_object_or_404(Airport, iata=airport_iata)
+        airport.delete()
+        messages.success(request, 'Airport was deleted successfully.')
+        return redirect(reverse('all_airports'))
+    except IntegrityError:
```
wking: Maybe catch the more-specific [`ProtectedError`][1] here?

[1]: https://docs.djangoproject.com/en/1.8/ref/exceptions/#django.db.models.ProtectedError

workshops/views.py
```diff
@@ -1035,3 +1071,16 @@ def _time_series(request, data, title):
...
+
+def _failed_to_delete(request, object, back=None):
+    context = {
+        'title': 'Failed to delete',
+        'back': back or object.get_absolute_url,
+        'object': object,
+        'refs': dict(),
+    }
+    for field in dir(object):
+        if field.endswith('_set') and not field.startswith('_'):
```
wking: The `_set` check assumes that nobody is setting [`related_name`][1] (more discussion [here][2]).  Even if that assumption was true now it would be brittle, but without #120 we need the current `related_name` for [`Site.organizer`][3].  Is there another way to find all the reverse relationships?

[1]: https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.ForeignKey.related_name
[2]: https://docs.djangoproject.com/en/1.8/topics/db/queries/#related-objects
[3]: https://github.com/swcarpentry/amy/blob/1ab2c6ca628f05b56359d497bf9d0c487ec98822/workshops/models.py#L277

pbanaszkiewicz:
@wking: I changed the way reverse foreign key fields are looked up, and switched from IntegrityError to ProtectedError.

Cheers!

pbanaszkiewicz:
@wking improved!

pbanaszkiewicz:
@wking ping

wking:
On Fri, Jun 12, 2015 at 01:44:12AM -0700, Piotr Banaszkiewicz wrote:
> I changed the way reverse foreign key fields are looked up…

It doesn't look like `.one_to_many` covers many-to-many or one-to-one
relationships, which also involve cross-table references.  What you
want is likely:

  [
    f for f in object._meta.get_fields()
    if (f.one_to_many or f.one_to_one or f.many_to_many) and f.auto_created
  ]  

See also [1].

[1]: https://docs.djangoproject.com/en/1.8/ref/models/meta/#migrating-from-the-old-api

pbanaszkiewicz:
@wking: Rebased && updated

wking:
On Thu, Jun 18, 2015 at 02:50:59PM -0700, Piotr Banaszkiewicz wrote:
> Rebased && updated

Looks good enough to merge ;).

pbanaszkiewicz:
Thanks!

PR #358, opened 2015-06-11
Improving pattern-matching and error reporting in bulk upload.

gvwilson:
1.  Report exactly which name field failed to match.
2.  Show blanks when there are no errors (rather than dashes).
3.  Modify pattern matching to match empty string in uploaded data vs. null in database to avoid spurious errors.

### DIFF
--- a/workshops/templates/workshops/person_bulk_add_results.html
+++ b/workshops/templates/workshops/person_bulk_add_results.html
@@ -51,7 +51,6 @@
                 {% endfor %}
                 </ul>
             {% else %}
-                {{ entry.errors|default_if_none:"&mdash;" }}
             {% endif %}
             </td>
         {% endwith %}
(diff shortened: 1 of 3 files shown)


workshops/templates/workshops/person_bulk_add_results.html
```diff
@@ -51,7 +51,6 @@
                 {% endfor %}
                 </ul>
             {% else %}
-                {{ entry.errors|default_if_none:"&mdash;" }}
```
wking: I kind of like the dash, but there's no clear motivation for it when it landed in 9e44195 (Bulk-edit individual fields on confirmation page, 2015-01-14), so I guess it's ok to remove.  If it goes, you should remove the `{% else %}` too.

wking:
I left a minor commit about cleaning up the `{% else %}`, but otherwise this looks good to me.

gvwilson:
Thanks - pushed fix.

PR #360, opened 2015-06-12
Migrating skills

### LINKED ISSUE #348: Repeated skill "Matlab"
We can see it on `Find instructors` page: there's `MATLAB` and `Matlab`.

To get rid of it we probably have to write a migration.

gvwilson:
1.  Replace Skill model with Lesson model.
2.  Link Qualification to Lesson.
3.  Add new Qualification entries by translating old skills to particular lessons.
4.  Updating forms and views to display.
5.  Updating tests to use Lesson instead of Skill.

Note that tests still fail because they are trying to apply migrations that have already been applied and the migrations assume that they can always be applied.  `workshops/migrations/0012_*.py` must be improved to handle this.

Fixes #348.


pbanaszkiewicz:
I have fixes for you: https://github.com/gvwilson/amy/pull/4

pbanaszkiewicz:
BTW I'm very unhappy with how long 0012* takes to migrate… @wking any idea how could we speed it up?

gvwilson:
How long does it take on your machine?  (And is it really that big a 
deal - once we do the migration in production, we can go and re-set the 
migrations directory to erase all these things.)

pbanaszkiewicz:
> How long does it take on your machine?

About 2-4 seconds.

> (And is it really that big a deal - once we do the migration in production, we can go and re-set the migrations directory to erase all these things.)

Yeah, I think we can do that.

pbanaszkiewicz:
Is this still WIP?

gvwilson:
I want to test it with data from the form I mailed people about today 
before I merge it and close it.

workshops/migrations/0012_auto_20150612_0807.py
```diff
@@ -0,0 +1,79 @@
...
+
+EXTRA_LEGACY_NAMES = ['MATLAB']
+
+
+def add_new_lesson_names(apps, schema_editor):
+    '''Add instances of Lesson named after lessons.'''
+    Lesson = apps.get_model('workshops', 'Lesson')
+    for (old_name, new_names) in TRANSLATE_NAMES.items():
+        for name in new_names:
+            Lesson.objects.create(name=name)
```
wking: You don't use `old_name` here, so you should probably start the loop with:

    for new_names in TRANSLATE_NAMES.values():

workshops/migrations/0012_auto_20150612_0807.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+TRANSLATE_NAMES = {
+    'Git': ['swc/git-novice'],
+    'Make': [],
```
wking: Don't we have a make lesson?

gvwilson: Not in a usable state - Mike Jackson has just offered us a rework that's 
much more usable, but it isn't ready for prime time yet.

wking:
On Mon, Jun 15, 2015 at 02:52:00AM -0700, Piotr Banaszkiewicz wrote:
> BTW I'm very unhappy with how long 0012* takes to migrate… @wking
> any idea how could we speed it up?

Not really.  It's hard to push more logic into the database layer with
the non-one-to-one mapping between old skills and new lessons.

gvwilson:
I propose that we live with the slowdown for now, and in 0.5, we get rid 
of the old migrations.

pbanaszkiewicz:
Yeah, if we update our development database with blurred data, we migrations are already applied, so no worries about that from me.

wking:
On Mon, Jun 15, 2015 at 04:35:16AM -0700, Greg Wilson wrote:
> And is it really that big a deal - once we do the migration in
> production, we can go and re-set the migrations directory to erase
> all these things.

At this stage, this is probably fine.  But blowing away the migration
logic every few months is going to make it hard for folks besides
Piotr that are managing AMY deploys.  That may be an empty set now,
but developing as if you had an existing usebase is good practice, and
it keeps you getting in the habit of cutting corners.  Why is a slow
migration a problem?  If it's used in the test suites, we should be
able to rework the test setup to avoid running the migrations without
having to remove the migrations themselves.

gvwilson:
Didn't realize we could avoid migrations in the test setup - that's 
clearly a better answer than erasing old migrations.

workshops/migrations/0012_auto_20150612_0807.py
wking: On Tue, Jun 16, 2015 at 10:06:16AM -0700, Greg Wilson wrote:
> > +TRANSLATE_NAMES = {
> > +    'Git': ['swc/git-novice'],
> > +    'Make': [],
> 
> Not in a usable state - Mike Jackson has just offered us a rework
> that's much more usable, but it isn't ready for prime time yet.

For example, I'm willing to teach the Make lesson however it gets
written (or to just teach Make using a one-off lesson based on how
I've taught Make in the past).  Do we want to use the WIP lesson as a
place-holder for folks like me?  Or do we want to re-populate that
qualification once we have an official Make lesson?

gvwilson: We'll repopulate when and as lessons ripen - otherwise, there's a 
boatload of potential topics to track.  (OOP?  Data viz?  GIS? Etc.)

pbanaszkiewicz:
> Didn't realize we could avoid migrations in the test setup

Yup, me too - until I did `./manage.py migrate --list` on a fresh development database.

workshops/migrations/0012_auto_20150612_0807.py
wking: On Tue, Jun 16, 2015 at 10:16:10AM -0700, Greg Wilson wrote:
> We'll repopulate when and as lessons ripen - otherwise, there's a 
> boatload of potential topics to track.  (OOP?  Data viz?  GIS? Etc.)

But we presumably already have good data for Make, and not for those
other cases.  But repopulating isn't the end of the world.

pbanaszkiewicz:
Shifting to v0.5 since #374 is scheduled for v0.5.

pbanaszkiewicz:
Suppressed by #395.

PR #363, opened 2015-06-13
Removing published flag

gvwilson:
`Event.published` was unnecessary and annoying, so this PR:

1.  Removes it from the model.
2.  Rewrites the generators for upcoming and unpublished events to use the logic we actually want.
3.  Refactors tests to put event generation in a common place.
4.  Modifies some tests to reflect the changes above.


workshops/models.py
```diff
@@ -227,9 +228,14 @@ def ongoing_events(self):
         return queryset
 
     def unpublished_events(self):
-        '''Return a queryset for events that are not yet published.'''
+        '''Return events without URLs that are upcoming or have unknown starts.
```
wking: To avoid catching folks by surprise, I'd suggest adding `help_text` to `Event.url` explaining that it shouldn't be set until an event is ready to be published.  In my limited experience, there's often a setup period while a partially-completed website is available online, but before you'd actually want to point anyone at it.

gvwilson: It might make more sense to change the name from "unpublished" to 
"incomplete" to better reflect the actual semantics. If so, should I put 
that into this PR or do it separately?

workshops/models.py
```diff
@@ -227,9 +228,14 @@ def ongoing_events(self):
...
 
-        return self.filter(published=False)
+        Events are ordered by slug and then by serial number.'''
+
+        future_without_url = (Q(start__gte=datetime.date.today()) & Q(url__isnull=True))
+        unknown_start = Q(start__isnull=True)
+        return self.filter(future_without_url | unknown_start)\
+                   .order_by('slug', 'id')
```
wking: From [PEP 8][1], recommended wrapping is by parentheses, brackets and braces, so maybe:

    return (self.filter(future_without_url | unknown_start)
                .order_by('slug', 'id'))

[1]: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

workshops/models.py
```diff
@@ -227,9 +228,14 @@ def ongoing_events(self):
         return queryset
 
     def unpublished_events(self):
-        '''Return a queryset for events that are not yet published.'''
+        '''Return events without URLs that are upcoming or have unknown starts.
 
-        return self.filter(published=False)
+        Events are ordered by slug and then by serial number.'''
+
+        future_without_url = (Q(start__gte=datetime.date.today()) & Q(url__isnull=True))
```
wking: No  need for the wrapping parens unless you're using them for line continuation.

wking: Actually, no need for the two `Q` objects at all.  Just:

    future_without_url = Q(start__gte=datetime.date.today(), url__isnull=True)

workshops/models.py
wking: And you'll also want to add something to this effect in the `help_text` for `Event.start` too, since you'll also need to have that set before the event is published.  A nice UI if we want to setup the right JavaScript would be to show a “published? {yes|no}” label, and have the “yes” case displayed dynamically when the edit form was configured in a way that would publish the site.  Then folks wouldn't have to track the underlying logic mentally.  But maybe that should wait until we have a better framework for JavaScript tooling.

workshops/test/base.py
```diff
@@ -1,8 +1,8 @@
 import cgi
 import traceback
+from datetime import date, timedelta
 import os
 import re
-import datetime
```
wking: Everyone else imports the full module, so why switch to `from … import …` here?

wking:
On Sat, Jun 13, 2015 at 07:57:10AM -0700, Greg Wilson wrote:
> `Event.published` was unnecessary and annoying…

+1 for more DRY :).  But we don't want to confuse people with
unintended side effects, so we need to make sure we're communicating
the backend logic to folks who are entering the information driving
that logic.

workshops/models.py
wking: On Sat, Jun 13, 2015 at 11:11:45AM -0700, Greg Wilson wrote:
> > @@ -227,9 +228,14 @@ def ongoing_events(self):
> >          return queryset
> >  
> >      def unpublished_events(self):
> > -        '''Return a queryset for events that are not yet published.'''
> > +        '''Return events without URLs that are upcoming or have unknown starts.
> 
> It might make more sense to change the name from "unpublished" to
> "incomplete" to better reflect the actual semantics. If so, should I
> put that into this PR or do it separately?

I dunno.  Unpublished has a clear consequence (“is it going to show up
on our public event list or not?”).  What's “incomplete” about?  For
example, we may want to show public information before we've locked in
the full complement of instructors.  That sounds like “published but
incomplete”.  So I'm unconvinced on the name change.

As for whether that lands in this PR or another, I don't really mind.
If we expect one issue (the logic change or the rename) to stick in
discussion for longer, it's better to split them apart (so the
non-controversial one can go through).  In this case, I expect us to
reach consensus on both the logic change and the rename quickly, so it
shouldn't matter.

pbanaszkiewicz:
Do admins need an "(un)published" indicator? Would they need to take some actions if an event was (un)published?

I'm +1 for the notion of "incomplete" event. If we can generate Eventbrite registration forms from AMY, or copy/send Google Docs surveys, then we probably don't want to send them to events that should not yet start.

So maybe "unfit/ready" instead of "(in)complete"?

Additionally, if you agree with me about Eventbrite events and GDocs surveys, then we need to decide what makes an event ready for publication. A slug? Daterange? Valid homepage? Not-null admin fee?

wking:
On Mon, Jun 15, 2015 at 03:15:22PM -0700, Piotr Banaszkiewicz wrote:
> Do admins need an "(un)published" indicator? Would they need to take
> some actions if an event was (un)published?

Only published events show up here [1], although I'm not actually sure
how that view is populated from AMY data.  The current export view
doesn't seem to be about events [2].

> I'm +1 for the notion of "incomplete" event. If we can generate
> Eventbrite registration forms from AMY, or copy/send Google Docs
> surveys, then we probably don't want to send them to events that
> should not yet start.

Yeah, these should also be tied to our “unpublished” logic (whatever
we want to call it).

> Additionally, if you agree with me about Eventbrite events and GDocs
> surveys, then we need to decide what makes an event ready for
> publication. A slug? Daterange? Valid homepage? Not-null admin fee?

I think the new logic is reasonably clear (if currently undocumented
in the UI).  See discussion here [3].

[1]: http://software-carpentry.org/workshops/
[2]: https://github.com/swcarpentry/amy/blob/v0.3/workshops/views.py#L836-L848
[3]: https://github.com/swcarpentry/amy/pull/363/files#r32372514

workshops/models.py
gvwilson: Mm - I've never liked this, since it's easy to confuse with "make it a tuple".

gvwilson:
@wking updated.

workshops/models.py
```diff
@@ -271,14 +277,14 @@ def uninvoiced_events(self):
...
-    published  = models.BooleanField(default=False)
     site       = models.ForeignKey(Site)
     tags       = models.ManyToManyField(Tag)
     organizer  = models.ForeignKey(Site, related_name='organizer', null=True, blank=True)
     start      = models.DateField(null=True, blank=True)
     end        = models.DateField(null=True, blank=True)
     slug       = models.CharField(max_length=STR_LONG, null=True, blank=True)
-    url        = models.CharField(max_length=STR_LONG, unique=True, null=True, blank=True)
+    url        = models.CharField(max_length=STR_LONG, unique=True, null=True, blank=True,
+                                  help_text='Setting this "publishes" the event.')
```
wking: It's [a bit more complicated][1] than just setting `Event.url`.  You should also document the effect of `Event.start` (both need to be set for publication).

[1]: https://github.com/swcarpentry/amy/pull/363/files#r32372570

workshops/models.py
```diff
@@ -227,9 +228,14 @@ def ongoing_events(self):
         return queryset
 
     def unpublished_events(self):
-        '''Return a queryset for events that are not yet published.'''
+        '''Return events without URLs that are upcoming or have unknown starts.
 
-        return self.filter(published=False)
+        Events are ordered by slug and then by serial number.'''
+
+        future_without_url =Q(start__gte=datetime.date.today(), url__isnull=True)
```
wking: Missing whitespace in `=Q`.

pbanaszkiewicz:
I'm going to merge this with my fixes:

* 'published' fields removed from `EventForm` and `EventFilter`
* undid `Site` model `on_delete` undo (it was `on_delete=models.PROTECT`, then rolled back to nothing, now I'm changing back to `on_delete=models.PROTECT`)
* removed migration 0017_* that was introduced because of above change (from `on_delete=…` to nothing)

Issue #371, opened 2015-06-17
Events with xx date placeholder are listed out of order


amyrbrown:
If I do a search on "northwestern" (for example), the returned list of events is like this:

- 2015-08-21-northwestern
- 2015-04-17-nw
- 2014-10-31-nw
- 2015-xx-xx-northwestern

The workshop with the xx placeholders in the slug is listed at the bottom, instead of somewhere in the 2015's (before? after? I don't think it matters.) Workshops with xx are always listed at the bottom.

pbanaszkiewicz:
So what order do you suggest? Ascending by slug name?

amyrbrown:
Thank you! For this and the other fixes. :)

Issue #369, opened 2015-06-17
Line wrapping on event notes (and other notes)


amyrbrown:
Would be nice to have line wrapping on notes fields so you don't end up with a wide field and horizontal scroll bar, as in https://amy.software-carpentry.org/workshops/event/2015-08-21-northwestern

gvwilson:
Switching to Markdown rendering will solve this (?).

pbanaszkiewicz:
> Switching to Markdown rendering will solve this (?).

Yes, since we'll display in `<div>` rather than `<pre>`.

Issue #368, opened 2015-06-17
Slug field should be top of New Event page


amyrbrown:
The slug field should be the first thing on the New Event page (after the "Published" checkbox) -- it's like the "title" of an event so it's the first thing a user has on their mind when they start creating an event (IMO).

pbanaszkiewicz:
Closing via #373.

Issue #353, opened 2015-06-09
Cannot require country for site


gvwilson:
The "edit site" interface currently requires a country to be specified for a site, but some sites (AAS, WiSE) don't have countries.

wking:
On Mon, Jun 08, 2015 at 05:40:48PM -0700, Greg Wilson wrote:
> The "edit site" interface currently requires a country to be
> specified for a site, but some sites (AAS, WiSE) don't have
> countries.

Sites should have to have countries.  AAS, WiSE, etc. should be hosts,
which shouldn't need countries.  See #120.

gvwilson:
This comes back to the host org/site of workshop split.  Until that is sorted out, I need to be able to add sites that don't have countries.

pbanaszkiewicz:
If this is important, let's have a quick fix (Site.country nullable, blank), and then leave Host/Site split for v0.5 (I like @wking's idea from https://github.com/swcarpentry/amy/pull/120#issuecomment-111562221) with a decent UI (ie. one that doesn't provide more work for admins).

If this is not that important, then let's only have Host/Site split for v0.5 with a decent UI.

Does this sound good, @gvwilson?

gvwilson:
+1 to quick fix.

pbanaszkiewicz:
I hope this is +1 for better fix for v0.5, too :-)

pbanaszkiewicz:
Closing via #373.

PR #373, opened 2015-06-18
Bunch of quickfixes

pbanaszkiewicz:
For:
* #313 
* #353 
* #369 
* #371
* #368

### DIFF
--- a/workshops/templates/workshops/all_airports.html
+++ b/workshops/templates/workshops/all_airports.html
@@ -5,18 +5,24 @@
 {% if all_airports %}
     <table class="table table-striped">
         <tr>
-	    <th>IATA</th>
-	    <th>full name</th>
-	    <th>latitude</th>
-	    <th>longitude</th>
-	</tr>
+            <th>IATA</th>
+            <th>full name</th>
+            <th>latitude</th>
+            <th>longitude</th>
+            <th class="additional-links"></th>
+        </tr>
     {% for airport in all_airports %}
         <tr>
             <td>{{ airport.iata }}</td>
             <td><a href="{% url 'airport_details' airport.iata %}">{{ airport.fullname }}</a></td>
             <td>{{ airport.latitude }}</td>
             <td>{{ airport.longitude }}</td>
-	</tr>
+            <td>
+                <a href="{% url 'airport_details' airport.iata %}" title="View {{ airport }}"><span class="glyphicon glyphicon-option-horizontal"></span></a>
+                &nbsp;
+                <a href="{% url 'airport_edit' airport.iata %}" title="Edit {{ airport }}"><span class="glyphicon glyphicon-pencil"></span></a>
+            </td>
+        </tr>
     {% endfor %}
     </table>
     <div class="pagination">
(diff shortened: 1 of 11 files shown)


workshops/templates/workshops/all_airports.html
```diff
@@ -5,18 +5,24 @@
...
-	    <th>IATA</th>
-	    <th>full name</th>
-	    <th>latitude</th>
-	    <th>longitude</th>
-	</tr>
+        <th>IATA</th>
+        <th>full name</th>
+        <th>latitude</th>
+        <th>longitude</th>
+        <th width="70"></th>
```
gvwilson: shouldn't width be set via CSS?

workshops/templates/workshops/all_events.html
```diff
@@ -4,49 +4,55 @@
...
+        <th>tags</th>
+        <th>instructors</th>
+        <th>slug</th>
+        <th>url</th>
+        <th>site</th>
+        <th>dates</th>
+        <th>attendance</th>
+        <th>admin fee</th>
+        <th>invoiced</th>
+        <th width="70"></th>
```
gvwilson: shouldn't width be set with CSS?

workshops/templates/workshops/all_tasks.html
```diff
@@ -5,18 +5,22 @@
...
         <tr>
-	    <th>event</th>
-	    <th>person</th>
-	    <th>role</th>
-	    <td></td>
-	</tr>
+            <th>event</th>
+            <th>person</th>
+            <th>role</th>
+            <th width="70"></th>
```
gvwilson: Again, set width via CSS?

workshops/views.py
```diff
@@ -817,23 +817,27 @@ def search(request):
                 sites = Site.objects.filter(
                     Q(domain__contains=term) |
                     Q(fullname__contains=term) |
-                    Q(notes__contains=term))
+                    Q(notes__contains=term)) \
+                    .order_by('fullname')
```
gvwilson: Does ordering by fullname involve a function call per sorted element?  If so, what's the speed hit?

workshops/views.py
```diff
@@ -817,23 +817,27 @@ def search(request):
...
+                    Q(github__contains=term)) \
+                    .order_by('family')
             if form.cleaned_data['in_airports']:
                 airports = Airport.objects.filter(
                     Q(iata__contains=term) |
-                    Q(fullname__contains=term))
+                    Q(fullname__contains=term)) \
+                    .order_by('iata')
         else:
             pass # FIXME: error message
```
gvwilson: Please fill in the error message.

workshops/views.py
pbanaszkiewicz: It's a QuerySet method https://docs.djangoproject.com/en/1.8/ref/models/querysets/#order-by

The slowdown shouldn't be noticeable, because databases are good at sorting.

workshops/views.py
pbanaszkiewicz: Going to remove it (not my code + in case of form error, the form displays appropriate message.)

pbanaszkiewicz:
@gvwilson all fixed!

gvwilson:
Thanks - go ahead and merge.

PR #377, opened 2015-06-19
Pagination error with filtering

pbanaszkiewicz:
Switching pages with pagination erased our filters from address, for example:

* page 1: `/workshops/events/?order=slug`
* page 2 should be: `/workshops/events/?order=slug&page=2`, but was: `/workshops/events/?page=2`

Additionally:
* a generic pagination template was used, so we're more DRY
* I decided to change ordering of events to DESC BY SLUG by default; I believe that was the order from before django-filters.

### DIFF
--- a/workshops/filters.py
+++ b/workshops/filters.py
@@ -13,7 +13,7 @@ class EventFilter(django_filters.FilterSet):
             'organizer',
             'invoiced',
         ]
-        order_by = ['slug', '-slug', 'start', '-start', 'end', '-end']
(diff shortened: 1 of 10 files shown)


gvwilson:
Unlikely to get to this before Monday (packing up the house to put stuff 
in storage).  Does this need to be deployed today?

pbanaszkiewicz:
I'd like to get this merged today since it's bugfixing one annoying bug incorporated with django-filters, so if you don't have time then I'll just merge it without review.

gvwilson:
Go for it.

PR #378, opened 2015-06-19
Display person's username

pbanaszkiewicz:
Username was missing on person details page.  This commit adds it on top
of the table with details.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -256,7 +256,7 @@ class PersonForm(forms.ModelForm):
         # don't display the 'password', 'user_permissions',
         # 'groups' or 'is_superuser' fields
         # + reorder fields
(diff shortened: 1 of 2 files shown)


gvwilson:
Thanks - it'll be editable automatically?

pbanaszkiewicz:
It is editable, but it's not in the same place as on the details page. Should I move it to the top?

gvwilson:
I think everything else is consistently ordered, so yes please.

On 2015-06-19 10:49 AM, Piotr Banaszkiewicz wrote:
>
> It is editable, but it's not in the same place as on the details page. 
> Should I move it to the top?
>
> —
> <https://github.com/swcarpentry/amy/pull/378#issuecomment-113536239>.
>

-- 

pbanaszkiewicz:
Done.

PR #380, opened 2015-06-20
Added `make migrations` command to README

carlosp420:
The instructions skipped this step to it was not possible to successfully run
the command `python3 manage.py createsuperuser` as the tables in the sqlite3
database were not complete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,14 +56,20 @@ before starting work on new features.
     ~~~
     $ make database
     ~~~
+    
+5.  Update the models in your local database:
 
-5.  Create an administrator account:
+    ~~~
+    $ make migrations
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
Thank you, @carlosp420. I have no idea how we missed that step…

wking:
On Sat, Jun 20, 2015 at 06:20:57AM -0700, Piotr Banaszkiewicz wrote:
> Thank you, @carlosp420. I have no idea how we missed that step…

I think we were planning on keeping the anonymized data up to date
with migrations ;).

gvwilson:
I fell behind on this - #387 brings the anonymized data up to date.

PR #381, opened 2015-06-20
Nicer pagination bar

carlosp420:
Applied simple modifications to the template `pagination.html`. There will be
shown up to 20 pagination links in cases where there are too many to show.
The pagination bar will show the closest 10 links (before and after) to the
current page.

These screen-shots show how it looks before and after the patch.

a

### DIFF
--- a/workshops/templates/pagination.html
+++ b/workshops/templates/pagination.html
@@ -17,7 +17,7 @@
          {% endif %}
 
          {% if current_query.urlencode != '' %}
-           <a href="?page={{ i }}&{{ current_query.urlencode }}">{{ i }}</a>
+           <a href="?page={{ i }}&amp;{{ current_query.urlencode }}">{{ i }}</a>
          {% else %}
            <a href="?page={{ i }}">{{ i }}</a>
          {% endif %}
--- a/workshops/templates/pagination.html
+++ b/workshops/templates/pagination.html
@@ -1,15 +1,32 @@
-<div class="pagination">
-  <span class="step-links">
+<nav>
+  <ul class="pagination">
      {% if objects.has_previous %}
-         <a href="?{{ prev_query.urlencode }}">previous</a>
+       <li>
+         <a href="?{{ prev_query.urlencode }}" aria-label="Previous">
+           <span aria-hidden="true">&laquo;</span>
+         </a>
+       </li>
      {% endif %}
 
-     <span class="current">
-         Page {{ objects.number }} of {{ objects.paginator.num_pages }}.
-     </span>
+     {% for i in objects.paginator.page_range %}
+       {% if i > objects.number|add:"-10" and i < objects.number|add:"11" %}
+         {% if objects.number == i %}
+           <li class="active">
+         {% else %}
+           <li>
+         {% endif %}
+         <a href="?page={{ i }}">{{ i }}</a>
+       </li>
+       {% endif %}
+     {% endfor %}
 
      {% if objects.has_next %}
-         <a href="?{{ next_query.urlencode }}">next</a>
+       <li>
+         <a href="?{{ next_query.urlencode }}" aria-label="Next">
+           <span aria-label="true">&raquo;</span>
+         </a>
+       </li>
+
      {% endif %}
-  </span>
-</div>
+  </ul>
+</nav>
--- a/workshops/templates/pagination.html
+++ b/workshops/templates/pagination.html
@@ -15,8 +15,14 @@
          {% else %}
            <li>
          {% endif %}
-         <a href="?page={{ i }}">{{ i }}</a>
-       </li>
+
+         {% if current_query.urlencode != '' %}
+           <a href="?page={{ i }}&{{ current_query.urlencode }}">{{ i }}</a>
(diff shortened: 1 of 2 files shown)


workshops/templates/pagination.html
```diff
@@ -1,15 +1,32 @@
...
-         Page {{ objects.number }} of {{ objects.paginator.num_pages }}.
-     </span>
+     {% for i in objects.paginator.page_range %}
+       {% if i > objects.number|add:"-10" and i < objects.number|add:"11" %}
+         {% if objects.number == i %}
+           <li class="active">
+         {% else %}
+           <li>
+         {% endif %}
+         <a href="?page={{ i }}">{{ i }}</a>
```
pbanaszkiewicz: This erases our filtering (see 5a560fb0499e00dbf03dcac0f763b4d7e3a1c292).

gvwilson:
I really like the look, though - thanks!

carlosp420:
doh! I will try to fix it.

pbanaszkiewicz:
@carlosp420: what do you think about a little different approach:

1. drop all `*_query` from `pagination` tag
2. create a new tag `set_page_query`
3. use it in templates in a following way: `<a href="?{% set_page_query i %}">{{ i }}</a>`

New tag's code:
```python
@register.simple_tag(takes_context=True)
def set_page_query(context, page):
    query = context['request'].GET.copy()
    query['page'] = page
    return query.urlencode()
```

pbanaszkiewicz:
@carlosp420 a few weeks ago I was wondering how to accomplish this kind of pagination:
http://pythonhosted.org/flask-ypaginate/_images/paginate-links.png

Notice how '…' are separating different ranges.

I came up with this algorithm today:
```python
#!/usr/bin/env python3

def sections(items, index):
    L = items[0:5]
    M = items[index-3:index+4] or items[0:index+1]
    R = items[-5:]
    L_s = set(L)
    M_s = set(M)
    R_s = set(R)

    D1 = L_s.isdisjoint(M_s)
    D2 = M_s.isdisjoint(R_s)

    if D1 and D2:
        print("L…M…R")
        pagination = L + ['...'] + M + ['...'] + R
    elif not D1 and D2:
        print("LM…R")
        pagination = sorted(L_s | M_s) + ['...'] + R
    elif D1 and not D2:
        print("L…MR")
        pagination = L + ['...'] + sorted(M_s | R_s)
    else:
        print("LMR")
        pagination = sorted(L_s | M_s | R_s)

    print(pagination)


items = list(range(1, 301))
sections(items, 4)
```

Could you somehow add it to this PR? Thanks.

gvwilson:
This feels like the sort of thing that ought to already be in the 
pagination package somewhere (?).

pbanaszkiewicz:
@gvwilson compare my code to [flask-ypaginate](https://github.com/yanhan/flask-ypaginate/blob/master/flaskext/ypaginate.py) package. In my opinion, my code is easier, but I'm biased…

There's a bunch of packages at https://www.djangopackages.com/grids/g/pagination/, but only 2 of them support Python 3.

[Django Endless Pagination](https://www.djangopackages.com/packages/p/django-endless-pagination/) is in a good shape but I doubt we want this sort of pagination.

gvwilson:
Your call.
G

pbanaszkiewicz:
Okay, then I'd like to see the more advanced pagination :-)

carlosp420:
@pbanaszkiewicz that new simple_tag code is better but all templates and views will need to be modified for this to work:
* a variable called `objects` will need to be passed in the context from the view so that it can be used in the pagination template.
* the templates `all_persons`, `all_sites`, etc, will need to be modified to accept the simple_tag instead of the inclusion tag.

A pagination bar with spacers between ranges is much better. Maybe your algorithm can be used to override the Django's pagination method for `page_range` ?

pbanaszkiewicz:
Hi @carlosp420,

1. No, you can use existing tag *and* new `set_page_query` and only change would be in pagination template.

2. Yeah, maybe we can provide our own pagination inheriting from Django and add our own method somewhere so that we can easily traverse through the output of that algorithm in the template.

Cheers,
Piotr

carlosp420:
> No, you can use existing tag and new set_page_query and only change would be in pagination template.

ok, this makes more sense now :smile:

pbanaszkiewicz:
Hi @carlosp420, can you update this PR?

carlosp420:
hi, will try to, in the next few days, hopefully.

Dr. Carlos Peña
Laboratory of Genetics
Department of Biology
University of Turku
20014 Turku
FINLAND

* Associate Editor: Revista peruana de Biología
http://is.gd/TwbW

* The Nymphalidae Systematics Group
http://nymphalidae.utu.fi/db.php

pbanaszkiewicz:
Hi @carlosp420,

I can finish this PR if you don't have time, so please let me know.

Cheers,
Piotr

carlosp420:
ok, it is better if you do it. Don't have much time at the moment!

pbanaszkiewicz:
Closing in favor of #396.

PR #382, opened 2015-06-21
Fix: success/error messages hovered by filter control

pbanaszkiewicz:
A left-side offset was added to the messages on fluid layouts.


gvwilson:
Works for me - please merge and deploy.

PR #387, opened 2015-06-22
Freshly-anonymized data for testing.

gvwilson:
But something's weird: the dashboard doesn't show *any* Unpublished events...  This is probably a bug in `scripts/anonymizer.py`.


pbanaszkiewicz:
Let's don't update the development database now. Let's wait for the schema changes in #395.

PR #395, opened 2015-06-29
Migrating skills (new)

pbanaszkiewicz:
This work extends what @gvwilson did in #360.

I'm submitting it separately because Greg's on holiday.

This is WIP: tests are missing for the upgrading profiles command.

The test run of this command on the data we gathered is promising, ie. most entries that were unable to add lay into one of these categories:

1. airport doesn't exist in our DB
2. user is not an instructor in our DB
3. user changed their email and provided a different name


pbanaszkiewicz:
@wking awaiting your feedback :-)

pbanaszkiewicz:
Merging. We will rework this if it doesn't do what Greg intends.

PR #399, opened 2015-07-01
Model versioning

### LINKED ISSUE #309: Add 'last updated' field to person, site, and event
We should keep track of when changes were last made to our key entries (person, site, and event), and display them where appropriate.

pbanaszkiewicz:
This fixes #309.

The solution:
* shows last 50 items changed on the dashboard
* shows changes made (WIP)
* uses [django-reversion](http://django-reversion.readthedocs.org/)

This is WIP:
* missing showing specific changes
* missing tests

EDIT: it's enabled only for Event and Person models right now.


pbanaszkiewicz:
This implementation is far from perfect, but that's the easiest we can have with `django-reversion`. It's most limitation is that it won't track changes made to related objects, for example: it won't track changes to event if someone changes that event tags (or rather we won't show that tags changed, because it's painful and doesn't work).  Similarly if person is awarded a new badge, or we change person's qualification.

PR #414, opened 2015-07-07
Instructors update fix for 0.5.1

pbanaszkiewicz:
Fixes for upgrade_instructor_profiles command
    
* lessons that don't match are translated into list of lessons (previously they were dropped)
* lessons that don't match are shown as errors
* we update non-instructor profiles, too
* non-existing domains are reported as warnings
* two run modes: either all entries must be correct to proceed, or all correct entries are upgraded
* airport has strong matching
    
Tests were fixed accordingly.


scripts/parse-instructor-info.py
```diff
@@ -37,7 +37,7 @@
         ('Programming with Python', 'swcarpentry', 'python-novice-inflammation', 'swc/python'),
         ('Programming with R', 'swcarpentry', 'r-novice-inflammation', 'swc/r'),
         ('Programming with MATLAB', 'swcarpentry', 'matlab-novice-inflammation', 'swc/matlab'),
-        ('Data Organization in Spreadsheets', 'datacarpentry', 'excel-ecology', 'dc/spreadsheets'),
+        ('Data Organization in Spreadsheets', 'datacarpentry', 'excel-ecology', 'dc/spreadsheet'),
```
gvwilson: We should use "spreadsheets" (the plural form).  What do we have to change elsewhere to make this consistent?

workshops/management/commands/upgrade_instructor_profiles.py
```diff
@@ -41,7 +41,7 @@
         ('Programming with MATLAB', 'swcarpentry',
          'matlab-novice-inflammation', 'swc/matlab'),
         ('Data Organization in Spreadsheets', 'datacarpentry', 'excel-ecology',
-         'dc/spreadsheets'),
+         'dc/spreadsheet'),
```
gvwilson: Right - this should be 'spreadsheets' (plural).

workshops/management/commands/upgrade_instructor_profiles.py
```diff
@@ -106,9 +150,8 @@ def translate(self, record):
             # normalize gender
             new_record['gender'] = self.translate_gender(new_record['gender'])
 
-            # normalize airport (just use first 3 letters of the input, all
-            # upper-case)
-            new_record['airport'] = new_record['airport'][0:3].upper()
+            # normalize airport (upper-case whole thing)
+            new_record['airport'] = new_record['airport'].upper()
```
gvwilson: Agreed.

workshops/test/test_commands.py
```diff
@@ -62,7 +72,7 @@ def test_lessons_translation(self):
                 'https://github.com/datacarpentry/sql-ecology/blob/gh-pages/'
                 'sql.md)',
                 [
-                    'dc/spreadsheets', 'dc/r', 'dc/sql',
+                    'dc/spreadsheet', 'dc/r', 'dc/sql',
```
gvwilson: spreadsheets plural

workshops/management/commands/upgrade_instructor_profiles.py
pbanaszkiewicz: Changing this would require a migration… Are you sure you want this?

gvwilson: Yup, we should use the plural.  Once we've done this migration and 
updated the database with instructor info, we can scrub old migrations.

PR #421, opened 2015-07-09
Add tasks from person edit

### LINKED ISSUE #391: Should be able to add a task from the 'Edit Person' page
We can add Awards from the 'Edit Person' page - should also be able to add tasks.

pbanaszkiewicz:
This fixes #391.

More in the commit messages.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -343,3 +343,19 @@ class PersonAwardForm(forms.ModelForm):
         model = Award
         fields = '__all__'
         widgets = {'person': HiddenInput}
+
+
+class PersonTaskForm(forms.ModelForm):
+
+    event = selectable.AutoCompleteSelectField(
+        lookup_class=lookups.EventLookup,
+        label='Event',
+        required=True,
+        help_text=AUTOCOMPLETE_HELP_TEXT,
+        widget=selectable.AutoComboboxSelectWidget,
+    )
+
+    class Meta:
+        model = Task
+        fields = '__all__'
+        widgets = {'person': HiddenInput}
--- a/workshops/static/css/amy.css
+++ b/workshops/static/css/amy.css
@@ -77,3 +77,9 @@ pre {
 body {
     padding-top: 60px;
 }
+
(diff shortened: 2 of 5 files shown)


pbanaszkiewicz:
I added the tests.

This implementation includes a compromise between UI clearness and usefulness. I'd like someone to decide if it's a change for good or for bad.

gvwilson:
It's a good start - thanks.  A couple of changes:

1.  I'd put the existing awards/tasks *above* the form to add more, not below.
2.  After adding a new award/task, the user should be returned to the same tab, not to the main "Person" tab.

Other than that, :+1:.

pbanaszkiewicz:
Hi Greg,

On Thu, Jul 9, 2015 at 7:23 PM, Greg Wilson <notifications@github.com>
wrote:

> It's a good start - thanks. A couple of changes:
>
>    1. I'd put the existing awards/tasks *above* the form to add more, not
>    below.
>
> I thought about this, and decided a while ago to put form above the
listing.

The reason was that some people (like yourself) have a ton of tasks related
to them, so if admin wanted to add a new task they'd have to scroll down to
the bottom.

Do you still want to shift forms and listings?

>
>    1. After adding a new award/task, the user should be returned to the
>    same tab, not to the main "Person" tab.
>
> It's a downside of using these tabs from Bootstrap, because we cannot
redirect to the specific tab.

The only solution I can think of is to change tabs based on the message
returned from Django messaging framework.

I'll try to work this out.

Cheers,
Piotr

pbanaszkiewicz:
@gvwilson switching tabs works (it won't get any better, I think; unless we switch plugins). Cheers :-)

gvwilson:
Ah, good thinking - yes, leave the form above the tasks.

pbanaszkiewicz:
Do you want to move the award form to the bottom? Pro: there's only a limited number of badges a person can get, so the listing will never be long. Con: inconsistent with the task form placement.

gvwilson:
Please keep it consistent with the task display (form, then listing).

PR #423, opened 2015-07-10
Fix: searching for "name lastname"

### LINKED ISSUE #319: search fails because we search fields separately
A search for 'Greg Wilson' turns up nothing, because 'Greg Wilson' isn't found in forenames, surnames, or site names.  A search for 'Wilson' finds Greg Wilson, plus other Wilsons, because 'Wilson' is found in a surname field.

pbanaszkiewicz:
This fixes #319.

The fix: if the term can be tokenized to 2 words (for example: `"Piotr Banaszkiewicz"` → `["Piotr", "Banaszkiewicz"]`), then the search looks for people who have `personal=FIRSTTERM, family=SECONDTERM` or `personal=SECONDTERM, family=FIRSTTERM`.

Additional improvement was made to the search yielding only one record. In this case, user is redirected immediately to that result's details page.

WIP: missing tests.

### DIFF
--- a/workshops/test/test_search.py
+++ b/workshops/test/test_search.py
@@ -28,22 +28,22 @@ class TestSearchSite(TestBase):
     def test_search_for_site_by_partial_name(self):
         response = self.client.post(reverse('search'),
                                     {'term' : 'Alpha',
-                                     'in_sites' : 'on'})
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
Tests were added.

pbanaszkiewicz:
Bump

PR #426, opened 2015-07-13
Import event data from URL

pbanaszkiewicz:
More details in commit messages.

### DIFF
--- a/workshops/test/test_util.py
+++ b/workshops/test/test_util.py
@@ -10,7 +10,11 @@ from django.test import TestCase
 from django.core.urlresolvers import reverse
 
 from ..models import Site, Event, Role, Person, Task
-from ..util import upload_person_task_csv, verify_upload_person_task
+from ..util import (
+    upload_person_task_csv,
+    verify_upload_person_task,
+    normalize_event_index_url
+)
 
 from .base import TestBase
 
@@ -321,3 +325,23 @@ Harry,,Potter,harry@hogwarts.edu,foobar,Instructor
         self.assertEqual(tasks_pre, tasks_post)
         self.assertEqual(users_pre, users_post)
         self.assertEqual(rv.status_code, 200)
+
+
+class TestEventURLNormalization(TestCase):
+    def setUp(self):
+        self.test_cases = [
+            'http://username.github.io/2015-07-13-City/',
+            'http://username.github.io/2015-07-13-City',
+            'http://username.github.io/2015-07-13-City/index.html',
+            'https://github.com/username/2015-07-13-City/',
+            'https://github.com/username/2015-07-13-City',
+            ('https://github.com/username/2015-07-13-City/blob/'
+             'gh-pages/index.html'),
+        ]
+
+        self.output = ('https://raw.githubusercontent.com/username/'
+                       '2015-07-13-City/gh-pages/index.html')
+
+    def test_normalization(self):
+        for url in self.test_cases:
+            assert normalize_event_index_url(url) == self.output
--- a/workshops/test/test_util.py
+++ b/workshops/test/test_util.py
@@ -382,7 +382,7 @@ eventbrite: 10000000
 ---
 """
         url = 'http://test.github.io/2015-07-13-test/'
-        notes = """VENUE: Euphoric State University
+        notes = """VENUE: Euphoric State University, Highway to Heaven 42, Academipolis, USA
 
 INSTRUCTORS: Hermione Granger, Harry Potter, Ron Weasley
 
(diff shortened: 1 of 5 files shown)


pbanaszkiewicz:
@gvwilson I went with your suggestion. Let me know if you want the AMY to set up instructors / helpers.

The admin attention is required anyway because they need to set up a site (and instructors/helpers after the event is created).

gvwilson:
Where do I go in the UI to launch this?
Thx,
G

On 2015-07-13 11:54 AM, Piotr Banaszkiewicz wrote:
>
> @gvwilson <https://github.com/gvwilson> I went with your suggestion. 
> Let me know if you want the AMY to set up instructors / helpers.
>
> The admin attention is required anyway because they need to set up a 
> site (and instructors/helpers after the event is created).
>
> —
> <https://github.com/swcarpentry/amy/pull/426#issuecomment-120976520>.
>

-- 

pbanaszkiewicz:
Hi Greg,

go to "New Event" and you should see the "Load from URL" button.

Cheers,
Piotr

pbanaszkiewicz:
Ping @gvwilson.

gvwilson:
I'm not going to be able to test this in the next four days - @r-gaia-cs, can you test this?  You'll need to make some dummy workshop websites...

r-gaia-cs:
> can you test this? You'll need to make some dummy workshop websites...

OK. @pbanaszkiewicz please ping me if you don't hear from me until your Saturday afternoon.

pbanaszkiewicz:
Thanks Raniere, but the deadline for v0.6 is today. I'm going to write a release blogpost and then test this PR some more, and then if I don't hear back from you I'll merge.

r-gaia-cs:
This is a UI suggestion. Can we have two forms? The first one for URL and the second one when URL isn't available? I hope that one image help what I want to have:

new-amy

workshops/test/test_util.py
```diff
@@ -321,3 +328,104 @@ def test_upload_existing_user_existing_task(self):
...
+address: Highway to Heaven 42, Academipolis
+country: USA
+language: us
+latlng: 36.998977, -109.045173
+humandate: Jul 13-14, 2015
+humantime: 9:00 - 17:00
+startdate: 2015-07-13
+enddate: 2015-07-14
+instructor: ["Hermione Granger", "Harry Potter", "Ron Weasley",]
+helper: ["Peter Parker", "Tony Stark", "Natasha Romanova",]
```
r-gaia-cs: LOL

pbanaszkiewicz:
@r-gaia-cs but there are two forms. The second one shows up in a modal.

The current workflow if the URL is available:

1. Click "Load…"
2. Paste URL in the new form, submit - the script should pre-fill most of the form
3. Add site
4. Submit

When URL is not available:

1. Fill up all details
2. Submit

workshops/test/test_util.py
pbanaszkiewicz: These are legitimate people in our tests. Greg came up with this, and I think it's brilliant :-)

workshops/util.py
```diff
@@ -315,3 +318,79 @@ def merge_persons(person_from, person_to):
 
     # removes tasks, awards, qualifications in a cascading way
     person_from.delete()
+
+
+class WrongEventURL(Exception):
+    pass
+
+
+def normalize_event_index_url(url):
```
r-gaia-cs: Can we make this opt-out?

pbanaszkiewicz: What do you mean by that?

r-gaia-cs: I know that we more or like that workshop pages are on GitHub Pages but if host A want to use http://web-site-of-host-a.com/swc-workshop? Could we support this? I will not block this pull request only by this but will be nice to have it option available.

pbanaszkiewicz: This would be a significant difficulty. The current code reaches to `raw.githubcontent.com` for the index file of the workshop. How would we read the metadata if the index file is not accessible?

One solution would be to add all necessary variables to the `<meta>` of the generated `index.html` (this solution requires changes to the `workshop-template`). But what about historical workshops? We'd have to support both versions…

r-gaia-cs:
> The current workflow if the URL is available:
>
> 1. Click "Load…"
> 2. Paste URL in the new form, submit - the script should pre-fill most of the form
> 3. Add site
> 4. Submit

This looks reasonable to me.

@pbanaszkiewicz This looks good for me know. :+1: to include it on today release.

workshops/util.py
r-gaia-cs: OK. Let worry about this in the future when GitHub close for business.

workshops/test/test_util.py
gvwilson: > > +instructor: ["Hermione Granger", "Harry Potter", "Ron Weasley",]
> > +helper: ["Peter Parker", "Tony Stark", "Natasha Romanova",]
>
> These are legitimate people in our tests. Greg came up with this, and 
> I think it's brilliant :-)
>

The question is, which team would you bet on in a fight?

workshops/util.py
gvwilson: We only support GitHub-hosted sites for now.

gvwilson: On 2015-07-17 3:30 PM, Raniere Silva wrote:
> OK. Let worry about this in the future when GitHub close for business.
...or gets acquired by Google...

PR #430, opened 2015-07-17
Added ordering field to Meta class for Lessons.

### LINKED ISSUE #422: Skills should be displayed in sorted order
Skills are currently displayed in a more-or-less random order (based I presume on the order in which they were added to the DB).  They should be sorted, either alphabetically (which would put all 'dc' before all 'swc') or by second part (e.g., by what's after the '/').

narayanaditya95:
Sorts all Lessons alphabetically.

Fixes #422.

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -391,6 +391,9 @@ class Lesson(models.Model):
     def __str__(self):
         return self.name
 
+    class Meta:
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
Thank you, @narayanaditya95, for your contribution.

Issue #412, opened 2015-07-06
add field of affiliation for instructors


tracykteal:
It's often useful to know what institute or company a person is at in a particular area. It sounds like we're collecting that data, but it would be good have that field available to view in Amy. It can be just a regular text field. 

I know there will not be uniformity of how people enter their university - some might enter UCLA and some University of California, Los Angeles, but as administrators having that information available to view gives us helpful information for scheduling.

pbanaszkiewicz:
We've recently gathered affiliation via instructor profile survey.

What I'd like to know is if we need to track people who are affiliated with [SCF Organizational Members](http://software-carpentry.org/scf/membership.html) or something more general (as in Tracy's example: UCLA, or AGH-UST as in my case).

gvwilson:
Organizational membership can change over time, so let's record every 
affiliation that people give us.

tracykteal:
I agree we should just have a general organizational affiliation category.

Ultimately it would also be good to track which instructors are at Organizational Member universities. But, as Greg says, that will change, so I'm not quite sure how to track that. 

I can see a use case where an Affiliate university wants to know who its current local instructors are when they're trying to schedule local workshops. Or we will want statistics on how many of our instructors are at member universities. Maybe we'll just have to go through periodically and categorize them ourselves.

pbanaszkiewicz:
Ref #375.

pbanaszkiewicz:
Closed via #433.

PR #433, opened 2015-07-17
Person affiliation field

pbanaszkiewicz:
This solves #412.

New field, `Person.affiliation`, was added. It's being displayed on "Find Instructors" page, too.


pbanaszkiewicz:
We'll need to update affiliation for all instructors after v0.6 release.

PR #437, opened 2015-07-20
Nonnegative administrative fee and attendance for Event

### LINKED ISSUE #435: Event's Admin fee and Attendance can't be negative numbers
We should enforce that the "Admin fee" and "Attendance" are non-negative.

pbanaszkiewicz:
This fixes #435.
Attendance is database-constrained, but admin_fee is only form-constrained. This means that if someone executes code `e.admin_fee = -1200` it will work, but `e.attendance = -36` won't.

So the admin_fee constraint applies only when adding/editing events via form.


r-gaia-cs:
@pbanaszkiewicz +1 to merge.

PR #438, opened 2015-07-20
Event start date must be earlier than it's end date

### LINKED ISSUE #436: Event's start date must be earlier than it's end date
We should enforce that the start date is early that the end date.

pbanaszkiewicz:
This fixes #436.


r-gaia-cs:
@pbanaszkiewicz +1 to merge.

PR #453, opened 2015-07-30
Add "Update from URL" functionality

### LINKED ISSUE #442: Add "Update from URL" button to Event
Event edit form should be extended by adding the possibility to update values from workshop webpage.

pbanaszkiewicz:
This fixes #442.

It's pure JS so it doesn't require tests.


gvwilson:
"It's pure JS so it doesn't require tests."

That's a surprising claim :-)

PR #463, opened 2015-08-12
Workshops API

pbanaszkiewicz:
This 3 new API endpoints add possibility to find:
- all past workshops
- currently running workshops
- upcoming workshops

I hope I didn't mess up ordering in `api/views.py` :-)

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -35,22 +35,25 @@ class EventSerializer(serializers.ModelSerializer):
 
     def get_humandate(self, obj):
         """Render start and end dates as human-readable short date."""
-        start = obj.start
-        end = obj.end
-        if start and not end:
-            return '{:%b %d, %Y}-???'.format(start)
-        elif end and not start:
-            return '???-{:%b %d, %Y}'.format(end)
-        elif not end and not start:
+        return EventSerializer.human_readable_date(obj.start, obj.end)
+
+    @staticmethod
+    def human_readable_date(date1, date2):
+        """Render start and end dates as human-readable short date."""
+        if date1 and not date2:
+            return '{:%b %d, %Y}-???'.format(date1)
+        elif date2 and not date1:
+            return '???-{:%b %d, %Y}'.format(date2)
+        elif not date2 and not date1:
             return '???-???'
 
-        if start.year == end.year:
-            if start.month == end.month:
-                return '{:%b %d}-{:%d, %Y}'.format(start, end)
+        if date1.year == date2.year:
+            if date1.month == date2.month:
+                return '{:%b %d}-{:%d, %Y}'.format(date1, date2)
             else:
-                return '{:%b %d}-{:%b %d, %Y}'.format(start, end)
+                return '{:%b %d}-{:%b %d, %Y}'.format(date1, date2)
         else:
-            return '{:%b %d, %Y}-{:%b %d, %Y}'.format(start, end)
+            return '{:%b %d, %Y}-{:%b %d, %Y}'.format(date1, date2)
 
     class Meta:
         model = Event
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 
-from workshops.models import Badge, Airport, Person
+from workshops.models import Badge, Airport, Person, Event
 
 
 class PersonUsernameSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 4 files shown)


api/serializers.py
```diff
@@ -27,3 +27,37 @@ class ExportInstructorLocationsSerializer(serializers.ModelSerializer):
...
+    def get_humandate(self, obj):
+        """Render start and end dates as human-readable short date."""
+        return EventSerializer.human_readable_date(obj.start, obj.end)
+
+    @staticmethod
+    def human_readable_date(date1, date2):
+        """Render start and end dates as human-readable short date."""
+        if date1 and not date2:
+            return '{:%b %d, %Y}-???'.format(date1)
+        elif date2 and not date1:
```
gvwilson: This should be flagged as an error - a one-day workshop should only have a start date.

api/serializers.py
```diff
@@ -27,3 +27,37 @@ class ExportInstructorLocationsSerializer(serializers.ModelSerializer):
...
+
+    def get_humandate(self, obj):
+        """Render start and end dates as human-readable short date."""
+        return EventSerializer.human_readable_date(obj.start, obj.end)
+
+    @staticmethod
+    def human_readable_date(date1, date2):
+        """Render start and end dates as human-readable short date."""
+        if date1 and not date2:
+            return '{:%b %d, %Y}-???'.format(date1)
```
gvwilson: Please put the date format in a variable somewhere and refer to that throughout.

api/serializers.py
```diff
@@ -27,3 +27,37 @@ class ExportInstructorLocationsSerializer(serializers.ModelSerializer):
...
+        elif date2 and not date1:
+            return '???-{:%b %d, %Y}'.format(date2)
+        elif not date2 and not date1:
+            return '???-???'
```
gvwilson: And if we have both?

api/serializers.py
```diff
@@ -27,3 +27,37 @@ class ExportInstructorLocationsSerializer(serializers.ModelSerializer):
...
+
+        if date1.year == date2.year:
+            if date1.month == date2.month:
+                return '{:%b %d}-{:%d, %Y}'.format(date1, date2)
+            else:
+                return '{:%b %d}-{:%b %d, %Y}'.format(date1, date2)
+        else:
+            return '{:%b %d, %Y}-{:%b %d, %Y}'.format(date1, date2)
```
gvwilson: OK, so why not just harvest the "humandate" field from the index.html file?

api/views.py
```diff
@@ -17,7 +19,13 @@ def get(self, request, format=None):
...
                                      format=format),
             'export-instructors': reverse('api:export-instructors',
-                                          request=request, format=format)
+                                          request=request, format=format),
+            'events-past': reverse('api:events-past',
+                                   request=request, format=format),
+            'events-ongoing': reverse('api:events-ongoing',
+                                      request=request, format=format),
+            'events-upcoming': reverse('api:events-upcoming',
+                                       request=request, format=format),
```
gvwilson: And will you add a fourth which is "all published events, unfiltered"?  I think that's what we want for building the website.

api/views.py
```diff
@@ -41,3 +49,28 @@ def get(self, request, format=None):
                                   .prefetch_related('person_set')
         serializer = ExportInstructorLocationsSerializer(airports, many=True)
         return Response(serializer.data)
+
+
+class ListEvents(APIView):
+    # I wanted to use ListAPIView, but it had problems with the way we test
+    # this code...
```
gvwilson: What problems?

api/views.py
```diff
@@ -41,3 +49,28 @@ def get(self, request, format=None):
...
+    queryset = None  # override this in the subclass
+
+    def get(self, request, format=None):
+        objects = self.queryset.all()
+        serializer = self.serializer_class(objects, many=True)
+        return Response(serializer.data)
+
+
+class PastEvents(ListEvents):
+    queryset = Event.objects.past_events().order_by('-start')
```
gvwilson: Didn't we already have code to do this (and the other event sets)?

api/serializers.py
pbanaszkiewicz: I thought about this and it seems good, and my reasons for not including this change seem now dumb.

api/serializers.py
pbanaszkiewicz: This is a utility function, it should not perform any validation.

By the way, this function is now only used for past/ongoing/upcoming events, so they do have a starting date. Only unpublished may not have any dates.

api/serializers.py
pbanaszkiewicz: Then the second main if-else kicks in (a few lines below).

api/serializers.py
pbanaszkiewicz: There would be two ways to do it and both have flaws:

1. Looking up humandate from index.html file of each workshop means a lot of requests to GitHub - something we wanted to cut out.
2. Storing humandate within AMY database: humandate does not follow one formatting + we can easily calculate it, so why store it?

So I'd not store humandate anywhere.

api/views.py
pbanaszkiewicz: I think I may need to rework it to only return all the published events, so that the `site` can calculate on its own which events are upcoming and which are historical.

So:
```
- past
- ongoing
- upcoming
+ published
```

Do you agree?

api/views.py
pbanaszkiewicz: The ListAPIView used pagination and with our fake request in tests (essentially `request = None`) it was throwing AttributeErrors (due to the fact that request was None and the ListAPIView class used pagination). Switching to APIView got rid of pagination and therefore we can fake requests in tests successfully.

api/views.py
pbanaszkiewicz: That's the code we're using here (`.past_events()`, `.ongoing_events()`, `.upcoming_events()`).

api/views.py
gvwilson: Sure.

api/views.py
gvwilson: +1

pbanaszkiewicz:
@gvwilson: important question. Right now we treat events as published if they have both starting date and URL.
What about location data? Do we want the main site to receive events from AMY without latitude/longitude and country (these three are required by the http://software-carpentry.org/workshops/index.html)?

gvwilson:
Yup - AMY ships everything it has, it's up to the consumer to decide 
what/how to filter.

pbanaszkiewicz:
@gvwilson in that case this PR is probably ready to merge. Give it a last look and merge if it's OK.

Issue #465, opened 2015-08-18
Normalize workshop URLs


gvwilson:
Right now (2015-08-18), workshop URLs are being stored as:

* `https://github.com/user/yyyy-mm-dd-site` (my preferred format)
* `http://github.com/user/yyyy-mm-dd-site/` (`http` instead of `https`, and a trailing slash)
* `http://user.github.io/yyyy-mm-dd-site` (the GitHub Pages URL rather than the repository URL)

URLs should be normalized to simplify search and processing (e.g., for the main web site).

pbanaszkiewicz:
The machinery in #467 provides us with `event.get_repository_url()` and `event.get_website_url()` methods.

If `event.url` is in one of recognized formats (either to repo or to website), these methods will provide the other URL.

So if `event.url` is `http://user.github.io/yyyy-mm-dd-site` the `event.get_repository_url()` will return `https://github.com/user/yyyy-mm-dd-site/` (we can drop the trailing slash if required), but `event.get_website_url` will return `event.url`.

We can change it so that:
1. `event.get_website_url` formats the `event.url` to `https://user.github.io/yyyy-mm-dd-site` even when `event.url` matches any other format for a website URL
2. `event.get_repository_url` formats the `event.url` to `https://github.com/user/yyyy-mm-dd-site` even when `event.url` matches any other format for a repository URL

One other option is to overwrite `event.url` upon save + force users to use specific format on input, but I'm against it because it requires migration of all existing events.

@gvwilson what do you think about this?

gvwilson:
I think we should:

1. Store the normalized form.
2. Put up a great big warning on the "save" page telling people when 
we've changed what they typed in.
3. Migrate existing records - there shouldn't be that many left in the 
database that aren't normalized.

pbanaszkiewicz:
If we want to enable workshop sites hosted outside of GitHub, we should:

1. Put all required variables into `<meta>` in workshop-template
2. Read from these `<meta>` tags instead of parsing `workshop/index.html` in workshop page validation
3. Use only workshop website URL in AMY, not repository URL

@gvwilson if you agree I could start working on this problem soon.

gvwilson:
Please assign it to me for thinking, but it's a low priority for now.

pbanaszkiewicz:
[...]

PR #467, opened 2015-08-18
Fix event url normalization

### LINKED ISSUE #466: URL normalization bug (?)
The Event page displays a website URL and a repository URL. Normally, it seems that when the repository URL is edited, the website URL is automatically updated - more specifically, if the repository URL is `https://github.com/site/yyyy-mm-dd-site`, the website URL becomes `http://site.github.io/yyyy-mm-dd-site`. However, this is not happening with https://amy.software-carpentry.org/workshops/event/2015-08-26-uw-madison (as of 2015-08-18).

pbanaszkiewicz:
This fixes #466.

It fixes normalization and moves it out of `templatetags/links.py` and into two `Event` model methods (tests provided, too — so this should be less erroneous).


workshops/models.py
```diff
@@ -317,6 +317,13 @@ class Event(models.Model):
     latitude = models.FloatField(null=True, blank=True)
     longitude = models.FloatField(null=True, blank=True)
 
+    REPO_REGEX = re.compile(r'https://github\.com/(?P<name>[^/]+)/'
```
gvwilson: Use `https?` in case they give us an `http://` URL instead of an `https://` URL.

workshops/models.py
```diff
@@ -317,6 +317,13 @@ class Event(models.Model):
...
+                            r'(?P<repo>[^/]+)/?')
+    REPO_SCHEMA = 'https://github.com/{name}/{repo}/'
```
gvwilson: `FORMAT` instead of `SCHEMA`.  (The latter makes me think of a database schema.)

workshops/models.py
```diff
@@ -317,6 +317,13 @@ class Event(models.Model):
...
+    WEBSITE_REGEX = re.compile(r'(http|https)://(?P<name>[^.]+)\.github\.'
+                               r'(io|com)/(?P<repo>[^/]+)/?')
```
gvwilson: As above, `https?` for the protocol should work.  Also, if they're giving us `name.github.xyz`, then `xyz` can only be `io` not `com`.

workshops/models.py
```diff
@@ -317,6 +317,13 @@ class Event(models.Model):
...
+    WEBSITE_SCHEMA = 'https://{name}.github.io/{repo}/'
```
gvwilson: As above for `FORMAT` vs `SCHEMA`.

workshops/models.py
```diff
@@ -329,6 +336,34 @@ def __str__(self):
     def get_absolute_url(self):
         return reverse('event_details', args=[self.get_ident()])
 
+    def get_repository_url(self):
+        """Return self.url formatted as it was repository URL.
+
+        Repository URL is: https://github.com/user/repository/
```
gvwilson: "Repository URL is as specified in REPO_FORMAT" (so that the doc string doesn't go stale).

workshops/models.py
```diff
@@ -329,6 +336,34 @@ def __str__(self):
...
+        If it doesn't match, the original URL is returned."""
+        try:
+            mo = self.WEBSITE_REGEX.match(self.url)
```
gvwilson: I would not use an exception here, since it's not clear that you're relying on `mo` being `None`. Instead, I would use:

~~~
mo = ...match...
if mo:
    return blah
else:
    return blarg
~~~

workshops/models.py
```diff
@@ -329,6 +336,34 @@ def __str__(self):
...
+            return self.REPO_SCHEMA.format(**mo.groupdict())
+        except (TypeError, AttributeError, KeyError):
```
gvwilson: ...because catching three different kinds of exceptions is a lot less clear than explicitly testing `mo`.

workshops/models.py
pbanaszkiewicz: I'll change to 'https?', but last time I checked, user subdomains were served from '*.github.com' too.

workshops/models.py
```diff
@@ -329,6 +336,34 @@ def __str__(self):
...
+            # TypeError: self.url is None
+            # AttributeError: mo is None (ie. it didn't match)
+            # KeyError: mo.groupdict doesn't supply required names to format
+            return self.url
+
+    def get_website_url(self):
+        """Return self.url formatted as it was website URL.
+
+        Website URL is: https://user.github.io/repository/
```
gvwilson: "Website URL is as specified in WEBSITE_FORMAT" (again, to avoid doc string going stale).

workshops/models.py
```diff
@@ -329,6 +336,34 @@ def __str__(self):
...
+        If it doesn't match, the original URL is returned."""
+        try:
```
gvwilson: As above for if/else instead of try/except.

workshops/models.py
gvwilson: didn't know that - please keep the `.com`.

pbanaszkiewicz: I think they'll switch that off in future. But I'll keep the `io|com`.

pbanaszkiewicz:
@gvwilson fixed :-)

pbanaszkiewicz:
@gvwilson bump

workshops/test/test_event.py
```diff
@@ -158,6 +158,44 @@ def test_delete_event_with_tasks_and_awards(self):
...
+        links = [
+            'http://user-name.github.com/repo-name',
+            'http://user-name.github.io/repo-name',
+            'https://user-name.github.com/repo-name',
+            'https://user-name.github.io/repo-name',
+            'http://user-name.github.com/repo-name/',
+            'http://user-name.github.io/repo-name/',
+            'https://user-name.github.com/repo-name/',
+            'https://user-name.github.io/repo-name/',
+        ]
```
gvwilson: What happens if the url is `http://wikipedia.org/blarg`?

workshops/test/test_event.py
```diff
@@ -158,6 +158,44 @@ def test_delete_event_with_tasks_and_awards(self):
...
+            assert event.get_repository_url() == REPO
+
+    def test_website_url(self):
+        test_host = Host.objects.all()[0]
+        links = [
+            'http://github.com/user-name/repo-name',
+            'http://github.com/user-name/repo-name/',
+            'https://github.com/user-name/repo-name',
+            'https://github.com/user-name/repo-name/',
+        ]
```
gvwilson: Same question: can you add tests for the case where the URL is completely wrong, and make sure we fail in a useful way?

pbanaszkiewicz: If the URL doesn't match any of our patterns, we return it in a non-altered form.
For example, 'http://wikipedia.org' will give you back 'http://wikipedia.org'.

But you're right, I'll add new tests.

gvwilson: Hm - but we still accept it without a warning?  Should we do that, or 
reject it?

pbanaszkiewicz: I think that would be the solution for #465, not #466.

gvwilson: Agreed - go ahead and merge this one.

Issue #464, opened 2015-08-13
Add read-only access


gvwilson:
1. Implement read-only access.
2. Provide read-only logins to all Steering Committee members.

pbanaszkiewicz:
I'd set it for v0.9, because to do this correctly we should start using group permissions. We have the ability, it's just that our views don't take permissions into account.

pbanaszkiewicz:
Closed via #472.

PR #472, opened 2015-08-25
Permissions

pbanaszkiewicz:
This PR implements #464.

Work so far:
* setting `@permission_required` decorator (in various forms) for all non-read-only views (for example `delete_host` or `person_add`)
* hiding elements of UI for users without specific permissions

So in order to give someone a read-only access to AMY, all we have to do is to enable them to log in.

If we want to have someone as an administrator (like Maneesha or Giacomo), we can either set them as superuser (that's how it's done now) or explicitly select all specific permissions for them.

Things to consider:
* adding permission groups 'administrator', 'steering committee'
* adding a 'can_login' field to `Person`

The permission groups may be just a more verbose way so say that someone belongs to administrators or steering committee.

The `person.can_login` would indicate that someone may be able to log into AMY – right now we kind of don't see that. It's related to `person.is_password_usable() == False`, but it's not possible to any admin to, for example, disable login for someone.

@gvwilson any thoughts?

### DIFF
--- a/workshops/templates/base.html
+++ b/workshops/templates/base.html
@@ -29,8 +29,12 @@
                 {% navbar_element "Badges" "all_badges" %}
                 {% navbar_element "Airports" "all_airports" %}
                 <li class="divider"></li>
+                {% if perms.workshops.add_person %}
                 {% navbar_element "Bulk add persons" "person_bulk_add" %}
+                {% endif %}
+                {% if perms.workshops.add_person and perms.workshops.delete_person %}
                 {% navbar_element "Merge persons" "person_merge" %}
+                {% endif %}
                 {% navbar_element "Find instructors" "instructors" %}
                 {% navbar_element "Debrief" "debrief" %}
                 <li class="divider"></li>
@@ -64,17 +68,29 @@
           </form>
 
           <ul class="nav navbar-nav navbar-right">
+            {% if perms.workshops.add_event or perms.workshops.add_host or perms.workshops.add_person or perms.workshops.add_airport %}
             <li class="dropdown">
               <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false"><span class="glyphicon glyphicon-plus" aria-hidden="true"></span> <span class="caret"></span></a>
               <ul class="dropdown-menu" role="menu">
+                {% if perms.workshops.add_event %}
                 {% navbar_element "New event" "event_add" %}
+                {% endif %}
+                {% if perms.workshops.add_host %}
                 {% navbar_element "New host" "host_add" %}
+                {% endif %}
+                {% if perms.workshops.add_person %}
                 {% navbar_element "New person" "person_add" %}
+                {% endif %}
+                {% if perms.workshops.add_airport %}
                 {% navbar_element "New airport" "airport_add" %}
+                {% endif %}
                 <li class="divider"></li>
+                {% if perms.workshops.add_person %}
                 {% navbar_element "Bulk add people" "person_bulk_add" %}
+                {% endif %}
               </ul>
             </li>
+            {% endif %}
             <li class="dropdown">
               <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false">Hi, {{ user.get_short_name }} <span class="caret"></span></a>
               <ul class="dropdown-menu" role="menu">
--- a/workshops/templates/base.html
+++ b/workshops/templates/base.html
@@ -29,12 +29,8 @@
                 {% navbar_element "Badges" "all_badges" %}
                 {% navbar_element "Airports" "all_airports" %}
                 <li class="divider"></li>
-                {% if perms.workshops.add_person %}
-                {% navbar_element "Bulk add persons" "person_bulk_add" %}
-                {% endif %}
-                {% if perms.workshops.add_person and perms.workshops.delete_person %}
-                {% navbar_element "Merge persons" "person_merge" %}
-                {% endif %}
+                {% navbar_element_permed "Bulk add persons" "person_bulk_add" "workshops.add_person" %}
+                {% navbar_element_permed "Merge persons" "person_merge" "workshops.delete_person" %}
                 {% navbar_element "Find instructors" "instructors" %}
                 {% navbar_element "Debrief" "debrief" %}
                 <li class="divider"></li>
@@ -68,29 +64,17 @@
           </form>
 
           <ul class="nav navbar-nav navbar-right">
-            {% if perms.workshops.add_event or perms.workshops.add_host or perms.workshops.add_person or perms.workshops.add_airport %}
             <li class="dropdown">
               <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false"><span class="glyphicon glyphicon-plus" aria-hidden="true"></span> <span class="caret"></span></a>
               <ul class="dropdown-menu" role="menu">
-                {% if perms.workshops.add_event %}
-                {% navbar_element "New event" "event_add" %}
-                {% endif %}
-                {% if perms.workshops.add_host %}
-                {% navbar_element "New host" "host_add" %}
-                {% endif %}
-                {% if perms.workshops.add_person %}
-                {% navbar_element "New person" "person_add" %}
-                {% endif %}
-                {% if perms.workshops.add_airport %}
-                {% navbar_element "New airport" "airport_add" %}
-                {% endif %}
+                {% navbar_element_permed "New event" "event_add" "workshops.add_event" %}
+                {% navbar_element_permed "New host" "host_add" "workshops.add_host" %}
+                {% navbar_element_permed "New person" "person_add" "workshops.add_person" %}
+                {% navbar_element_permed "New airport" "airport_add" "workshops.add_airport" %}
                 <li class="divider"></li>
-                {% if perms.workshops.add_person %}
-                {% navbar_element "Bulk add people" "person_bulk_add" %}
-                {% endif %}
+                {% navbar_element_permed "Bulk add people" "person_bulk_add" "workshops.add_person" %}
               </ul>
             </li>
-            {% endif %}
             <li class="dropdown">
               <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false">Hi, {{ user.get_short_name }} <span class="caret"></span></a>
               <ul class="dropdown-menu" role="menu">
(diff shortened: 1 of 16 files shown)


gvwilson:
@wking do you have time to look at this one?  You know a lot more about 
this than I do.

workshops/templates/base.html
```diff
@@ -29,8 +29,12 @@
                 {% navbar_element "Badges" "all_badges" %}
                 {% navbar_element "Airports" "all_airports" %}
                 <li class="divider"></li>
+                {% if perms.workshops.add_person %}
```
gvwilson: Instead of showing/not showing the menu element, can you gray out the menu element if permissions aren't there? It's good UI design to have everything always in the same place (muscle memory), just disable the things that people can't get to.

workshops/templates/base.html
```diff
@@ -29,8 +29,12 @@
...
                 {% navbar_element "Bulk add persons" "person_bulk_add" %}
+                {% endif %}
+                {% if perms.workshops.add_person and perms.workshops.delete_person %}
```
gvwilson: As above.

workshops/templates/base.html
```diff
@@ -64,17 +68,29 @@
           </form>
 
           <ul class="nav navbar-nav navbar-right">
+            {% if perms.workshops.add_event or perms.workshops.add_host or perms.workshops.add_person or perms.workshops.add_airport %}
```
gvwilson: Again, show them but gray them out. Also, is there an easy way to combine multiple atomic permissions into a single permission so that you can say "if perms.workshops.add_anything" ?  Because spelling out the entire range of atomic permissions in a template strikes me as error-prone.

workshops/templates/workshops/airport.html
```diff
@@ -11,7 +11,11 @@
 </table>
 
 <div class="clearfix">
+  {% if perms.workshops.change_airport %}
```
gvwilson: Again, gray out instead of not showing.

workshops/views.py
```diff
@@ -136,6 +135,19 @@ def as_view(cls, **kwargs):
         view = super(LoginRequiredMixin, cls).as_view(**kwargs)
         return login_required(view)
 
+
+class PermissionRequiredMixin(object):
+    """
+    Mixin for allowing only users with specific permissions to access the view.
+    """
+    perms = ''  # permission name or a list of them
```
gvwilson: Shouldn't this be an empty list??

workshops/views.py
```diff
@@ -595,6 +627,12 @@ class PersonPermissions(LoginRequiredMixin, UpdateViewContext):
 def person_password(request, person_id):
     user = get_object_or_404(Person, pk=person_id)
 
+    # Either the user requests change of their own password, or someone with
+    # permission for changing person does.
+    if not ((request.user == user) or
```
gvwilson: Found this a bit difficult to parse, but can't think of a simpler way to write it.

gvwilson:
Implementation looks good, but I would much rather have inaccessible 
menu items grayed out than disappear in the templates.  (Can you define 
a helper function to use in the templates to do this?)

workshops/templates/base.html
pbanaszkiewicz: Yep, it's probably possible. Will check in an hour.

workshops/templates/base.html
pbanaszkiewicz: I don't think there's a sane way to do this. Or it's just undocumented.

workshops/views.py
pbanaszkiewicz: Yeah, in one version I was trying out with XOR, but it turned out to be bad.

pbanaszkiewicz:
@gvwilson sure, I'll make it a custom template tag.

workshops/templates/base.html
gvwilson: Huh - post a question to Stack Overflow?  Seems like there *ought* to be 
a point where you could inject something that would add your own Boolean 
to perms.workshops before the template is rendered. (But it's a very low 
priority...)

pbanaszkiewicz: With "gray-out" we don't need to check this long chain of permissions. We'll only check individual URL.

gvwilson: Nice.

pbanaszkiewicz:
@gvwilson added some changes so that the items will be grayed-out.
The only thing impossible to gray-out was glyphicon links (pencils) to
object-edit views. They remain hidden for now.

I'll add permission groups in next commit.

pbanaszkiewicz:
@gvwilson:

* added groups 'administrators' and 'steering committee' through migration. I'll assign folks once this is on the server.
* enabled admin interface at `/workshops/admin/`. It's only possible to manage groups - I thought it'd easier than writing whole interface just for them. The only people to access that page will be me and you, ie. superusers (current administrators will be revoked superuser status, and assigned to 'administrators' group instead).

gvwilson:
Ready for re-review?

pbanaszkiewicz:
Yes, I think so.

workshops/templates/workshops/airport.html
```diff
@@ -11,7 +11,15 @@
...
   <p class="edit-object pull-left"><a href="{% url 'airport_edit' airport.iata %}" class=" btn btn-primary">Edit</a></p>
+  {% else %}
+  <p class="edit-object pull-left"><a href="{% url 'airport_edit' airport.iata %}" class=" btn btn-primary disabled">Edit</a></p>
+  {% endif %}
```
gvwilson: What about this instead:

~~~
<p class="..."><a href="{% url ... %}" class="btn btn-primary {%if not perms.workshops.change_airport %}disabled{% endif %}">Edit</a></p>
~~~

That cuts it from 5 lines to 1 and makes the intention clearer by removing a lot of duplication.  (I had to scan the line a couple of times to convince myself there was only one significant difference.)

workshops/templates/workshops/airport.html
```diff
@@ -11,7 +11,15 @@
...
+  {% if perms.workshops.delete_airport %}
   <p class="delete-object pull-right"><a href="{% url 'airport_delete' airport.iata %}" onclick='return confirm("Are you sure you wish to remove \"{{ airport }}\"?")' class="btn btn-danger">Delete airport</a></p>
+  {% else %}
+  <p class="delete-object pull-right"><a href="{% url 'airport_delete' airport.iata %}" class="btn btn-danger disabled">Delete airport</a></p>
+  {% endif %}
```
gvwilson: Again, is there a way to move the conditional inline to eliminate duplication?

workshops/templates/workshops/all_airports.html
```diff
@@ -9,7 +9,11 @@
 {% endblock %}
 
 {% block content-fluid %}
+    {% if perms.workshops.add_airport %}
     <p><a href="{% url 'airport_add' %}" class="btn btn-success">New airport</a></p>
+    {% else %}
+    <p><a href="{% url 'airport_add' %}" class="btn btn-success disabled">New airport</a></p>
+    {% endif %}
```
gvwilson: Same as above (move conditional inline to eliminate duplication) - I'll skip duplicating this comment below :-)

workshops/templatetags/navigation.py
```diff
@@ -4,6 +4,28 @@
 register = template.Library()
 
 
+def make_element(title, url, active=False, disabled=False):
+    screen_reader = ''
```
gvwilson: Docstring please - I don't know what this does.

workshops/templatetags/navigation.py
```diff
@@ -12,23 +34,38 @@ def navbar_element(context, title, url_name):
     is later reversed into proper URL.
     """
     url = reverse(url_name)
-    return navbar_element_url(context, title, url)
+    active = context['request'].path == url
```
gvwilson: I don't understand why this test works?

workshops/templatetags/navigation.py
```diff
@@ -12,23 +34,38 @@ def navbar_element(context, title, url_name):
...
+    Works like `navbar_element`, but also disables if user doesn't have
+    permissions.
+
+    `perms` can be a comma-separated string of permissions or a single
+    permission.
     """
-    request = context['request']
+    url = reverse(url_name)
+    active = context['request'].path == url
+    disabled = True
```
gvwilson: Why bother setting a default? (See comment below.)

workshops/templatetags/navigation.py
```diff
@@ -12,23 +34,38 @@ def navbar_element(context, title, url_name):
...
 
-    active = ""
-    screen_reader = ""
+    # check permissions
+    perms_ctx = context['perms']
+    perms = perms.split(',')
+    perms = map(lambda x: x in perms_ctx, perms)
+    if all(perms):
+        disabled = False
```
gvwilson: Why not just `disabled = not all(perms)` and get rid of the default setting to `True` above?

gvwilson:
Back to @pbanaszkiewicz

workshops/templates/workshops/airport.html
pbanaszkiewicz: I'm against these changes.

Here are the reasons:

1. These are logical paths in the templates, and it's easier to change them in future if you want to take a different action if some conditions are (not) met
2. You're easier to spot these paths if they are marked like: `{% if %} \n blah \n {% endif %}`
3. Highlighting Django template / Jinja syntax within HTML attributes is flawed (another reason why it's so hard to spot these inline, in-attr conditionals)
4. Too long lines (even with wrapping) make it harder to understand the code

So, generally, I'm opting for code duplication in sake of its' clearness.

gvwilson: I'll defer to you - if you're happy with the rest of this, please merge.

pbanaszkiewicz: I'll add some other things you requested and get it back to you for final approval :)

workshops/templatetags/navigation.py
pbanaszkiewicz: It's the same as:
```python
active = True if context['request'].path == url else False
```

pbanaszkiewicz:
@gvwilson the ball is in your court now.

workshops/templatetags/navigation.py
gvwilson: But doesn't `context['request'].path == url` return a Boolean?

pbanaszkiewicz: Yes, it does. If the `url` and the `request.path` are the same, then the navigation element should be highlighted as active.

gvwilson:
:+1:

pbanaszkiewicz:
All right, so I'll merge this and work on rebasing the other PR (sigh…).

gvwilson:
"No good deed goes unpunished."

PR #481, opened 2015-08-30
Adding target to Makefile to force pip upgrade

gvwilson:


pbanaszkiewicz:
Okay, don't be mad, but we could also upgrade bower's components here. `bower update` will work.

PR #482, opened 2015-08-31
Prevent users from entering 'http://' or '/' as part of Host.domain

### LINKED ISSUE #468: Illegal characters in Host.domain
Putting `http://domain.com/` in `Host.domain` breaks the server, because we don't match for these characters in `urls.py` (only `\w` regex).

I think we could make a validator that disallows `\W` characters.

pbanaszkiewicz:
This fixes #468.
Includes regression test.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -419,9 +419,10 @@ class HostForm(forms.ModelForm):
     domain = forms.CharField(
         max_length=Host._meta.get_field('domain').max_length,
         validators=[
-            RegexValidator('[^\w\.-]+', inverse_match=True,
-                           message='Do NOT enter "http://" part or the '
-                                   'trailing slash.')
+            RegexValidator(
+                '[^\w\.-]+', inverse_match=True,
+                message='Please enter only the domain (such as "math.esu.edu")'
+                        ' without a leading "http://" or a trailing "/".')
         ],
     )
 
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -1,4 +1,5 @@
 from django import forms
+from django.core.validators import RegexValidator
 from django.forms import HiddenInput, CheckboxSelectMultiple
 
 from crispy_forms.helper import FormHelper
@@ -9,7 +10,7 @@ from django_countries.fields import CountryField
 from selectable import forms as selectable
 
 from workshops.models import (
-    Award, Event, Lesson, Person, Task, KnowledgeDomain, Airport,
+    Award, Event, Lesson, Person, Task, KnowledgeDomain, Airport, Host,
 )
 from workshops import lookups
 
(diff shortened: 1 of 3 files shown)


workshops/forms.py
```diff
@@ -412,3 +413,18 @@ class Meta:
...
         fields = '__all__'
         widgets = {'person': HiddenInput}
+
+
+class HostForm(forms.ModelForm):
+    domain = forms.CharField(
+        max_length=Host._meta.get_field('domain').max_length,
+        validators=[
+            RegexValidator('[^\w\.-]+', inverse_match=True,
+                           message='Do NOT enter "http://" part or the '
```
gvwilson: Change the message to "Please enter only the domain (such as math.esu.edu) without a leading http:// or a trailing /"

workshops/test/test_host.py
```diff
@@ -47,3 +47,23 @@ def test_host_delete(self):
...
                 Host.objects.get(domain=host_domain)
+
+    def test_host_invalid_chars_in_domain(self):
+        """Ensure users can't put wrong characters in the host's domain field.
+
+        Invalid characters are any that match `[^\w\.-]+`, ie. domain is
+        allowed only to have alphabet-like chars, dot and dash.
+
+        The reason for only these chars lies in `workshops/urls.py`.  The regex
+        for the host_details URL has `[\w\.-]+` matching...
```
gvwilson: So can we share the URL-matching pattern in a constant somewhere that everyone else refers to?

workshops/forms.py
```diff
@@ -412,3 +413,18 @@ class Meta:
         model = Task
         fields = '__all__'
         widgets = {'person': HiddenInput}
+
+
+class HostForm(forms.ModelForm):
+    domain = forms.CharField(
+        max_length=Host._meta.get_field('domain').max_length,
+        validators=[
+            RegexValidator('[^\w\.-]+', inverse_match=True,
```
gvwilson: Does this regex duplicate one elsewhere in the code?  The comment on the test below suggests so.

pbanaszkiewicz: It matches the inverse, so the regex in `urls.py` is `[\w\.-]+`, here it's `[^\w\.-]+`.

workshops/test/test_host.py
pbanaszkiewicz: Not sure if that's a good idea, since we have two different regexes (see https://github.com/swcarpentry/amy/pull/482/files#r38395883).

workshops/forms.py
gvwilson: Right, but why match inverse?  Why not match exact, so that the regexes 
can be the same?

pbanaszkiewicz: `RegexValidator` throws `ValidationError` when it doesn't match and `inverse_match == False`.
It throws `ValidationError` when it does match and `inverse_match == True`.

Now, I want the error to show up when any of these chars is encountered: `:/` (and others, but let's keep the example short).

So if I used `[\w]+` and `inverse_match == False` then:
* `http` would pass
* `http://` would pass
* `:/` wouldn't pass

I only want to show errors if the wrong characters are found in the `Host.domain`. If I used `[\w]+` instead of `[^\w]+` it would mean "show error when *good* chars are in the Host.domain".

gvwilson: gotcha.

pbanaszkiewicz:
@gvwilson I updated the error message.

gvwilson:
:+1:

PR #494, opened 2015-09-10
Add more options to Event.invoiced

### LINKED ISSUE #469: Allow "N/A" & other options for "Invoiced"
Looking at amy/workshops/models.py it looks like this defined as bool.
However, there are cases when the actual use case is not bool.  Invoiced = No/False may be the case because we need to invoice but we haven't yet, but it may also be the case because they don't need to be invoiced because they are self-organized or other arrangements were made.  We need to differentiate among  these situations.

pbanaszkiewicz:
Actually the Event.invoiced field was migrated to text field and renamed
`invoice_status` since it doesn't answer the question
"Was the workshop's host invoiced?", but rather shows the status of
workshop's invoicing.

This commit provides:
* all necessary migrations
* display changes (`invoiced|yesno` -> `get_invoice_status_display`
  mostly)
* test updates

This fixes #469.

### DIFF
--- a/workshops/filters.py
+++ b/workshops/filters.py
@@ -74,7 +74,7 @@ class EventFilter(django_filters.FilterSet):
             'tags',
             'host',
             'administrator',
-            'invoiced',
+            'invoice_status',
         ]
         order_by = ['-slug', 'slug', 'start', '-start', 'end', '-end']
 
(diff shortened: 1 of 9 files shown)


gvwilson:
If @maneesha is happy with the workflow and user interface, :+1: to merge.

maneesha:
looks good.

PR #496, opened 2015-09-11
'Online' country

### LINKED ISSUE #492: Representing online courses
1. Create a "country" for the World-Wide Web in AMY.
2. Add a flag (icon to be chosen) to represent that "country" on the main website.
3. Make sure the two agree on the 2-letter ISO code for "online" :-)

pbanaszkiewicz:
This fixes #492.

The solution is:
* new country 'Online' (ISO code: `W3`)
* new flag for that country (oblate black-white globe)
* migration for events that have 'online' tag; migration is setting the correct country for these events


gvwilson:
Flag goes in the 'site' repository, not in the AMY repository, and we'll need it in four sizes.  We'll also need an eu.png in each size given that you're adding it as well. Otherwise, looks good.

pbanaszkiewicz:
We use flags in AMY, too.

European Union was added because I saw some historical events using it.

I'll send more flags later today.

gvwilson:
Really?  Where?

pbanaszkiewicz:
Now there're gone, so maybe I'll get rid of EU.

pbanaszkiewicz:
@gvwilson awaiting review.

PR #498, opened 2015-09-15
Checking PDF certificates against database

gvwilson:
Adds a new management command to check awards listed in database against certificates stored on disk.

### DIFF
--- a/workshops/management/commands/check_certificates.py
+++ b/workshops/management/commands/check_certificates.py
@@ -1,0 +1,56 @@
+import sys
+import os
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management.base import BaseCommand, CommandError
+from workshops.models import Award, Badge, Person
+
+
+class Command(BaseCommand):
+    args = '/path/to/certificates'
+    help = 'Report inconsistencies in PDF certificates.'
+
+    def handle(self, *args, **options):
+        '''Main entry point.'''
+
+        if len(args) != 1:
+            raise CommandError('Usage: check_certificates /path/to/dir')
+        path_to_root = args[0]
+
+        awards = self.get_awards()
+        for (name, badge) in awards:
+            db_people = self.get_db_people(badge)
+            cert_path = os.path.join(path_to_root, name)
+            if not os.path.isdir(cert_path):
+                print('No directory {0}'.format(name))
+            else:
+                file_people = self.get_file_people(cert_path)
+                self.report_missing('database but not disk', name, db_people - file_people)
+                self.report_missing('disk but not database', name, file_people - db_people)
+
+    def get_awards(self):
+        '''Get all available awards as list of lower-case name and badge pairs.'''
+
+        return [(b.name.lower(), b) for b in Badge.objects.all()]
+
+    def get_db_people(self, badge):
+        '''Get set of usernames of all people with the given badge.'''
+
+        return set([a.person.username for a in Award.objects.filter(badge=badge)])
+
+    def get_file_people(self, path):
+        '''Get names of all people with the given certificate.'''
+
+        return set([os.path.splitext(e)[0]
+                    for e in os.listdir(path)
+                    if e.endswith('.pdf')])
+
+    def report_missing(self, title, kind, items):
+        '''Report missing items.'''
+        if items:
+            print('{0} {1}'.format(kind, title))
+            for i in sorted(list(items)):
+                try:
+                    p = Person.objects.get(username=i)
+                    print(' {0}: {1}'.format(i, p))
+                except ObjectDoesNotExist:
+                    print(' {0}'.format(i))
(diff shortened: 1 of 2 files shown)


workshops/management/commands/check_certificates.py
```diff
@@ -0,0 +1,56 @@
+import sys
+import os
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management.base import BaseCommand, CommandError
+from workshops.models import Award, Badge, Person
+
+
+class Command(BaseCommand):
+    args = '/path/to/certificates'
```
pbanaszkiewicz: It's waaay better to use `optparse` capabilities of Django. Take for example https://github.com/swcarpentry/amy/blob/develop/workshops/management/commands/upgrade_instructor_profiles.py#L75

You'd use `add_parameter` method and it would:

1. Give you a very nice help msg
2. Validate parameters for you

workshops/management/commands/check_certificates.py
```diff
@@ -0,0 +1,56 @@
...
+    help = 'Report inconsistencies in PDF certificates.'
+
+    def handle(self, *args, **options):
+        '''Main entry point.'''
+
+        if len(args) != 1:
+            raise CommandError('Usage: check_certificates /path/to/dir')
```
pbanaszkiewicz: This will be gone when you switch to optparse.

workshops/management/commands/check_certificates.py
```diff
@@ -0,0 +1,56 @@
...
+            db_people = self.get_db_people(badge)
+            cert_path = os.path.join(path_to_root, name)
+            if not os.path.isdir(cert_path):
+                print('No directory {0}'.format(name))
+            else:
+                file_people = self.get_file_people(cert_path)
+                self.report_missing('database but not disk', name, db_people - file_people)
+                self.report_missing('disk but not database', name, file_people - db_people)
+
+    def get_awards(self):
```
pbanaszkiewicz: This actually gets you badges, not awards.

workshops/management/commands/check_certificates.py
```diff
@@ -0,0 +1,56 @@
...
+        '''Get all available awards as list of lower-case name and badge pairs.'''
+
+        return [(b.name.lower(), b) for b in Badge.objects.all()]
+
+    def get_db_people(self, badge):
+        '''Get set of usernames of all people with the given badge.'''
+
+        return set([a.person.username for a in Award.objects.filter(badge=badge)])
```
pbanaszkiewicz: There's a better way: https://docs.djangoproject.com/en/1.8/ref/models/querysets/#values-list

```python
return set(Award.objects.filter(badge=badge).values_list('person.username', flat=True))
```

workshops/management/commands/check_certificates.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def report_missing(self, title, kind, items):
+        '''Report missing items.'''
+        if items:
+            print('{0} {1}'.format(kind, title))
+            for i in sorted(list(items)):
+                try:
+                    p = Person.objects.get(username=i)
+                    print(' {0}: {1}'.format(i, p))
+                except ObjectDoesNotExist:
```
pbanaszkiewicz: Not sure about this, but it may be better to use `Person.DoesNotExist` exception.

pbanaszkiewicz:
@gvwilson I think there aren't many changes, mostly switch to optparse and values_list.

gvwilson:
Thanks, will do.

workshops/management/commands/check_certificates.py
gvwilson: This line doesn't work as written (for me):
~~~
  File "/Users/gvwilson/anaconda/envs/py3k/lib/python3.3/site-packages/django/db/models/options.py", line 552, in get_field
    return self.fields_map[field_name]
KeyError: 'person.username'
~~~

pbanaszkiewicz: Interesting. You're right, there's an error there, but when I tried this line out, mine Python and Django installation threw: `django.core.exceptions.FieldError: Cannot resolve keyword 'person.username' into field. Choices are: awarded, badge, badge_id, event, event_id, id, person, person_id`.

Anyway, there should be `'person__username'` instead of `'person.username'`.

gvwilson: Nice - got it.

workshops/management/commands/check_certificates.py
pbanaszkiewicz: Is this line still necessary after the switch to optparse?

gvwilson: Fixed.

pbanaszkiewicz:
All right, this looks good. Again, I'll merge once the v0.9.5 is done.

PR #499, opened 2015-09-15
Removing code to check workshop URLs

gvwilson:
We are no longer storing workshop URLs in the `site` repo, so we don't need to check them against the database.


pbanaszkiewicz:
Looks okay. I'll merge once we're done with v0.9.5.

PR #502, opened 2015-09-22
Update settings

pbanaszkiewicz:
This PR:

* changes default logging settings in order to silence SuspiciousOperation errors that were spamming our mailboxes
* adds updated email address to admins' mailing list - this change is present in production, I just want to add it to the codebase and maybe release v0.9.7 after that?

### DIFF
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -107,6 +107,32 @@ TEMPLATES = [
     }
 ]
 
+LOGGING = {
+    'version': 1,  # necessary
+
+    # merge with default settings instead of overwriting
+    'disable_existing_loggers': False,
+
+    'handlers': {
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
I'm closing this PR without merging.

Reason: with current [fail2ban](http://www.fail2ban.org/wiki/index.php/Main_Page) settings, we get only 1 email notification about wrong HTTP_HOST (previously it was like up to 17 emails per night).

If 64b834d was merged, we'd never see any notifications and would never know that there's something to block (for example attackers change their strategy and we don't get notifications about their newest attempts).

To sum up: with fail2ban we keep up to 1 email per IP per day, but if the attackers change their strategy we'll be notified.

And I'll cherry-pick 7cb9979 into `develop`.

PR #504, opened 2015-09-24
Making form purpose clearer in titles.

gvwilson:
The forms for requesting SWC and DC workshops didn't say which kind of workshop it was - this adds that. It also make the title of the instructor info update form clearer.

Question: why are the SWC and DC workshop request forms classes, while the instructor info update form is a function?


pbanaszkiewicz:
> Question: why are the SWC and DC workshop request forms classes, while the instructor info update form is a function?

Instructor's update view doesn't have to be inherited. Originally SWC and DC workshop request views were functions too, but they shared **a lot** of the same code.

PR #506, opened 2015-09-24
Use separate templates for SWC/DC workshop requests

### LINKED ISSUE #505: Forms are not styled with AMY logo or other logos
The workshop request forms and the instructor info update form should display the AMY logo and the SWC or DC logos.

pbanaszkiewicz:
The intent is to add organizations' logos to respective workshop request
pages.

This fixes #505.

The things to consider:
* scale DC image down?
* add AMY logo too?


gvwilson:
What about passing the logo name in the context?  That way, one template would still serve for both forms.

pbanaszkiewicz:
@gvwilson I decided to use template inheritance for a number of reasons:

1. Similarity to view inheritance
2. Django views in most cases use similar approach with template name in a variable, instead of different contexts

But of course passing logo name in a context is just another approach, and I don't think any of them is worse.

gvwilson:
Your call.

pbanaszkiewicz:
@gvwilson I'd stay with template inheritance.

What about DC logo? Is it too big or not?

gvwilson:
It seems pretty big - how big is the SWC logo?

pbanaszkiewicz:
The logos are the same as on this page: http://swcarpentry.github.io/instructor-retreat-2015/

Except that DC's logo is scaled down, but you can click on the image and select "View image" from context menu.

SWC logo is in natural size.

gvwilson:
Hm - I think Data Carpentry's logo has been squished vertically (at 
least, the text looks funny).  I'll check with Tracy.

pbanaszkiewicz:
@gvwilson you're right, it's squished, but the original logo (https://swcarpentry.github.io/instructor-retreat-2015/img/data-carpentry-banner.png) isn't that bad. And that's the logo I used in the form. You can compare the size with SWC logo (https://swcarpentry.github.io/instructor-retreat-2015/img/software-carpentry-banner.png) in your browser.

PR #509, opened 2015-09-28
Automatically count event attendees

### LINKED ISSUE #508: Get attendance count from list of learners
See https://amy.software-carpentry.org/workshops/event/2015-08-27-ualberta for example.  List of learners was imported into AMY but attendance count is still null.

Can we have something where attendance = null by default, can be updated by direct user input OR count(learners), whichever is greater.

(Also I've been interpreting attendance to be only learners, not helpers/instructors/etc)

pbanaszkiewicz:
This fixes #508.

### DIFF
--- a/workshops/test/test_util.py
+++ b/workshops/test/test_util.py
@@ -98,6 +98,7 @@ class CSVBulkUploadTestBase(TestBase):
                                         fullname='Test Host')
 
         Role.objects.create(name='Instructor')
+        Role.objects.create(name='learner')
         Event.objects.create(start=datetime.now(),
                              host=test_host,
                              slug='foobar',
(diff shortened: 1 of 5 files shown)


workshops/util.py
```diff
@@ -227,6 +235,13 @@ def create_uploaded_persons_tasks(data):
             except ObjectDoesNotExist as e:
                 raise ObjectDoesNotExist('{0} (for {1})'.format(str(e), row))
 
+    for event in events:
+        # count attendees
+        learners = event.task_set.filter(role__name='learner').count()
+        # update attendees count if number of learner tasks > event.attendance
+        Event.objects.filter(pk=event.pk, attendance__lt=learners) \
+                     .update(attendance=learners)
```
gvwilson: Does this work if event.attendance is NULL as well?

workshops/views.py
```diff
@@ -853,6 +853,12 @@ def event_edit(request, event_ident):
                     ),
                 )
 
+                # if event.attendance is lower than number of learners, then
+                # update the attendance
+                learners = event.task_set.filter(role__name='learner').count()
+                Event.objects.filter(pk=event.pk, attendance__lt=learners) \
+                             .update(attendance=learners)
+
```
gvwilson: Is this the same logic as in util.py?  Can it be consolidated so that there's one place where we do this calculation and update?  Because it's highly likely to evolve.

workshops/util.py
pbanaszkiewicz: No, it doesn't - `attendance` needs to be numeric (ie. non-null). You think it should work for `attendance_isnull=True`?

gvwilson: If you can guarantee that attendance is always non-NULL... otherwise, yeah.

pbanaszkiewicz:
@gvwilson updated.

workshops/test/test_util.py
```diff
@@ -329,6 +330,42 @@ def test_upload_existing_user_existing_task(self):
         self.assertEqual(users_pre, users_post)
         self.assertEqual(rv.status_code, 200)
 
+    def test_attendance_increases(self):
+        """
+        Check if uploading tasks with role "learner" increase event's
+        attendance.
+        """
+        foobar = Event.objects.get(slug="foobar")
+        assert foobar.attendance is None  # testing for None case
```
gvwilson: I think the comment is redundant :-)

gvwilson:
:+1: - is there a migration to update the attendance counts of existing events, or do we do that by hand after applying this change?

pbanaszkiewicz:
I honestly did not think about a migration. You're right, there should be one for the historical events.

gvwilson:
Can you please add, and then I'll re-review?
Thakns,
Greg

On 2015-09-29 2:01 PM, Piotr Banaszkiewicz wrote:
>
> I honestly did not think about a migration. You're right, there should 
> be one for the historical events.
>
> —
> <https://github.com/swcarpentry/amy/pull/509#issuecomment-144051309>.
>

-- 

pbanaszkiewicz:
@gvwilson sure, added :)

gvwilson:
:+1:

PR #510, opened 2015-09-29
API: use website URLs

### LINKED ISSUE #501: Wrong link being sent to main website
(Amended) use event.get_website_url to fetch the URL for a workshop when building the main website in order to prevent problems of the kind described below.

----

1. Go to http://software-carpentry.org/workshops/index.html#future
2. Look at the Sept 28-29 workshop at Oxford
3. The link is to the GitHub repository, *not* to the GitHub Pages website.
4. But if you look in AMY at https://amy.software-carpentry.org/workshops/event/2015-09-28-oxford, the website URL and repository URL fields look correct.

Same issue shows up with several other workshops (e.g., Sept 17-18 at Western University Sydney links to the repo, not to the gh-pages site).

pbanaszkiewicz:
This fixes #501 by changing `Event.get_website_url()` and `Event.get_repository_url()` into properties `Event.website_url` and `Event.repository_url`.


workshops/models.py
```diff
@@ -481,7 +481,7 @@ class Event(models.Model):
     REPO_FORMAT = 'https://github.com/{name}/{repo}'
     WEBSITE_REGEX = re.compile(r'https?://(?P<name>[^.]+)\.github\.'
                                r'(io|com)/(?P<repo>[^/]+)/?')
-    WEBSITE_FORMAT = 'https://{name}.github.io/{repo}'
+    WEBSITE_FORMAT = 'https://{name}.github.io/{repo}/'
```
gvwilson: Provisionally, but once we move to your "info in meta headers" idea, we'll need to accept a wider range of URLs (?).

api/test/test_events.py
```diff
@@ -104,12 +104,12 @@ def setUp(self):
...
                 'humandate': 'Aug 28-29, 1993',
-                'latitude': 3,
-                'longitude': -2,
+                'latitude': 3.,
+                'longitude': -2.,
                 'venue': 'University',
                 'address': 'On the street',
                 'country': 'US',
-                'url': 'http://url1/',
+                'url': 'https://user.github.io/repository/',
```
gvwilson: Are there tests as well for invalid URLs being rejected?  Or converted?

workshops/templates/workshops/all_events.html
```diff
@@ -45,9 +45,9 @@
             </a>
           {% endif %}
         </td>
-        <td {% if not event.get_repository_url %}class="warning"{% endif %}>
-          {% if event.get_repository_url %}
-          <a href="{{ event.get_repository_url }}">{{ event.get_repository_url|cut:"https://github.com/" }}
+        <td {% if not event.repository_url %}class="warning"{% endif %}>
+          {% if event.repository_url %}
+          <a href="{{ event.repository_url }}">{{ event.repository_url|cut:"https://github.com/" }}
```
gvwilson: Urk - no.  If we're going to massage URLs for display, we should do that in the view class, not in the template.  (Have a helper function to add another element to the context which is the value displayed?)  Reasoning is 1. logic all in one place and 2. we may allow a wider range of URLs in the very near future.

workshops/templates/workshops/event.html
```diff
@@ -15,8 +15,8 @@
...
+  <tr><td>Website URL:</td><td colspan="2">{{ event.website_url|default:"—"|urlize }}</td></tr>
+  <tr><td>Repository URL:</td><td colspan="2">{{ event.repository_url|default:"—"|urlize }} {% if event.repository_url %}<a href="{% url 'validate_event' event.get_ident %}" class="btn btn-primary btn-xs pull-right" id="validate_event">validate event</a>{% else %}<a class="btn btn-danger btn-xs pull-right" id="error_event_url" href="#" data-toggle="popover" data-trigger="focus" title="Validation error" data-content="Cannot validate an event without URL pointing to the GitHub repository, e.g.: <code>https://github.com/swcarpentry/2015-05-24-training</code>">Error</a>{% endif %}</td></tr>
```
gvwilson: +1.

workshops/test/test_event.py
```diff
@@ -187,14 +187,14 @@ def test_website_url(self):
             'https://github.com/user-name/repo-name',
             'https://github.com/user-name/repo-name/',
         ]
-        WEBSITE = 'https://user-name.github.io/repo-name'
+        WEBSITE = 'https://user-name.github.io/repo-name/'
```
gvwilson: Hm - do we absolutely require the trailing / ?  Do we have tests for what happens when it's not present?  Does the web interface reject URLs lacking the trailing / ?

workshops/templates/workshops/all_events.html
pbanaszkiewicz: I'm against cutting the URL in this place anyway, but I believe it was historical code so I left it intact. Since you're against then I'll remove it.

workshops/models.py
pbanaszkiewicz: It's just a formatting. So it works like this: 
1. try to match known regex against some URL
2. Success? Format it with `WEBSITE_FORMAT` or `REPO_FORMAT` (depending on whether it's `Event.website_url` or `Event.repository_url`)
3. Failure? Just return the original URL

This way, since most of the URLs are from GitHub, we can match & transform most of them, but for the links we cannot match we still have the original URL.

api/test/test_events.py
pbanaszkiewicz: These are tests for API returning what we want. The tests for `website_url` and `repository_url` are in `workshops.test.test_event.TestEvent.test_website_url`, `workshops.test.test_event.TestEvent.test_repository_url` and `workshops.test.test_event.TestEvent.test_wrong_repository_website_urls`.

workshops/test/test_event.py
pbanaszkiewicz: 1. We have tests for lack of trailing slash (or its presence), see https://github.com/swcarpentry/amy/pull/510/files#r40770937
2. In this case it's only a representation that we store in the `WEBSITE` variable.
3. The web interface accepts URLs with both trailing slashes and without, but tries to represent them with the trailing slash present
4. The rationale for adding the trailing slash is that GitHub redirects from `https?://user.github.io/repo` to `https?://user.github.io/repo/`, but it does not do that with repositories in `https://github.com/user/repo` format.

pbanaszkiewicz:
@gvwilson Rebased + added some slight changes.

gvwilson:
:+1:

PR #511, opened 2015-10-01
Add back the EU country, since it's used in some hosts

pbanaszkiewicz:
This commits adds back the "fake" EU country. Apparently Europe's flag is already present in the django-countries, it's just the country that's not available.


gvwilson:
Is the flag icon already there?  Either way, :+1:

pbanaszkiewicz:
Yes, apparently the EU flag is included in django-countries.

gvwilson:
Ship it.

PR #528, opened 2015-10-09
Switch for new/closed requests

### LINKED ISSUE #523: View archived requests
We can now create a new event in AMY based on a workshop request.  Once we do that though, the workshop request is no longer visible.  If I missed entering in some detail in creating the event, the workshop request is no longer visible. 

I know it's in my email but it would be nice to see it right from AMY.

pbanaszkiewicz:
This fixes #523.

New "filter switch" was added on both event requests and profile update requests lists. They allow users to quickly switch between closed and open requests.

### DIFF
--- a/workshops/templates/workshops/all_eventrequests.html
+++ b/workshops/templates/workshops/all_eventrequests.html
@@ -1,6 +1,10 @@
 {% extends "workshops/_page.html" %}
 
 {% block content %}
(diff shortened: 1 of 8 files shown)


pbanaszkiewicz:
@gvwilson I kinda need this reviewed and merged :-) Thanks!

gvwilson:
:+1:

PR #527, opened 2015-10-08
Renaming problems report prior to adding another

gvwilson:
Changing the name 'problems' to 'workshops_with_problems' so that we can add an 'instructors_with_problems' report as well.

PR #530, opened 2015-10-10
Reporting instructors with problems

gvwilson:
Displaying instructors whose location we don't know. (Currently doesn't display anyone, which leads me to suspect that my query isn't actually pulling up all instructors...)


pbanaszkiewicz:
This PR and #527 contains the same commits. Let me make sure all of #527 is in here and if so I'll just close it.

pbanaszkiewicz:
Okay, one commit (c3697931539137905d1f160dc608ee23f60040d9) is missing. Let me try to fix that…

pbanaszkiewicz:
Okay, the change from c369793 was introduced in ad21e5622f68d2d5e8399b631d4a20532a1a9765.

Issue #518, opened 2015-10-05
Add created by & last modified by to event


maneesha:
Each  page (event, person, etc) should include 

Created by username on YYYY-MM-DD
Last modified by username on YYYY-MM-DD

gvwilson:
+1

pbanaszkiewicz:
We're kind of are able to track this data down with django-reversion.

PR #533, opened 2015-10-10
Making home page lay out more cleanly

gvwilson:


pbanaszkiewicz:
Is this what admins want?

gvwilson:
It's certainly what *I* want - even at high res, my screen can't show 
four columns without items wrapping.

pbanaszkiewicz:
@gvwilson I'm thinking about how to do this better, because with your changes the "recently changed" will become invisible. Maybe http://getbootstrap.com/components/#nav-tabs for Current/Uninvoiced/Unpublished would work?

gvwilson:
Sure - check with Maneesha?

maneesha:
I've only used this on one screen and have had no problems with the layout. I can try on some other monitors and see what this looks like.  

However, I've never used the rightmost column (Recently Changed), so I'd be OK removing that if the screen is too crowded.  Much more valuable would be to see "Created by user on date" and "Last updated by user on date" on each page (see #518)

gvwilson:
Thanks @maneesha - @pbanaszkiewicz, how about we accept this change 
(which moves the event log down on the home page) and then prioritize 
#518 for the next release?

pbanaszkiewicz:
@maneesha @gvwilson in this case I suggest moving the Log out of the dashboard to its own page. I'll submit a PR for that.

In the long run we may have customizable dashboards or dashboard layouts specific to each user group (superadmin, admin, steering committee).

I'll also start working on #518 soon.

pbanaszkiewicz:
Superseded by #535.

PR #542, opened 2015-10-14
Profile update request view: fix couple of issues

### LINKED ISSUE #538: Change Instructor profile request display to use 3 columns
"Has instructor badge" is getting overwritten from "yes" to "NA"

I haven't accepted any instructor profile updates yet because of this.

pbanaszkiewicz:
1. Move fields not being saved to DB to the bottom of the form
2. Separate these fields with long horizontal row saying
   "These aren't saved"
3. Change "N/A" to "<em>Not available</em>"
4. Add some '|default:"—"' to the fields that didn't have this filter

This fixes #538 but in a different way (see attachment).

screenshot from 2015-10-14 21 02 35

This PR will cause conflict once #528 is merged.

### DIFF
--- a/workshops/templates/workshops/profileupdaterequest.html
+++ b/workshops/templates/workshops/profileupdaterequest.html
@@ -38,16 +38,16 @@
   <tbody>
     <tr><th>Personal:</th>
         <td>{{ new.personal }}</td>
-        <td>{% if old %}{{ old.personal }}{% endif %}</td></tr>
+        <td>{% if old %}{{ old.personal|default:"—" }}{% endif %}</td></tr>
     <tr><th>Family:</th>
         <td>{{ new.family }}</td>
-        <td>{% if old %}{{ old.family }}{% endif %}</td></tr>
+        <td>{% if old %}{{ old.family|default:"—" }}{% endif %}</td></tr>
     <tr><th>Email:</th>
         <td>{{ new.email|urlize }}</td>
-        <td>{% if old %}{{ old.email|urlize }}{% endif %}</td></tr>
+        <td>{% if old %}{{ old.email|default:"—"|urlize }}{% endif %}</td></tr>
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
@maneesha @gvwilson I changed your concept of 3 columns a little bit. Since only 3 fields were not overlapping, I moved them to sort-of a new table down below. See the attachment above and tell me if you like this or not.

@gvwilson the "—" corresponds to empty value being stored, while "*Not available*" means "that model doesn't have this field so we can't compare". For example, with `Person` model we don't store ORCID ID.

gvwilson:
Will the order of field display be the same for all persons, all the time, with this change? Having fields move up or down the page based on availability of information will be really hard to track.

pbanaszkiewicz:
@gvwilson the order will stay the same on all profile update request details pages (ie. the ones with table from screen shot), unless someone changes that table in template.

pbanaszkiewicz:
@gvwilson rebased the branch.

pbanaszkiewicz:
@gvwilson could you review?

workshops/templates/workshops/profileupdaterequest.html
```diff
@@ -118,6 +109,16 @@
         {% endif %}
         </td>
     </tr>
+    <tr><th colspan="3" class="text-center">Additional information (not being saved to the database)</th></tr>
```
gvwilson: Why "not being saved"?

pbanaszkiewicz: Because neither ORCID ID or Occupation is being stored in Person record.

The "Has instructor badge" is rather a property that can only be calculated from Person record.

gvwilson: Gotcha.  We should store ORCID and occupation eventually.  Other than 
that, happy to merge this.

PR #543, opened 2015-10-14
Small typo in README

timtomch:
Corrected small typo.


pbanaszkiewicz:
@timtomch good catch! Thanks a lot for providing a fix.

PR #551, opened 2015-10-20
Reporting how often people have taught

gvwilson:
Report who has taught how often in downloadable CSV format.  Only remaining issue is the use of 'award.person.airport.country' in the template has to be wrapped with 'if' because 'award.person.airport' may not exist, so we sometimes get 'xxx-unset-variable-xxx' - haven't been able to find a filter that'll catch this and display '-' instead.


pbanaszkiewicz:
@gvwilson still WIP?

pbanaszkiewicz:
Closed via #552.

PR #555, opened 2015-10-21
Complete events

### LINKED ISSUE #471: Bool field for events=closed
Can we have a T/F field so that we can mark events as "closed" -- meaning that there's no more work to be done with them... surveys done, invoicing done, etc.

pbanaszkiewicz:
Fixes #471.

The implementation adds a new field, `Event.completed` — it will be easier to do filtering in all-events view than when using Tag to do this task.

For example, with current approach we can display all events, events complete, or events incomplete. With Tag-approach we could only filter events incomplete, or complete.

I decided to not change the dashboard at all. My thinking is admins will only mark events as completed once they're not ongoing, nor upcoming, nor uninvoiced, nor unpublished.

If you would rather have `Event.complete` than `Event.completed`, just let me know.

### DIFF
--- a/workshops/filters.py
+++ b/workshops/filters.py
@@ -68,6 +68,10 @@ class EventFilter(django_filters.FilterSet):
     ]
     status = EventStateFilter(choices=STATUS_CHOICES)
 
+    invoice_status = django_filters.ChoiceFilter(
+        choices=((None, '---------'), ) + Event.INVOICED_CHOICES,
+    )
+
     class Meta:
         model = Event
         fields = [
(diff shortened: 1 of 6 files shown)


workshops/filters.py
```diff
@@ -68,6 +68,10 @@ class EventFilter(django_filters.FilterSet):
     ]
     status = EventStateFilter(choices=STATUS_CHOICES)
 
+    invoice_status = django_filters.ChoiceFilter(
+        choices=((None, '---------'), ) + Event.INVOICED_CHOICES,
```
gvwilson: The long dash now appears 3 times in filters.py - time to stick it in a constant at the top of the file?

gvwilson:
One minor comment, then :+1+

PR #564, opened 2015-11-05
Hotfix: wrong date formatting on event's details page

pbanaszkiewicz:
Events without start date caused `"".format()` to error out because it received `None` object instead of `datetime`.

The fix was merged to the master and deployed so that admins can continue to work.

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -564,12 +564,14 @@ class Event(models.Model):
         return self.invoice_status == 'not-invoiced'
 
     def get_invoice_form_url(self):
+        from .util import universal_date_format
+
         query = {
             'entry.823772951': self.venue,  # Organization to invoice
             'entry.351294200': 'Workshop administrative fee',  # Reason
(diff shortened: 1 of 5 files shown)


workshops/models.py
```diff
@@ -569,7 +569,8 @@ def get_invoice_form_url(self):
             'entry.351294200': 'Workshop administrative fee',  # Reason
 
             # Date of event
-            'entry.1749215879': '{:%Y-%m-%d}'.format(self.start),
+            'entry.1749215879': ('{:%Y-%m-%d}'.format(self.start)
+                                 if self.start else ''),
```
gvwilson: I'd create a helper function format_date in util.py and call it with `format_date(self.start)`, then look for other places where dates are formatted.

pbanaszkiewicz:
@gvwilson added that helper function you asked for.

gvwilson:
:+1:

PR #566, opened 2015-11-05
Reset password form

### LINKED ISSUE #560: Add password reset form
More and more people are forgetting their passwords. We need a way to reset them.

pbanaszkiewicz:
This fixes #560.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -346,9 +346,9 @@ class PersonForm(forms.ModelForm):
         # don't display the 'password', 'user_permissions',
         # 'groups' or 'is_superuser' fields
         # + reorder fields
-        fields = ['username', 'personal', 'middle', 'family', 'may_contact',
-                  'email', 'gender', 'airport', 'affiliation', 'github',
-                  'twitter', 'url', 'notes', 'lessons', 'domains']
+        fields = ['username', 'personal', 'middle', 'family', 'is_active',
+                  'may_contact', 'email', 'gender', 'airport', 'affiliation',
+                  'github', 'twitter', 'url', 'notes', 'lessons', 'domains']
 
 
 class PersonPermissionsForm(forms.ModelForm):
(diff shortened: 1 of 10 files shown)


workshops/forms.py
```diff
@@ -346,9 +346,9 @@ class Meta:
         # don't display the 'password', 'user_permissions',
         # 'groups' or 'is_superuser' fields
         # + reorder fields
-        fields = ['username', 'personal', 'middle', 'family', 'may_contact',
-                  'email', 'gender', 'airport', 'affiliation', 'github',
-                  'twitter', 'url', 'notes', 'lessons', 'domains']
+        fields = ['username', 'personal', 'middle', 'family', 'is_active',
+                  'may_contact', 'email', 'gender', 'airport', 'affiliation',
+                  'github', 'twitter', 'url', 'notes', 'lessons', 'domains']
```
gvwilson: Why not add `is_active` to the end of the list to make a smaller diff?

gvwilson:
Welcome back @wking.

workshops/forms.py
wking: On Sun, Nov 08, 2015 at 09:04:54AM -0800, Greg Wilson wrote:
> Why not add `is_active` to the end of the list to make a smaller
> diff?

This list also sets the field order in forms, but I don't see an
obvious location for `is_active` in the person form.  One the one
hand, it's fairly important (can the person log in or reset their
password?), which argues for an earlier listing.  On the other hand,
we're unlikely to edit it manually, which argues for a later listing.

pbanaszkiewicz: Alternatively we could add `is_active` to the permissions form, since it's used mostly by admins.

pbanaszkiewicz:
@wking updated.

wking:
On Mon, Nov 09, 2015 at 07:45:43AM -0800, Piotr Banaszkiewicz wrote:
> @wking updated.

3e4030b looks good to me.  I'll leave it to someone with deploy access
to merge it when they're ready to push it to production.

pbanaszkiewicz:
@wking merging to `develop` now; when v1.2 is done, `develop` will be merged into `master` and deployed.

PR #567, opened 2015-11-07
Rework logic in EventQuerySet's unpublished and published filters

### LINKED ISSUE #547: Empty workshop titles showing up
At 09:28 UK time on 2015-10-16, two workshops were showing up blank in the "upcoming" list on the web site:
screen shot 2015-10-16 at 9 28 18 am
I *think* this was because "Venue" was blank for 2015-11-11-northwestern and 2015-11-11-nyu-med; I've added venue names for both, but:

1. AMY should complain about missing venue names (or we should use something else as the title of the event when displaying on the website).

2. Our build script should check for empty venue names and... skip the workshop?  Report an error?  Something.

pbanaszkiewicz:
Fix #547.

Now unpublished and published are in opposition to each other.

Previously unpublished showed only events in future or without starting
date, now:
* unpublished shows anything WITHOUT country OR venue OR starting date
  OR URL,
* published shows everything WITH country AND venue AND starting date
  AND URL.

Maybe in future we'll get a feature to "inverse" a QuerySet - then the
EventManager.published_events() will be implemented based on
unpublished_events().

This commit also adds tests for EventQuerySet.(un)published_events
methods.

EventQuerySet.published_events gets tested by API tests since they use
this method.

EventQuerySet.unpublished_events required a new test, so I placed it
near tests for upcoming/past events.


pbanaszkiewicz:
The downside of reworking `EventQuerySet.unpublished_events` is that now in the "Unpublished" on the dashboard appears all past events that don't have an URL (about 20-30), including TTT-online events.

I think we should manually add some URLs to them (because I think the new logic for unpublished_events is better than before). @maneesha @gvwilson disagree with me?

gvwilson:
If we can find the URLs, we should add them, but some events might not 
have URLs.

pbanaszkiewicz:
@gvwilson I think this task should be done by admins / event organizers. If no URL is available, I'd put at least a space in that field.

wking:
On Sat, Nov 07, 2015 at 09:10:55AM -0800, Greg Wilson wrote:
> If we can find the URLs, we should add them, but some events might
> not have URLs.

In #547 there's talk of requiring URLs before events are publishable
[1], and we seem to have required URLs since 08e022c4 (Removing the
'published' flag from events, 2015-06-13, #363).  But if we want to
support published events without URLs, I'd rather update the “is this
event published” logic to just be events with both a start date and
sufficient venue/address information.  But do online TTT events have a
venue/address?  When are they publishable?  Maybe it's too complicated
for a universal rule, and we need to roll back #363?

[1]: https://github.com/swcarpentry/amy/issues/547#issuecomment-149252624

wking:
On Sat, Nov 07, 2015 at 08:32:36AM -0800, Piotr Banaszkiewicz wrote:
> Maybe in future we'll get a feature to "inverse" a QuerySet…

Like this [1]?

> Also, `Q` objects can be negated using the `~` operator, allowing
> for combined lookups that combine both a normal query and a negated
> (`NOT`) query:

[1]: https://docs.djangoproject.com/en/1.8/topics/db/queries/#complex-lookups-with-q-objects

pbanaszkiewicz:
@wking:

1. Essentially I'd like to negate `.filter(Q()…)` to have `.exclude(Q()…)`. Or if you imagine results from a query as a subset Q (Q ⊂ A, where A is a set of all records), then I want to have Q' (Q ∪ Q' = A).

2. TTT events are special. I'm not sure if Greg wants to have them published and visible amongst "normal" workshops.

@gvwilson the new logic shows public events if they have URL, starting date, country, and venue, so some events (especially those without location info, like TTT) won't show up. Is this what you want?

wking:
On Wed, Nov 11, 2015 at 08:38:57AM -0800, Piotr Banaszkiewicz wrote:
> 1. Essentially I'd like to negate `.filter(Q()…)` to have
>    `.exclude(Q()…)`. Or if you imagine results from a query as a
>    subset Q (Q ⊂ A, where A is a set of all records), then I want to
>    have Q' (Q ∪ Q' = A).

You should be able to achieve that with:

  @staticmethod
  def _unpublished_filter():
      …
      return Q(start__isnull=True) | Q(url__isnull=True) | …

  def unpublished_events(self):
      return self.filter(self._unpublished_filter)…

And either:

  def published_events(self):
      return self.exclude(self._unpublished_filter)…

Or:

  def published_events(self):
      return self.filter(~self._unpublished_filter)…

> 2. TTT events are special. I'm not sure if Greg wants to have them
>    published and visible amongst "normal" workshops.

TTT are more likely to be locationless (because they happen online)
and URL-less (for reasons I'm not clear on), but those are both things
that *could* happen for other events.  We could have TTT-specific
logic on top of (or as an alternative to) location/URL/…-specific
logic, but I'm not clear enough on the intended business logic to say.

gvwilson:
I'll defer to @wking on this one... Reassign?

pbanaszkiewicz:
@gvwilson please make a decision: do you want TTT events in the list of workshops on the main site or not?

pbanaszkiewicz:
WIP: @maneesha wants to add lat/long to the "unpublished" conditions.

gvwilson:
Yes please - TTT should be part of the regular workshops list.

pbanaszkiewicz:
@gvwilson then either should TTT be treated separately (since they often lack URL, or venue, or address, or lat/long) or will someone put this data into AMY?

gvwilson:
I'll put data into AMY - most will be online events, but we're now able 
to handle that.  (Are we able to handle that?  If we require lat/long 
for an event to be published, and online events don't have them...)

wking:
On Fri, Nov 13, 2015 at 06:51:03AM -0800, Piotr Banaszkiewicz wrote:
> WIP: @maneesha wants to add lat/long to the "unpublished" conditions.

This will mean “we never publish online-only events”, which is
probably stricter than we want to be.  If we do want to have physical
events that don't get published until we have a lat/lon, we will
probably have to revert #363.

gvwilson:
We definitely need to publish online-only events (to include instructor 
training).

maneesha:
Didn't we decide to put online events somewhere in the middle of the ocean?
I have to dig to see if I can find that email/github conversation.

maneesha:
Here it is - the Pole of Inaccessibility!

https://github.com/swcarpentry/amy/blob/b155f90b263b1fe553382a0d1264b8c0f5a7b65f/workshops/migrations/0040_add_country_to_online_events.py

wking:
On Fri, Nov 13, 2015 at 11:53:02AM -0800, maneesha sane wrote:
> Here it is - the Pole of Inaccessibility!

That probably needs a mention in the location help_text for
Event.country and EventRequest.country.  And a rule to set lat/lon on
Event.save() if that is the assigned country (the migration isn't
updating *new* Events assigned to that country).

pbanaszkiewicz:
@wking reworked some bits a little.

The solution for online events I took is: suggest admins to put Pole of Inaccessibility coordinates in lat/long (via help_text).

pbanaszkiewicz:
@wking take a look now. I changed what you suggested + added JS for selecting online country.

pbanaszkiewicz:
@wking can you please review this one? #580 is depending on it (JS changes).

workshops/migrations/0060_published_help_text.py
```diff
@@ -0,0 +1,64 @@
...
+
+def add_address_to_online_events(apps, schema_editor):
+    """Set Event.address if empty for online events."""
+    Event = apps.get_model('workshops', 'Event')
+    Tag = apps.get_model('workshops', 'Tag')
+    online = Tag.objects.get(name='online')  # should be created via 0040_*
+
+    Event.objects.filter(tags__in=[online]) \
+                 .filter(models.Q(address=None) | models.Q(address='')) \
+                 .update(address='Internet')
```
wking: Are we still keeping an online tag in parallel with the online country / address / lat / lon?  That doesn't seem very DRY :p.  If we need the online country / address / lat / long for the (un)published logic, maybe we can drop the online tag?  In any case, this migration should also be setting the country, lat, and lon.  And you'll want another update call that gets everything with the W3 country and forces the address, lat, and lon to the values from 0040.  I wouldn't bother about the `Q` filters in either update query, because a bit of extra clobbering isn't a big deal for a one-off migration, and it ensures we don't miss some logic and leave an online event with the wrong values.

wking: Also, 0040 was setting `Event.venue` to `Internet` but  this one is setting `Event.address` to `Internet`.  Since we need both to publish, we probably want an additional data migration that sets `venue` to `Internet` for all existing online events.

workshops/static/online_country.js
```diff
@@ -0,0 +1,42 @@
...
+
+  if (new_country == 'W3') {
+    prev_venue = $(id_venue).val();
+    prev_address = $(id_address).val();
+    prev_latitude = $(id_latitude).val();
+    prev_longitude = $(id_longitude).val();
+    $(id_venue).val('Internet');
+    $(id_address).val('Internet');
+    $(id_latitude).val('-48.876667');
+    $(id_longitude).val('-123.393333');
```
wking: This approach assumes all oneline-event updates will go through an AMY UI using this JavaScript and that admins won't tweak the magic values after the JavaScript here populates the field values.  I'd rather make `Event.country` the canonical source of online-ness, and have `Event.save()` *force* all the associated fields to their magic online values, because I have more faith in `Event.save()` being in the update path ;).

workshops/static/online_country.js
```diff
@@ -0,0 +1,42 @@
...
+    $(id_venue).prop('disabled', true);
+    $(id_address).prop('disabled', true);
+    $(id_latitude).prop('disabled', true);
+    $(id_longitude).prop('disabled', true);
```
wking: Ah, this keeps admins from tweaking after the JavaScript fills in the values, so I think keep the JavaScript here, but also add the `Event.save()` logic (e.g. folks could use cURL to POST, or be saving events from a Django shell, or …).

workshops/templates/workshops/workshop_issues.html
```diff
@@ -36,7 +36,7 @@
         {% if event.mailto_ %}
           <a href="mailto:{{event.mailto_}}?subject={% filter urlencode %}Missing location for workshop {{event.get_ident}}{% endfilter %}&body={% filter urlencode %}Hi,
 
-Can you please send us the location where the {{ event.get_ident }} workshop took place?
+Can you please send us the location where the {{ event.get_ident }} workshop took place? We need a country, address, venue, and latitude/longitude.
```
wking: This probably needs a caveat like “(unless your event happened only online, in which case, that's all you need to tell us)”.  @maneesha, do you have an opinion on the wording?

workshops/views.py
```diff
@@ -1376,18 +1376,23 @@ def workshop_issues(request):
...
-               Q(venue=None) | Q(venue__exact='') |
-               Q(address=None) | Q(address__exact='') |
-               Q(start__gt=F('end')))
+    events = Event.objects.past_events().filter(
+        Q(attendance=None) | Q(attendance=0) |
+        Q(country=None) |
+        Q(venue=None) | Q(venue__exact='') |
+        Q(address=None) | Q(address__exact='') |
+        Q(latitude=None) | Q(longitude=None) |
+        Q(start__gt=F('end'))
```
wking: You can replace a number of these by using `Event.objects.past_events().unpublished_events().filter(…)` (the `venue=None` case and some others should be illegal because `Event.venue` is not nullable).  I was initially concerned that the overlap between `unpublished_events` and this filter set might be a coincidence, but the only reason we'd care about those fields for past events is probably so we can publish them (“hey, look at all the events we've held!”), so I think the overlap is meaningful and the `unpublished_events` replacement is fair.

wking:
On Thu, Nov 19, 2015 at 01:13:53PM -0800, Piotr Banaszkiewicz wrote:
> @wking can you please review this one?

Sorry, thanks for the bump.  I left a few more comments inline, but
4b5dec4 looks pretty close to me.

workshops/migrations/0060_published_help_text.py
pbanaszkiewicz: The intent of this migration was to fill up `venue` for all online events, since the `0040` migration didn't set it.

I disagree about the DRY – we keep tags for to easily show the type of the event (DC / SWC / LC / online / TTT / stalled [coming soon]).

I don't want to put another constraint on the admins (event has 'online' tag? Therefore it's location must be in 'Internet'!) because there can be instances of events that may have a different location (like streamed events?).

The `0040` migration did set `venue` because I couldn't think of a good value for that field (I still can't - I reuse 'Internet' for `address` here).

I agree that this migration should also set the `address` for non-empty addresses for events with W3 country. I'll update shortly.

workshops/static/online_country.js
pbanaszkiewicz: Sure!

workshops/templates/workshops/workshop_issues.html
pbanaszkiewicz: I don't think we want to add conditionals to the email. Now that I think about it, we might not want to change the wording at all - more text == less reading.

@maneesha can you please review this change?

pbanaszkiewicz:
@wking thanks a lot. I added the changes + removed something I just noticed (567f1d7 - I didn't want to make a separate PR for that).

So I think we're waiting for Maneesha's green light.

workshops/templates/workshops/workshop_issues.html
maneesha: I'd want to understand why we are collecting this for past workshops.  Longitudinally I think we want to track workshops geographically i.e. how many in Florida? how many in south east US? etc.  But the question is how specific do we need to get? @jduckles ?
 
Obviously for upcoming workshops we want to know the exact location down to the street address so people who want to attend can find it.

For past workshops, can we just have the host - i.e., "University of Miami" - or does it matter in the long term exactly what building on Miami's campus it took place?  

We have no workshops currently missing location so I don't have a real test case (https://amy.software-carpentry.org/workshops/reports/workshop_issues).

pbanaszkiewicz: @maneesha after introduction of this PR, the list of workshops without location info is long (~40?). It's because we changed some rules:

* start date, URL, country, venue, address, latitude and longitude must be in place for the workshop to be considered published
* we're not looking for upcoming workshops only, but for all we ever had (this mainly introduces the big number of workshops with missing all location data)

Why all these changes? Because we didn't have clear rules in all places (see the list below) + because of https://github.com/swcarpentry/amy/issues/547 + because of related https://github.com/swcarpentry/site/issues/1130:

* API endpoint published events
* workshop issues page
* dashboard

So summing up: with this PR we're cleaning up some a small mess that was covering loops in our data and logic.

pbanaszkiewicz:
@wking Travis is failing because it's doing a merge. I need to rebase this PR.

wking:
On Fri, Nov 20, 2015 at 09:11:57AM -0800, Piotr Banaszkiewicz wrote:
> I need to rebase this PR.

Rebase away :).

workshops/migrations/0060_published_help_text.py
wking: On Fri, Nov 20, 2015 at 08:30:09AM -0800, Piotr Banaszkiewicz wrote:
> … there can be instances of events that may have a different
> location (like streamed events?).

Ah, this is a good reason to keep the ‘online’ tag (so we have a way
to mark streamed events).  However, events in the W3 country will
*always* be online, so I'd make sure we set that tag for such events
in Event.save() (you've set the rest of those fields in 90198e8, and
I'd suggest you want similar updates in this migration for all
existing W3 events).

workshops/templates/workshops/workshop_issues.html
wking: On Fri, Nov 20, 2015 at 08:56:17AM -0800, maneesha sane wrote:
> I'd want to understand why we are collecting this for past
> workshops.  Longitudinally I think we want to track workshops
> geographically i.e. how many in Florida? how many in south east US?
> etc.  But the question is how specific do we need to get? @jduckles
> ?

I think the motivation is:

1. Publishing more past workshops makes it easier to show folks how
   busy we are, or where we have been active in the past.
2. This PR requires the additional fields before an event is published
   [1,2].
3. We have to collect the newly-required information to avoid
   unpublishing old events that we used to publish.

If that reasoning makes sense, that means we need to be specific
enough to pass the new, stricter 'published' filter.

[1]: https://github.com/swcarpentry/amy/pull/567#issuecomment-156452337
[2]: https://github.com/swcarpentry/amy/pull/567/files#r45492564

pbanaszkiewicz:
@wking rebased

PR #575, opened 2015-11-16
Making titles on instructor issue page more descriptive

gvwilson:
Having been away from this for a few weeks, I didn't know what the first table was showing me, so I've titled both.

### DIFF
--- a/workshops/templates/workshops/instructor_issues.html
+++ b/workshops/templates/workshops/instructor_issues.html
@@ -1,7 +1,7 @@
 {% extends "workshops/_page.html" %}
 
 {% block content %}
-<h2>Instructors</h2>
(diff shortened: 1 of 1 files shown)


maneesha:
Looks good to me.
Can we automate sending emails to everyone on these lists?
Also can we just dismiss everyone who was in instructor training 2014 or earlier?  If they haven't been in touch in over a year then I think we can safely assume they're not interested.

gvwilson:
The problem is that some people will have incomplete instructor training 
from (for example) 2014, but then also be in an event in 2015 (taking a 
second run at it).  I need a way to mark the earlier events as "closed", 
and then we can subtract them from this problem report.

Issue #539, opened 2015-10-13
Generate a list of SCF members


gvwilson:
According to our rules, everyone who has taught at least once in the last two years is a member, as is anyone who has been specifically appointed a member for other services. We need a report page in AMY that generates this list. Any chance of getting it sooner rather than later?

maneesha:
Can we add a "member" field (bool) to the db?
Finding out who's taught at least 1x in the past two years is a simple SQL query & this can automatically be set to True.
We can then also manually check off anyone who's been specially appointed a member.

gvwilson:
I don't think we need a field: we can do the query and merge the result 
with people who have a 'member' badge for special work to regenerate the 
members list on the fly.

maneesha:
Of course... I forgot about the badges.

maneesha:
In that case... we can *view* everyone with a member badge if you go to "Badges" under "More" but we should run an update query on that for everyone who's taught in the past 2 yrs.... right now we only have 29 members.

gvwilson:
Yup, I'm going to construct a page for this - should have it in the next 
release.

pbanaszkiewicz:
@gvwilson @maneesha what's current policy on members? If someone taught 2 years ago, will he be still considered a member in 6 months?

gvwilson:
From http://software-carpentry.org/scf/governance.html :

~~~
2.  The Membership is made up of:

     a.  Every qualified instructor who has taught at least twice in the
         past two calendar years.

     b.  Anyone who has done 30 days or more work for the SCF in the past
         calendar year.

     c.  Anyone who has, in the opinion of the Steering Committee, made
         a significant contribution in the past year.
~~~

pbanaszkiewicz:
@gvwilson but time changes, so if someone taught exactly 2 years ago, tomorrow they won't be a member?

gvwilson:
In 2015, "previous two calendar years" means any of 2013 or 2014. If you 
last taught on Jan 1, 2013, you stop being a member at midnight on Dec 
31, 2015.

maneesha:
So if someone last taught Dec 31, 2013, they'll also stop being a member on Dec 31, 2015?

gvwilson:
Yup.  Anything less than a three year gap, they're still officially a 
member.  But they have to have taught twice in that interval.

pbanaszkiewicz:
@gvwilson what's the status on this issue?

pbanaszkiewicz:
[...]

PR #577, opened 2015-11-16
Starting to work on a report of SCF members

gvwilson:
Work in progress: will address #539.

Lots of problems with this approach right now - inefficient query, `queryset` is being evaluated for every request, etc. etc.


api/views.py
```diff
@@ -45,6 +47,15 @@ class ExportInstructorLocationsView(ListAPIView):
     serializer_class = ExportInstructorLocationsSerializer
 
 
+class ExportMembersView(ListAPIView):
+    """Show everyone who qualifies as an SCF member."""
+    permission_classes = (IsAuthenticatedOrReadOnly, )
+    paginator = None  # disable pagination
+
+    queryset = get_members()
```
pbanaszkiewicz: This line gets evaluated on every request because of this line: https://github.com/swcarpentry/amy/pull/577/files#diff-ef5bc9252746919e203b86c32d3d437eR14

pbanaszkiewicz: I think that:

* it can be fixed by making `get_queryset(self)` method that return `get_members()`
* fixing it will fix all or almost all of the failing tests.

Makefile
```diff
@@ -102,6 +102,10 @@ published :
 serve : bower_components workshops/git_version.py
 	${MANAGE} runserver
 
+## serve_now    : run a server now
```
pbanaszkiewicz: Why not use `make serve`?

workshops/util.py
```diff
@@ -471,3 +472,31 @@ def update_event_attendance_from_tasks(event):
...
+    instructor_role = Role.objects.get(name='instructor')
+    implicit_members = []
+    for i in instructors:
+        if i.id not in member_ids:
+            tasks = i.task_set.filter(role=instructor_role)
+            for t in tasks:
+                if t.event.start and (t.event.start >= cutoff):
+                    implicit_members.append(i)
+                    member_ids.add(i.id)
+                    break
```
pbanaszkiewicz: Okay, here are Django queries I came up with:

```python
def get_members():
    '''Get everyone who is a member of the Software Carpentry Foundation.'''

    today = datetime.date.today()
    cutoff = today - 2 * datetime.timedelta(days=365)

    member_badge = Badge.objects.get(name='member')
    instructor_badge = Badge.objects.get(name='instructor')
    instructor_role = Role.objects.get(name='instructor')

    # Everyone who is an explicit member.
    qs1 = Person.objects.filter(badges__in=[member_badge]).distinct()

    # Everyone who qualifies by having taught recently.
    qs2 = Person.objects.filter(task__role=instructor_role, task__event__start__gte=cutoff, badges__in=[instructor_badge]).distinct()

    queryset = qs1 | qs2
    return queryset
```

Now can someone validate that this yields correct results?

Makefile
gvwilson: Because I was doing development offline, and (ironically) you can only 
run a local server if you're online, because it tries to do a bower update.

workshops/util.py
gvwilson: I'll write some tests to check this - thanks.

api/views.py
gvwilson: Does the framework know to call `get_queryset`, or will we have to do 
that explicitly somewhere? And is it a regular method or a class method?

gvwilson: Will that line trigger evaluation of `get_queryset` once we implement 
it?  And are the querysets for other things like `ExportBadgesView` also 
being triggered on every HTTP request?

pbanaszkiewicz: `get_queryset` should be used if you want to provide, for example, different querysets based on logged in user (http://www.django-rest-framework.org/api-guide/generic-views/#attributes). It's just a different way of showing to the framework which queryset should be used.

If we end up using my proposition (https://github.com/swcarpentry/amy/pull/577/files#r44923073) then we don't need to change `ExportMembersView` at all.

QuerySets are lazy objects and if we switch to my proposition then nothing will get evaluated unless the `ExportMembersView` is visited.

gvwilson:
Ready for another look from @pbanaszkiewicz

pbanaszkiewicz:
I'll continue @gvwilson's work in another PR.

PR #580, opened 2015-11-17
Change event tag finding, parsing and validating

### LINKED ISSUE #465: Normalize workshop URLs

### LINKED ISSUE #571: Bugs with "Import/Update from URL"
1. "Import from URL" has wrong `country` field ID in JavaScript
2. Something's wrong with for example this event: https://amy.software-carpentry.org/workshops/event/2015-02-09-ucdavis (it uses remote that redirects to somewhere else, plus when we use the remote after redirection we get some errors). The officially saved URL is https://datacarpentry.github.io/2015-02-09-ucdavis/
3. Fix spaces around URLs

pbanaszkiewicz:
1. No more @swcarpentry/workshop-template/bin/check.py dependency.
2. Less restrictive tests (in some cases), for example:
  * no testing if list of instructors (or helpers) is comma separated
  * no testing of contact email format.
3. Finding / normalizing (parsing) / validating functions don't use
   network stack (python-requests), but rather work on provided data.
4. No more need for URL normalization so it was removed.
5. Some code for URL normalization is still intact in
   `Event.repository_url`, `Event.website_url` (it will probably end up
   being removed).
6. This PR will be accompanied with related PR to
   @swcarpentry/workshop-template that will bring `<meta>` tags to the
   workshop's websites.

This fixes #465.
This fixes #571.


pbanaszkiewicz:
Corresponding workshop-template PR: https://github.com/swcarpentry/workshop-template/pull/267

workshops/test/test_util.py
```diff
@@ -376,195 +374,184 @@ def test_attendance_increases(self):
...
+            'latitude': 36.998977,
+            'longitude': -109.045173,
+            'reg_key': 10000000,
+            'instructors': ['Hermione Granger', 'Ron Weasley'],
+            'helpers': ['Peter Parker', 'Tony Stark', 'Natasha Romanova'],
+            'contact': 'hermione@granger.co.uk, rweasley@ministry.gov',
+        }
+        self.assertEqual(expected, parse_tags_from_event_website(tags))
+
+    def test_parsing_tricky_tags(self):
```
wking: Probably better to have separate tests for each tricky case, and then have more descriptive names for each case.  If we have a single tricky-tags test we only get a single tricky case for each tag.

workshops/test/test_util.py
```diff
@@ -376,195 +374,184 @@ def test_attendance_increases(self):
...
-
-        self.assertEqual(rv['country'], 'US')
+        errors = validate_tags_from_event_website(tags)
+        assert len(errors) == 7
+        assert all([error.startswith('Invalid value') for error in errors])
+
+    def test_validating_missing_tags(self):
+        tags = {}
+        errors = validate_tags_from_event_website(tags)
+        assert all([error.startswith('Missing') for error in errors])
```
wking: You probably also want to assert that you have at least one error.

workshops/test/test_util.py
```diff
@@ -376,195 +374,184 @@ def test_attendance_increases(self):
...
+            'instructor': 'FIXME',
+            'helper': 'FIXME',
+            'contact': 'FIXME',
+        }
+        errors = validate_tags_from_event_website(tags)
+        assert all([
+            error.startswith('Placeholder value "FIXME"')
+            or error.startswith('Invalid value')
+            for error in errors
+        ])
```
wking: This also needs a check to make sure there is at least one error.

wking:
I haven't followed this all the way down to see what happens when you
post a two-character country that doesn't exist in the database (for
example), but I think this is the direction we should be heading in.
I left a few comments about tests, but I'm happy with this landing
whenever you like.

I'm not sure how the transition will work though.  Maybe you need to
keep both sets of code until swcarpentry/workshop-template#267 lands
and gains sufficient traction?  It should be easy to detect its
presence (“is there a slug meta?”).  For future-proofing, you might
want a workshop-template-version meta, so we can gracefully handle
future migrations to the template's metadata.

pbanaszkiewicz:
Hello @wking,

> I haven't followed this all the way down to see what happens when you post a two-character country that doesn't exist in the database (for example), but I think this is the direction we should be heading in.

I didn't do any content-interpretation validation, so for example:

* `latitude == 200` is invalid, but will pass the validation
* `country == "XY"` is invalid, but will pass the validation
* `startdate=="2015-05-05"` and `enddate="2014-05-05"` are invalid, but will pass the validation

> I left a few comments about tests, but I'm happy with this landing whenever you like. 

You were right, I separated tests and then added some more cases.

> I'm not sure how the transition will work though. Maybe you need to keep both sets of code until swcarpentry/workshop-template#267 lands and gains sufficient traction?

I didn't care about transition, but then I started thinking… The finding-parsing feature will be used by admins when they do "Import from URL" or "Update from URL". Especially the latter is going to be used against workshops with old `_layout`. I think I do need to keep the old code somewhere… I'll mark this WIP for now, until I figure something out.

wking:
On Thu, Nov 19, 2015 at 12:20:25AM -0800, Piotr Banaszkiewicz wrote:
> I didn't do any content-interpretation validation, so for example:
> 
> * `latitude == 200` is invalid, but will pass the validation
> * `country == "XY"` is invalid, but will pass the validation
> * `startdate=="2015-05-05"` and `enddate="2014-05-05"` are invalid,
>    but will pass the validation

Those should all be raising errors somewhere before being written to
the database (in the country case, that might get all the way to the
database before raising a “foreign key does not exist” error of some
sort).  I'm fine defering the check to the database (although I don't
know if there's SQL to limit the range of a REAL type for lat/lon),
but we should make sure somewhere in the stack is checking for valid
values, and that invalid values will generate understandable error
messages.  If we have to add our own validation checks {Model}.save()
is probably a reasonable place, raising ValueErrors with
understandable messages.

> > I'm not sure how the transition will work though. Maybe you need
> > to keep both sets of code until swcarpentry/workshop-template#267
> > lands and gains sufficient traction?
> 
> I didn't care about transition, but then I started thinking… The
> finding-parsing feature will be used by admins when they do "Import
> from URL" or "Update from URL". Especially the latter is going to be
> used against workshops with old `_layout`. I think I do need to keep
> the old code somewhere… I'll mark this WIP for now, until I figure
> something out.

I'd recommend shifting the parsing into a sub-package
(workshops.event.import.v1, workshops.event.import.v2, …?), with
version-agnostic code in workshops/event/import/__init__.py.  Each
version module would define:

* get_url(event), to get the URL that will return the tag-containing
  file.
* check_version(content), to take the content of the get_url URL and
  see if it supports the implemented version
  (e.g. workshops.event.import.v1.check_version(content) will return
  True iﬀ the content looks like the v1 content).
* parse_tags(content), will return the tags parsed from the content,
  but not perform any validation.

Then workshops/event/import/__init__.py would define:

* get_tags(event), which would do something like:

  for version_module in [v2, v1]:
      get_url = getattr(version_module, 'get_url')
      url = get_url(event=event)
      content = fetch_and_decode(url=url)
      if check_version(content=content):
          parse_tags = getattr(version_module, 'parse_tags')
          return parse_tags(content=content)
  raise ValueError('event {} does not match a known version'.format(event))

and the import button would feed the standardized get_tags output into
its updated Event (or JSON response, or wherever the data is being
consumed).

pbanaszkiewicz:
@wking:

> Those should all be raising errors somewhere before being written to
the database (in the country case, that might get all the way to the
database before raising a “foreign key does not exist” error of some
sort).  I'm fine defering the check to the database (although I don't
know if there's SQL to limit the range of a REAL type for lat/lon),
but we should make sure somewhere in the stack is checking for valid
values, and that invalid values will generate understandable error
messages.  If we have to add our own validation checks {Model}.save()
is probably a reasonable place, raising ValueErrors with
understandable messages.

I agree, but let me point out something:

* validation via `validate_tags_from_event_website` is meant only for display purposes (ie. "just show me if there's something wrong with the event")
* start > end is validated via form's `clean_end`

pbanaszkiewicz:
@wking:

> Those should all be raising errors somewhere before being written to
the database (in the country case, that might get all the way to the
database before raising a “foreign key does not exist” error of some
sort).  I'm fine defering the check to the database (although I don't
know if there's SQL to limit the range of a REAL type for lat/lon),
but we should make sure somewhere in the stack is checking for valid
values, and that invalid values will generate understandable error
messages.  If we have to add our own validation checks {Model}.save()
is probably a reasonable place, raising ValueErrors with
understandable messages.

I agree, but let me point out something:

* validation is meant only for display purposes ("just show me if something is wrong with the event") – I wonder if that feature is used at all,
* the actual validation happens with form submission, for example `start < end` is checked in form's `clean_end` method, and country is validated through by enforcing a choice from a list.
* I admit there's no check for lat/long…

> I'd recommend shifting the parsing into a sub-package (workshops.event.import.v1, workshops.event.import.v2, …?), with version-agnostic code in workshops/event/import/__init__.py. Each version module would define:

Huh, it does sound good, but I don't plan on changing workshop-template this often.

wking:
On Thu, Nov 19, 2015 at 12:39:19PM -0800, Piotr Banaszkiewicz wrote:
> > Those should all be raising errors somewhere…
> 
> I agree, but let me point out something:
>
> * validation is meant only for display purposes ("just show me if
>   something is wrong with the event") – I wonder if that feature is
>   used at all,
> * the actual validation happens with form submission, for example
>   `start < end` is checked in form's `clean_end` method, and country
>   is validated through by enforcing a choice from a list.

We *can* punt on validation until the admin copies things over and
submits the new/updated event.  But we have all the data there after
fetching, so validating and flagging errors in the display we show the
admin adding/editing an event will make their lives a bit easier.
Since it's nice-but-not-critical, I'm fine punting on it for this PR
(there's enough else up in the air ;).

> > I'd recommend shifting the parsing into a sub-package
> > (workshops.event.import.v1, workshops.event.import.v2, …?), with
> > version-agnostic code in workshops/event/import/__init__.py. Each
> > version module would define:
>
> Huh, it does sound good, but I don't plan on changing
> workshop-template this often.

Does it seem much more difficult than writing the code to handle the
pre- and post- swcarpentry/workshop-template#267 events?  It seems
similar to me, and puts us in a better place in case we do have future
workshop-template meta changes (although we didn't land a
workshop-template-version meta in swcarpentry/workshop-template#267 so
the check_version implementations may start getting sticky [1]).

[1]: https://github.com/swcarpentry/amy/pull/580#issuecomment-157512526

pbanaszkiewicz:
@wking okay, rebased + updated the JS. I think this is in the review phase now.

workshops/test/test_util.py
```diff
@@ -376,195 +376,326 @@ def test_attendance_increases(self):
...
+<meta name="startdate" content="2015-07-13" />
+<meta name="enddate" content="2015-07-14" />
+<meta name="country" content="us" />
+<meta name="venue" content="Euphoric State University" />
+<meta name="address" content="Highway to Heaven 42, Academipolis" />
+<meta name="latlng" content="36.998977, -109.045173" />
+<meta name="language" content="us" />
+<meta name="invalid" content="invalid" />
+<meta name="instructor" content="Hermione Granger, Ron Weasley" />
+<meta name="helper" content="Peter Parker, Tony Stark, Natasha Romanova" />
```
wking: These should be [pipe-delimited now][1], no?

[1]: https://github.com/swcarpentry/workshop-template/commit/1e19a53c2fa00fe1becacbab916cc70d25262b3f

workshops/test/test_util.py
```diff
@@ -376,195 +376,326 @@ def test_attendance_increases(self):
...
+            'address': '',
+            'latitude': None,
+            'longitude': None,
+            'reg_key': None,
+            'instructors': [],
+            'helpers': [],
+            'contact': '',
+        }
+        for latlng, (latitude, longitude) in tests:
+            with self.subTest(latlng=latlng):
```
wking: Very nice :).

workshops/views.py
```diff
@@ -1017,14 +1046,50 @@ def event_delete(request, event_ident):
...
+
+            if response.status_code == 200:
+                # don't throw errors for pages we fall back to
+                content = response.text
+                tags = find_tags_on_event_index(content)
+
+        # normalize (parse) them
+        tags = parse_tags_from_event_website(tags)
+
+        # TODO: change JavaScript for handling import/update from URL
```
wking: Is this TODO for this PR, or is it TODO for some point after this PR lands?

workshops/test/test_util.py
pbanaszkiewicz: Fixed!

workshops/views.py
pbanaszkiewicz: It was fixed in " Update JS for importing/updating events from URL " faaf571c863a8c9e74c25b49ff04e037e7f46548 but I forgot to remove the TODO. Let me remove this line with that commit (rebase on the way).

workshops/test/test_util.py
pbanaszkiewicz: Thanks!

workshops/views.py
```diff
@@ -1017,14 +1046,50 @@ def event_delete(request, event_ident):
...
+            # there are no HTML tags, so let's try the old method
+            url = generate_url_to_event_index(url)
+
+            # fetch page
+            response = requests.get(url)
+
+            if response.status_code == 200:
+                # don't throw errors for pages we fall back to
+                content = response.text
+                tags = find_tags_on_event_index(content)
```
wking: I'd prefer [a more explicit versioning strategy][1], but this is enough for now.  Adding an explicit versioning strategy is something that gets more expensive with time, but it doesn't get *that* much more expensive ;).

[1]: https://github.com/swcarpentry/amy/pull/580#issuecomment-158125458

workshops/views.py
```diff
@@ -1017,14 +1046,50 @@ def event_delete(request, event_ident):
...
+            'Request for "{0}" returned status code {1}.'
+            .format(url, e.response.status_code)
+        )
+
+    except (requests.exceptions.ConnectionError,
+            requests.exceptions.Timeout):
+        raise SuspiciousOperation('Network connection error.')
+
+    except KeyError:
+        raise SuspiciousOperation('Missing or wrong "url" POST parameter.')
```
wking: There's a lot of code between the `try` and this `except`, so I'm concerned that you'll get false catches of *other* `KeyError`s here.  In general, it's good practice to use very explicit exceptions (e.g. `requests.exceptions.ConnectionError`) and/or make the code covered by a try/except block as small as possible.

wking:
On Sat, Nov 21, 2015 at 09:10:45AM -0800, Piotr Banaszkiewicz wrote:
> @wking okay, rebased + updated the JS. I think this is in the review
> phase now.

There's too much in here for me to feel like I've done a thorough
review, but nothing is glaring at me ;).  I left a few more comments
for the tip at faaf571 (and I see you've already fixed the pipe
splitting in 53eef72).  But again, I'm not the one paying the bills,
using the system, or managing the deploy, so whenever it feels good
enough to you, go ahead and land it ;).

pbanaszkiewicz:
@wking thanks a lot for the review!

PR #581, opened 2015-11-20
Add 'stalled' tag

### LINKED ISSUE #545: Flag people who have done instructor training but are unlikely to complete
We now have a way to find people who have done instructor training but have not yet completed; what we need is a way to mark some of them as "don't worry about it, they're unlikely to ever finish". This could be a property of the Task object, so that we could filter by it when selecting who to display.

### LINKED ISSUE #549: Note when giving up on workshop counts
We have space in each event to record how many people attended. 
#514 already asks to break this out to track attrition - some people have actually been providing that.

Some, however, don't respond at all after 3-4 requests at which point I give up asking.  I want to record this but I'm not sure how.  Any ideas?

pbanaszkiewicz:
For use in cases when admins lost contact with event's host, or for TTT
events that are no longer running.

Fixes #549.
Fixes #545.

### DIFF
--- a/workshops/migrations/0062_add_stalled_unresponsive_tags.py
+++ b/workshops/migrations/0062_add_stalled_unresponsive_tags.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+def add_stalled_unresponsive_tags(apps, schema_editor):
+    """Add "stalled" and "unresponsive" tags."""
+    Tag = apps.get_model('workshops', 'Tag')
+    Tag.objects.create(
+        name='stalled',
+        details='Events with lost contact with the host or TTT events that '
+                'aren\'t running.',
+    )
+    Tag.objects.create(
+        name='unresponsive',
+        details='Events whose hosts and/or organizers aren\'t going to send '
+                'attendance data',
+    )
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('workshops', '0061_merge'),
+    ]
+
+    operations = [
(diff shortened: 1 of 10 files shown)


gvwilson:
What's the UI for adding a tag?  I.e., where and how do I do it once 
these is merged?

pbanaszkiewicz:
In here:
screenshot from 2015-11-20 18 47 15

(Event edit form)

gvwilson:
+1

pbanaszkiewicz:
@gvwilson review, please and thanks!

pbanaszkiewicz:
@gvwilson I just thought about something. Do you only want a "stalled" tag, or do you want an interface for viewing people from stalled events too?

If you only need to see all the events marked as "stalled" (and perhaps "TTT"), you'll be able to do so on All Events page thanks to filters.

pbanaszkiewicz:
@gvwilson "Instructors with Issues" ( ;) ) now shows stalled trainees too.

gvwilson:
I'm having trouble following the logic here - if Jane did instructor training at an event in January 2014 (now marked as "we're giving up"), and also took part in instructor training in July 2015 (which is *not* marked as "we're giving up"), but does not yet have an instructor badge, she should show up as "we need to contact her to see if she's going to complete".  On the other hand, if she only took part in the January 2014 event, she should *not* show up as "we need to contact her".  Does this PR do that?

pbanaszkiewicz:
@gvwilson:
>  […] Does this PR do that?

Probably not yet, so let me implement that!

pbanaszkiewicz:
@gvwilson can you please check if this is what you want?

gvwilson:
:+1:

PR #583, opened 2015-11-21
Add filtering to published events API endpoint

### LINKED ISSUE #529: Add event organizer info to the API
Compute Canada would like to be able to use the API to pull all the events it is hosting and then use this information to populate website.

Might be nice to have the EventBrite IDs there too.

pbanaszkiewicz:
Filtered values:
* administrator organization
* host

Both take IDs.

Additionally a new value is being returned by the API: event's
eventbrite ID.

This fixes #529.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -51,4 +51,18 @@ class PublishedEvents(ListAPIView):
     paginator = None  # disable pagination
 
     serializer_class = EventSerializer
-    queryset = Event.objects.published_events()
+
+    def get_queryset(self):
+        """Optionally restrict the returned event set to events hosted by
+        specific host or administered by specific admin."""
+        queryset = Event.objects.published_events()
+
+        administrator = self.request.query_params.get('administrator', None)
+        if administrator is not None:
+            queryset = queryset.filter(administrator__pk=administrator)
+
+        host = self.request.query_params.get('host', None)
+        if host is not None:
+            queryset = queryset.filter(host__pk=host)
+
+        return queryset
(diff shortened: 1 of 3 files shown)


wking:
On Sat, Nov 21, 2015 at 08:14:21AM -0800, Piotr Banaszkiewicz wrote:
> Filtered values:
> * administrator organization
> * host

Looks good to me.  I'd recommend documenting the API.  At work, we use
Swagger [1], development looks like [2], and the rendered UI looks
like [3].  But having docs/api.md with a list of Markdown descriptions
would work too.  Anything that gives folks who are consuming AMY's API
(but not actually working on AMY's code) an overview of whats
available and details on how the endpoints work.

[1]: http://swagger.io/
[2]: https://github.com/azurestandard/api-spec
[3]: http://petstore.swagger.io/?url=https%3A%2F%2Frawgit.com%2Fazurestandard%2Fapi-spec%2Fmaster%2Fpublic.json

pbanaszkiewicz:
@wking 

We have a browsable API and I extended it to show query parameters too:
screenshot from 2015-11-22 09 49 59

This metadata is available after you click "OPTIONS" in http://127.0.0.1:8000/api/v1/events/published/ or https://amy.software-carpentry.org/api/v1/events/published/.

api/views.py
```diff
@@ -14,6 +15,21 @@
 )
 
 
+class QueryMetadata(SimpleMetadata):
+    """Additionally include info about query parameters."""
+
+    def determine_metadata(self, request, view):
+        print('doing something')
```
pbanaszkiewicz: Ooops, should not land this in.

gvwilson: :-)

pbanaszkiewicz: Good thing I don't have a habit of using swear words in debug-printing anymore :-)

PR #585, opened 2015-11-22
Fix: react to more tricky cases of latitude/longitude/eventbrite_id

### LINKED ISSUE #584: "Import from URL" does not seem to set country
Imported two workshops this morning via their URLs (which were http://huguesfontenelle.github.io/2015-12-02-Oslo/ and http://karinlag.github.io/2015-11-20-NVI/). In both cases we got address and lat/long, but *not* country (which should be "Norway").

pbanaszkiewicz:
This fixes #584.

Cases:
* tag is missing from the website - we get `None`
* tag has string content
* tag has empty string content

Tests for these cases were added.

Another change:
slug is now returned in `event_import` view when using the old, workshop
index.html method.

### DIFF
--- a/workshops/util.py
+++ b/workshops/util.py
@@ -431,18 +431,26 @@ def parse_tags_from_event_website(tags):
     try:
         latitude, _ = tags.get('latlng', '').split(',')
         latitude = float(latitude.strip())
-    except ValueError:
+    except (ValueError, TypeError, AttributeError):
+        # value error: can't convert string to float
+        # type error: can't convert None to float
+        # attribute error: object doesn't have "split" or "strip" methods
         latitude = None
     try:
         _, longitude = tags.get('latlng', '').split(',')
         longitude = float(longitude.strip())
-    except ValueError:
+    except (ValueError, TypeError, AttributeError):
+        # value error: can't convert string to float
+        # type error: can't convert None to float
+        # attribute error: object doesn't have "split" or "strip" methods
         longitude = None
 
     try:
         reg_key = tags.get('eventbrite', '')
         reg_key = int(reg_key)
-    except ValueError:
+    except (ValueError, TypeError):
+        # value error: can't convert string to int
+        # type error: can't convert None to int
         reg_key = None
 
     try:
--- a/workshops/util.py
+++ b/workshops/util.py
@@ -373,7 +373,7 @@ def generate_url_to_event_index(website_url):
     results = regex.match(website_url)
     if results:
         return template.format(**results.groupdict())
-    return None
+    raise ValueError('URL does not match Event.WEBSITE_REGEX.')
 
 ALLOWED_TAG_NAMES = [
     'slug', 'startdate', 'enddate', 'country', 'venue', 'address',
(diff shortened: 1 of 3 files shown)


workshops/util.py
```diff
@@ -431,18 +431,26 @@ def parse_tags_from_event_website(tags):
     try:
         latitude, _ = tags.get('latlng', '').split(',')
         latitude = float(latitude.strip())
-    except ValueError:
+    except (ValueError, TypeError, AttributeError):
+        # value error: can't convert string to float
+        # type error: can't convert None to float
```
wking: How do you get that far?  If `tags.get(…)` gives you `None`, the `split(…)` will raise an `AttributeError`.  If `tags.get(…)` doesn't give you `None`, I don't see how you'd get a `None` later.

Also, I don't think you need the `strip()`, since:

    >>> float(' 123 ')
    123.0

but that's not code you're touching in this PR, and it shouldn't hurt anything to have the `strip()` call.

workshops/views.py
```diff
@@ -1061,16 +1062,19 @@ def event_import(request):
 
         if 'slug' not in tags:
             # there are no HTML tags, so let's try the old method
-            url = generate_url_to_event_index(url)
+            index_url = generate_url_to_event_index(url)
 
             # fetch page
-            response = requests.get(url)
+            response = requests.get(index_url)
```
wking: Not something you're changing in this PR, but I just noticed that `generate_url_to_event_index` will return `None` instead of raising a `ValueError` or something if the passed URL doesn't match the expected regexp.  We probably want to make the function raise an exception when it can't find the raw source, and then catch and handle that exception in here.

workshops/views.py
```diff
@@ -1061,16 +1062,19 @@ def event_import(request):
...
             if response.status_code == 200:
                 # don't throw errors for pages we fall back to
                 content = response.text
                 tags = find_tags_on_event_index(content)
 
+                if 'slug' not in tags:
+                    tags['slug'] = urlparse(url).path.replace('/', '')
+
         # normalize (parse) them
         tags = parse_tags_from_event_website(tags)
```
wking: Also tangential to this PR, but we probably want to check for and handle the case where we get up to `parse_tags_from_event_website` but don't actually have a tag dictionary yet (e.g. we got a non-200 status code for the old method).

workshops/views.py
```diff
@@ -1061,16 +1062,19 @@ def event_import(request):
...
-            response = requests.get(url)
+            response = requests.get(index_url)
 
             if response.status_code == 200:
                 # don't throw errors for pages we fall back to
                 content = response.text
                 tags = find_tags_on_event_index(content)
 
+                if 'slug' not in tags:
+                    tags['slug'] = urlparse(url).path.replace('/', '')
```
wking: This assumes more about the URL than I'd like.  Can we do something like:

    Event.WEBSITE_REGEX.match(url).group('repo')

?  We can leave it to `generate_url_to_event_index` to notice and complain about URLs that don't match `WEBSITE_REGEX`.

workshops/util.py
pbanaszkiewicz: You're right, I'll fix that!

workshops/views.py
pbanaszkiewicz: This PR is a good place for introducing a fix to that.

workshops/views.py
pbanaszkiewicz: `parse_tags_from_event_website` is insensitive to empty dictionaries, see for example `test_parsing_empty_tags` in `test_util.py`.

workshops/views.py
pbanaszkiewicz: Sure!

pbanaszkiewicz:
@wking: added fix.

workshops/views.py
wking: On Mon, Nov 23, 2015 at 01:39:48PM -0800, Piotr Banaszkiewicz wrote:
> `parse_tags_from_event_website` is insensitive to empty
> dictionaries, see for example `test_parsing_empty_tags` in
> `test_util.py`.

“We didn't find any tags, because that URL 404ed” is going to be more
useful to admins than “We didn't get any tags, good luck!” ;).

PR #586, opened 2015-11-22
Reporting members of SCF

### LINKED ISSUE #539: Generate a list of SCF members

pbanaszkiewicz:
This work is based on @gvwilson work in #577.

This fixes #539.

Comments welcome!


api/views.py
```diff
@@ -60,6 +65,46 @@ class ExportInstructorLocationsView(ListAPIView):
...
+
+    serializer_class = PersonNameEmailSerializer
+
+    def get_queryset(self):
+        earliest_default, latest_default = default_membership_cutoff()
+
+        earliest = self.request.query_params.get('earliest', None)
+        if earliest is not None:
+            try:
+                earliest = datetime.datetime.strptime(earliest, '%Y-%m-%d') \
```
gvwilson: shouldn't this use universal_date_format?

api/views.py
```diff
@@ -60,6 +65,46 @@ class ExportInstructorLocationsView(ListAPIView):
...
+                                            .date()
+            except ValueError:
+                earliest = earliest_default
+        else:
+            earliest = earliest_default
+
+        latest = self.request.query_params.get('latest', None)
+        if latest is not None:
+            try:
+                latest = datetime.datetime.strptime(latest, '%Y-%m-%d').date()
```
gvwilson: Shouldn't this use universal_date_format?

api/views.py
pbanaszkiewicz: No, this is date *parsing*, not formatting.

api/views.py
pbanaszkiewicz: No, this is date *parsing*, not formatting.

workshops/util.py
```diff
@@ -551,3 +551,32 @@ def update_event_attendance_from_tasks(event):
...
+        task__event__start__lte=latest
+    ).distinct()
+
+    # Merge the two sets.
+    return explicit | implicit
+
+
+def default_membership_cutoff():
+    "Calculate a default cutoff dates for members finding with `get_members`."
+    earliest = datetime.date.today() - 2 * datetime.timedelta(days=365)
```
wking: I haven't looked up a reference, but I think @gvwilson wanted this to be based on calendar years (starting January 1st) with year jumps as we roll over a new-year, instead of a sliding window like this.

workshops/util.py
```diff
@@ -551,3 +551,32 @@ def update_event_attendance_from_tasks(event):
...
+    latest = datetime.date.today()
```
wking: With a cutoff setup for calendar years, I'm not sure how to handle `latest`.  Do we project out to the end of the current year, and see what events an instructor is planning to teach at?  Do we project out to a particular event (e.g. whenever we plan on holding elections)?  To figure out the business logic around `earliest` and `latest`, we might want to consider:

1. The current time, since AMY's knowledge of instructors at past events is presumably more reliable than AMY's knowlege of instructors at future events.
2. When we care about applying the measurement.  I may be interested in the current set of members, or maybe I'm preparing for an election on January 10th, and want an estimate at the member set then.  Or maybe I want the set of instructors who will still be members on January 10th if everybody stops teaching today.

workshops/util.py
```diff
@@ -551,3 +551,32 @@ def update_event_attendance_from_tasks(event):
...
+    instructor_badge = Badge.objects.get(name='instructor')
+    instructor_role = Role.objects.get(name='instructor')
+
+    # Everyone who is an explicit member.
+    explicit = Person.objects.filter(badges__in=[member_badge]).distinct()
+
+    # Everyone who qualifies by having taught recently.
+    implicit = Person.objects.filter(
+        task__role=instructor_role,
+        badges__in=[instructor_badge],
```
wking: No need for the single-entry list here.  Just use:

    badges=instructor_badge,

(although it's always good to test things like that, since I may be misremembering).

workshops/util.py
gvwilson: I can modify the default start date once I've checked the exact 
definition with the Steering Committee, but having a query parameter to 
set the "earliest" date allows us to get the answer even without that.

workshops/util.py
gvwilson: Being able to set "latest" via a query parameter will let us do whatever 
the Steering Committee decides without having to write code.

gvwilson:
:+1: from me

workshops/util.py
wking: On Mon, Nov 23, 2015 at 09:09:42AM -0800, Greg Wilson wrote:
> > +    earliest = datetime.date.today() - 2 * datetime.timedelta(days=365)
> > +    latest = datetime.date.today()
> 
> Being able to set "latest" via a query parameter will let us do
> whatever the Steering Committee decides without having to write
> code.

This offloads “implement the official logic” from Django's Python to
the UI (or the user?).  It seems like we should be centralizing that
logic to avoid looking it up every time we need to run the query,
which means *someone* will be writing code for it.

gvwilson: +1 to centralizing once we have a clear ruling on how exactly to 
interpret the rules.  We can deploy this feature now and use query 
parameters to get the answer I need to have by the end of the month.

wking: On Mon, Nov 23, 2015 at 09:41:27AM -0800, Greg Wilson wrote:
> +1 to centralizing once we have a clear ruling on how exactly to
> interpret the rules.  We can deploy this feature now and use query
> parameters to get the answer I need to have by the end of the month.

Ah, query-parameters as a stop-gap seems reasonable to me.

workshops/util.py
pbanaszkiewicz: You're right:
```python
>>> list(Person.objects.filter(badges__in=[member])) == list(Person.objects.filter(badges=member))
True
```

I didn't know about this "shortcut". However I don't think there's any significant advantage in going this way, so I'll merge without changing this line.

wking: On Tue, Nov 24, 2015 at 08:59:30AM -0800, Piotr Banaszkiewicz wrote:
> I didn't know about this "shortcut". However I don't think there's
> any significant advantage in going this way, so I'll merge without
> changing this line.

It's probably something that the SQL engine will optimize out, but for
dumb engines that don't I expect the singular comparison would be more
efficient.  Not a big deal for us, since we're not hitting any
performance bottlenecks that I know of.

PR #589, opened 2015-11-24
Search in Event's URL, contact, venue and address fields

### LINKED ISSUE #477: Allow search in event URLS
Search should look at event URLs as well as everything else (because that's often an easy way to find workshops).

pbanaszkiewicz:
Fixes #477.


pbanaszkiewicz:
@gvwilson if you can think of more places we should search in, let me know.

pbanaszkiewicz:
@gvwilson bump.

gvwilson:
:+1:

PR #590, opened 2015-11-24
Add more enumeration choices for Event.invoice_status

### LINKED ISSUE #587: More enumeration values for invoicing status
1.  "not invoiced for historical reasons" (because some workshops took place before we started charging)
2.  "not invoiced because of partnership" (which will cover many train-the-trainer events).

@maneesha can you think of others?

pbanaszkiewicz:
Retire events older than 2014, ie. change their invoice status from
unknown to "not invoice for historical reasons" and mark them as
completed.

Fixes #587.


pbanaszkiewicz:
@gvwilson bump.

gvwilson:
:+1:

PR #592, opened 2015-11-27
Event.mailto: new property for grabbing event's contact emails

### LINKED ISSUE #553: Modify "send us attendance figures" email to include host (if any)
The mailto: list for the message asking for attendance figures should include the hosts as well as the instructors.

pbanaszkiewicz:
Used on event's details page and on workshop issues page.

This commit also contains:
* new utility function `find_emails` for finding emails in text (it's
  being used to find emails in Event.contact)
* "mailto:..." content of 'href' attributes was factored out into new
  template (since we don't like repeating ourselves). It's being used by
  event.html, workshop_issues.html templates.

This fixes #553.

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -585,6 +585,30 @@ class Event(AssignmentMixin, models.Model):
         """Indicate if the event has been invoiced or not."""
         return self.invoice_status == 'not-invoiced'
 
+    @property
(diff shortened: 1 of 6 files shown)


pbanaszkiewicz:
@gvwilson bump

Issue #573, opened 2015-11-13
Need ability to edit airport code in Instructor Profile Update


maneesha:
https://amy.software-carpentry.org/workshops/profile_update/36

This instructor has mistyped their airport code (TUC, should be TUS)  but I can't edit it and I can't accept the profile update as is.

Issue #532, opened 2015-10-10
Redisplay person record after approving instructor information


gvwilson:
After I approve an instructor profile update request, I should see the person's record, not the list of other records waiting to be updated - I want to check that the information is correct.

PR #593, opened 2015-11-27
Profile update request fix + redirection to user's profile after approving

pbanaszkiewicz:
This fixes multiple small issues we had about profile update requests machinery: #532, #573.


pbanaszkiewicz:
@gvwilson bump

PR #595, opened 2015-11-30
Removing gvwilson from admin cc list

gvwilson:


pbanaszkiewicz:
I'm thinking about disabling "Invalid HTTP_HOST header" notifications, they're not useful anymore.

gvwilson:
Your call.

PR #600, opened 2015-12-03
Tracking memberships

### LINKED ISSUE #546: Indicate host membership status
Can we note for each host whether they're a member & info on terms of membership agreement?  Dates, how many workshops they get at no cost, etc.

pbanaszkiewicz:
This fixes #546.

The current approach is that there's only one membership per host, ie. if old membership fades out and foundation gets a new one we won't be able to track the historic one. (It's a one-to-one relationship in the database.)

Status is WIP because we one-to-one approach may change (I'm awaiting response https://github.com/swcarpentry/amy/issues/546#issuecomment-161602640)


pbanaszkiewicz:
@wking ready for a review.

pbanaszkiewicz:
I need to rebase this PR because Travis thinks there's an issue (conflicting migration). Why is everyone using `develop` when they shouldn't?!

PR #607, opened 2015-12-06
New instructor badge: DC Instructor

### LINKED ISSUE #596: Tagging instructors as DC/SWC instructors
There's no such distinction yet.

pbanaszkiewicz:
This fixes #596 by:
* changing original 'instructor' badge into 'swc-instructor'
* adding a new 'dc-instructor' badge
* changing all logic behind original 'instructor' badge to work with multiple badges ending in 'instructor'.

### DIFF
--- a/workshops/management/commands/report_all_instructor_activity.py
+++ b/workshops/management/commands/report_all_instructor_activity.py
@@ -16,7 +16,7 @@ class Command(BaseCommand):
         if len(args) != 0:
             raise CommandError('Usage: report_all_instructor_activity')
 
-        instructor_badges = Badge.objects.filter(name__endswith='instructor')
+        instructor_badges = Badge.objects.instructor_badges()
 
         result = []
         all_awards = Award.objects.filter(badge__in=instructor_badges)
--- a/workshops/management/commands/report_all_instructor_activity.py
+++ b/workshops/management/commands/report_all_instructor_activity.py
@@ -16,10 +16,10 @@ class Command(BaseCommand):
         if len(args) != 0:
             raise CommandError('Usage: report_all_instructor_activity')
 
-        instructor_badge = Badge.objects.get(name='instructor')
+        instructor_badges = Badge.objects.filter(name__endswith='instructor')
 
         result = []
-        all_awards = Award.objects.filter(badge=instructor_badge)
+        all_awards = Award.objects.filter(badge__in=instructor_badges)
         for award in all_awards:
             person = award.person
             if not person.email:
--- a/workshops/management/commands/upgrade_instructor_profiles.py
+++ b/workshops/management/commands/upgrade_instructor_profiles.py
@@ -290,11 +290,12 @@ class Command(BaseCommand):
                                           entry['family']))
 
             # check if the person really is an instructor
-            instructor_badge = Badge.objects.get(name='instructor')
-            if person and instructor_badge not in person.badges.all():
+            instructor_badges = Badge.objects.filter(
+                name__endswith='instructor')
+            if person and not (instructor_badges & person.badges.all()):
                 # it's not an error, because we want to have that person in the
                 # database even though they aren't certified instructor
-                warnings.append('This person does not have an instructor'
+                warnings.append('This person does not have any instructor'
                                 ' badge')
 
             # cache airport IATA codes
--- a/workshops/management/commands/upgrade_instructor_profiles.py
+++ b/workshops/management/commands/upgrade_instructor_profiles.py
@@ -290,8 +290,7 @@ class Command(BaseCommand):
                                           entry['family']))
 
             # check if the person really is an instructor
-            instructor_badges = Badge.objects.filter(
-                name__endswith='instructor')
+            instructor_badges = Badge.objects.instructor_badges()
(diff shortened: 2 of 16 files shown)


amy/settings.py
```diff
@@ -131,6 +131,7 @@
     'rest_framework',
     'api',
     'captcha',
+    # 'debug_toolbar',
```
gvwilson: ??

workshops/management/commands/report_all_instructor_activity.py
```diff
@@ -16,10 +16,10 @@ def handle(self, *args, **options):
         if len(args) != 0:
             raise CommandError('Usage: report_all_instructor_activity')
 
-        instructor_badge = Badge.objects.get(name='instructor')
+        instructor_badges = Badge.objects.filter(name__endswith='instructor')
```
gvwilson: Mmm... this feels fragile, but I can't think of a cleaner way to do it.

workshops/management/commands/upgrade_instructor_profiles.py
```diff
@@ -290,11 +290,12 @@ def check_entry(self, entry):
                                           entry['family']))
 
             # check if the person really is an instructor
-            instructor_badge = Badge.objects.get(name='instructor')
-            if person and instructor_badge not in person.badges.all():
+            instructor_badges = Badge.objects.filter(
```
gvwilson: OK, this really does feel fragile. Can we build a set of strings that are the names of instructor-ish badges and use that everywhere instead? (What happens when we get a badge called 'mentor' or 'trainer'?)

workshops/templates/workshops/all_persons.html
```diff
@@ -24,7 +24,7 @@
       </tr>
     {% for person in all_persons %}
       <tr>
-        <td class="text-center">{% if instructor in person.badges.all %}✓{% endif %}</td>
+        <td class="text-center">{% if person.pk in instructors %}✓{% endif %}</td>
```
gvwilson: I like this change.

workshops/templates/workshops/instructors.html
```diff
@@ -25,6 +27,13 @@
         <td>{{ p.num_taught }}</td>
         <td>{% if p.airport %}<a href="{{ p.airport.get_absolute_url }}">{{ p.airport }}</a>{% else %}—{% endif %}</td>
         <td>{% if p.airport %}{{ p.airport.country.name }}{% else %}—{% endif %}</td>
+        <td>{% for badge in p.badges.all %}
+        {% if badge.name == 'swc-instructor' %}
+          {% bootstrap_tag 'SWC' %}
+        {% elif badge.name == 'dc-instructor' %}
+          {% bootstrap_tag 'DC' %}
+        {% endif %}
```
gvwilson: No way to do a keyed lookup for this?

workshops/test/base.py
```diff
@@ -86,9 +86,14 @@ def _setUpAirports(self):
     def _setUpBadges(self):
         '''Set up badge objects.'''
 
-        self.instructor = Badge.objects.create(name='instructor',
-                                               title='Software Carpentry Instructor',
-                                               criteria='Worked hard for this')
+        self.swc_instructor, _ = Badge.objects.get_or_create(
+            name='swc-instructor',
+            defaults=dict(title='Software Carpentry Instructor',
+                          criteria='Worked hard for this'))
```
gvwilson: I think we can come up with clearer criteria :-)

amy/settings.py
pbanaszkiewicz: Yes, see https://github.com/pbanaszkiewicz/amy/commit/5e776de2f2418effee904f407b80917a4646e088 for details. I use `debug_toolbar` application quite often and I am just bored adding and removing this line. (Un)commenting is just easier.

workshops/test/base.py
pbanaszkiewicz: It's only in the tests.

workshops/management/commands/upgrade_instructor_profiles.py
pbanaszkiewicz: We can use a hard-coded list of badge names, like `INSTRUCTOR_BADGES = ['dc-instructor', 'swc-instructor']`.

workshops/templates/workshops/instructors.html
pbanaszkiewicz: There's a way, but it will only work for '*-instructor' badges. For others, a new tag (like `bootstrap_tag`) should be introduced.

pbanaszkiewicz:
@gvwilson I fixed your objections, can you check please?

amy/settings.py
gvwilson: I understand, but please don't check this into the code base. (Can we add a command-line config variable so that when you launch AMY locally for testing, debug_toolbar is enabled?)

gvwilson:
OK to merge after one more correction (don't check in commented-out code).

amy/settings.py
pbanaszkiewicz: Well `debug_toolbar` is supposed to run if and only if `DEBUG=True` in the settings. Are you fine with adding it to the requirements and enabling (==uncommenting that line)?

PR #610, opened 2015-12-08
Updating command used to poke instructors about their activity and profiles

gvwilson:
Need to update activity and profile information in advance of January's election.


gvwilson:
This was used successfully in the 2015-12-10 mailout - @pbanaszkiewicz I think it's ready to merge.

pbanaszkiewicz:
Yup, I received the email and it looked correctly.

PR #614, opened 2015-12-08
Indication of wrong URL

pbanaszkiewicz:
This fixes #618. See commit messages for details.

### DIFF
--- a/workshops/templates/workshops/event_create_form.html
+++ b/workshops/templates/workshops/event_create_form.html
@@ -7,32 +7,7 @@
 <button type="button" class="btn btn-default" id="import_url_button">Import from URL</button>
 {% crispy form %}
 
-<!-- Event import URL modal -->
-<div class="modal fade" id="import_url_modal" tabindex="-1" role="dialog" aria-labelledby="import_url_modal_label" aria-hidden="true">
-  <div class="modal-dialog">
-    <div class="modal-content">
-      <div class="modal-header">
-        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">×</span></button>
-        <h4 class="modal-title" id="import_url_modal_label">Import event data from URL</h4>
-      </div>
-      <div class="modal-body">
-        <div class="alert alert-danger hidden" id="error_message"></div>
-        <form id="import_url_form">
-          {% csrf_token %}
-          <div class="form-group">
-            <label for="event_import_url" class="control-label">Event URL:</label>
-            <input type="url" id="event_import_url" name="url" required class="form-control" />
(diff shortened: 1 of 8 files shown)


workshops/templates/workshops/event_create_form.html
```diff
@@ -16,12 +16,13 @@
...
       </div>
       <div class="modal-body">
+        <div class="alert alert-danger hidden" id="error_message"></div>
         <form id="import_url_form">
           {% csrf_token %}
           <div class="form-group">
             <label for="event_import_url" class="control-label">Event URL:</label>
             <input type="url" id="event_import_url" name="url" required class="form-control" />
-            <p class="help-block">Either URL to the event's website or to it's repository <code>index.html</code> will work.</p>
+            <p class="help-block">Use URL to the event's website, for example <code>http://swcarpentry.github.io/workshop-template/</code>.</p>
```
gvwilson: "Use the URL of the event's website, e.g., XXX, and not the URL of its Git repository."

workshops/templates/workshops/event_edit_form.html
```diff
@@ -16,12 +16,13 @@
...
         </div>
         <div class="modal-body">
+          <div class="alert alert-danger hidden" id="error_message"></div>
           <form id="update_url_form">
             {% csrf_token %}
             <div class="form-group">
               <label for="event_update_url" class="control-label">Event URL:</label>
               <input type="url" id="event_update_url" name="url" required="required" class="form-control" value="{{ object.url }}" />
-              <p class="help-block">Either URL to the event's website or to it's repository <code>index.html</code> will work.</p>
+              <p class="help-block">Use URL to the event's website, for example <code>http://swcarpentry.github.io/workshop-template/</code>.</p>
```
gvwilson: As above (can this message be put in a constant somewhere rather than duplicated?)

workshops/util.py
```diff
@@ -366,6 +366,12 @@ def merge_persons(person_from, person_to):
     person_from.delete()
 
 
+class WrongWorkshopURL(ValueError):
+    def __str__(self):
+        return ('Event\'s URL doesn\'t match Github website format '
+                '"http://user.github.com/2015-12-08-workshop".')
```
gvwilson: What if people aren't using GitHub?

pbanaszkiewicz: This error is raised only if we fall back to the old method, which requires GitHub website URL, and only if we don't get that URL format.

gvwilson: Thanks - worth a docstring.

pbanaszkiewicz:
@gvwilson fixed

Issue #619, opened 2015-12-14
Instructor Profile Update - matching record not found even though it exists


maneesha:
Possible bug?  Or maybe records don't match because of whitespace?  If so can we always strip leading/trailing whitespace?

https://amy.software-carpentry.org/workshops/profile_update/103  gives the warning that no record matches name or email but 

https://amy.software-carpentry.org/workshops/person/361 is a name match

maneesha:
Another example:
Profile update request:  https://amy.software-carpentry.org/workshops/profile_update/94
Existing record:  https://amy.software-carpentry.org/workshops/person/8755

gvwilson:
We're supposed to be stripping spaces - @pbanaszkiewicz, can you please 
dig into this one?

pbanaszkiewicz:
I'll just:

1. Force `.strip()` upon save of both Person and ProfileUpdateRequest (this will fix #612 and #619)
2. Update production by stripping spaces from Person and ProfileUpdateRequest (not in migration)

Stripping will only apply to first, middle and last names, and email, since these are the crucial fields for us.

pbanaszkiewicz:
Closing via #620.

PR #620, opened 2015-12-14
Fix: strip spaces from names and email in Person, ProfileUpdateRequest

### LINKED ISSUE #612: Strip spaces from personal/middle/family names during import
I just removed leading and trailing spaces from the personal, middle, and family names of about 50 people. My guess is that these came in during import from CSV; we should strip these. We should also check that multiple internal spaces are compressed to one (e.g., "Luc  Andre" becomes "Luc Andre" with only one internal space).

pbanaszkiewicz:
Spaces are stripped upon saving the object.

This fixes #612, #619.


gvwilson:
:+1:.

Issue #450, opened 2015-07-24
Add possibility of managing tags


pbanaszkiewicz:
Tags may become a powerful feature for admins and for scenarios we hadn't thought about before.

gvwilson:
Agreed, but low priority for now.

pbanaszkiewicz:
Closed via #621.

PR #621, opened 2015-12-14
Add basic models into admin management interface

pbanaszkiewicz:
This way we can adjust tags or lessons without adding migrations.


pbanaszkiewicz:
This fixes #450.

PR #625, opened 2015-12-15
Fix: disable inputs for online country if it was preselected

### LINKED ISSUE #591: Location fields aren't disabled by default when visiting an online Event
as above.

pbanaszkiewicz:
Error was: location inputs were not disabled if someone entered an edit
page for event which country was 'Online'.
They were only being disabled/enabled when changing the country to/from
'Online'.

This commit fixes this error.
It fixes #591 too.


gvwilson:
:+1:

PR #627, opened 2015-12-16
Fake database instead of anonymized production database

### LINKED ISSUE #534: Improve development database (use [auto generated] fixtures of some sort)
The data that is loaded in the development database is pleasingly non identifying, but tries so hard to not break privacy that it breaks brains. 

I have downloaded and installed the test data as recommended. It's very hard to get an idea of what the software does because the names and emails are so appallingly random. Maybe try [fake factor](https://pypi.python.org/pypi/fake-factor) instead? Also, [a good wordlist can be found here](http://web.archive.org/web/20090918202746/http://tothink.com/mnemonic/wordlist.html) to make random words (like for "slug").

pbanaszkiewicz:
Pros:
* better security
* we can [add our own](http://fake-factory.readthedocs.org/en/latest/#how-to-create-a-provider) Providers to the faker
* we can easily test the `fake_database` command
* IMHO easier maintenance

Cons:
* still this doesn't remove maintenance completely

This fixes #534.


pbanaszkiewicz:
One TODO: change private methods (`_name`) into public ones. Not sure why I made them private in the first place.

pbanaszkiewicz:
@gvwilson maybe you could take a look at this PR?

PR #629, opened 2015-12-20
Cut scripts and commands

### LINKED ISSUE #605: Review scripts and commands
We probably don't need a lot of these scripts or commands. Check which ones are being used, and get rid of others.

(I'd like to not have any commands or scripts, because it's harder to maintain them.)

TODO:
* [X] remove `test-command-line-upload.sh` (done in #608)
* [X] remove `parse-instructor-info.py` (done in #608)
* [ ] update anonymizer or whole dev database (see #534)
* [ ] remove `mail-all-activity.py` and integrate content into `workshops/management/commands/report_all_instructor_activity.py` (see #629)
* [ ] examine other management commands in workshops/management/commands/* and remove those that are out of date or no longer needed (see #629)

pbanaszkiewicz:
This fixes #605. See commit messages for details.


scripts/mail-all-activity.py
```diff
@@ -1,181 +0,0 @@
-#!/usr/bin/env python
```
gvwilson: Why is this one being cut? We used it just last week to confirm people's activity in the run-up to the election. Can't integrate into AMY - the HTTP request would time out.

workshops/management/commands/report_all_instructor_activity.py
```diff
@@ -1,57 +0,0 @@
-import sys
```
gvwilson: How does this relate to manage/commands/instructor_activity.py added above?

workshops/management/commands/report_incomplete_instructors.py
```diff
@@ -1,41 +0,0 @@
-import datetime
```
gvwilson: Do we have command-line equivalents of the various API calls?

scripts/mail-all-activity.py
pbanaszkiewicz: `mail-all-activity.py` is being merged with `report_all_instructor_activity.py` command into `instructors_activity` command.

Commands are run outside of HTTP request (with `./manage.py instructors_activity`), so there are no timeouts. The downside is that if we want to use Django mailing system, we need to supply some settings via envvars.

workshops/management/commands/report_all_instructor_activity.py
pbanaszkiewicz: `mail-all-activity.py` is being merged with `report_all_instructor_activity.py` command into `instructors_activity` command.

workshops/management/commands/report_incomplete_instructors.py
pbanaszkiewicz: No, we had some files that were getting data from API, but they were removed.

`report_incomplete_instructors.py` was quite out of date (the logic wasn't nearly as good as our `Instructor issues` view).

scripts/mail-all-activity.py
gvwilson: 1. Does this mean we have to run manage.py on the server that hosts the 
production database?
2. Where are the required env vars documented?

pbanaszkiewicz: 1. No, we can run this pretty much anywhere (as long as we have the production database downloaded), including the server from which the last bunch of emails was sent out.
2. Here: https://github.com/swcarpentry/amy/blob/develop/amy/settings.py#L62-L72

gvwilson: OK, if the email settings are documented somewhere so that we can find 
them next time, then +1 to merge.

pbanaszkiewicz: 1. Would a `.sh` script that sets the required envvars on the production server work?
2. From where did you send the email last time? The `mail` command used in that old script is still supported in this new one: https://github.com/swcarpentry/amy/pull/629/files#diff-60f31abade2bde4c5ee02a99d9781df4R96

gvwilson: >  1. Would a |.sh| script that sets the required envvars on the
>     production server work?
>

Just docs would be fine - I worry that when we need this, we'll spend 
half an hour figuring out where to find the required vars.

>  1. From where did you send the email last time? The |mail| command
>     used in that old script is still supported in this new one:
>     https://github.com/swcarpentry/amy/pull/629/files#diff-60f31abade2bde4c5ee02a99d9781df4R96
>
>

Logged into software-carpentry.org's server and ran the script from the 
shell.

pbanaszkiewicz:
@gvwilson: I rebased + added the docs you wanted: https://github.com/pbanaszkiewicz/amy/blob/cut-scripts-commands/docs/sending_mails.md

Is it good to merge yet?

gvwilson:
:+1:

PR #630, opened 2015-12-20
Bulk upload: suggest names, generate usernames

### LINKED ISSUE #526: Need better name merging
When we upload data from CSV, we often wind up with duplicate records.  This may happen because we don't always have email addresses to match on, but whatever the cause, we should match names as well to suggest matches.

### LINKED ISSUE #623: Bulk upload must display/allow editing of user ID
The form displayed for editing and confirmation during bulk upload does not show the user name that AMY is constructing, but data may be rejected because either the user's first or last name includes special characters, which cannot be used in user names.  The page should display the user name that AMY has matched on or constructed, and allow it to be edited.

pbanaszkiewicz:
This (hopefully) fixes #623 (see https://github.com/swcarpentry/amy/issues/623#issuecomment-166152983). And this fixes #526.

For details, see commit messages.

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -223,7 +223,10 @@ class Person(AbstractBaseUser, PermissionsMixin):
     github      = models.CharField(max_length=STR_MED, unique=True, null=True, blank=True)
     twitter     = models.CharField(max_length=STR_MED, unique=True, null=True, blank=True)
     url         = models.CharField(max_length=STR_LONG, null=True, blank=True)
-    username    = models.CharField(max_length=STR_MED, unique=True)
+    username = models.CharField(
+        max_length=STR_MED, unique=True,
+        validators=[RegexValidator(r'^[\w\.]+$', flags=re.A)],
+    )
     notes = models.TextField(default="", blank=True)
     affiliation = models.CharField(max_length=STR_LONG, default='', blank=True)
 
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -209,7 +209,7 @@ class Person(AbstractBaseUser, PermissionsMixin):
     )
 
     # These attributes should always contain field names of Person
-    PERSON_UPLOAD_FIELDS = ('personal', 'middle', 'family', 'email')
+    PERSON_UPLOAD_FIELDS = ('personal', 'family', 'email')
     PERSON_TASK_EXTRA_FIELDS = ('event', 'role')
     PERSON_TASK_UPLOAD_FIELDS = PERSON_UPLOAD_FIELDS + PERSON_TASK_EXTRA_FIELDS
 
--- a/workshops/templates/workshops/person_bulk_add_results.html
+++ b/workshops/templates/workshops/person_bulk_add_results.html
@@ -28,7 +28,7 @@
                 <span>{{ entry.family|default:"&mdash;" }}</span>
                 <input type="text" name="family" value="{{ entry.family|default:"" }}" class="hidden">
             </td>
-            <td>
+            <td {% if not entry.person_exists %}class="editable"{% endif %}>
                 <span>{{ entry.username|default:"&mdash;" }}</span>
                 <input type="text" name="username" value="{{ entry.username|default:"" }}" class="hidden">
             </td>
--- a/workshops/templates/workshops/person_bulk_add_results.html
+++ b/workshops/templates/workshops/person_bulk_add_results.html
@@ -9,7 +9,6 @@
     <table class="table table-striped" id="bulk-add-results">
         <tr>
             <th>personal</th>
-            <th>middle</th>
             <th>family</th>
             <th>email</th>
             <th>event</th>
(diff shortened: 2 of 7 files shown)


workshops/models.py
```diff
@@ -209,7 +209,7 @@ class Person(AbstractBaseUser, PermissionsMixin):
     )
 
     # These attributes should always contain field names of Person
-    PERSON_UPLOAD_FIELDS = ('personal', 'middle', 'family', 'email')
```
gvwilson: Why remove middle name?

workshops/templates/workshops/person_bulk_add_results.html
```diff
@@ -25,13 +25,13 @@
...
             <td class="editable">
-                <span>{{ entry.middle|default:"&mdash;" }}</span>
-                <input type="text" name="middle" value="{{ entry.middle|default:"" }}" class="hidden">
-            </td>
-            <td class="editable">
                 <span>{{ entry.family|default:"&mdash;" }}</span>
                 <input type="text" name="family" value="{{ entry.family|default:"" }}" class="hidden">
             </td>
+            <td>
+                <span>{{ entry.username|default:"&mdash;" }}</span>
```
gvwilson: Is it possible to populate the username field with the username that AMY will automatically generate?  In almost all cases, we'll want to use this (but only "almost").

gvwilson:
Usernames are sometimes also used as filenames (e.g., for the _include file in the main website with the person's biography, and as the stem of the PDF certificate name for instructors). Since some Unix command-line tools still have trouble with non-7-bit characters, we want usernames to be 7-bit, e.g., lacera.ana instead of laçera.aña. Since the mapping from accented or non-Latin characters to Latin characters is tricky, it seems simpler to allow editing of the computer-generated username during upload.

workshops/models.py
pbanaszkiewicz: Because, as stated in commit message for ee2b234, it seems like a common mistake for admins to, for example, insert last name into middle name field.

workshops/templates/workshops/person_bulk_add_results.html
pbanaszkiewicz: Right now the field is populated with one of the values:

1. output from `create_username` function if there's no matching person in the database
2. matched person's username

pbanaszkiewicz:
@gvwilson:
> Usernames are sometimes also used as filenames (e.g., for the _include file in the main website with the person's biography, and as the stem of the PDF certificate name for instructors). Since some Unix command-line tools still have trouble with non-7-bit characters, we want usernames to be 7-bit, e.g., lacera.ana instead of laçera.aña. Since the mapping from accented or non-Latin characters to Latin characters is tricky, it seems simpler to allow editing of the computer-generated username during upload.

This is convincing. I could add a regex check for non-ASCII characters.

I'll also introduce this change to the way username field works:

* if there's no matching person (== we'll add a new person and a new task to the database), the field will be editable
* if there is a matching person (== we'll only add a new task to that person), the field will not be editable

Sounds good?

pbanaszkiewicz:
@gvwilson updated!

workshops/templates/workshops/person_bulk_add_results.html
```diff
@@ -25,13 +25,13 @@
...
-                <input type="text" name="middle" value="{{ entry.middle|default:"" }}" class="hidden">
-            </td>
-            <td class="editable">
                 <span>{{ entry.family|default:"&mdash;" }}</span>
                 <input type="text" name="family" value="{{ entry.family|default:"" }}" class="hidden">
             </td>
+            <td {% if not entry.person_exists %}class="editable"{% endif %}>
+                <span>{{ entry.username|default:"&mdash;" }}</span>
+                <input type="text" name="username" value="{{ entry.username|default:"" }}" class="hidden">
+            </td>
```
gvwilson: So when I do bulk upload, is the editable table pre-populated with machine-generated usernames? If so, where is that done?

workshops/util.py
```diff
@@ -167,18 +170,60 @@ def verify_upload_person_task(data):
...
+            item['username'] = person.username
+            item['person_exists'] = True
 
-        if (event and not role) or (role and not event):
-            errors.append("Must have both: event ({0}) and role ({1})"
-                          .format(event, role))
+        else:
+            # force a newly created username
+            if not item.get('username'):
+                item['username'] = create_username(personal, family)
```
gvwilson: So this is done during upload?

gvwilson:
OK, if I understand the logic correctly now, :+1: to merge - but man, this is becoming complicated...

workshops/templates/workshops/person_bulk_add_results.html
pbanaszkiewicz: Yes, and it's done in `verify_upload_person_task`: after the data is uploaded and read from CSV into Pythonic dict.

workshops/util.py
pbanaszkiewicz: `verify_upload_person_task` is called multiple times, for example upon clicking 'Verify' button. See `person_bulk_add_confirmation` view for details.

pbanaszkiewicz:
@gvwilson:
> OK, if I understand the logic correctly now, :+1: to merge - but man, this is becoming complicated...

Yep, every removed bit counts, even something as small as ditch of `middle` field.

PR #637, opened 2015-12-30
API: events' tags

### LINKED ISSUE #636: API: add filtering by workshop type in published events
We need to grab DC-only, or SWC-only published events. Probably there's no need to return type of the event in the structure, just filtering.

pbanaszkiewicz:
This fixes #636 by adding:

* query param for filtering events by tags
* serializer for event tags

### DIFF
--- a/api/test/test_events.py
+++ b/api/test/test_events.py
@@ -3,6 +3,7 @@ import json
 from unittest.mock import patch
 
 from django.core.urlresolvers import reverse
+from django.http import QueryDict
 from rest_framework import status
 from rest_framework.test import APITestCase
 
@@ -147,7 +148,8 @@ class TestListingPublishedEvents(APITestCase):
             },
         ]
 
-    @patch.object(PublishedEvents, 'request', query_params={}, create=True)
+    @patch.object(PublishedEvents, 'request', query_params=QueryDict(),
+                  create=True)
     def test_serialization(self, mock_request):
         # we're mocking a request here because it's not possible to create
         # a fake request context for the view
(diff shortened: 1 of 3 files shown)


api/test/test_events.py
```diff
@@ -144,10 +147,12 @@ def setUp(self):
                 'url': 'https://user.github.io/repository/',
                 'contact': 'sb@sth.edu',
                 'eventbrite_id': None,
+                'tags': [],
             },
         ]
 
-    @patch.object(PublishedEvents, 'request', query_params={}, create=True)
+    @patch.object(PublishedEvents, 'request', query_params=QueryDict(),
```
gvwilson: why this change?

api/views.py
```diff
@@ -141,6 +148,7 @@ def get_query_params_description(self):
             'administrator': 'ID of the organization responsible for admin '
                              'work on events.',
             'host': 'ID of the organization hosting the event.',
+            'tag': 'Events\' tag. You can use this parameter multiple times.',
```
gvwilson: I would write as `"Events; tags(s). You can use this parameter multiple times."` (double quote to avoid escaping the single quote, then a bit of grammar)

gvwilson:
minor question and minor suggestion, then go ahead and merge.

api/test/test_events.py
pbanaszkiewicz: Because the view started using `query_params.getlist` which is specific to `QueryDict`, not "normal" `dict`. This change is compatible with `dict` and prevents AttributeError.

Issue #582, opened 2015-11-21
Let's register teaching languages for instructors


karinlag:
I was speaking to two of our french speaking instructors yesterday, and that got me wondering whether we ask people to register which language they would be comfortable teaching in. If we do, I though that could be used to select a first group of instructors to ask in case of workshops in countries where another language than English would be preferred.

pbanaszkiewicz:
Hello @karinlag.

We don't store information about languages that instructors feel comfortable teaching in.

@jduckles, @maneesha, @gvwilson do you feel we need that info too?

Thanks,
Piotr

pbanaszkiewicz:
Let me just add that we do, in fact, ask for workshop language in workshop request form.

We also provide a way for admins to find instructors from specific countries (I believe there's a correlation between used language and country person lives in :-) ).

If we start storing language info then we should:

1. store event language in events
2. store instructors language in their profile
3. ask for the language in profile update form

gvwilson:
+1 - we've already been asked to get this for trainees so that they can 
do their mentoring and mini-lesson in their preferred (human) language.

karinlag:
On 11/21/2015 12:12 PM, Piotr Banaszkiewicz wrote:
> Let me just add that we do, in fact, ask for workshop language in
> workshop request form.
>
> We also provide a way for admins to find instructors from specific
> countries (I believe there's a correlation between used language and
> country person lives in :-) ).
>
> If we start storing language info then we should:

Just to have it mentioned, the two french speaking instructors that 
spurred this both live in Norway.

Karin

jduckles:
+1 I think it should allow for more than a single language, some instructors might be comfortable teaching in multiple languages. This is essentially the same as how we track the lessons someone is comfortable teaching it, we just want to track the languages they're comfortable teaching in as well. As Greg says, there may be utility in tracking the "preferred" language as well.

wking:
On Sat, Nov 21, 2015 at 07:15:19AM -0800, Jonah Duckles wrote:
> +1 I think it should allow for more than a single language, some
> instructors might be comfortable teaching in multiple
> languages. This is essentially the same as how we track the lessons
> someone is comfortable teaching it, we just want to track the
> languages they're comfortable teaching in as well. As Greg says,
> there may be utility in tracking the "preferred" language as well.

Here's how this works for HTML:

* Language(s) of some content's intended audience with [1]:

    Content-Language = 1#language-tag

  Like:

    Content-Language: mi, en

[...]

Issue #626, opened 2015-12-16
Database: clear migrations, move data migrations → fixtures


pbanaszkiewicz:
We have a number of data migrations that input new values into empty database. Try creating a fresh new database:

``` shell
$ rm db.sqlite3
$ ./manage.py migrate
$ make superuser
$ make serve
```

Now log in with `admin:admin` credentials and go to http://127.0.0.1:8000/workshops/admin/ to see some of the values that we add with migrations.

Since we're switching to fake-generated "fixtures" (see #534 for details), I'd like also to get rid of all migrations that provide initial values for some of the models, and instead use fixtures for that.

I think we can safely use fixtures because these initial values don't change so often.

Also, I think removing all migrations (or squashing them while dropping data migrations) needs to be done carefully. It may require a release solely for this change (so that no new migrations are added).

pbanaszkiewicz:
I'm deferring this for now, because with Django 1.9 the migrations are **so fast** it doesn't matter… It would still be a nice thing, but for now it's just fine.

narayanaditya95:
@pbanaszkiewicz I am starting to reconsider this proposal as we progress towards adding PyData specific objects to the database.
Having fixtures would make our tests independent of the initial data we intend to populate our database with.

pbanaszkiewicz:
@narayanaditya95 there are advantages to using fixtures, but personally I don't want to touch migrations to fix our previous errors anytime soon.

chrismedrela:
Side note: squashing migrations doesn't speed up applying them. I've squashed first 87 migrations here (https://github.com/chrismedrela/amy/commit/4e5a29a02e877827397a50d44a630609cb40386b) and there is no improvement -- applying migrations still takes ~30 seconds on my machine.

pbanaszkiewicz:
I don't think we have manpower to do the switch right now, so I'm postponing until I don't know when.

gvwilson:
:+1:

maneesha:
@pbanaszkiewicz  is this something that we still need to consider?

pbanaszkiewicz:
@maneesha no, I don't feel like we need this transition anymore. Data migrations work just fine.

Issue #341, opened 2015-06-05
Make fields in-place editable


gvwilson:
We're currently using forms for almost everything, which is very old-fashioned - it would really (really) help if events, sites, persons, etc., could be edited in-place without form submission. This is pretty much at the top of Amy-the-person's wishlist.

pbanaszkiewicz:
This is pretty big. @amyrbrown how would this feature help you with your work?

amyrbrown:
The "notes" field is the one I edit most often -- it would be awesome if I could just click in it and edit it (and only it) instead of clicking the "edit" button, which makes everything editable.

Alternately maybe just a button to edit the notes field alone -- kind of like you can edit the title of a Github issue without editing the whole thing?

(I wouldn't say this is at the top of my list, but it would be really really nice. :) )

wking:
On Fri, Jun 05, 2015 at 09:00:41AM -0700, Piotr Banaszkiewicz wrote:
> This is pretty big.

The little “Django API + Angular frontend” bird is singing in my head
;).

pbanaszkiewicz:
> The "notes" field is the one I edit most often -- it would be awesome if I could just click in it and edit it (and only it) instead of clicking the "edit" button, which makes everything editable and is pretty slow.

If it's the edit OBJECT webpage loading slowly, then we have a solution in v0.3; I don't think it's running on the production yet, though.

> Alternately maybe just a button to edit the notes field alone -- kind of like you can edit the title of a Github issue, or just a comment, without editing the whole thing?

That's definitely an easy way to solve this issue, and I like it.

> The little “Django API + Angular frontend” bird is singing in my head ;).

That's the whole release on its own, and I'd like to not go there if we can :-) but yeah, it may be inevitable at some point. (Not saying it's a bad idea, though! It's just a big thing.)

narayanaditya95:
Can we use a wysiwyg editor on the notes field. [django-wysiwyg](https://github.com/pydanny/django-wysiwyg/) is stated to work with django-crispy-forms. It is worth a try or going the AngularJS way is the better option in the long run.

pbanaszkiewicz:
@narayanaditya95 [annotating with dates](https://github.com/swcarpentry/amy/issues/299) is more important than wysiwyg.

gvwilson:
Helpful, but not essential.

PR #640, opened 2016-01-02
WIP: workshops/models.py: Add LanguageTag and LanguageQualification models

wking:
Based on [a very restricted subset][1] of [RFC 5646][2].  This will
allow us to match instructors (and potentially students) with events
in an appropriate language.

As of the current tip (63fa782), this stores language ↔ person
associations, but doesn't currently allow you to edit weights (the
database has weights, but the UI doesn't provide a way to edit them,
and will clobber with 1s on person-edit).

There's currently no support for @pbanaszkiewicz's [requested][3]:

* store event language in events
* ask for the language in profile update form

I'm not familiar enough with the latter to know how to handle it.  Do
I have to do anything besides updating ProfileUpdateRequest?  I looked
around briefly for a template that we email instructors, since that
probably needs to be updated to list the languages we know about, but
couldn't find anything.  Can anyone give me some pointers or a refence
to docs or previous similar work?

I haven't stored event languages, because I expect there's not much
use for this information after the event is setup.  The EventRequest
form now uses a many-to-many language table, so we don't have to parse
user-entered languages.  And once the EventRequest is translated into
an Event, the language bits have already taken care of themselves
(unless we want to list them on out upcoming events feed?)

So this is a bit of a work-in-progress, but I think it's far enough
along to get some feedback ;).

[1]: https://github.com/swcarpentry/amy/issues/582#issuecomment-158678235
[2]: https://tools.ietf.org/html/rfc5646
[3]: https://github.com/swcarpentry/amy/issues/582#issuecomment-158628490

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -361,6 +361,14 @@ class PersonForm(forms.ModelForm):
         widget=selectable.AutoComboboxSelectWidget,
     )
 
+    languages = selectable.AutoCompleteSelectMultipleField(
+        lookup_class=lookups.LanguageLookup,
+        label='Languages',
+        required=False,
+        help_text=AUTOCOMPLETE_HELP_TEXT,
+        widget=selectable.AutoComboboxSelectMultipleWidget,
+    )
+
     class Meta:
         model = Person
         # don't display the 'password', 'user_permissions',
@@ -369,6 +377,15 @@ class PersonForm(forms.ModelForm):
         fields = ['username', 'personal', 'middle', 'family', 'may_contact',
                   'email', 'gender', 'airport', 'affiliation', 'github',
                   'twitter', 'url', 'notes', 'lessons', 'domains']
+
+    def __init__(self, instance=None, **kwargs):
+        kwargs['instance'] = instance
+        if instance:
+            if 'initial' not in kwargs:
+                kwargs['initial'] = {}
+            if 'languages' not in kwargs['initial']:
+                kwargs['initial']['languages'] = instance.languages.all()
+        super(PersonForm, self).__init__(**kwargs)
 
 
 class PersonPermissionsForm(forms.ModelForm):
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -110,6 +110,13 @@ class InstructorsForm(forms.Form):
                                              widget=CheckboxSelectMultiple(),
                                              required=False)
 
+    language = selectable.AutoCompleteSelectField(
+        lookup_class=lookups.LanguageLookup,
+        label='Language',
+        required=False,
+        widget=selectable.AutoComboboxSelectWidget,
+    )
+
     INSTRUCTOR_BADGE_CHOICES = (
         ('swc-instructor', 'Software Carpentry Instructor'),
         ('dc-instructor', 'Data Carpentry Instructor'),
@@ -165,6 +172,7 @@ class InstructorsForm(forms.Form):
                 ),
                 css_class='panel panel-default ',
             ),
+            'language',
             'gender',
             'lessons',
             'instructor_badges',
--- a/workshops/lookups.py
+++ b/workshops/lookups.py
@@ -66,8 +66,18 @@ class AirportLookup(ModelLookup):
     )
 
 
+@login_required
+class LanguageLookup(ModelLookup):
+    model = models.Language
+    search_fields = (
+        'name__icontains',
+        'language__icontains'
+    )
+
+
 registry.register(EventLookup)
 registry.register(HostLookup)
 registry.register(PersonLookup)
 registry.register(AdminLookup)
 registry.register(AirportLookup)
+registry.register(LanguageLookup)
--- a/workshops/migrations/0068_auto_20160102_1502.py
+++ b/workshops/migrations/0068_auto_20160102_1502.py
@@ -1,8 +1,60 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import email.parser
+import urllib.request
+
 from django.db import migrations, models
 from django.conf import settings
+
+
+def add_languages(
+        apps, schema_editor,
+        url='http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry'):
+    """Populate the Languages table.
+
+    [1] lists IANA as the registry maintainer, [2] looks like that
+    registry, and [3] describes the entry-format in the registry.
+
+    [1]: https://tools.ietf.org/html/rfc5646#section-2.2.1
+    [2]: http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry
+    [3]: https://tools.ietf.org/html/rfc5646#section-3.1
+    """
+    with urllib.request.urlopen(url=url) as response:
+        payload_bytes = response.read()
+        charset = response.headers.get_content_charset()
+        payload = payload_bytes.decode(charset)
+    Language = apps.get_model('workshops', 'Language')
+    record = []
+    for line in payload.splitlines():
+        if line == '%%':
+            add_language(record=record, Language=Language)
+            record = []
+        else:
+            record.append(line)
+    add_language(record=record, Language=Language)
+
+
+def add_language(record, Language):
+    if len(record) == 0:
+        return
+    # https://tools.ietf.org/html/rfc5646#section-3.1.2
+    if len(record) == 1 and record[0].startswith('File-Date:'):
+        return
+    # https://docs.python.org/3/library/email.parser.html#parser-class-api
+    # to handle line-wrapping
+    # https://tools.ietf.org/html/rfc5646#section-3.1.1
+    fields = email.parser.Parser().parsestr('\r\n'.join(record))
+    if fields['Type'] != 'language':
+        return
+    if len(fields['Subtag']) > 2:
+        # skip these until we need them
+        # https://github.com/swcarpentry/amy/issues/582#issuecomment-159506884
+        return
+    Language.objects.get_or_create(
+        name=fields['Description'],
+        language=fields['Subtag'],
+    )
 
 
 class Migration(migrations.Migration):
@@ -34,4 +86,5 @@ class Migration(migrations.Migration):
             name='languages',
             field=models.ManyToManyField(to='workshops.Language', through='workshops.LanguageQualification', blank=True),
         ),
+        migrations.RunPython(add_languages),
     ]
--- a/workshops/migrations/0068_auto_20160102_1502.py
+++ b/workshops/migrations/0068_auto_20160102_1502.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from django.conf import settings
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('workshops', '0067_person_username_regexvalidator'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Language',
+            fields=[
+                ('id', models.AutoField(serialize=False, primary_key=True, verbose_name='ID', auto_created=True)),
+                ('name', models.CharField(max_length=40, help_text='Description of this language tag in English')),
+                ('language', models.CharField(max_length=10, help_text='Primary language subtag.  https://tools.ietf.org/html/rfc5646#section-2.2.1')),
+            ],
+        ),
+        migrations.CreateModel(
+            name='LanguageQualification',
+            fields=[
+                ('id', models.AutoField(serialize=False, primary_key=True, verbose_name='ID', auto_created=True)),
+                ('weight', models.FloatField(default=1, help_text='https://tools.ietf.org/html/rfc7231#section-5.3.1')),
+                ('language', models.ForeignKey(to='workshops.Language')),
+                ('person', models.ForeignKey(to=settings.AUTH_USER_MODEL)),
+            ],
+        ),
+        migrations.AddField(
+            model_name='person',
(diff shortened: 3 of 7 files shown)


pbanaszkiewicz:
ref #582

pbanaszkiewicz:
@wking thank you for the contribution. I went through your work and below are my comments.

### The weights issue

The problem with 0÷1 grading is that no-one says what "1" and what "0" means. Does "1" mean I can speak English fluently, or does it mean I finished English studies?

Remember that people tend to under-evaluate their capabilities when self-assessing (what @gvwilson trains instructors to avoid).

So if we keep the weight system, I'd force `LanguageQualification.weight` to use predefined values (choices). This way we can help users self-assess.

For example we could use these values:
* 1 - I'm comfortable being a lead instructor for an event
* 0.75 - I'm comfortable being a helper during the event
* 0.5 - I'd rather not speak to the audience

They aren't perfect and assume that stress and discomfort when public speaking are related to lower language capabilities (which may or may not be true).

### Relationship

It's not explicitly stated, but I'd like to have a relation `Person` ↔ `Language`, with this design probably with an intermediate model `LanguageQualification`.

If we, however, dropped weights (which I am in favor of personally), then:

* we lose intermediate model (huge win for programming efforts)
* and we lose UI for setting weights,
* and we lose N more fields users would have to fill out.

### Default languages

1. I'm against getting languages from the internet
2. I'm trying to get rid of data-import migrations (see #626)

Instead of using IANA registry (dd3c641f01f52ba76e5aa681e8c7a7d7ef7f5458 " workshops/migrations/0068: Add Language-import data migration"), I'd just copy Django's list:
https://github.com/django/django/blob/59f861fcb49a43acfbf2f75e683a7cb971600f52/django/conf/global_settings.py (it's shorter!)

### Event's languages

> I haven't stored event languages, because I expect there's not much
> use for this information after the event is setup. The EventRequest
> form now uses a many-to-many language table, so we don't have to parse
> user-entered languages. And once the EventRequest is translated into
> an Event, the language bits have already taken care of themselves
> (unless we want to list them on out upcoming events feed?)

This is not true. When a new EventRequest comes in, admins turn it into an unfinished Event. It stays that way until all details are sorted out (location, dates, instructors, money, etc.).

So I want to store event's languages with the event, and also populate EventForm's languages when EventRequest is being accepted.

wking:
On Sun, Jan 03, 2016 at 01:25:46PM -0800, Piotr Banaszkiewicz wrote:
> For example we could use these values:
> * 1 - I'm comfortable being a lead instructor for an event
> * 0.75 - I'm comfortable being a helper during the event
> * 0.5 - I'd rather not speak to the audience

Sounds good to me, in which case we may want to change the …__gte=0.5
to …_gt=0.5.  But I otherwise I think this is a UI/docs decision that
doesn't need backend changes.

> They aren't perfect…

I think this is going to be true of any system ;).  I wouldn't worry
about finessing it too much.

> If we, however, dropped weights (which I am in favor of personally),
> then:
> 
> * we lose intermediate model (huge win for programming efforts)

But programming efforts are completed now, no?  Excepting the
weight-setting UI, which we obviously can't have without a
weight-storage field.

> * and we lose UI for setting weights,

Agreed.

> * and we lose N more fields users would have to fill out.

?  Setting weights (assuming we had a UI) sounds like one more
optional radio/select per language you speak.  I'm not sure how many
languages our instructors speak, but I'd be surprised if anyone speaks
so many that setting weights would laborious.  And if the instructor
is comfortable enough with an effective default weight of one (what
you get if there's nowhere to store weights), then they don't need to
bother configuring their weights.  So I see weights causing a marginal
increase in user effort, but only if the user opts into sharing that
level of detail.

> ### Default languages
> 
> 1. I'm against getting languages from the internet

I don't see a way to avoid that ;), and in this case the IANA URL is
the canonical source for RFC 5646 language tags.

> 2. I'm trying to get rid of data-import migrations (see #626)

I can change my data migration to a fixture-creating script.  Then we
can re-run the script if/when we want to update the fixture to pull in
IANA changes.

> Instead of using IANA registry …, I'd just copy Django's list …
> (it's shorter!)

Shorter because many languages don't have out-of-the-box Django
translations [1].  The IANA approach scales well: you need a langauge
we don't have listed?  Just bump up the subtag size limit to three
characters.  You need a language that the IANA doesn't list?  Register
it with IANA [2], and we'll pick it up once it's accepted.  I'm not
sure how the Django approach scales.  If you need a new language,
write a Django translation?  Or we just add it by hand?

> So I want to store event's languages with the event, and also
> populate EventForm's languages when EventRequest is being accepted.

Ok.  I feel like having Event and a separate EventRequest (and Person
and a separate ProfileUpdateRequest is causing more duplication than I
want to code/test ;).  But I guess that gets back to #70, where I'd
just let event-requestors create an Event or edit their Person
directly, with permissions controlling sensitive fields (e.g. need an
admin to finish the Event off for publication, or to badge a person).
I'll look at writing an Event CRUD API to mock something like that up.

So my proposed plan for this PR:

* Convert the IANA-pulling data migration to a fixture-generator.
* Commit a fixture generated by that script.
* Add Event.language (so multi-language events would get multiple
  Event entries?) which is initialized from EventRequest.langauge for
  EventRequest-based Events.

Do I need to do more around ProfileUpdateRequest?  I'm still lost on
that front.

[1]: https://github.com/django/django/blob/59f861fcb49a43acfbf2f75e683a7cb971600f52/django/conf/global_settings.py#L48
[2]: http://tools.ietf.org/html/rfc5646#section-3.5

pbanaszkiewicz:
@wking:
> Shorter because many languages don't have out-of-the-box Django
translations [1].  The IANA approach scales well: you need a langauge
we don't have listed?  Just bump up the subtag size limit to three
characters.  You need a language that the IANA doesn't list?  Register
it with IANA [2], and we'll pick it up once it's accepted.  I'm not
sure how the Django approach scales.  If you need a new language,
write a Django translation?  Or we just add it by hand?

But do we need to have 200 languages in the database when our users only claim to speak less than 30 of them? Or will we ever have a workshop in Interlingua, Esperanto or Ido?

Also related issue I've just found: IANA registry's logic is more complex than your script's logic. For example: some entries have multiple descriptions (like `cu`) and some are deprecated (see `ik`, `id`).

> So my proposed plan for this PR:
> 
>  * Convert the IANA-pulling data migration to a fixture-generator.
>  * Commit a fixture generated by that script.
>  * Add Event.language (so multi-language events would get multiple Event entries?) which is initialized from EventRequest.langauge for EventRequest-based Events.

We're not switching to fixtures yet, so for now can you please:

1. Write a management command (not script!) for getting data from IANA
2. Hardcode that data in a data migration
3. Add Event.languages

> Do I need to do more around ProfileUpdateRequest? I'm still lost on that front.

1. Add `languages` M2M field - that's the easy part
2. Figure out what should we do about a widget for the form - that's the hard part.

But before that, I want to hear second opinion on including weights into the system. If we end up using them, then we'd have to somehow come up with a proper, easy to use widget (and from what I remember this is not a trivial task). Probably some JS would be needed as well.

However, if we dropped the weights, our forms would be significantly easier to do.

-----

@gvwilson should we include weights in relations between person and their languages to indicate how comfortable a person is speaking in these languages?

For example I'd give 1 to Polish but something around 0.8 to English and 0.1 to German.

We can use hardcoded values to help users decide what level of fluency they have:

> * 1 - I'm comfortable being a lead instructor for an event
> * 0.75 - I'm comfortable being a helper during the event
> * 0.5 - I'd rather not speak to the audience

wking:
On Sun, Jan 03, 2016 at 03:21:59PM -0800, Piotr Banaszkiewicz wrote:
> But do we need to have 200 languages in the database when our users
> only claim to speak less than 30 of them? Or will we ever have a
> workshop in Interlingua, Esperanto or Ido?

Probably not ;).  But I'd rather not be in the business of deciding
which subset of IANA languages we need.  Autocomplete should cover us
here.

> Also related issue I've just found: IANA registry's logic is more
> complex than your script's logic. For example: some entries have
> multiple descriptions (like `cu`)…

Good point.  In this case, email.parser.Parser collects all the
entries, but __getitem__ returns a single entry [1].  That should be
fine for us.  Supporting multiple descriptions doesn't seem to be
worth the trouble, but if we want to concat (with semicolon
separators?) and push into a single field for auto-completion, we
could do that with get_all [2].

> … and some are deprecated (see `ik`, `id`).

I think this is also more trouble than it's worth for the level of
data cleanliness I expect us to have.  Ideally, somebody else would
have written a conformant RFC 5646 Python package… and maybe they have
[3]?  We could use that instead of my current Language table, with a
character field containing the full tag instead of a primary-key
pointing into Language.

> 1. Write a management command (not script!) for getting data from IANA
> 2. Hardcode that data in a data migration

These could both be replaced by a language-tag [3] dependency, if we
want to go that way.

> 3. Add Event.languages

Is the plural intended?  If so, we probably want to update
EventRequest as well, since it's currently using a singular field
name.

> > Do I need to do more around ProfileUpdateRequest? I'm still lost on that front.
> 
> 1. Add `languages` M2M field - that's the easy part
> 2. Figure out what should we do about a widget for the form - that's the hard part.

The M2M autocomplete widgets I'm using already (e.g. in the
person-detail form) seem straigtforward enough.  If we can relax the
login requirement for lookups [4] or use a UI based on [5] (with
client-side autocompletion), it shouldn't be too bad to do something
similar with language-tags.

> But before that, I want to hear second opinion on including weights
> into the system. If we end up using them, then we'd have to somehow
> come up with a proper, easy to use widget (and from what I remember
> this is not a trivial task). Probably some JS would be needed as
> well.

Agreed.  I'm personally still a fan of an API (in Django, or whatever)
and separate JavaScript UI (in a JavaScript framework like Angular),
because it makes this sort of problem trivial [6].

[1]: https://docs.python.org/3/library/email.message.html#email.message.Message.__getitem__
[2]: https://docs.python.org/3/library/email.message.html#email.message.Message.get_all
[3]: https://github.com/OnroerendErfgoed/language-tags
[4]: https://github.com/wking/swc-amy/commit/435f7e8945b7f1b878c2c622746bc743132a07f4#diff-e94887926e2b74167200503eca189e19R69
[5]: https://www.npmjs.com/package/language-tags
[6]: https://github.com/swcarpentry/amy/issues/341#issuecomment-109637799

pbanaszkiewicz:
@wking:

> Ideally, somebody else would have written a conformant RFC 5646 Python package… and maybe they have [3]?

+10 to using a 3rd party Python provider, however you may have troubles listing all included language tags.

> Supporting multiple descriptions doesn't seem to be worth the trouble, but if we want to concat (with semicolon separators?)

`language-tags` returns a list of descriptions:

```python
>>> print(tags.language('nl').description);
['Dutch', 'Flemish']
```

I'd concat them.

>> 3. Add Event.languages
>
> Is the plural intended?  If so, we probably want to update
> EventRequest as well, since it's currently using a singular field
> name.

No, sorry, that was my mistake. Event can have only one language.

> The M2M autocomplete widgets I'm using already (e.g. in the
> person-detail form) seem straigtforward enough.  If we can relax the
> login requirement for lookups [4] or use a UI based on [5] (with
> client-side autocompletion), it shouldn't be too bad to do something
> similar with language-tags.

Okay, I checked and indeed M2M autocomplete widget works excellent (it just needs some better styling).

And we can have relaxed permission requirements for language tag lookup.

Anyway, let's see if [`language-tags`](https://github.com/OnroerendErfgoed/language-tags) works okay for us and what does @gvwilson say about weights.

pbanaszkiewicz:
@wking since [Greg is focusing on other projects](https://github.com/swcarpentry/amy/pull/641#issuecomment-168829179) (and [others](https://github.com/swcarpentry/amy/issues/582#issuecomment-158778448) [agreed](https://github.com/swcarpentry/amy/issues/582#issuecomment-158780364) share my opinion) I'll make a design decision here: let's wipe out the `LanguageQualification` model (and therefore weights). It will save us from using an intermediary model for M2M relation with Person, and we won't have to think about UI.

wking:
On Mon, Jan 04, 2016 at 02:55:53PM -0800, Piotr Banaszkiewicz wrote:
> … let's wipe out the `LanguageQualification` model (and therefore weights).

Ok, I'll post a not here once I've rerolled to drop them and converted
Language over to language-tags.

pbanaszkiewicz:
@wking any progress here?

wking:
On Mon, Feb 01, 2016 at 09:20:59AM -0800, Piotr Banaszkiewicz wrote:
> @wking any progress here?

No, I haven't had time to implement the planned reroll.  Maybe this
week?

narayanaditya95:
I agree to most of the suggestions and the code that exist here.

>  let's wipe out the LanguageQualification model (and therefore weights). It will save us from using an intermediary model for M2M relation with Person, and we won't have to think about UI.

+1 for that. Lets make it easier for the instructors to rate their ability to deliver a workshop in a particular language.

> Ideally, somebody else would have written a conformant RFC 5646 Python package… and maybe they have [3]?

How often is [official registry](http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) updated? I don't think the the list of languages with two-lettered tags (the popular languages we are interested in) receive additions/modification regularly that we should use a separate Python package to track changes.  
The migration that is included seems the right solution to me.
Opinions?

I will be pushing a PR by tonight which will probably borrow a lot of code from this PR.

gvwilson:
I feel very strongly that we should get rid of weights and just have 
"person X can teach in language Y".  After all, we don't ask people to 
rate their proficiency in lessons - we trust them when they say "I can 
do this".

pbanaszkiewicz:
We're ditching this one in favor of #798.

PR #641, opened 2016-01-04
workshop/forms.py: change some fields' widgets to TextInput

### LINKED ISSUE #559: Remove slider widget from attendance count & invoice amount
Just let it accept a inputted number.
I've never used it, and actually had it cause problems when my scroll wheel activates it and changes the value.

pbanaszkiewicz:
This fixes #559 by changing default widgets to TextInput for these
fields:

* admin_fee
* attendance
* latitude
* longitude.

Unfortunately the `admin_fee` field's declaration is more important
than Meta.widgets content so the widget for admin_fee had to be
overwritten in the declaration.

Let me just add that the reason for this change is that under MacOS the HTML5 widgets can change value upon scrolling over them. This is not the case for Linux, and certainly not for Windows, but our admins use MacOSX so…

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -2,7 +2,7 @@ import re
 
 from django import forms
 from django.core.validators import RegexValidator
-from django.forms import HiddenInput, CheckboxSelectMultiple
+from django.forms import HiddenInput, CheckboxSelectMultiple, TextInput
 
 from captcha.fields import ReCaptchaField
 from crispy_forms.helper import FormHelper
@@ -258,7 +258,7 @@ class EventForm(forms.ModelForm):
     )
 
     admin_fee = forms.DecimalField(min_value=0, decimal_places=2,
-                                   required=False)
+                                   required=False, widget=TextInput)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


gvwilson:
I'll try to get to this, but we need to find someone else who can review 
these from now on - I have to focus on instructor training.

wking:
On Mon, Jan 04, 2016 at 02:20:02PM -0800, Greg Wilson wrote:
> I'll try to get to this, but we need to find someone else who can
> review these from now on - I have to focus on instructor training.

I'm fine just leaving it to @pbanaszkiewicz to push PRs like this, and
then merge them himself after a few days of quiet (if he feels happy
with them).  Especially for small changes like this, if it works for
him, it's probably fine to merge.  The only areas that *need* other
feedback are issues (where AMY users have been chiming in).  As long
as we agree on what AMY should be doing, how it goes about doing it
doesn't need consensus building unless there are more folks than Piotr
frequently interacting with the code.

pbanaszkiewicz:
> I'm fine just leaving it to @pbanaszkiewicz to push PRs like this, and then merge them himself after a few days of quiet (if he feels happy with them). Especially for small changes like this, if it works for him, it's probably fine to merge.

That's what I was doing and that's what I'm going to do for this issue.

Issue #603, opened 2015-12-06
Using underscores instead of periods in user IDs


gvwilson:
See https://github.com/swcarpentry/website/issues/90.

pbanaszkiewicz:
Okay, I looked into this and there's a huge impact: we're basically changing users' login data.

What I suggest we should do is return an underscored name in the API instead - that's the only way the website talks with AMY, right?

gvwilson:
But how many users have logins?

pbanaszkiewicz:
Well everyone has a username, and 17 people have login access.

gvwilson:
Easy to regenerate user names, and better to do it now with 17 people 
than a year from now with more.  If this cuts one custom build step out 
of the website, I think it's worth doing.

pbanaszkiewicz:
Closing via #645.

PR #642, opened 2016-01-05
API: replace dots with underscores in usernames

pbanaszkiewicz:
This is a non-violent change: it addresses the #603 issue only on the
representation layer, ie. API call results, and not in the database.

The reason for not changing the database is that we'd change users'
logins with this change - not something we want to do.


gvwilson:
:-1: (at least) - this will be a documentation and maintenance nightmare. I'd rather keep the custom build step on `website`...  If we only have 17 people with logins right now, I think it's reasonable to tell them that their user IDs have changed.

PR #660, opened 2016-01-22
Adding maintainer and trainer badges

gvwilson:
Once this is merged and deployed, @gvwilson will give maintainers and trainers their awards.

### DIFF
--- a/workshops/migrations/0070_maintainer_trainer_badge.py
+++ b/workshops/migrations/0070_maintainer_trainer_badge.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+def add_maintainer_badge(apps, schema_editor):
+    Badge = apps.get_model('workshops', 'Badge')
+    Badge.objects.create(name='maintainer', title='Maintainer', criteria='Maintainer of Software or Data Carpentry lesson')
+    Badge.objects.create(name='trainer', title='Trainer', criteria='Teaching instructor training workshops')
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
Looks okay, but I wonder if we should enable editing of badges in the special Django admin.

It's available at /workshops/admin and allows for adding/changing/removing models that carry one field (such as Tag, AcademicLevel, Role, DataAnalysisLevel, KnowledgeDomain).

gvwilson:
:+1: from me.  Do we have something that populates a brand-new database 
with standard values (e.g., all our standard badges)?  Good to have as 
docs if nothing else.

pbanaszkiewicz:
@gvwilson we have the [`fake_database.py`](https://github.com/swcarpentry/amy/blob/develop/workshops/management/commands/fake_database.py) command which is responsible for populating our development database with some sane values, but it doesn't create badges. We can add it, though.

The only badges being created are `swc-instructor` and `dc-instructor` - probably in some migration.

@gvwilson do you suggest to add a method for creating badges to the `fake_database` command?

In future I want AMY to have such default data in the fixtures instead of migrations (see #626), but I don't know if we ever get to that point…

gvwilson:
+1 to all the badges.  No opinion on default data in fixtures - dunno 
enough about Django...

pbanaszkiewicz:
@gvwilson added your suggestions. Don't worry about fixtures just yet, there's a long way before we'll have them (if ever).

gvwilson:
Thanks Piotr.  ETA on deploying the maintainer and trainer badges to the 
production database?

pbanaszkiewicz:
Feb 1st, one week from now. Let me know if you need it earlier.

gvwilson:
Feb 1 works for me.
Thx,
G

PR #671, opened 2016-01-27
Fetch people to be warned of instructor event closing

gvwilson:
Helps with managing instructor training.

### DIFF
--- a/workshops/management/commands/instructors_last_call.py
+++ b/workshops/management/commands/instructors_last_call.py
@@ -14,12 +14,16 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
 
         # Every learner who was at this event but isn't an instructor.
-        slug = options['event']
-        event = Event.objects.get(slug=slug)
-        learner = Role.objects.get(name='learner')
-        ttt = Tag.objects.get(name='TTT')
-        instructor_badges = Badge.objects.instructor_badges()
-        trainees = Task.objects.filter(event=event, role=learner).exclude(person__badges__in=instructor_badges)
+        try:
+            slug = options['event']
+            event = Event.objects.get(slug=slug)
+            learner = Role.objects.get(name='learner')
+            ttt = Tag.objects.get(name='TTT')
+            instructor_badges = Badge.objects.instructor_badges()
+            trainees = Task.objects.filter(event=event, role=learner).exclude(person__badges__in=instructor_badges)
+        except ObjectDoesNotExist as e:
+            print(str(e), file=sys.stderr)
+            sys.exit(1)
 
         # Report.
         people = list(set([(t.person.family, t.person.personal, t.person.email) for t in trainees]))
--- a/workshops/management/commands/instructors_last_call.py
+++ b/workshops/management/commands/instructors_last_call.py
@@ -20,13 +20,13 @@ class Command(BaseCommand):
             learner = Role.objects.get(name='learner')
             ttt = Tag.objects.get(name='TTT')
             instructor_badges = Badge.objects.instructor_badges()
-            trainees = Task.objects.filter(event=event, role=learner).exclude(person__badges__in=instructor_badges)
+            trainees = Person.objects.filter(task__event=event, task__role=learner).exclude(badges__in=instructor_badges).distinct()
         except ObjectDoesNotExist as e:
             self.stderr.write(str(e))
             sys.exit(1)
 
         # Report.
-        people = list(set([(t.person.family, t.person.personal, t.person.email) for t in trainees]))
-        people.sort()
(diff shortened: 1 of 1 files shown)


workshops/management/commands/instructors_last_call.py
```diff
@@ -0,0 +1,28 @@
...
+            'event', help='Event slug',
+        )
+
+    def handle(self, *args, **options):
+
+        # Every learner who was at this event but isn't an instructor.
+        slug = options['event']
+        event = Event.objects.get(slug=slug)
+        learner = Role.objects.get(name='learner')
+        ttt = Tag.objects.get(name='TTT')
```
pbanaszkiewicz: You're not catching errors that may be raised when there's no event, role or tag.

workshops/management/commands/instructors_last_call.py
```diff
@@ -0,0 +1,28 @@
...
+        instructor_badges = Badge.objects.instructor_badges()
+        trainees = Task.objects.filter(event=event, role=learner).exclude(person__badges__in=instructor_badges)
+
+        # Report.
+        people = list(set([(t.person.family, t.person.personal, t.person.email) for t in trainees]))
+        people.sort()
```
pbanaszkiewicz:
```python
        trainees = trainees.order_by('family', 'personal', 'email')
        trainees = trainees.distinct()
        people = trainees.values_list('family', 'personal', 'email')
```

A better way is to use Django's QuerySet capabilities.

workshops/management/commands/instructors_last_call.py
```diff
@@ -0,0 +1,28 @@
...
+        for (family, personal, email) in people:
+            print('{0} {1} <{2}>'.format(personal, family, email))
```
pbanaszkiewicz: Use [`self.stdout.write(…)`](https://docs.djangoproject.com/en/1.9/howto/custom-management-commands/#module-django.core.management), it's way better for testing.

pbanaszkiewicz:
3 minor things, then it's ready for merge.

workshops/management/commands/instructors_last_call.py
gvwilson: Dunno how to do this.

workshops/management/commands/instructors_last_call.py
gvwilson: Is there a self.stderr.write(...) as well?

pbanaszkiewicz: 99% sure there is.

workshops/management/commands/instructors_last_call.py
pbanaszkiewicz: @gvwilson you can replace lines 24-25 with code I provided above and you should be good (unless there's an error somewhere…)

pbanaszkiewicz:
@gvwilson merge if you're happy with the code, but personally I'd still use QuerySet capabilities for sorting and displaying values list (https://github.com/swcarpentry/amy/pull/671#discussion_r51034659).

workshops/management/commands/instructors_last_call.py
gvwilson: Doesn't work for me - `trainees` is a bunch of `Task` objects, which don't have fields `personal`, `family`, and `email` - those are in the `Person` objects. What's the missing step?

pbanaszkiewicz: Hmmm, try switching this:
```python
        trainees = Task.objects.filter(event=event, role=learner).exclude(person__badges__in=instructor_badges)
```

to this:
```python
        trainees = Person.objects.filter(task__event=event, task__role=learner).exclude(badges__in=instructor_badges)
```

You may need to add `.distinct()` at the end.

gvwilson: Done - check?

workshops/management/commands/instructors_last_call.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+
+    def handle(self, *args, **options):
+
+        # Every learner who was at this event but isn't an instructor.
+        try:
+            slug = options['event']
+            event = Event.objects.get(slug=slug)
+            learner = Role.objects.get(name='learner')
+            ttt = Tag.objects.get(name='TTT')
```
pbanaszkiewicz: You're not using this variable, maybe you wanted to make sure that the fetched event has this tag?

Anyway, I checked the code and it's working correctly, so go ahead and merge.

gvwilson: Removed TTT usage, reorganized code so that only things that might 
return 'no such object' are inside the try/catch.

PR #675, opened 2016-01-30
Add report of completion rates for instructor trainings

### LINKED ISSUE #667: Reporting completion percentages for instructor training
Please create a management command that reports:

* instructor training event slug
* total # of attendees
* # of attendees who got their instructor badge from that event
* # of attendees who have an instructor badge (from that event or any other)

for each instructor training event, so that we can see completion rates (both from the event and eventually).

pbanaszkiewicz:
The management command reports:
* training tags,
* badge used for lookups,
* number of learners,
* how many of them got a badge for that training,
* how many of them got a badge for other training,
* how many of them have no badge.

This fixes #667.

### DIFF
--- a/workshops/management/commands/trainings_completion_rates.py
+++ b/workshops/management/commands/trainings_completion_rates.py
@@ -1,0 +1,53 @@
+from django.core.management.base import BaseCommand
+from django.template.loader import get_template
+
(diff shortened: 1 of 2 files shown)


gvwilson:
Thanks,
Greg

PR #688, opened 2016-02-04
Filtered badge view

### LINKED ISSUE #685: Sort list of awarded badges by date
I would like an option to sort https://amy.software-carpentry.org/workshops/badge/swc-instructor by award date (either most recent first or last) instead of alpha by recipient, so that I can see who's been awarded something recently.

pbanaszkiewicz:
This fixes #685 by turning `badge_details` view into filtered view.

Filtered views have not only the obvious ability to filter, but they can also change ordering.

### DIFF
--- a/workshops/filters.py
+++ b/workshops/filters.py
@@ -13,6 +13,7 @@ from workshops.models import (
     EventRequest,
     Tag,
     Task,
+    Award,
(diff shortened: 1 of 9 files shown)


gvwilson:
Nice!

PR #690, opened 2016-02-04
Remove erroring DRF URLs

### LINKED ISSUE #682: No reverse match for rest_framework namespace
The error for a very strange reason shows when accessing these URLs:
https://github.com/swcarpentry/amy/blob/develop/api/urls.py#L57

I wasn't able to get rid of it; it's not being used at all, so maybe it should be removed…?

pbanaszkiewicz:
For unknown reason they threw NoReverseMatch.

This fixes #682 by removing the URLs (actually removing the inclusion of
these URLs).

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -52,10 +52,4 @@ urlpatterns = [
     url('^', include(todos_router.urls)),
 ]
 
-# for login-logout functionality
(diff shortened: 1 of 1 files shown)


gvwilson:
:+1:

PR #706, opened 2016-02-20
New persons merging

### LINKED ISSUE #673: Rework persons merging
Soon #672 will be accepted; it's a different way of doing 2-objects merge, and I like it way more.

Let's rewrite person merging in the same way as events merging works. It doesn't have to follow DRY, but it will be bonus points if we could reuse the underlying method for merging.

pbanaszkiewicz:
This incorporates the same mechanics as were introduced in events merging recently.

Fixes #673.


gvwilson:
Looks good (there's a lot of duplication in the forms, but I guess 
that's inevitable).  I won't be able to set up and run an instance 
locally for testing until I'm back in Toronto next week.

pbanaszkiewicz:
@gvwilson sure. I'll probably merge it soon and just fix if any errors should appear.

PR #711, opened 2016-02-25
Debrief: rename to "List Instructors by Time Period"

### LINKED ISSUE #704: Debrief Report Changes
1.  Rename to "List of Instructors by Time Period" - we are using this not just for debriefing sessions but also for pre-workshop mentoring sessions and possibly other reasons as well.

2.  Option to export list as csv  (event, full name, email, # of workshops taught)

pbanaszkiewicz:
This fixes #704.

I'm not the happiest with the way this worked: in order to use CSV I decided to leverage API, but I didn't want to build table with JavaScript, so the solution here uses API to generate CSV and "normal" website to generate the table.


api/views.py
```diff
@@ -41,6 +41,7 @@
     WorkshopsOverTimeSerializer,
     InstructorsOverTimeSerializer,
     InstructorNumTaughtSerializer,
+    InstructorsByTime,
```
gvwilson: So we have `InstructorsOverTimeSerializer` and `InstructorsByTime` ?

api/views.py
```diff
@@ -313,35 +314,37 @@ def instructor_num_taught(self, request, format=None):
...
-        """Workshops, instructors, and missing data in specific periods."""
-        start_default, end_default = self._default_start_end_dates()
-
-        start = self.request.query_params.get('start', None)
         if start is not None:
             try:
                 start = datetime.datetime.strptime(start, '%Y-%m-%d').date()
             except ValueError:
-                start = start_default
+                start = start_of_year
```
gvwilson: What basis is there for defaulting to the start of the year?

api/views.py
```diff
@@ -313,35 +314,37 @@ def instructor_num_taught(self, request, format=None):
...
-            start = start_default
+            start = start_of_year
 
-        end = self.request.query_params.get('end', None)
         if end is not None:
             try:
                 end = datetime.datetime.strptime(end, '%Y-%m-%d').date()
             except ValueError:
-                end = end_default
+                end = end_of_year
```
gvwilson: Same question as above - why is this a sensible default?

gvwilson:
I share your unease with the approach, but I can't think of a cleaner way off the top of my head :-(

api/views.py
pbanaszkiewicz: Yeah, it should perhaps be `InstructorsByTimePeriodSerializer`.

`InstructorsOverTimeSerializer` counts how many instructors there are.

api/views.py
pbanaszkiewicz: That's the default used for `all_activity_over_time`.

PR #715, opened 2016-03-01
Including the event in the CSV listing

gvwilson:
So that we can figure out who trained them.


pbanaszkiewicz:
Looks OK, thanks!

Issue #659, opened 2016-01-20
Include "notes" field on instructor profile update form


maneesha:
On this form:
https://amy.software-carpentry.org/workshops/update_profile/

Can we also have  "notes" field that is just for the admin's info?  I often have people completing the profile update form and then emailing me with some other notes.  If I could just see this all in one place I would not have to match up the profile update request with the email.

PR #717, opened 2016-03-02
Included "notes" field on instructor profile update form

shubhsingh594:
Fixes https://github.com/swcarpentry/amy/issues/659

in this commit i am not rewriting the old notes with the new one.
as mentioned in this issue, the new notes are for the info of the admin.

PR #718, opened 2016-03-02
 Included notes field on instructor profile update form

shubhsingh594:
Fixes https://github.com/swcarpentry/amy/issues/659


pbanaszkiewicz:
Thanks for your PR, @shubhsingh594. Unfortunately it's still missing "Notes" field on the form accessible by normal users (see https://github.com/swcarpentry/amy/pull/717#issuecomment-191087974). Please update accordingly.

Your change will break tests, so you'll have to fix them too.

shubhsingh594:
@pbanaszkiewicz  hi,

i noticed that forms.ProfileUpdateRequestFormNoCaptcha , is using ' exclude ' instead of fields. so Notes should get included automatically . i also checked and i am able to use the notes textfield, by going to this url - http://127.0.0.1:8000/workshops/update_profile/

pbanaszkiewicz:
@shubhsingh594 I was wrong, you're right; the field appears where it should be.

Please wrap the field's contents with `<pre>` tag and I'll merge this.

Thanks & sorry!

pbanaszkiewicz:
Thanks a lot for your contribution, @shubhsingh594!

Issue #705, opened 2016-02-18
Email addresses should not be case sensitive


maneesha:
For example, these pairs should have been flagged as duplicates:

https://amy.software-carpentry.org/workshops/person/8955   and
https://amy.software-carpentry.org/workshops/person/8957

or

https://amy.software-carpentry.org/workshops/person/46
https://amy.software-carpentry.org/workshops/person/47

(I am going to merge them soon -- just wanted others to see these first)

pbanaszkiewicz:
Unfortunately according to specs, the part before `@` is case-sensitive.

What about I add matching of similar emails to #624?

gvwilson:
I think we should disregard the spec in this case and do 
case-insensitive matching - it's highly unlikely that 
'Greg@wherever.com' and 'gREG@wherever.com' are different people...

narayanaditya95:
@pbanaszkiewicz @gvwilson 
My solution: Convert the part of email before the '@' to lower case in the `save()` method of the model layer. Run a migration that applies the same logic to all rows in the database.

Is this solution acceptable?

Reference: [17561](https://code.djangoproject.com/ticket/17561)

pbanaszkiewicz:
@narayanaditya95 are there any field types that support case insensitivity? Event database-dependant?

narayanaditya95:
> Event database-dependant?

Can you elaborate on this?

pbanaszkiewicz:
>> Event database-dependant?
> Can you elaborate on this?

Maybe e.g. PostgreSQL can enforce case-insensitive uniqueness on some fields? Maybe it's supported by Django / some Django app?

narayanaditya95:
I will get back to you.

narayanaditya95:
@pbanaszkiewicz 
We can enforce case-insensitive uniqueness on the email field.

```sql
CREATE UNIQUE INDEX case_insensitive_email ON workshop_person (lower(email));
```
However, this enforces case insensitivity on the entire email field. To enforce it on the substring before the last `@` character, we will have to resort to regex.
The above functional index would force Django to throw `django.db.IntegrityError` whenever we try to save a person with the same email of different case.  
**Note:** We will still have to run a migration on the table to ensure that such a unique constraint can be applied on the table, i.e. we will have to filter/merge all emails that are case-insensitive.

narayanaditya95:
@pbanaszkiewicz 

[...]

PR #719, opened 2016-03-03
Fix for issue #705

nikhilweee:
Added a view in `Reports -> List duplicates` to list down people with same emails (Issue #705 )
Also, the existing view showed all people instead of `None` if no duplicates were found. Also fixed that.
Please review.

### DIFF
--- a/workshops/templates/workshops/duplicates.html
+++ b/workshops/templates/workshops/duplicates.html
@@ -30,4 +30,21 @@
   {% else %}
   <p>None.</p>
   {% endif %}
+
+  <h3>Persons with the same emails</h3>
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
@nikhilweee thanks for your submission. Unfortunately your solution doesn't solve the problem in way Greg was suggesting.

We'd rather have a constraint on the database level, and if this isn't possible then on model level.

I'd like you to submit the fix for empty result set, though (in new PR).

Thanks!

nikhilweee:
Thanks for your review, @pbanaszkiewicz!
I'll submit the fix asap!

Issue #617, opened 2015-12-11
Record participation in hackathons


gvwilson:
1.  Start recording hackathons as events of type 'hackathon'.
2.  Record participants.
3.  Add historical information.

This will allow us to track conversion rates (how many hackathon participants become lesson authors and/or instructors).

pbanaszkiewicz:
This probably means "add a new tag 'hackaton'"?

gvwilson:
Yup.

PR #721, opened 2016-03-03
 added hackathon tags

shubhsingh594:
Fixes https://github.com/swcarpentry/amy/issues/617


pbanaszkiewicz:
@gvwilson are you satisfied with this solution?

pbanaszkiewicz:
Thanks for the fix, @shubhsingh594!

Issue #656, opened 2016-01-19
Generate certificates from AMY


gvwilson:
We are presently generating PDF certificates for instructors using a small Python script in the swcarpentry/certification repository (which in turn relies on Inkscape). Doing this by hand means that certificates are always out of step with the list of instructors in the database. We should instead:
1. Have AMY generate certificates.
2. Have it check that a certificate for a user and award already exists.

This will save me a lot of time...

pbanaszkiewicz:
So #343 is meant for attendees certificates?

gvwilson:
Yup - I want to separate generating certificates that we have already 
agreed to create (instructors of both kinds, eventually organizers as 
well) from generating certificates for attendees (still under debate).  
The former is a higher priority.

gvwilson:
I've updated bin/certificates.py in the https://github.com/swcarpentry/certification repository - please use that as a starting point for certificate generation.  (I still really dislike the need for Inkscape for SVG-to-PDF conversion, but don't have anything better.)

nikhilweee:
@gvwilson @pbanaszkiewicz : I'm willing to work on this issue.
As far as I can understand, here is what I've figured out how:

I'll add an option under the `More` tab to generate certificates.
This will then open up a view similar to `Persons` where you can select the recipients. This view will also show whether a certificate has already been generated. The data to be printed will be fetched from the badges awarded to a person, which has the `Date of award` and `Certified by` fields required in the certificate.
Once you confirm and click generate, the certificates will be generated where and available for download.

Responses awaited!
I'll start working based on your inputs.

gvwilson:
This would be great:

1. @pbanaszkiewicz, @nikhilwee has just modified the certificate 
generation program in https://github.com/swcarpentry/certification to 
use Cairo to generate PDFs rather than Inkscape, so we've got the tools 
we need.

2. If we're going to generate PDF certificates in AMY, we should also 
provide a way to view them in AMY (click on certificate link, new tab 
opens, show the PDF) so that people can see what's been generated and 
fix it if necessary.

3. Deleting a badge should automatically delete its PDF certificate.

4. When a PDF certificate is generated, we should auto-populate the form 
from the badge info, but allow that info to be edited before the PDF is 
generated - e.g., some certificates have "Person One and Person Two" as 
the instructors' names, rather than just one person's name.

How does this sound?

Cheers,
Greg

nikhilweee:
@gvwilson : 
[...]

PR #724, opened 2016-03-13
Issue #656: Generate certificates from AMY

nikhilweee:
Dear @gvwilson @pbanaszkiewicz,
As discussed in Issue #656, here is my implementation to integrate certificate generation through AMY
- The `Edit Person` page now has a form to award certificates to people for various badges. The form supports multiple instructors in the `Awarded By` field.
- Once a certificate entry has been saved, The`Edit Person` page has options to download a pdf certificate, or to delete the certificate entry.
- The certificate pdf is generated dynamically upon each request, and is not stored on the system.
- The `Person Summary` page also lists any available certificates, and has options for downloading them.

Please apply new migrations before testing.
Awaiting your responses!


gvwilson:
Thanks very much, Nikhil - this is a good start.  A couple of thoughts:

1. Certificates shouldn't be regenerated each time, but rather should be 
stored.  (We can then link to them from the main web site.)

2. If I try to generate a certificate, and one already exists for that 
person-badge combination, I should be asked to confirm overwrite.

@pbanaszkiewcz, what do you think?

nikhilweee:
Hi Greg,

The way this currently works is that the certificate details are stored in the database, and every time you intend to download a certificate, it is generated on the spot and presented for viewing or download.

>Certificates shouldn't be regenerated each time, but rather should be
stored.

I chose not to store certificates on the server for the following reasons:
* I think this is not something that needs to be accessed very often, so there aren't going to be too many requests to generate certificates, and hence we can afford that much of computation
* Not storing certificates makes managing them very easy. You only have to worry about certificate details on the database, and not about file management issues, like 'overwriting' a certificate, or cleaning up the server often

>(We can then link to them from the main web site.)

You can provide links to the certificate even now. Once you click on download, you are redirected to the certificate's url that is unique, and can be shared around easily

>If I try to generate a certificate, and one already exists for that
person-badge combination, I should be asked to confirm overwrite.

I'll add it in the next revision.

Please let me know if you have different thoughts. I'll be happy to discuss them

Thanks,
Nikhil

gvwilson:
Hm - OK, I personally like having the PDF stored under version control, 
but I can be persuaded.  @pbanaszkiewicz, your call - which approach do 
you prefer?

pbanaszkiewicz:
Hi @nikhilweee and @gvwilson,

1. I'd like to keep the request-response cycle as short as possible – so no generating on the fly, if we can;
2. I'd like to serve certificates from a different domain than AMY – to avoid potential code execution attacks.

I don't know if we need version control here, but these two points (↑) are quite important.

Question to @gvwilson: is there a possibility that one person has more than one certificate for one badge?

If not, then we don't need a new model; we can extend `Badge` to include `certificate` file field, maybe with some timestamps.

Cheers

certification/README.md
```diff
@@ -0,0 +1,2 @@
+# certification
+Software Carpentry certifications
```
pbanaszkiewicz: If this comes from our other repository, could you add a link? We may also need to think about updating this dependency (git submodules anyone? @gvwilson thoughts?)

workshops/admin.py
```diff
@@ -19,3 +20,4 @@
 admin.site.register(Lesson)
 admin.site.register(KnowledgeDomain)
 admin.site.register(Badge)
+admin.site.register(Certificate)
```
pbanaszkiewicz: Why add it to admin?

workshops/models.py
```diff
@@ -1177,6 +1177,22 @@ def __str__(self):
...
+
+    person     = models.ForeignKey(Person)
+    badge      = models.ForeignKey(Badge)
+    awarded    = models.DateField()
+    awarded_by = models.ManyToManyField(
+        Person, blank=True, related_name='certified_set')
+
+    def get_awarded_by_names(self):
+        names = [x.get_full_name() for x in self.awarded_by.all()]
+        print(names)
```
pbanaszkiewicz: You missed `print` here (but I think this model will go anyway, so…)

gvwilson:
I think one person can only have one award of a particular badge, so 
there should only be one certificate for any (person, badge) combination.

certification/README.md
gvwilson: Death to submodules!

nikhilweee:
I'm sorry I'll be busy for two days owing to my mid-semester examinations. I'll have a look at these changes ASAP after that. I hope you don't mind!

pbanaszkiewicz:
@nikhilweee that's okay, university comes first.

workshops/admin.py
nikhilweee: That was just for testing purposes, I still need to clean up and add tests, sorry about that!

certification/README.md
nikhilweee: This is a stripped down version of the script, modified to fit with the current workflow. However, I'll add the link anyway. It deserves credits!

nikhilweee:
@pbanaszkiewicz,
As @gvwilson states [here](https://github.com/swcarpentry/amy/issues/656#issuecomment-194799680)
>some certificates have "Person One and Person Two" as
the instructors' names, rather than just one person's name.

This is the reason I created a new model for certificates

>I'd like to keep the request-response cycle as short as possible – so no generating on the fly, if we can.

2-1 - Okay, I lost! Then I guess I'll generate the certificates the first time a request is made and then serve this generated certificate for subsequent requests.

>I'd like to serve certificates from a different domain than AMY – to avoid potential code execution attacks.

I'm not sure how can I work on this. I mean, a different domain means that's not under my control, right? I can only place a redirect. How can *I* sort it out completely?

rgaiacs:
My 2 pennies: From my experience, replace some string from a SVG can be problematic because text wrap isn't support by some libraries and if the user input a name longer than you expect the final result will be ugly. I prefer a LaTeX approach instead of SVG because it doesn't have the problem describe before. Using LibreOffice maybe works but I never try change something inside a ODT using Python.

certification/README.md
pbanaszkiewicz: Since this is one kind of dependency we can't automate installing/updating by normal tools, please add a note to `docs/` to mark where it comes from.

pbanaszkiewicz:
@nikhilweee:

> > some certificates have "Person One and Person Two" as  the instructors' names, rather than just one person's name.

> This is the reason I created a new model for certificates

Can you elaborate how this would help?

> I guess I'll generate the certificates the first time a request is made and then serve this generated certificate for subsequent requests.

I think it's better to write a management command that does it once outside WWW server.

> I'm not sure how can I work on this. I mean, a different domain means that's not under my control, right? I can only place a redirect. How can I sort it out completely?

Well, for one there's `STATIC_URL` (or `MEDIA_URL`) which *maybe* can be configured to use a different domain.

I can set up a domain for the static files (+certs) once we have everything in AMY working.

pbanaszkiewicz:
@rgaiacs that's very important. @nikhilweee how well does the string interpolation work with CairoSVG?

rgaiacs:
blank

Template that I used in the past.

long-name

Some learners could have a very long name and part of their name goes behind our logo.

four-instructors

For a workshop with four instructors the names of the instructors goes outside the printing area.

nikhilweee:
@pbanaszkiewicz,

>Can you elaborate how this would help?

The new model uses a `ManyToManyField` and hence can have more than one person in the `awarded by` field. The `Award` model was the closest to what I wanted, but that had a `ForeignKey`, which meant only one person can award the certificate. That's why I created a new model.

>I think it's better to write a management command that does it once outside WWW server.

Okay, so the way I get it, I think the 'generation' of certificates would be done manually, by the management command, and the link to a certificate 


>Well, for one there's STATIC_URL (or MEDIA_URL) which maybe can be configured to use a different domain.
>I can set up a domain for the static files (+certs) once we have everything in AMY working.

THIS is what I was talking about. I'm aware of the STATIC_URL and MEDIA_URL settings. Since I'm not an admin, I cannot set up the official new domain for serving static files. That has to be done by you, right?

@rgaiacs,

You are absolutely right, the overflow problem exists even with the current template, but I guess that can be worked around easily if I add a line break if the filler text is too long. I'll try this out, and hopefully that should solve it.

nikhilweee:
@rgaiacs : you're right, text wrap isn't supported in svg directly - so unfortunately what I thought doesn't work out. 
How about using different templates according to the length of the text? The other template would be used if `{{name}}` overflows.

gvwilson:
It would be nice to have a fully-general solution, but I think we have 
higher priorities - https://github.com/swcarpentry/amy/issues/656 would 
be great to get in.

pbanaszkiewicz:
@gvwilson but this is a solution to #656?

@nikhilweee I think the only reasonable way is to generate SVG and PDF (and maybe JPG/PNG thumbnail), and allow anyone responsible for generating these certificates to download SVG (so that they can fix it manually) and re-upload…

certification/README.md
nikhilweee: I'm adding the script and the required templates to amy itself, that way nothing needs to be installed separately. Should I add it to the docs?

gvwilson: Everything should be documented (though we're rather lax about doing 
that ourselves sometimes).  @pbanaszkiewicz can you review the PR when 
it comes in?

nikhilweee:
@pbanaszkiewicz : 
Okay, so to summarize the conversation so far, this is how I should work it out:
* The models structure is fine
* Certificates should be generated via a management command which should prompt before override
* Word overflow should be ignored for now
Please correct me if I am wrong.

nikhilweee:
Hey @pbanaszkiewicz @gvwilson 
I've separated certificates generation. Please have a look. A short guide is located in `docs/certificates.md`

pbanaszkiewicz:
@nikhilweee apologies — I'm really busy… I'll try to take a look during the weekend. Thanks for your contributions!

gvwilson:
I'll try to have a look on Friday - thanks!

certification/code/certificates.py
```diff
@@ -0,0 +1,70 @@
...
+'''
+
+import sys
+import os
+import re
+import tempfile
+import cairosvg
+from django.conf import settings
+
+#---------------------------------------------------------------
```
pbanaszkiewicz: This line contains one of PEP8 violations. This is not a big deal, but we try to follow PEP8 rules in AMY.

You can probably install PEP8 plugin to your editor of choice to help you validate Python syntax against PEP8.

certification/code/certificates.py
```diff
@@ -0,0 +1,70 @@
...
+
+def generate(args):
+    '''Process a single entry and returns their certificate'''
+    root_dir = os.path.dirname(__file__)
+    template_path = construct_template_path(root_dir, args.badge_type)
+    return create_certificate(template_path, args.params, args.cert_id)
+
+
+def create_certificate(template_path, params, cert_id):
+    '''Creates and returns a single certificate.'''
```
pbanaszkiewicz: PEP257:
> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

certification/code/certificates.py
```diff
@@ -0,0 +1,70 @@
...
+    with open(template_path, 'r') as reader:
+        template = reader.read()
+    check_template(template, params)
+
+    for key, value in params.items():
+        pattern = '{{' + key + '}}'
+        template = template.replace(pattern, value)
+
+    tmp = tempfile.NamedTemporaryFile(suffix='.svg', delete=False)
+    # tmp = open(str(params.id) + '.svg', 'wb')
```
pbanaszkiewicz: Please remove this code if you're not using it.

certification/code/certificates.py
```diff
@@ -0,0 +1,70 @@
...
+    tmp.write(bytes(template, 'utf-8'))
```
pbanaszkiewicz: Some folks [were wondering](http://stackoverflow.com/questions/7585435/best-way-to-convert-string-to-bytes-in-python-3) which should be used: `bytes(templates, 'utf-8')` or `template.encode('utf-8')` and generally they preferred the latter.

certification/code/certificates.py
```diff
@@ -0,0 +1,70 @@
...
+    filename = os.path.join(settings.CERTIFICATES_DIR, str(cert_id) + '.pdf')
+    cairosvg.svg2pdf(url=tmp.name, dpi=90, write_to=filename)
+    return
```
pbanaszkiewicz: You can return what `cairosvg.svg2pdf` returns.

certification/code/certificates.py
```diff
@@ -0,0 +1,70 @@
...
+
+    with open(template_path, 'r') as reader:
+        template = reader.read()
+    check_template(template, params)
+
+    for key, value in params.items():
+        pattern = '{{' + key + '}}'
+        template = template.replace(pattern, value)
+
+    tmp = tempfile.NamedTemporaryFile(suffix='.svg', delete=False)
```
pbanaszkiewicz: Actually I don't think we need a separate '.svg' file – [docs](http://cairosvg.org/documentation/#python) claim that you can just use a bytestring as a first argument.

certification/code/certificates.py
```diff
@@ -0,0 +1,70 @@
...
+def create_certificate(template_path, params, cert_id):
+    '''Creates and returns a single certificate.'''
+
+    with open(template_path, 'r') as reader:
+        template = reader.read()
+    check_template(template, params)
+
+    for key, value in params.items():
+        pattern = '{{' + key + '}}'
+        template = template.replace(pattern, value)
```
pbanaszkiewicz: If this script is going to be called from Django, then you can use Django template system.

workshops/views.py
```diff
@@ -707,6 +716,28 @@ def person_edit(request, person_id):
...
+            if certificate_form.is_valid():
+                certificate = certificate_form.save()
+
+                messages.success(
+                    request,
+                    '{person} was awarded a {badge} certificate.'.format(
+                        person=str(person),
+                        badge=certificate.badge.title,
+                    ),
+                    extra_tags='certificates',
```
pbanaszkiewicz: This is a trick used by other tabs ("Awards", "Tasks") so that when e.g. PersonAwardForm is submitted and new award is added, the page switches back to "Awards" tab.

For some reason JS doesn't work with "certificates" tab.

pbanaszkiewicz:
@nikhilweee 
> * The models structure is fine
> * Certificates should be generated via a management command which should prompt before override
> * Word overflow should be ignored for now
>
> Please correct me if I am wrong.

You are correct.

What's missing (from my point of view):

1. thumbnails
2. download option (and hence generation upon manually adding a certificate)

But let's give @gvwilson a chance to express his opinion as well.

workshops/management/commands/certisync.py
```diff
@@ -0,0 +1,67 @@
...
+            filename = os.path.join(settings.CERTIFICATES_DIR, x)
+            os.remove(filename)
+            print('Deleted ' + filename)
+
+    def generate(self, **options):
+        """Generate certificate PDFs from the database"""
+
+        from certification.code import certificates
+
+        class Arguments:
```
pbanaszkiewicz: That's a very peculiar structure to use.

Please rework certificate generation so that you don't need to pass a class object to it.

certification/code/certificates.py
nikhilweee: This script is called when you run the `python manage.py certisync` command.
WIth regards to your comments on this file, this script it is essentially a port of the certificates script found in the swcarpentry/certification repo, so I just modified it to make it work with the current workflow, but I think you're right, I'll rework on the script for better integration

nikhilweee:
@pbanaszkiewicz 
>What's missing (from my point of view):
* thumbnails
* download option (and hence generation upon manually adding a certificate)

When you create a certificate from Amy, a download button instantly appears on the person's profile page and on the certificates tab on the edit profile page. And as we agreed upon, to generate certificates run the `python manage.py certisync` command.
As far as thumbnails are concerned, I don't think we need them, because they'll all look the same in a 32x32 image!
Let me know your views on the same.

workshops/views.py
nikhilweee: I've fixed it, will show up in the next commit. Thanks!

certification/code/certificates.py
nikhilweee: In this case `cairosvg.svg2pdf` just creates a pdf from the template. I think the return statement is not needed here. Thanks

certification/code/certificates.py
nikhilweee: I'm really sorry for that!

certification/code/certificates.py
nikhilweee: Whoopsie! I'll correct that. Thanks for pointing out

pbanaszkiewicz: PEP8 and PEP257 are some general notes on a style, but they're very helpful when you're going to write more Python and contribute to different projects, because many of them may impose PEP8-syntax style.

In AMY we don't really follow all PEP257 rules (docstrings are often in triple apostrophes `'''` instead of triple quote marks `"""`; multiline doctsrings don't always have ending `"""` on a separate line), but we do try to follow the language rule (https://github.com/swcarpentry/amy/pull/724#discussion_r59117147).

certification/code/certificates.py
pbanaszkiewicz: No big deal! You don't have to apologize.

nikhilweee:
I've responded to @pbanaszkiewicz's comments. Please have a look

certification/code/certificates.py
pbanaszkiewicz: Are you sure `cairosvg.svg2pdf` doesn't return a bool value with meaning "yes I was able to convert that file to PDF" or "no, something wrong happened"?

Because if it does, you may want to send that output to whatever called the `create_certificate` function.

certification/code/certificates.py
pbanaszkiewicz: I'm tempted to make a Django application module out of `certification`. Any thoughts if that's a good or a bad change to make?

certification/code/certificates.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python
+
+'''
+Generate a certificate from a template.
+
+* Requires the python package 'cairosvg' to be installed.
```
pbanaszkiewicz: We need that in `requirements.txt` as well.

pbanaszkiewicz:
Okay a few points I want to cover here.

#### certificate generation

If we agreed to only use `certisync` command, then I think we need a cronjob to run it every 30 minutes.

We also need a feedback information to let users know that this newly added certificate will generate in up to 30 minutes.

Finally I think the `Certificate` model should have two additional fields:
1. `certificate_pdf`
2. `certificate_thumbnail`

and both of them should be populated by the `certisync`.

These fields will let us have a feedback information from certisync to the Django templates and therefore we'll be able to "enable" or "disable" download button.

#### thumbnails

They don't have to be 32×32. They can easily be higher resolution, e.g. 600px width with height scaled to that width, but I want them because it's way faster to peek at the certificate by looking at the thumbnail than it is to download the cert and open in separate PDF reader.


Additional note: the code won't run if the `certification/certificates` directory doesn't exist (it doesn't by default). Please add it to git (with a dummy file).

certification/code/certificates.py
nikhilweee: That's really a nice thought, but I think that should be avoided for now, mainly because it's deeply integrated with the workshops app. We're only generating certificates for the workshops app, plus it relies on a model which is heavily dependent on the workshops app.

certification/code/certificates.py
nikhilweee: gotcha!

certification/code/certificates.py
nikhilweee: Yes, `svg2pdf` returns `None` if the `write_to` argument is supplied. And this is what the docs say:
>If the `write_to` argument is provided (filename or file-like object), the output is written there. Otherwise, the function returns a byte string.

nikhilweee:
@pbanaszkiewicz: 
>We also need a feedback information to let users know that this newly added certificate will generate in up to 30 minutes.

Currently if the certificate is not yet generated and you click on the download button, a "Please check back later" page pops up. I'll have the page display a message to check back in 30 minutes once you set up the cronjob.

>I want them because it's way faster to peek at the certificate by looking at the thumbnail than it is to download the cert and open in separate PDF reader.

I've already taken care of that! The PDFs are served with a `Content-DIsposition: inline` header that makes the browser open the pdf inline rather than having to download it.

certification/code/certificates.py
pbanaszkiewicz: I agree, it wouldn't be a reusable app. Our `api` app is not reusable, but it is separated out from the `workshops` app.

I was thinking about doing the same for certification app.

But let's leave this for now, we'll do it later if the need arises.

workshops/views.py
```diff
@@ -1463,6 +1496,26 @@ def award_delete(request, award_id, person_id=None):
...
+    """Delete a certificate. This is used on the person edit page."""
+    certi = get_object_or_404(Certificate, pk=certificate_id)
+    badge_name = certi.badge.name
+    certi.delete()
+
+    messages.success(request, 'Certificate was deleted successfully.',
+                     extra_tags='certificates')
+
+    if person_id:
+        # if a second form of URL, then return back to person edit page
```
pbanaszkiewicz: Is the other URL used anywhere?

workshops/views.py
```diff
@@ -2680,3 +2733,31 @@ def duplicates(request):
...
+    except FileNotFoundError:
+        messages.error(
+            request,
+            'This certificate is not yet available for download. Please check back later.',
+        )
+        return render(request, 'workshops/certificate_error.html')
+
+    response = HttpResponse(binary, content_type='application/pdf')
+    response['Content-Disposition'] = 'inline; filename="{}.pdf"'\
+        .format(certificate.person.get_full_name())
```
pbanaszkiewicz: Filename should contain badge name too (can be short, ie. 'swc' / 'swc-instructor').

pbanaszkiewicz:
@nikhilweee:

> Currently if the certificate is not yet generated and you click on the download button, a "Please check back later" page pops up. I'll have the page display a message to check back in 30 minutes once you set up the cronjob.

No, the "Download" button must be greyed out if no certificate is ready for download. Additionally, upon hover or right next to said button should be infobox saying when the certificate will be available.

> I've already taken care of that! The PDFs are served with a `Content-DIsposition: inline` header that makes the browser open the pdf inline rather than having to download it.

That's an inelegant replacement for thumbnails, but leave it for now. If users start complain, we'll add thumbnails.

### TODO

Okay, let's work towards closing this issue. Here's what's necessary do to before merging:

* add `Certificate.download_ready` boolean field that's set to `True` upon certificate generation
* clean up admin
* add tests against: `certisync` command, new views
* resolve my comment about second URL form
* change downloaded file name

Thanks for your contribution, cheers!

workshops/views.py
nikhilweee: Yes, it's used in the `Person Edit` page in the `certificates` tab

nikhilweee:
@pbanaszkiewicz ,
I've worked upon your comments, please have a look.
I'll work on the tests as soon as you approve these changes.

pbanaszkiewicz:
@nikhilweee thanks!

I'm quite busy right now, but I'll try my best to provide you with feedback before weekend. Sorry for all these delays!

nikhilweee:
@pbanaszkiewicz

No problem! I'll work on the tests meanwhile.

pbanaszkiewicz:
@nikhilweee: (+1) to the changes made so far.

I don't have any comments, only one note: in multiple places you use string concatenation, while more pythonic way would be (a few examples):
```python
# is
str(10) + '.pdf'
# should be
'{}.pdf'.format(10)

# is
'Message ' + param
# should be
'Message {}'.format(param)
```

These are not crucial issues and I'm happy to merge the code with them - as soon as tests show that we didn't miss any error.

Cheers!

pbanaszkiewicz:
@nikhilweee any progress on this issue?

pbanaszkiewicz:
Let's make this essential for the v1.7.

pbanaszkiewicz:
@gvwilson how strong is your need for this feature to land in the next release?

gvwilson:
Not needed in the next release - other things are more important.

narayanaditya95:
Closed in favour of #1123

PR #726, opened 2016-03-17
Fix: 500 when submitting incomplete person match form

pbanaszkiewicz:
The report email came during the night: incomplete form submitted to find a matching person didn't work correctly (caused server 500 error).

The error was: `form` variable was unbound in the local scope (which was a result of wrong steps order in the person matching logic).


gvwilson:
:+1: to merge - what was the user action that produced the incomplete form?

pbanaszkiewicz:
User entered text "ingi" into the autocomplete field and pressed [enter]. This caused the form to not find anyone and later, in the maze of logical conditions in the code, `form` variable was never assigned (because in original code it was assigned after the statement that was causing an exception).

PR #731, opened 2016-03-18
change options to require at least two instructors per wkshp

maneesha:
We don't want anyone to think one instructor is an option (even for <20 learners)


pbanaszkiewicz:
I think this change triggers a migration when you run `./manage.py makemigrations`. Unless @maneesha adds it to this PR, I'll add it later.

maneesha:
I'll leave it to you @pbanaszkiewicz

PR #734, opened 2016-03-24
Linking to data privacy policy

gvwilson:


pbanaszkiewicz:
+1 I was thinking about this the privacy policy some time ago.

PR #740, opened 2016-03-31
Fix: event_import AttributeError and uncaught request exception

pbanaszkiewicz:
This PR fixes two bugs related to event importing:

1. missing `url` GET param when accessing `event_import` view caused 500
2. AttributeError when imported event had no instructors/helpers listed


gvwilson:
:+1:

PR #748, opened 2016-04-09
Pass url within the response from `event_import` view.

### LINKED ISSUE #746: "Update from URL" should also update URL
If the "update from URL" function is used and a new url is entered in the popup (ie, url url hadn't already been entered for that event),  then date, location, etc. updates as expected. 

The url field itself should also update.

narayanaditya95:
The event_import view function did not pass the url as a
parameter in the JSON response.
Fixes #746

@pbanaszkiewicz Is this solution acceptable?

### DIFF
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -1196,6 +1196,7 @@ def event_import(request):
 
         # normalize (parse) them
         tags = parse_tags_from_event_website(tags)
+        tags['url'] = url
 
         return JsonResponse(tags)
 


pbanaszkiewicz:
@narayanaditya95 looks good, please add a test to `test_event.TestEventImport` and ping me back.

pbanaszkiewicz:
@narayanaditya95 I checked again and we don't need to return URL back; JavaScript code in `update_from_url.js` handles updating URL correctly.

I'm closing this for now, but the bug will stay open until Maneesha decides what to do next.

Thank you for your contribution, Aditya!

PR #752, opened 2016-04-21
Cronify GitHub api

### LINKED ISSUE #562: Alert for recently changed github pages
This is an issue more for self organized workshops than traditional ones.
Sometimes a workshop's instructors change.  The organizers make the change on the github page but don't necessarily tell me.  Since I don't know I can't then make the change in instructors in AMY.

So when the workshop is over, the right people are not invited to the debriefing session.  And then long term we don't have accurate records on who taught what.

Can we have an alert in AMY for recently changed github pages?  Maybe something that runs once a week or so?

pbanaszkiewicz:
This fixes #562 by extending `Event` object to hold event's website repository data and by adding a command used by cronjob to lookup changes to events' websites.
Users see notification if changes were discovered for a workshop (with small summary of what changed). Then users can review the changes (against current version), then accept or dismiss them.

Proposed cronjob entry:

```
0  0  *  *  1 cd /home/amy/amy_site/amy && /home/amy/amy_site/venv/bin/python ./manage.py check_for_workshop_websites_updates --token=GITHUBTOKEN
```


workshops/management/commands/check_for_workshop_websites_updates.py
```diff
@@ -0,0 +1,244 @@
...
+from workshops.util import (
+    fetch_event_tags,
+    parse_tags_from_event_website,
+    WrongWorkshopURL,
+)
+
+
+def datetime_match(string):
+    """Convert string date/datetime/time to date/datetime/time."""
+    try:
```
gvwilson:
~~~
for format in self.TIME_FORMATS:
  try:
    v = datetime.datetime.strptime(string, format)
    break
  except ValueError:
    pass
~~~

or something like that?

workshops/management/commands/check_for_workshop_websites_updates.py
```diff
@@ -0,0 +1,244 @@
...
+        v = datetime.datetime.strptime(string, '%H:%M:%S.%f')
+        return v.time()
+    except ValueError:
+        pass
+
+    return string
+
+
+def datetime_decode(obj):
+    """Recursively call for each iterable, and try to decode each string."""
```
gvwilson: Why all this complexity? Can we not simply poke the page author if parsing fails?

workshops/management/commands/check_for_workshop_websites_updates.py
```diff
@@ -0,0 +1,244 @@
...
+        """Load initial data into event's repository and tag information."""
+        event.repository_last_commit_hash = branch.commit.sha
+        tags = self.get_event_tags(event.url)
+        event.repository_tags = self.serialize(tags)
+        event.tag_changes_detected = ''
+        event.tags_changed = False
+        event.save()
+
+    def handle(self, *args, **options):
+        """Run."""
```
gvwilson: I think you win the award for the shortest docstring.

workshops/templates/workshops/event.html
```diff
@@ -6,6 +6,17 @@
 {% load tags %}
 
 {% block content %}
+
+{% if event.tags_changed %}
```
gvwilson: why `tags_changed` ? This tests if *anything* has changed, doesn't it?

workshops/management/commands/check_for_workshop_websites_updates.py
pbanaszkiewicz: I'll change that.

workshops/management/commands/check_for_workshop_websites_updates.py
pbanaszkiewicz: Because we have to serialize using JSON (pickle is unsafe), and we have to (de)serialize datetime/date/time objects. That's the easiest method…

workshops/templates/workshops/event.html
pbanaszkiewicz: Yes, if there are any changes detected (e.g. event website was updated and new instructors were set, or new contact details were specified) the `event.tags_changed` will be set to `True`, and the notification in this template will show up.

gvwilson: So is it just looking for changes to tags (as the name suggests) or to 
changes to any field?

pbanaszkiewicz: To the tags as in metatags stored in index.html of the workshop, but I see your point here: there's a name conflict with `Event.tags`.

gvwilson: So call 'event.data_changed' ?  Or 'event.details.changed' ?

pbanaszkiewicz: Can `event.metatags_changed` work? We'd call *tags appearing in index.html of each workshop* metatags.

gvwilson: Why use the word "tags" at all?  a) It's confusing, since we actually 
have things called "tags", and b) the information isn't actually stored 
as tags: it's either values (in YAML) or meta elements (in the generated 
HTML).

pbanaszkiewicz: These things were called tags long before, see `find_tags_on_event_index`, `find_tags_on_event_website`, `parse_tags_from_event_website`, `validate_tags_from_event_website`, `fetch_event_tags` in `util.py`.

I don't know how else you'd these key-value pairs if not tags. I like metatags since they provide structured context (metadata?) about events.

Or maybe metadata would be fine?

gvwilson: Let's go with 'tags' for now, and file an issue to rename everything to 
'metadata' (might be a good first task for Chris).  I'm over-thinking 
this...

pbanaszkiewicz: Okay. Do you have any other objections? Otherwise this is ready to merge.

gvwilson: Go for it.

PR #753, opened 2016-04-24
Foreign Keys in revisions pages for object_diff

narayanaditya95:
Adds support for showing changes in FKs on Tags, Membership, Airports and Hosts revision pages.


pbanaszkiewicz:
Good work, @narayanaditya95, but I have a few comments.

1. `relation_diff` can't be simply plugged into `event_diff.html` and other pre-made diff templates
2. please add comments to `relation_diff` - I spent 30 minutes trying to understand what it does
3. `left` and `right` correspond to previous/current version of the object. But what happens when you reach way back and compare object that was created one year ago? It may happen that PK values saved in these versions correspond to object that no longer exist.

Cheers,
Piotr

narayanaditya95:
> relation_diff can't be simply plugged into event_diff.html and other pre-made diff templates
Yes, I have noticed that. We do not pass the `field` array to the `event_diff` and `person_diff` templates and so we do not have the ability to plug in `relation_diff`.
I suggest that we gradually shift to using `object_diff` for all models. The `relation_diff` only support ForeignKeys for now and does not support ManyToMany fields. When `relation_diff` can handle m2m relations too, we make the transition. Are there concerns with the use of `object_diff` for Event and Person Models.

>please add comments to relation_diff - I spent 30 minutes trying to understand what it does
Apologies. I should have. Doing it.

pbanaszkiewicz:
> Yes, I have noticed that. We do not pass the `field` array to the `event_diff` and `person_diff` templates and so we do not have the ability to plug in `relation_diff`.
> 
> I suggest that we gradually shift to using `object_diff` for all models. The `relation_diff` only support ForeignKeys for now and does not support ManyToMany fields. When `relation_diff` can handle m2m relations too, we make the transition. Are there concerns with the use of `object_diff` for Event and Person models?

Originally there was only `object_diff.html`, but, due to the issue with handling relationship fields, it was replaced with `event_diff.html` for events and `person_diff.html` for persons.

If you can make `object_diff.html` work with events and persons, then we can drop `event_diff.html` and `person_diff.html`. Right now `object_diff.html` doesn't work with sample event ("Host matching query does not exist." in `relation_diff` for left relation).

I think that sorting out issues in `relation_diff` would help us drop `{event,person}_diff.html`.

> Apologies. I should have. Doing it.

No problems, that's what feedback is for.

> I failed to account for this possibility. How do we approach this? If the related model has revisions in it, we can pull out the deleted object and hyperlink the label to that particular revision. Otherwise, we simply neglect it.
I think we should also add hyperlinks to the labels. The name doesn't give enough information about the related object.

Since this part of AMY is rarely visited, we probably don't have to pull non-existent objects from django-reversion.

I'd stick to this procedure:
1. grab PK
2. object(pk=PK) exists? Excellent, let's use it
3. object(pk=PK) not found? Then let's only show its PK, not string representation

+1 for links.

narayanaditya95:
>I'd stick to this procedure:
1. grab PK
2. object(pk=PK) exists? Excellent, let's use it
3. object(pk=PK) not found? Then let's only show its PK, not string representation

This sounds the optimal way without much effort..

> +1 for links.

Not all models have an absolute URL for each object, like the `Role` model.
What do we link in such cases?

pbanaszkiewicz:
> Not all models have an absolute URL for each object, like the `Role` model.
What do we link in such cases?

Use the string representation.

narayanaditya95:
@pbanaszkiewicz 
I have replaced the `{person/event}_diff` templates by the generic `object_diff` template.

Few concerns:
- The person revision pages show redundant fields like `password`. Should they be explicitly filtered out (like we do for the `id` field).
- I should add tests for each model under revision, right?

workshops/templatetags/diff.py
```diff
@@ -9,3 +10,52 @@
 @register.simple_tag
 def semantic_diff(left, right, field):
     return mark_safe(generate_patch_html(left, right, field, cleanup='semantic'))
+
+
+@register.simple_tag
+def relation_diff(left, right, field):
+    if field.many_to_one or field.one_to_one:
+        # left.field_dict[field.name] is an integer or does not exist
```
narayanaditya95: This is too verbose. :/

workshops/templatetags/diff.py
```diff
@@ -9,3 +10,52 @@
...
+            left_PKs = []
+        if right.field_dict.get(field.name):
+            right_PKs = [right.field_dict.get(field.name)]
+        else:
+            right_PKs = []
+    else:
+        left_PKs = left.field_dict.get(field.name,[])
+        right_PKs = right.field_dict.get(field.name,[])
+    # Relations that exist only in the current version
+    additions = field.related_model.objects.filter(
```
narayanaditya95: Are there better ways to achieve the same filter or this is good?

narayanaditya95: Bug: Does not show objects that have been deleted.
I'm working on a fix for this.

pbanaszkiewicz:
> Few concerns:
> - The person revision pages show redundant fields like `password`. Should they be explicitly filtered out (like we do for the `id` field).

I like that these fields are shown, because often we recorded a single change to one of fields not showed, and the revision view would look like no changes took effect.

> - I should add tests for each model under revision, right?

Tests are good. I'd not test `relation_diff` against all models, only one that contains M2M and many-to-one/one-to-one/one-to-many fields.

narayanaditya95:
@pbanaszkiewicz 
The PR is up for final check and subsequent merge.

I have refactored the `TestRevisions` class at `test_diff.py` so that it has different tests that may/may not fail individually. Also, since we make no distinction between `Person` and `Event` models now, having duplicated tests is a redundancy.
The current tests test FK on the field `host` and M2M on the field `tags`.

Please review.

workshops/test/test_diff.py
```diff
@@ -10,60 +10,72 @@ class TestRevisions(TestBase):
...
+            ),
+            html=True
+        )
+        # Green label for assigned tag
+        self.assertContains(rv,
+            '<a class="label label-success" href="#">+{}</a>'.format(
+                self.tag
+            ),
+            html=True
+        )
```
pbanaszkiewicz: What about grey label for something that didn't change?

pbanaszkiewicz:
@narayanaditya95 one minor thing and this is ready for merge.

narayanaditya95:
Please re-evaluate.

narayanaditya95:
@pbanaszkiewicz 
On the first revision of an object (the time it is created), the object is compared to itself.
`views.py L1914`
```python
    except IndexError:
        # first revision for an object
        previous_version = current_version
        obj_prev = obj
```
Hence the revision page does not indicate any addition, it shows that there was no addition/deletion for that partcular revision.
Maybe we should add support for `None` objects in `previous_object`. What do you think about this?

pbanaszkiewicz:
@narayanaditya95 sorry about the delay :(

Good catch, I think `previous_version` should be `None` and in the template we should mark revision for a newly created object, ie. by displaying "This is a new object, it doesn't have any previous revisions" or similar.

narayanaditya95:
`generate_patch_html` strictly requires two revision versions to be compared.

I tried creating an empty `Version` object for the same object but `generate_patch_html` always complains that that `field_dict` attribute be populated for a patch to be generated.

I ended up simply mentioning the fact that it is the first revision for an object in the `object_diff` template.

workshops/templates/workshops/object_diff.html
```diff
@@ -8,6 +8,10 @@
   <a href="{{ object.get_absolute_url }}" class="btn btn-primary">View newest</a>
 </p>
 <p>Changed on {{ revision.date_created|date:'M j, P' }} by {{ revision.user|default:'Unknown user' }}.</p>
+{% if previous_version == current_version %}
+<p>This is a new object, it doesn't have any previous revisions.</p>
+{% endif %}
+<p>
```
pbanaszkiewicz: Is `<p>` necessary here?

narayanaditya95: Thank you for pointing that out.

pbanaszkiewicz:
Thanks a lot for your contribution, @narayanaditya95!

PR #756, opened 2016-04-28
Allow users without `add_todoitem` permission to view an event

### LINKED ISSUE #754: Permission `add_todoitem` required to view event
People without ability of adding TODO items should be able to view the events.

narayanaditya95:
This fixes #754.

### DIFF
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -937,7 +937,6 @@ def all_events(request):
 
 
 @login_required
-@permission_required('workshops.add_todoitem', raise_exception=True)
 def event_details(request, event_ident):
     '''List details of a particular event.'''
     try:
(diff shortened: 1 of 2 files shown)


workshops/views.py
```diff
@@ -937,7 +937,6 @@ def all_events(request):
 
 
 @login_required
-@permission_required('workshops.add_todoitem', raise_exception=True)
```
narayanaditya95: I think, POST requests that are intentionally generated to simulate form submissions would now end up creating TODOs.
Is that right?

pbanaszkiewicz: Yes. It's important to only process the `SimpleTodoForm` when user has `workshops.add_todoitem` permission.

narayanaditya95: `if request.method == "POST" and request.user.has_perm('workshops.add_todoitem'):`
Is this fine (this returns the event details page) or should we return HTTP403 when the user doesn't have the permission?

pbanaszkiewicz: This is good for now (but please mark with a comment why we check perms in the middle of the view). If we have more forms in that view in future we'll change that conditional.

narayanaditya95:
@pbanaszkiewicz Ready to be merged.

pbanaszkiewicz:
Thanks, @narayanaditya95!

PR #762, opened 2016-05-05
Display SWC/DC instructor status in names

### LINKED ISSUE #759: Always display badges with people's names
Always display icons for SWC/DC instructor badges alongside people's names, so that (for example) when we're viewing the list of people who participated in an instructor training workshop, we can see which ones are now qualified and now.

From a conversation with @tracykteal

gvwilson:
Closes #759.

1.  Modified `BadgeQuerySet.instructor_badges` to take extra positional argument(s) specifying which badges to interrogate. (Default behavior unchanged.)
2.  Modifying display of all persons to distinguish Software Carpentry and Data Carpentry instructors.
3.  Modifying display of tasks associated with Event to distinguish Software Carpentry and Data Carpentry instructors.

This is getting pretty hacky:

1.  Collecting all SWC and DC instructors, then doing an `in` test, seems inefficient.
2.  Manually specifying the `'swc-instructor'` and `'dc-instructor'` labels seems clumsy too - we should add fields to the `Badge` class to identify badges that are instructor-related and filter on that.

@chrismedrela, can you think of ways to improve this?

### DIFF
--- a/workshops/templates/workshops/all_persons.html
+++ b/workshops/templates/workshops/all_persons.html
@@ -2,6 +2,7 @@
 
 {% load crispy_forms_tags %}
 {% load pagination %}
+{% load tags %}
 
 {% block sidebar %}
     <h3>Filter</h3>
@@ -25,8 +26,8 @@
       </tr>
     {% for person in all_persons %}
       <tr>
-        <td class="text-center">{% if person.pk in dc_instructors %}Ⓓ{% endif %}</td>
-        <td class="text-center">{% if person.pk in swc_instructors %}Ⓢ{% endif %}</td>
+        <td class="text-center">{% if person.pk in dc_instructors %}{% bootstrap_tag "DC" %}{% endif %}</td>
+        <td class="text-center">{% if person.pk in swc_instructors %}{% bootstrap_tag "SWC" %}{% endif %}</td>
         <td><a href="{% url 'person_details' person.id %}">{{ person.get_full_name }}</a></td>
         <td>{% if person.email %}<a href="mailto:{{ person.email }}">{{ person.email }}</a>{% else %}—{% endif %}</td>
         <td>
--- a/workshops/templates/workshops/all_persons.html
+++ b/workshops/templates/workshops/all_persons.html
@@ -17,14 +17,16 @@
 {% if all_persons %}
     <table class="table table-striped">
       <tr>
-        <th width="50">Instructor?</th>
+        <th width="25">DC?</th>
+        <th width="25">SWC?</th>
         <th>Name</th>
         <th>Email</th>
         <th class="additional-links"></th>
       </tr>
     {% for person in all_persons %}
       <tr>
-        <td class="text-center">{% if person.pk in instructors %}✓{% endif %}</td>
+        <td class="text-center">{% if person.pk in dc_instructors %}Ⓓ{% endif %}</td>
+        <td class="text-center">{% if person.pk in swc_instructors %}Ⓢ{% endif %}</td>
         <td><a href="{% url 'person_details' person.id %}">{{ person.get_full_name }}</a></td>
         <td>{% if person.email %}<a href="mailto:{{ person.email }}">{{ person.email }}</a>{% else %}—{% endif %}</td>
         <td>
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -1159,8 +1159,14 @@ class BadgeQuerySet(models.query.QuerySet):
 
     INSTRUCTOR_BADGES = ('dc-instructor', 'swc-instructor')
 
-    def instructor_badges(self):
-        """Filter for instructor badges only."""
+    def instructor_badges(self, *extras):
+        """Filter for instructor badges."""
+
+        # Caller has specified which badge(s).
+        if extras:
+            return self.filter(name__in=extras)
+
+        # Caller hasn't specified, so look for instructor badges.
         return self.filter(name__in=self.INSTRUCTOR_BADGES)
 
 
--- a/workshops/templates/workshops/event.html
+++ b/workshops/templates/workshops/event.html
@@ -135,8 +135,8 @@
   </tr>
   {% for t in tasks %}
   <tr>
-    <td class="text-center">{% if t.person.pk in dc_instructors %}Ⓓ{% endif %}</td>
-    <td class="text-center">{% if t.person.pk in swc_instructors %}Ⓢ{% endif %}</td>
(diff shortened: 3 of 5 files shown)


workshops/templates/workshops/all_persons.html
```diff
@@ -17,14 +17,16 @@
...
+        <th width="25">SWC?</th>
         <th>Name</th>
         <th>Email</th>
         <th class="additional-links"></th>
       </tr>
     {% for person in all_persons %}
       <tr>
-        <td class="text-center">{% if person.pk in instructors %}✓{% endif %}</td>
+        <td class="text-center">{% if person.pk in dc_instructors %}Ⓓ{% endif %}</td>
+        <td class="text-center">{% if person.pk in swc_instructors %}Ⓢ{% endif %}</td>
```
pbanaszkiewicz: I'd rather use 1 column and `bootstrap_tag` Django template tag defined in `workshops/templatetags/tags.py` which makes these nice labels:
screenshot from 2016-05-05 07 47 38

workshops/templates/workshops/all_persons.html
```diff
@@ -17,14 +18,16 @@
 {% if all_persons %}
     <table class="table table-striped">
       <tr>
-        <th width="50">Instructor?</th>
+        <th width="25">DC?</th>
+        <th width="25">SWC?</th>
```
pbanaszkiewicz: You can make it one column "Instructor badges".

workshops/templates/workshops/all_persons.html
```diff
@@ -17,14 +18,16 @@
...
         <th>Name</th>
         <th>Email</th>
         <th class="additional-links"></th>
       </tr>
     {% for person in all_persons %}
       <tr>
-        <td class="text-center">{% if person.pk in instructors %}✓{% endif %}</td>
+        <td class="text-center">{% if person.pk in dc_instructors %}{% bootstrap_tag "DC" %}{% endif %}</td>
+        <td class="text-center">{% if person.pk in swc_instructors %}{% bootstrap_tag "SWC" %}{% endif %}</td>
```
pbanaszkiewicz: I thought about using `person.badges.instructor_badges()` but this yields additional query for every person in the list…

pbanaszkiewicz:
You can consider merging two columns into one, but otherwise +1 from me. @chrismedrela ?

workshops/templates/workshops/all_persons.html
gvwilson: I find two columns a lot easier to read. What I'd really like to do is:

1.  Mark particular badges as "instructor-flavored".
2.  Loop over that set to auto-construct the columns for things like this.

but that can wait...

gvwilson:
@chrismedrela please merge if you're happy with this.

chrismedrela:
@gvwilson, what about a bit different approach -- adding `is_swc_instructor` and `is_dc_instructor` properties to `Person` model?

And one minor issue: do you want two columns (`dc?` and `swc?` instead of `instructor badge`) in `Find Instructors` view?

gvwilson:
On 2016-05-08 9:22 AM, chrismedrela wrote:

> @gvwilson <https://github.com/gvwilson>, what about a bit different 
> approach -- adding |is_swc_instructor| and |is_dc_instructor| 
> properties to |Person| model?
>

and then is_library_carpentry_instructor? 
is_digital_humanities_carpentry_instructor?

> And one minor issue: do you want two columns (|dc?| and |swc?| instead 
> of |instructor badge|) in |Find Instructors| view?
>
Ah, yes - good catch.  I'll amend the PR.

-- 
Dr Greg Wilson
Director of Instructor Training
Software Carpentry Foundation

chrismedrela:
Ok, so what about `has_badge(name_of_badge)` method? We cannot pass parameters in django templates, but we could hack it with a custom filter:
```
{% if t.person|has_badge:"dc" %}{% bootstrap_tag "DC" %}{% endif %}
```
And this solution doesn't require any changes in the view code, especially adding `swc_instructors` and `dc_instructors` to the context dictionary.

gvwilson:
I like it - @pbanaszkiewicz ?

pbanaszkiewicz:
I don't like this approach as it would trigger additional 2N queries…

chrismedrela:
@pbanaszkiewicz it wouldn't trigger additional queries if we `prefetch_related` badges

pbanaszkiewicz:
> @pbanaszkiewicz it wouldn't trigger additional queries if we `prefetch_related` badges

Not sure how you want to implement `has_badge`, but I tried some very straightforward implementation and it caused additional queries (with `Person.badges` prefetched).

chrismedrela:
@pbanaszkiewicz, yes, you're right. Unfortunately, I don't see any solution that: 1) generates O(1) queries and, at the same time, 2) doesn't require manually specifying `swc/dc_instructor` labels.

@gvwilson, what about:
- reverting back to the previous version of `instructor_badges`, because `.instructor_badges('swc-instructor')` can be replaced by `.filter(name='swc-instructor')`
- annotating people with  `is_dc_instructor` and `is_swc_instructor` fields; example query (not the simplest query, but it's only one query and IMO it involves less hackery):
```python
Person.objects.defer('notes') # notes are too large
              .annotate(is_swc_instructor=Sum(Case(When(badges__name='swc-instructor', then=1), default=0, output_field=IntegerField())),
                        is_dc_instructor=Sum(Case(When(badges__name='dc-instructor', then=1), default=0, output_field=IntegerField())))
```
- using ```{% if p.is_dc_instructor %}{% bootstrap_tag "DC" %}{% endif %}``` inside templates. IMO It's a bit easier to understand what `p.is_dc_instructor` means rather than `p.pk in dc_instructors`.
- removing `swc/dc_instructors` from context dictionaries

Here is the implementation: https://github.com/chrismedrela/amy/commit/93b4b519e080589966b4e0fc00e1e5801246af61

pbanaszkiewicz:
@chrismedrela, @gvwilson:

1. +1 to reverting to old `instructor_badges` (we talked about this with Chris the other day)
2. annotation: @chrismedrela if you're sure this doesn't cause additional queries then +1, otherwise let's stick to `person in dc_instructors` format.

@gvwilson will you update?

chrismedrela:
@pbanaszkiewicz Yes, it requires O(1) queries, however the query I proposed results in subqueries.

I've updated this PR -> #777 (I don't know how you can amend this PR without creating a new one) .

pbanaszkiewicz:
You cannot amend PR unless you're the author, I'm afraid. You can merge with additional commits if you do it by hand, though.

I'm closing this one in favor of #777.

PR #768, opened 2016-05-07
Fix: wrong variable name for DC workshop request form

### LINKED ISSUE #760: Change logo on Data Carpentry workshop request form
On the Data Carpentry workshop request form, there's the Software Carpentry logo. Can we change that to the Data Carpentry logo?

It's the DC1_logo from here https://github.com/datacarpentry/logos/

pbanaszkiewicz:
The variable for storing template name was named 'form_template'
but it should have been 'template_name'.
Now the correct template is served.

This fixes #760.

### DIFF
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -2071,7 +2071,7 @@ class SWCEventRequestConfirm(TemplateView):
 class DCEventRequest(SWCEventRequest):
     form_class = DCEventRequestForm
     page_title = 'Request a Data Carpentry Workshop'
-    form_template = 'forms/workshop_dc_request.html'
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
@tracykteal I fixed the error, but we still use a different logo than the one you pointed to. We use https://github.com/datacarpentry/logos/blob/master/DC2_logo.png without the bottom line text, but you pointed to https://github.com/datacarpentry/logos/blob/master/DC1_logo.png — do you want me to switch the logo, too?

tracykteal:
Thanks a lot. That DC2 logo without the bottom line text is fine.

PR #770, opened 2016-05-07
Fix: remove duplicated names from AdminLookup query

### LINKED ISSUE #764: Names show up multiple times in assignment pulldown
1. Go to an event.
2. Try to assign to assign to someone other than yourself.
3. Selection dialog with pulldown appears so that you can choose person.
4. Some names (currently Greg Wilson and Tracy Teal, possibly others) show up multiple times in that list.

pbanaszkiewicz:
Query was matching either superusers or people from administrators
group, but if someone was in both categories they ended up twice on
the list.

Switching query to distinct mode fixed this error.

This fixes #764.


gvwilson:
:+1:

PR #771, opened 2016-05-08
Fix: example of website URL for update from URL form

### LINKED ISSUE #757: Update from URL should use repo OR site
> Use the URL of the event's website, e.g. http://swcarpentry.github.io/workshop-template/, and not the URL of its GitHub repository.

Often the site is not hosted at username.githib.io (for many different reasons).   It would be easier if import from repo was an option.

pbanaszkiewicz:
This was confusing to admins as if they could not use other URLs
than `username.github.io/workshop`.

This fixes #757.


pbanaszkiewicz:
@gvwilson, @maneesha: Please let me know if you want some other solution here.

gvwilson:
If @maneesha is happy, I'm happy.

maneesha:
ok by me

PR #772, opened 2016-05-08
Improved instructors' emails list generation in instructor_by_date view

### LINKED ISSUE #766: Bug in email list generation
In workshops/templates/workshops/instructors_by_date.html:25, we may wind up with a dangling comma if the last person doesn't have an email address. And we should test that we're allowed to contact them. See #765 for a better solution.

chrismedrela:
Two bugs were fixed:
    
- There might be dangling commas if we don't know email of some
  instructors, that is "john@example.com,,,bob@example.com," could be
  generated. See #766 issue.
    
- Instructors who don't want to be contacted (that is, may_contact
  is False) were included in the list.


gvwilson:
Needs more descriptive title (people aren't going to know what #766 or 
#772 are as they read it).

chrismedrela:
@gvwilson, done.

gvwilson:
:+1: @pbanaszkiewicz can you please merge?

pbanaszkiewicz:
Thanks, @chrismedrela!

PR #774, opened 2016-05-09
Delete broken iata link in airport_details view

chrismedrela:
For example, when airport iata is CVG, it generated an invalid link to `http://cvg`. This patch makes the airport iata a static text instead of a hyperlink.


pbanaszkiewicz:
Thanks! I never noticed this broken link.

PR #775, opened 2016-05-10
Add trainings dashboard

chrismedrela:
The training dashboard is available in the nav bar: more -> Trainings. The dashboard displays list of recent instructor training along with number of all participants and number of participants who has become SWC/DC instructors.

### DIFF
--- a/workshops/management/commands/fake_database.py
+++ b/workshops/management/commands/fake_database.py
@@ -1,4 +1,4 @@
-from datetime import timedelta
+from datetime import timedelta, datetime
 import itertools
 import random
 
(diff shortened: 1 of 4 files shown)


gvwilson:
@pbanaszkiewicz do you want commits squashed?

pbanaszkiewicz:
@gvwilson, @chrismedrela:

On one hand with these commits I can see how @chrismedrela was thinking, discovering existing code and changing his mind.

On the other hand, last year @wking decided with my code that I should rather incorporate increasing changes (ie. without "changing mind and reverting" part) and I think @chrismedrela should do the same for two reasons:

1. to get to know `git rebase -i` and `git reset` better
2. to follow the "increasing changes" strategy in future PRs

See https://twitter.com/pbanaszkiewicz/status/603604700290748417 from that memorable rebase I was doing :-)

So @chrismedrela should probably end up with 2-3 commits:

1. new `all_trainings` view
2. corresponding template
3. changes to the fake database generation script.

Cheers

chrismedrela:
@pbanaszkiewicz, it's much easier to reset branch to the state before any changes and, then, gradually commit changes. This is actually how I did it.

pbanaszkiewicz:
@chrismedrela if you don't want a challenge then sure, go with the soft reset :P

PR #776, opened 2016-05-12
Extend "Find Instructors" to "Find Workshop Staff"

### LINKED ISSUE #773: Find instructors function only finding badged instructors?
In the More -> Find Instructors function, it seems like it's only finding people who are certified. That default makes sense, since instructors require certification. 

Sometimes we want to find all associated people in a given region though, to potentially contact as helpers or let them know a workshop is going on. Is there a way to search for all People associated with a given airport code?

pbanaszkiewicz:
This fixes #773 by adding option to search for people who aren't instructors but perhaps have been helpers or organizers of some workshops.

This PR also reorganizes the search form and fixes some tests.


pbanaszkiewicz:
@tracykteal I gave up on adding "person is a in-progress trainee" filter to the form, but if you're certain you need it please let me know.

pbanaszkiewicz:
WIP: rework conflicts + add "person is an in-progress trainee" filter.

pbanaszkiewicz:
@gvwilson @chrismedrela please review, thanks!

pbanaszkiewicz:
@gvwilson @chrismedrela please note even though I complained about `Person.objects.filter(task__[...], task__[...])` code in the email, I still used it here because number of false-positives is (I hope) low.

workshops/forms.py
```diff
@@ -88,17 +88,8 @@ class BootstrapHelperFormsetInline(BootstrapHelper):
...
     '''Represent instructor matching form.'''
-
-    latitude = forms.FloatField(label='Latitude',
-                                min_value=-90.0,
-                                max_value=90.0,
-                                required=False)
-    longitude = forms.FloatField(label='Longitude',
-                                 min_value=-180.0,
-                                 max_value=180.0,
-                                 required=False)
```
gvwilson: why remove this? we may not have the airport of the host site in our DB.

workshops/forms.py
```diff
@@ -127,13 +118,21 @@ class InstructorsForm(forms.Form):
...
+        required=False, label='Is an in-progress instructor trainee')
+
     def __init__(self, *args, **kwargs):
         '''Build form layout dynamically.'''
-        super(InstructorsForm, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
 
         # dynamically build choices for country field
-        only = Airport.objects.distinct().values_list('country', flat=True)
-        only = [c for c in only if c]
```
gvwilson: Can't you filter empty values in the DB query?

pbanaszkiewicz: That's the reason why I removed these lines :-)

workshops/forms.py
```diff
@@ -146,49 +145,33 @@ def __init__(self, *args, **kwargs):
...
-        cleaned_data = super(InstructorsForm, self).clean()
+        cleaned_data = super().clean()
         airport = cleaned_data.get('airport')
-        lat = cleaned_data.get('latitude')
-        long = cleaned_data.get('longitude')
         country = cleaned_data.get('country')
 
-        sum = bool(airport) + bool(lat and long) + bool(country)
-        # user can specify only one: either airport, or lat&long, or country
+        sum = bool(airport) + bool(country)
```
gvwilson: This logic is pretty hard to follow (and I use the word "logic" loosely).

workshops/forms.py
pbanaszkiewicz: It was suggested by Tracy (and I feel like she's right) that these fields aren't used. I can understand that it's cumbersome to, for example, find a place on Google Maps, get the coordinates and enter them in this form.

workshops/templates/workshops/workshop_staff.html
```diff
@@ -13,21 +13,25 @@
   {% if persons %}
     <table class="table table-striped">
       <tr>
-	<th>DC?</th>
-	<th>SWC?</th>
+        <th>instr. badges</th>
```
gvwilson: I still prefer separate columns, one per badge, so that they line up.

workshops/forms.py
gvwilson: Then let's provide an easier way for them to enter a location - 
seriously, if I need to search for instructors near Moscow, what am I 
supposed to do right now?

workshops/templates/workshops/workshop_staff.html
pbanaszkiewicz: I can understand that, Greg, but I had to switch to other way of checking badges since it was causing Django to count wrong the "taught" column.

workshops/forms.py
pbanaszkiewicz: If there are instructors near Moscow then they must have an airport associated with them (ie. one of Moscow airports), therefore you can search by that airport.

Ref #271

workshops/templates/workshops/workshop_staff.html
gvwilson: If Maneesha and Giacomo are happy, I'm happy.

workshops/forms.py
gvwilson: That's not the use case I'm thinking of.  Suppose the host institution 
is in Moscow, but we don't have any instructors there. How do I find 
instructors *near* Moscow? The airport probably isn't in our database, 
hence not an option in the pulldown.

chrismedrela:
@pbanaszkiewicz, I'll review tomorrow.

chrismedrela:
@pbanaszkiewicz, I'm really sorry for that, but I cannot make the review today. This is the first thing that I'll do tomorrow.

workshops/forms.py
pbanaszkiewicz: Is that really that popular scenario?

@maneesha any strong feelings about leaving/removing lat/long search from "Find instructors" form (now "Find Workshop Staff" form)?

workshops/forms.py
pbanaszkiewicz: Fixed to proper "logic" (ie. XOR on 2 booleans) in a rebase.

pbanaszkiewicz:
@chrismedrela it's weekend, you're not supposed to be working during weekends, so…

workshops/forms.py
gvwilson: Hi Piotr,

It isn't a frequent scenario, but it is one that we have no other way to 
handle.

Thanks,

Greg

workshops/forms.py
```diff
@@ -146,49 +145,32 @@ def __init__(self, *args, **kwargs):
...
 
-        sum = bool(airport) + bool(lat and long) + bool(country)
-        # user can specify only one: either airport, or lat&long, or country
-        if sum != 1:
-            raise forms.ValidationError('Must specify an airport, or latitude'
-                                        ' and longitude, or a country.')
+        # user can specify only one: either airport or country
+        if not (bool(airport) ^ bool(country)):
+            raise forms.ValidationError('Must specify an airport OR a '
+                                        'country.')
```
chrismedrela: User can specify only one *or none*, i.e. searching for all SWC instructors is a valid query.

workshops/templates/workshops/workshop_staff.html
```diff
@@ -13,21 +13,25 @@
...
       </tr>
     {% for p in persons %}
       <tr>
-	<td class="text-center">{% if p.is_dc_instructor %}{% bootstrap_tag "DC" %}{% endif %}</td>
-	<td class="text-center">{% if p.is_swc_instructor %}{% bootstrap_tag "SWC" %}{% endif %}</td>
+        <td>{% for badge in p.badges.all %}
+        {% if badge in instructor_badges %}
+          {% bootstrap_tag badge.name|cut:"-instructor"|upper %}
+        {% endif %}
+        {% endfor %}</td>
```
chrismedrela: Indentation.

workshops/views.py
```diff
@@ -1628,41 +1628,52 @@ def all_trainings(request):
...
         num_taught=Count(
             Case(
-                When(
-                    task__role__name='instructor',
-                    then=Value(1)
-                ),
+                When(task__role__name='instructor', then=Value(1)),
+                output_field=IntegerField()
+            )
+        ),
```
chrismedrela: @gvwilson, what about an instructor who taught at one SWC workshop and two TTT? Do we want to display `1`, `2` or `3`?

workshops/views.py
```diff
@@ -1628,41 +1628,52 @@ def all_trainings(request):
...
+    people = Person.objects.filter(airport__isnull=False) \
+                           .select_related('airport') \
+                           .prefetch_related('badges', 'lessons')
+
+    trainees = Task.objects.filter(event__tags=TTT) \
+                           .filter(role__name='learner') \
+                           .filter(person__airport__isnull=False) \
+                           .exclude(event__tags=stalled) \
+                           .exclude(person__badges__in=instructor_badges) \
+                           .values_list('person__pk', flat=True)
```
chrismedrela: I guess that `trainees` should include a person who is not certified instructor, attended one stalled TTT event and one non-stalled TTT event? Could you add a test for that?

workshops/views.py
```diff
@@ -1672,48 +1683,57 @@ def instructors(request):
...
+            if data['was_organizer']:
+                people = people.filter(num_organizer__gte=1)
+
+            if data['is_in_progress_trainee']:
+                people = (
+                    people.filter(task__event__tags=TTT,
+                                  task__role__name='learner')
+                    .exclude(task__event__tags=stalled)
+                    .exclude(badges__in=instructor_badges)
+                )
```
chrismedrela: This excludes a person who is not a certified instructor, attended one stalled TTT event and one non-stalled TTT event.

chrismedrela:
@pbanaszkiewicz, looks good for me except for the minor issues that I raised.

gvwilson:
Good review - thanks.

workshops/templates/workshops/workshop_staff.html
pbanaszkiewicz: There's nothing wrong with this indentation :-) the removed code used wrong indentation (tabs).

workshops/forms.py
pbanaszkiewicz: That was a requirement inherited from previous code, but I guess we could remove that restriction.

workshops/views.py
pbanaszkiewicz: A person who is not a certified instructor, who attended one stalled TTT event and one not stalled TTT event will have two `Task` objects. While we get rid of the first one `Task(person=person, event=stalled_TTT_event, role=learner)` by excluding stalled events, the second one `Task(person=person, event=running_TTT_event, role=learner)` will still show up and therefore that person's ID will end up in the `trainees` list.

workshops/views.py
pbanaszkiewicz: Hmmm I think this is a little harder to fix, but let me try. I'll ping you later for a re-review of this part.

workshops/views.py
```diff
@@ -1672,48 +1683,65 @@ def instructors(request):
...
+
+            if data['was_organizer']:
+                people = people.filter(num_organizer__gte=1)
+
+            if data['is_in_progress_trainee']:
+                q = Q(task__event__tags=TTT) & ~Q(task__event__tags=stalled)
+                people = (
+                    people.filter(q, task__role__name='learner')
+                    .exclude(badges__in=instructor_badges)
+                )
```
pbanaszkiewicz: @chrismedrela can you confirm this is what you meant (ref https://github.com/swcarpentry/amy/pull/776#discussion-diff-63297666)?

workshops/templates/workshops/workshop_staff.html
chrismedrela: I was thinking about sth like that:
```html
        <td>
          {% for badge in p.badges.all %}
            {% if badge in instructor_badges %}
              {% bootstrap_tag badge.name|cut:"-instructor"|upper %}
            {% endif %}
          {% endfor %}
        </td>
```
This is very minor issue, so feel free to ignore it.

workshops/views.py
chrismedrela: Could you add a test for that in `test_searching_trainees`?
```python
        # Ironman
        Task.objects.create(person=self.ironman, event=e1, role=learner)
        Task.objects.create(person=self.ironman, event=e2, role=learner)

        ...

        self.assertEqual(list(response.context['persons']), [self.blackwidow, self.ironman])
```
This test fails (ironman is missing). However, if you execute the following code:
```python
        q = Q(task__event__tags=TTT) & ~Q(task__event__tags=stalled)
        print(Person.objects.filter(airport__isnull=False)
                            .select_related('airport')
                            .prefetch_related('badges', 'lessons')
                              # .annotate(
                              #       num_taught=Count(
                              #           Case(
                              #               When(task__role__name='instructor', then=Value(1)),
                              #               output_field=IntegerField()
                              #           )
                              #       ),
                              #       num_helper=Count(
                              #           Case(
                              #               When(task__role__name='helper', then=Value(1)),
                              #               output_field=IntegerField()
                              #           )
                              #       ),
                              #       num_organizer=Count(
                              #           Case(
                              #               When(task__role__name='organizer', then=Value(1)),
                              #               output_field=IntegerField()
                              #           )
                              #       )
                              #   )
                            .filter(q, task__role__name='learner')
                            .exclude(badges__in=Badge.objects.instructor_badges()))
```
you will get correct result (ironman is included). But when you uncomment the `annotate`, ironman is missing in the result. Looks like a bug in Django ORM?

pbanaszkiewicz: I'm afraid we hit Django ORM limitations. For example, this annotation will yield doubled results if we provide argument for default value (even if it's set to zero).

What's worse is that this is a common situation for trainees to be in at least one stalled event…

I'm marking this new test as `expectedFailure` and adding a bug issue so we don't forget…

PR #778, opened 2016-05-13
Notification about changed metatags

### LINKED ISSUE #769: Notify admins about changed metatags
If cronjob discovers updated metatags, there should be some notification popup for admins.

This is a continuation from #752 and #562.

pbanaszkiewicz:
This fixes #769.

This PR:
1. fixes `check_for_workshop_websites_updates.py` command a little (real-test run showed that `PyGithub` module throws some errors)
2. adds notification to the dashboard for admin whose events have changed metatags.

The machinery for viewing/accepting/dismissing changed metatags was already in place from previous PR (#752).

### DIFF
--- a/workshops/management/commands/check_for_workshop_websites_updates.py
+++ b/workshops/management/commands/check_for_workshop_websites_updates.py
@@ -81,7 +81,7 @@ def datetime_decode(obj):
 
 
 class Command(BaseCommand):
-    help = 'Check if events have had updated their metadata.'
+    help = 'Check if events have had their metadata updated.'
 
     def add_arguments(self, parser):
         parser.add_argument(
--- a/workshops/management/commands/check_for_workshop_websites_updates.py
+++ b/workshops/management/commands/check_for_workshop_websites_updates.py
@@ -81,7 +81,7 @@ def datetime_decode(obj):
 
 
 class Command(BaseCommand):
-    help = ''
+    help = 'Check if events have had updated their metadata.'
 
     def add_arguments(self, parser):
         parser.add_argument(
@@ -93,8 +93,13 @@ class Command(BaseCommand):
         parser.add_argument(
             '--init', action='store_true', help='Run for the first time.',
         )
+        parser.add_argument(
+            '--cutoff-days', default=180, type=int,
+            help='Age (in days) of the oldest events that can be checked.  '
+                 'Default: 180'
+        )
 
-    def get_events(self):
+    def get_events(self, cutoff_days=180):
         """Get all active events.
 
         This method is used for getting all events that should be checked
@@ -103,7 +108,7 @@ class Command(BaseCommand):
 
         # events as old as 2014 are still marked as active, so we impose age
         # limit of half a year
-        half_a_year = datetime.timedelta(days=182)
+        half_a_year = datetime.timedelta(days=cutoff_days)
         events = events.filter(start__gte=datetime.date.today() - half_a_year)
         return events
 
@@ -209,11 +214,12 @@ class Command(BaseCommand):
         token = options['token']
         initial_run = options['init']
         slug = options['slug']
+        cutoff_days = options['cutoff_days']
 
         g = Github(token)
 
         # get all events
-        events = self.get_events()
+        events = self.get_events(cutoff_days)
 
         if slug:
             events = events.filter(slug=slug)
--- a/workshops/management/commands/check_for_workshop_websites_updates.py
+++ b/workshops/management/commands/check_for_workshop_websites_updates.py
@@ -3,6 +3,7 @@ import datetime
 from functools import partial
 import json
 import socket
+import sys
 
 from django.core.management.base import BaseCommand
 from github import Github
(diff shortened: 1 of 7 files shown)


workshops/management/commands/check_for_workshop_websites_updates.py
```diff
@@ -97,6 +99,11 @@ def get_events(self):
         This method is used for getting all events that should be checked
         against up-to-date data."""
         events = Event.objects.active().filter(url__isnull=False)
+
+        # events as old as 2014 are still marked as active, so we impose age
+        # limit of half a year
+        half_a_year = datetime.timedelta(days=182)
```
gvwilson: Whoa - hardwired??

workshops/management/commands/check_for_workshop_websites_updates.py
```diff
@@ -228,8 +235,17 @@ def handle(self, *args, **options):
...
+                print('Timeout when accessing {} repo'.format(event.slug))
+
             except WrongWorkshopURL:
                 print('Wrong URL for {}'.format(event.slug))
 
             except requests.exceptions.RequestException:
                 print('Network error when accessing {}'.format(event.slug))
+
+            except Exception as e:
+                print('Unknown error ({}): {}'.format(event.slug, e))
```
gvwilson: where does this print statement appear? who will see it? what happens afterward? does this halt and catch fire?

workshops/management/commands/check_for_workshop_websites_updates.py
pbanaszkiewicz: Yes. Cuts off plenty of old events who still aren't marked as "completed" even though it's been few years.

workshops/views.py
```diff
@@ -1307,6 +1314,32 @@ def event_invoice(request, event_ident):
...
+
+    if assigned_to == 'me':
+        events = events.filter(assigned_to=request.user)
+
+    elif assigned_to == 'noone':
+        events = events.filter(assigned_to=None)
+
+    elif assigned_to == 'all':
+        # no filtering
+        pass
```
gvwilson: Is there no fall-through 'else'?

workshops/management/commands/check_for_workshop_websites_updates.py
gvwilson: I understand what it does, but hard-wiring half a year into the command 
seems like a bad idea.  Command-line option with default, maybe?

pbanaszkiewicz: Oh yeah, that's a nice idea :D

workshops/management/commands/check_for_workshop_websites_updates.py
pbanaszkiewicz: Yeah, I'll change to stderr but this mostly doesn't matter. The output will be logged by cronjob to some file.

We won't throw fire extinguishers at GitHub API if these errors happen, and they happen because:
1. event's URL point to wrong repository or the repo was removed (I believe that's the case for most of the errors), but anyway it's nothing to worry about
2. PyGithub has some issues with one event's repository - again, not our burden.

workshops/views.py
pbanaszkiewicz: No need to, since it's nothing is like setting 'all', and 'all' is set by default by `assignment_selection`.

pbanaszkiewicz: So we really only care about `'me'` or `'noone'` values, anything else gets you the original list of events (not filtered).

gvwilson: I'm mostly thinking that if we don't recognize the value of 
'assigned_to', we should halt with an internal error.

pbanaszkiewicz: Throw errors because someone accidentally typed something into their address bar?

pbanaszkiewicz:
@gvwilson updated, can you check? Thanks

workshops/views.py
gvwilson: Throws errors because we added an option in one part of the code and 
forgot to update this one - I always use an "else: fail" when handling 
enumerations in case someone expands it and misses some code.

workshops/management/commands/check_for_workshop_websites_updates.py
```diff
@@ -78,7 +81,7 @@ def datetime_decode(obj):
 
 
 class Command(BaseCommand):
-    help = ''
+    help = 'Check if events have had updated their metadata.'
```
gvwilson: 'Check if events have had their metadata updated' - your grammar is *almost* right, but *shrug* English...

workshops/management/commands/check_for_workshop_websites_updates.py
```diff
@@ -90,13 +93,23 @@ def add_arguments(self, parser):
         parser.add_argument(
             '--init', action='store_true', help='Run for the first time.',
         )
+        parser.add_argument(
+            '--cutoff-days', default=182, type=int,
+            help='Age (in days) of the oldest events that can be checked.  '
+                 'Default: 182 (half a year)'
+        )
```
gvwilson: You know, pretty much everyone I know would have rounded this to 180 :-)

workshops/management/commands/check_for_workshop_websites_updates.py
```diff
@@ -228,8 +242,21 @@ def handle(self, *args, **options):
...
             except WrongWorkshopURL:
-                print('Wrong URL for {}'.format(event.slug))
+                print('Wrong URL for {}'.format(event.slug), file=sys.stderr)
 
             except requests.exceptions.RequestException:
-                print('Network error when accessing {}'.format(event.slug))
+                print('Network error when accessing {}'.format(event.slug),
+                      file=sys.stderr)
+
+            except Exception as e:
```
gvwilson: This is equivalent to the "else: fail(...)" discussion we're having - you know what exceptions ought to be able to come out, but you catch 'Exception' in case something new arises.

gvwilson:
Minor comments - merge when you're ready, and thanks.

workshops/management/commands/check_for_workshop_websites_updates.py
pbanaszkiewicz: *have had their metadata updated*? *had their metadata updated*?

workshops/management/commands/check_for_workshop_websites_updates.py
pbanaszkiewicz: Fixed in the rebase.

workshops/views.py
pbanaszkiewicz: Fixed in other places too.

PR #783, opened 2016-05-15
Use linear interpolation to eliminate backward bends on charts

### LINKED ISSUE #779: Must-fix: chart of instructors vs. time is unusable
I cannot show https://amy.software-carpentry.org/workshops/reports/instructors_over_time to people: the backward bends in the curve make it unbelievable. This has been raised before, but it must be fixed.
screen shot 2016-05-15 at 5 26 26 pm

chrismedrela:
If using linear interpolation is not an option, we can set `interpolate_tension` to a high value (i.e. 0.95).

EDIT (@pbanaszkiewicz): this fixes #779.

### DIFF
--- a/workshops/templates/workshops/time_series.html
+++ b/workshops/templates/workshops/time_series.html
@@ -25,6 +25,7 @@
                 width: 800,
                 height: 600,
                 target: document.getElementById('chart'),
+                interpolate: 'linear',
                 x_accessor: 'date',
                 y_accessor: 'count'
             });


gvwilson:
Maybe I'm being slow, but we have non-decreasing data - can we not just 
do this as a line connecting a bunch of points?

narayanaditya95:
@gvwilson
Considering that we have non-decreasing data points, the `linear` interpolation mode does give us a bunch of lines.


I would also suggest considering the 'basis` interpolation mode as it produces a smooth curve without causing backward "bends".
screenshot from 2016-05-20 23-59-11

gvwilson:
-1 on basis interpolation - our data is actually pointwise. The linear diagram looks better, but is suspiciously smooth (the ramp on the right side makes it look like we had two data points, one in April and one today).  Is there a parameter to control the jagginess of the interpolation?
Thanks,
Greg

narayanaditya95:
In the example, there are no data points between the end of March and the start of May.
Can you explain what exactly is jagginess? I did google it but couldn't understand how it applies in the case of linear interpolation.

gvwilson:
Hm - I thought some people had qualified in that time, and that the 
curve was therefore smoother than it should be.  Can you please post an 
image of what instructors vs. time looks like for Jan 2012 - May 2016?

narayanaditya95:
I faked a database wherein 100 instructors (100 will probably be way off the actual number) were randomly added over a course of four years.

The graph plotted using linear interpolation looks like:
screenshot from 2016-05-22 20-57-59

On the other hand, the same dataset when plotted with the default interpolation looks like:
screenshot from 2016-05-22 20-53-13

gvwilson:
Thanks - we'll go with linear. (Also, nice work on the 
randomly-generated data.)

pbanaszkiewicz:
Thanks @chrismedrela and @narayanaditya95 for this collaboration.

PR #788, opened 2016-05-17
Rename tags -> metadata

chrismedrela:
Renaming tags -> metadata when it refers to metadata from event website, *not* tags like `TTT`, `stalled` etc.

Note that this PR alters `help_text` of some `Event`'s columns. Could this result in loss of data?

@gvwilson, @pbanaszkiewicz, this PR is ready for review.


gvwilson:
That's a lot of commits. Normally, I try to structure things so that each commit is a logical change, regardless of whether it affects one file or a hundred - the idea is that the commit moves the software from one working state to the next (and can be undone to move the software back). In this case, I think that most of these commits *couldn't* be undone without breakage, so I'd squash them.

workshops/management/commands/check_for_workshop_websites_updates.py
```diff
@@ -137,25 +137,25 @@ def load_from_github(self, github, repo_url, default_branch='gh-pages'):
...
         changes = []
 
         # compare commit hashes
         if branch.commit.sha != event.repository_last_commit_hash:
-            # Hashes differ? Update commit hash and compare stored tags
+            # Hashes differ? Update commit hash and compare stored metadata
             event.repository_last_commit_hash = branch.commit.sha
 
-            tags_new = self.get_event_tags(event.url)
+            new_metadata = self.get_event_metadata(event.url)
```
gvwilson: I would probably use `metadata_new` and `metadata_old` to be consistent with other usage.

workshops/management/commands/check_for_workshop_websites_updates.py
```diff
@@ -170,30 +170,30 @@ def detect_changes(self, branch, event, save_tags=False):
...
-                if save_tags:
-                    # we may not want to update the tags
-                    event.repository_tags = self.serialize(tags_new)
+                if save_metadata:
+                    # we may not want to update the metadata
+                    event.repository_metadata = self.serialize(new_metadata)
 
-                event.tag_changes_detected = "\n".join(changes)
-                event.tags_changed = True
+                event.metadata_changes_detected = "\n".join(changes)
```
gvwilson: This is probably just grammar, but I would expect a variable called `metadata_changes_detected` to be a Boolean ("have metadata changes been detected, yes or no?") rather than a long string.  Maybe `metadata_all_changes` ?

workshops/models.py
```diff
@@ -683,18 +683,18 @@ class Event(AssignmentMixin, models.Model):
...
         help_text='Event\'s repository last commit SHA1 hash')
-    repository_tags = models.TextField(
+    repository_metadata = models.TextField(
         blank=True, default='',
-        help_text='JSON-serialized tags from event\'s website')
-    tag_changes_detected = models.TextField(
-        blank=True, default='', help_text='List of detected tag changes')
-    tags_changed = models.BooleanField(
+        help_text='JSON-serialized metadata from event\'s website')
+    metadata_changes_detected = models.TextField(
```
gvwilson: As above.

workshops/test/base.py
```diff
@@ -315,7 +315,7 @@ def _parse(self, response, save_to=None):
             with open(save_to, 'w') as writer:
                 writer.write(content)
 
-        # Report unfilled tags.
+        # Report undefined variables in templates.
```
gvwilson: Is this undefined variables of all types, or undefined-but-required metadata?

workshops/util.py
```diff
@@ -476,113 +476,113 @@ def generate_url_to_event_index(website_url):
     raise WrongWorkshopURL()
 
 
-def find_tags_on_event_index(content):
-    """Given workshop's raw `index.html`, find and take YAML tags that
+def find_metadata_on_event_index(content):
```
gvwilson: I would change this to `find_metadta_on_event_homepage`, since the word "index" can mean many things (e.g., a database index). Very small point...

workshops/util.py
```diff
@@ -476,113 +476,113 @@ def generate_url_to_event_index(website_url):
...
+        return filtered_metadata
 
     except (ValueError, yaml.scanner.ScannerError):
         # can't unpack or header is not YML format
         return dict()
 
 
-def find_tags_on_event_website(content):
-    """Given website content, find and take <meta> tags that have
+def find_metadata_on_event_website(content):
```
gvwilson: OK, how is this different from `find_metadata_on_event_index`? The names seem interchangeable.

gvwilson:
A few small comments on naming - the big issue is squashing the commits so that we can undo/redo consistently in a single step. Once that's done, :+1: to merge. @pbanaszkiewicz ?

pbanaszkiewicz:
@chrismedrela thanks a lot for this PR.

@gvwilson, @chrismedrela I promised Greg that I release v1.5.4 last week (sorry Greg :( ) but I'm so busy that I won't get to the remaining PRs until tomorrow, and after v1.5.4 I'll review this one and #783.

workshops/test/base.py
chrismedrela: Undefined variables of all types.

workshops/util.py
chrismedrela: `find_metadata_on_event_index/homepage` extracts it from the YAML metadata that is at the beginning of raw `index.html` file while `find_metadata_on_event_website` extracts it from `html  > head > meta` tags

gvwilson: OK - I can't think of better names.

chrismedrela:
@gvwilson, all tests pass after each commit and each commit is a logical change, because each one renames one or few related pieces of code (like a function and its local variables, or a class). 

IMO lots of small commits is an advantage -- if you discover a bug in a future, you can narrow down where the bug lives by bisecting commits. If we squash everything, you could end up searching for the bug in one huge commit, which could take much more time. So, I'd rather leave commits as they are.

I've introduced all the changes that you proposed in your review (except for squashing commits) and rebased the branch.

gvwilson:
If all tests pass after each change, I'm good without squashing - please 
go ahead and merge if @pbanaszkiewicz is happy as well.

gvwilson:
@pbanaszkiewicz can you please merge?

pbanaszkiewicz:
Thanks a lot for this refactoring, @chrismedrela. Good job!

Having multiple smaller commits vs a few bigger commits is a matter of style, however making so many small migrations is only going to have a negative impact for example on the tests.

When you run the tests they take some time to start up, that's when the migrations are run. In Django v1.8 it was far worse than v1.9 since the startup could take as much as a minute…

I'll try to squash the migrations on my end and then merge, but should I fail I'll let @chrismedrela know so that he can squash them by direct manipulation.

pbanaszkiewicz:
@chrismedrela unfortunately I cannot merge this PR.

You seem to have started this PR before #778 got merged. You have to rebase this PR against `develop` (I don't recommend git-pulling develop into this PR - that's not the proper way).

And if you're on it, please add one commit with (django-)squashed migrations. I used this command:

```
./manage.py squashmigrations workshops 0087_renaming_column_repository_tags 0092_alter_help_text_of_metadata_changed
```

You can then remove the old migrations 0087..0092.

chrismedrela:
@pbanaszkiewicz, do you want:
- multiple commits and one migration
- or one commit and one migration?

pbanaszkiewicz:
@chrismedrela if we're talking about the rebase then you don't have to additionally squash the commits, I'm most interested in the outcome.

The squashed migration and the removal of migrations 0087..0092 can happen in one additional commit.

chrismedrela:
@pbanaszkiewicz, done. Note that I renamed `events_tag_changed` to `events_metadata_changed` (and related local variables, urls etc.).

pbanaszkiewicz:
Merging with a couple of really small fixes. Thanks @chrismedrela!!!

PR #798, opened 2016-05-19
Language model

### LINKED ISSUE #582: Let's register teaching languages for instructors

narayanaditya95:
I have added the ability to record the languages an instructor can deliver a workshop in. 
Also, admins can find instructors by the languages they can teach at the "Find Instructor" page.

This fixes #582.

To-Do:
- [x] Add Language model and migrations

- [x] Allow autocomplete lookups on Languages.

- [x] Allow filtering of instructors by a set of languages.

- [x] Migrate EventRequest models' `language` attribute from `CharField` to `ForeignKey`/`M2M'?


narayanaditya95:
@gvwilson @pbanaszkiewicz @chrismedrela
Please review the progress till now. Are there any concern on the features that have been implemented?

workshops/migrations/0087_auto_20160519_0739.py
```diff
@@ -0,0 +1,54 @@
...
+# Generated by Django 1.9.6 on 2016-05-19 12:39
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+import requests
+
+
+def populate_languages(apps, schema_editor,
+        url='https://raw.githubusercontent.com/mattcg/language-subtag-registry/master/data/json/registry.json'):
```
narayanaditya95: The [language-tag](https://github.com/OnroerendErfgoed/language-tags/) package uses this as the underlying data.
Plus it's neatly serialized to JSON.

gvwilson: Please cache a copy of this locally - we have no control over external resources (they can be renamed or deleted without warning), and I'd like to be able to run tests offline (I spend more time than I'd like on planes).

workshops/migrations/0087_auto_20160519_0739.py
```diff
@@ -0,0 +1,54 @@
...
+    """Populate the Languages table.
+
+    [1] lists IANA as the registry maintainer, [2] looks like that
+    registry, and [3] describes the entry-format in the registry.
+
+    [1]: https://tools.ietf.org/html/rfc5646#section-2.2.1
+    [2]: http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry
+    [3]: https://tools.ietf.org/html/rfc5646#section-3.1
+    """
+    languages = requests.get(url).json()
```
gvwilson: As above, please cache the data locally to make the application more robust.

workshops/models.py
```diff
@@ -478,6 +482,24 @@ def __str__(self):
...
+
+    https://tools.ietf.org/html/rfc5646
+    """
+    name = models.CharField(
+        max_length=STR_MED,
+        help_text='Description of this language tag in English')
+    subtag = models.CharField(
+        max_length=STR_SHORT,
+        help_text=
+            'Primary language subtag.  '
```
gvwilson: why "subtag"?

workshops/migrations/0087_auto_20160519_0739.py
narayanaditya95: Okay I will.

workshops/models.py
narayanaditya95: The IANA calls them "subtags" and not tags.
See <http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry>

workshops/migrations/0088_auto_20160520_0626.py
```diff
@@ -0,0 +1,46 @@
...
+def migrate_language(apps, schema_editor):
+    """Convert EventRequest.language string to key"""
+    EventRequest = apps.get_model('workshops', 'EventRequest')
+    Language = apps.get_model('workshops', 'Language')
+    for request in EventRequest.objects.all():
+        # Get the most precisely matching languages
+        language = Language.objects.filter(name__icontains=request.language)\
+                .order_by(Length('name')-len(request.language)).first()
+        if not language:
+            raise ValueError('no languages matching {} (for {})'.format(
```
narayanaditya95: Should I populate them with the default langauge "English" instead?

narayanaditya95:
@pbanaszkiewicz @gvwilson @pbanaszkiewicz 
The PR is up for review. I have addressed the issues @gvwilson posted.

The CI fails, although they pass locally on my computer. Debugging.

workshops/forms.py
```diff
@@ -624,6 +638,13 @@ class SWCEventRequestForm(forms.ModelForm):
                   '-to-admin-fee.html" target="_blank">Look up administration '
                   'fees</a>.',
     )
+    language = selectable.AutoCompleteSelectField(
+        lookup_class=lookups.LanguageLookup,
+        label='Language',
+        required=False,
+        initial=Language.objects.get(name='English'),
```
narayanaditya95: Added a default on the form rather than the model as I couldn't add a FK default.
Is this an issue?

workshops/models.py
gvwilson: OK.

workshops/migrations/0088_auto_20160520_0626.py
gvwilson: Yes please.

workshops/forms.py
gvwilson: @pbanaszkiewicz is there a way to add a foreign key default?

workshops/lookups.py
```diff
@@ -96,8 +96,23 @@ class AirportLookup(ModelLookup):
...
 
 
+class LanguageLookup(ModelLookup):
+    model = models.Language
+    search_fields = (
+        'name__icontains',
+        'subtag__icontains',
+    )
+
+    def get_query(self, request, term):
```
gvwilson: Hm - I never know what to do about this. Alphabetical makes them easier to find, but means scrolling for the common case. By frequency reduces scrolling but increases reading time.  Everything with > 100 entries first (alpha), then everything else (also alpha), is a good compromise, but at this point, I'm definitely over-engineering :-)

workshops/templates/workshops/person.html
```diff
@@ -100,6 +100,17 @@
     {% else %}
     <h5>No lessons.</h5>
     {% endif %}
+
+    {% if languages %}
+    <h5>Languages:</h5>
+    <ul>
+      {% for l in languages %}
```
gvwilson: I would use 'for lang in languages' since 'l' looks too much like a '1'.

gvwilson:
Minor comments, but looking good - you'll need to merge in from 'develop' and re-push. @pbanaszkiewicz are you OK with this one?

workshops/lookups.py
narayanaditya95: We can sort the first 10 most popular languages alphabetically while the rest are listed by their frequency.

workshops/templates/workshops/person.html
narayanaditya95: I was adhering to the naming convention that was followed around this code. Will make a change.

gvwilson:
@pbanaszkiewicz please merge if it looks good

workshops/templates/workshops/person.html
pbanaszkiewicz: Using `for p in persons` is bad (and we're sinners since we used similar code in AMY), but `for l in languages` is worse (as @gvwilson points out). It should be `for L in languages`, but since `lang` is such a short word it's way better to use it instead of `l` or `L`.

narayanaditya95: Okay. :)

workshops/lookups.py
```diff
@@ -96,8 +96,23 @@ class AirportLookup(ModelLookup):
     )
 
 
+class LanguageLookup(ModelLookup):
```
pbanaszkiewicz: Shouldn't this be login required?

pbanaszkiewicz: Or maybe not, it's supposed to be accessible to non-logged in users, right?

narayanaditya95: Yes. On the workshop request page.

narayanaditya95:
@pbanaszkiewicz I have rewritten the git history and squash redundant commits.

workshops/migrations/0088_auto_20160520_0626.py
```diff
@@ -0,0 +1,46 @@
...
+
+def migrate_language(apps, schema_editor):
+    """Convert EventRequest.language string to key"""
+    EventRequest = apps.get_model('workshops', 'EventRequest')
+    Language = apps.get_model('workshops', 'Language')
+    english = Language.objects.get(name='English')
+    for request in EventRequest.objects.all():
+        # Get the most precisely matching languages
+        language = Language.objects.filter(name__icontains=request.language)\
+                .order_by(Length('name')-len(request.language)).first()
```
pbanaszkiewicz: Can't test because of this `Length` which is unknown.

narayanaditya95: Failed to notice. :(

narayanaditya95:
@pbanaszkiewicz Ping.

pbanaszkiewicz:
@narayanaditya95 you must rebase this branch against `develop` since some conflicts are present.

narayanaditya95:
@pbanaszkiewicz Resolved conflicts.

workshops/forms.py
```diff
@@ -630,6 +644,12 @@ class SWCEventRequestForm(forms.ModelForm):
                   '-to-admin-fee.html" target="_blank">Look up administration '
                   'fees</a>.',
     )
+    language = selectable.AutoCompleteSelectField(
+        lookup_class=lookups.LanguageLookup,
+        label='Language',
+        required=False,
+        widget=selectable.AutoComboboxSelectWidget,
```
pbanaszkiewicz: No `help_text` here?

workshops/views.py
```diff
@@ -1794,6 +1796,9 @@ def workshop_staff(request):
                 people = people.filter(q, task__role__name='learner') \
                                .exclude(badges__in=instructor_badges)
 
+            if data['languages']:
+                people  = people.filter(languages__in=data['languages'])
+
```
pbanaszkiewicz: I think there are some tests for `workshop_staff` searching, please extend them to cover this addition too.

pbanaszkiewicz:
@narayanaditya95 excellent work so far. Apart from the two minor comments, I think we need to add languages to the `Event` model. All workshops have some language listed (see for example https://github.com/leszektarkowski/2015-10-03-Krakow-techspace/blob/gh-pages/index.html#L7), and it's being read by the import event machinery – it's just not saved anywhere because there is no field for the language yet.

workshops/forms.py
narayanaditya95: `PersonCreateForm` inherits from `PersonForm` the langauge**s** field. 
The form at `workshops/persons/add` shows language**s** as a field.
Are you referring to the language field here in the `SWCEventRequest` form? I am not sure why I did not opt for a `ForeignKey` to the `Langauge` model on the model `EventRequest`. Do you want me to switch to M2M to `Language` model on the `EventRequest` model?

workshops/models.py
```diff
@@ -872,11 +894,11 @@ class EventRequest(AssignmentMixin, models.Model):
                   'to accommodate those requests.',
         verbose_name='Preferred workshop dates',
     )
-    language = models.CharField(
-        max_length=STR_LONG,
+    language = models.ForeignKey(
```
narayanaditya95: @gvwilson @pbanaszkiewicz 
Should this be a `M2M` instead?

workshops/forms.py
pbanaszkiewicz: @narayanaditya95 sorry, the way GH displayed the diff it looked as if `PersonCreateForm` was implementing `language` field.

I think for now let's leave EventRequests monolingual, if someone wants to run workshop in other languages they can always add it to the notes.

Unless you implement `Event.languages`, well, then `EventRequest` should have `languages` too.

narayanaditya95:
@pbanaszkiewicz 
I have added tests.
Can an event have more than one language (comma-separated maybe) mentioned in the front-end matter in the index.html of website template?
If we do not have more than one language for an event, a ForeignKey to Language model on the Even model should suffice.

pbanaszkiewicz:
@narayanaditya95 I don't think people comma-separate languages in the index.html - if they do, that's a very very rare occasion. I'd stick to only one language per event until admins tell us they need M2M :-)

pbanaszkiewicz:
Alright, things are looking good. I think we're missing:

* `Event.language` FK to `Language`
* perhaps import / update from URL machinery to match language? Unless we don't cannot match languages by the data people are supposed to put in

narayanaditya95:
- [x] Event.language FK to Language

- [x] perhaps import / update from URL machinery to match language? Unless we don't cannot match languages by the data people are supposed to put in

So, I initiate a search on the language field based on the `data.language` returned by the `event_import` response. However since `data.language` is only a two-letter word, it generates a lot of search results as searches are done against both the `name` and the `subtag` field.
For eg, `en` generates about 13 results wherein the position of English will depend on the its popularity against the other 12.
We also cannot switch to restricting the search to just the `subtag` (although `en` matched perfectly to English). A user input for Polish must then be `pl` which we cannot expect.

pbanaszkiewicz:
>  We also cannot switch to restricting the search to just the `subtag` (although `en` matched perfectly to English).

Why?

> A user input for Polish must then be `pl` which we cannot expect.

Why?

Isn't `subtag` compatible with ISO 639-1?

pbanaszkiewicz:
@narayanaditya95 if the `subtag` is ISO 639-11 compatible, then we can use [this trick](http://django-selectable.readthedocs.io/en/latest/advanced.html#chained-selection) to only select entries with matching subtag. This requires two additional lines in `LanguageLookup.get_query`.

The disadvantage of this approach is that we have to remember to clear the prepared query subtag.

Also: does the user have to manually click on selected language (even if it's the only one matching) in order for it to be selected? Can't we somehow automate this?

Anyway, as soon as we can match by the subtag I'm going to merge this.

narayanaditya95:
Hi @pbanaszkiewicz 

> Is subtag not compatible with ISO 639-1?

Yes, it is.

> @narayanaditya95 if the subtag is ISO 639-11 compatible, then we can use this trick to only select entries with matching subtag. This requires two additional lines in LanguageLookup.get_query.

Correct me if I am mistaken. You propose we pass in a query string (say `subtag=True`) from the UI. This would allow us to only filter based on the subtag. Right?

>The disadvantage of this approach is that we have to remember to clear the prepared query subtag.

Clear when? I couldn't follow.

>Also: does the user have to manually click on selected language (even if it's the only one matching) in order for it to be selected? Can't we somehow automate this?

Currently, yes. I will browse through the API docs to find a way to automate it.

pbanaszkiewicz:
Hi,

On Fri, Jun 17, 2016 at 8:56 AM, Aditya Narayan <notifications@github.com>
wrote:
>
> @narayanaditya95 <https://github.com/narayanaditya95> if the subtag is
> ISO 639-11 compatible, then we can use this trick to only select entries
> with matching subtag. This requires two additional lines in
> LanguageLookup.get_query.
>
> Correct me if I am mistaken. You propose we pass in a query string (say
> subtag=True) from the UI. This would allow us to only filter based on the
> subtag. Right?
>
Yes, we can pass additional params via query string (more like
`?subtag=en`). It's possible to do by:

$('#id_city_0').djselectable('option', 'prepareQuery', newParameters);

As you can see, this is embedded in djselectable jQuery functionality.


> The disadvantage of this approach is that we have to remember to clear the
> prepared query subtag.
>
> Clear when? I couldn't follow.
>
Try manually setting the `prepareQuery` -- you'll see it stays "on" the
whole time after you set it. You'll not be able to change language if you
want…


> Also: does the user have to manually click on selected language (even if
> it's the only one matching) in order for it to be selected? Can't we
> somehow automate this?
>
> Currently, yes. I will browse through the API docs to find a way to
> automate it.
>
I was looking for it myself and I didn't find it, but then I didn't find
the 'prepareQuery' option either. Hopefully it is possible.

Cheers,
Piotr

narayanaditya95:
Phew, finally done.
Not the best of my codes, but it works. :)

workshops/static/import_from_url.js
```diff
@@ -9,6 +9,21 @@ function import_from_url(url) {
     $("#id_end").val(data.end);
     $("#id_reg_key").val(data.reg_key);
     $("#id_url").val(url);
+
+    var elem = $("id_language_0");
+    // query only using the subtag
+    elem.djselectable('option', 'prepareQuery', function(q) { q.subtag = true });
```
pbanaszkiewicz: Why `q.subtag = true` instead of `q.subtag = data.language`?

pbanaszkiewicz: Alright, I see what you did in the `get_query`.

pbanaszkiewicz:
@narayanaditya95 good job with finding how to click and close the selectable widget :-)

I only need you to rebase against develop and add a merging migration. Ping me so that I can merge this PR.

narayanaditya95:
@pbanaszkiewicz 
Rebased and rewritten git history. 
Good to be merged.

pbanaszkiewicz:
@narayanaditya95 thanks a lot, we finally have this PR in :-) good job!

PR #801, opened 2016-05-19
Speedup 2x tests

chrismedrela:
On my machine, this patch speedup tests execution from 90 seconds to 50 secs.

### DIFF
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -12,6 +12,7 @@ https://docs.djangoproject.com/en/1.7/ref/settings/
 import os
 import json
 
+import sys
 from django.utils.translation import ugettext_lazy as _
 
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
@@ -277,3 +278,8 @@ LOGGING = {
         },
     },
 }
+
+if DEBUG and 'test' in sys.argv:
+    PASSWORD_HASHERS = (
+        'django.contrib.auth.hashers.MD5PasswordHasher',
+    )


pbanaszkiewicz:
Whaaaaat. What improvement are you seeing? What do Django folks say about switching the hasher for tests?

chrismedrela:
With this patch, it doesn't take so long to run tests. From my point of view, this is an improvement. :)

Switching hasher for tests is a tip from [Django doc](https://docs.djangoproject.com/en/1.9/topics/testing/overview/#speeding-up-the-tests).

chrismedrela:
@pbanaszkiewicz, could we merge this with the next release? This'll make my life much easier.

pbanaszkiewicz:
I'm going to merge it soon, just let me test it first.

pbanaszkiewicz:
Indeed this brings decent speedup: 27s vs 17s for me.

Merging with single change (comment about that conditional in the settings).

@chrismedrela If you want further speed up the tests, you can try the new multithreaded mode available since Django v1.9 I believe.

Thanks for this contribution!

chrismedrela:
@pbanaszkiewicz, do you mean executing `python3 manage.py test --parallel=4`? On my machine, I get `NotImplementedError: subtests aren't supported at this time` error from `multiprocessing` library.

pbanaszkiewicz:
@chrismedrela yes, but I didn't claim it was painless :-)

narayanaditya95:
Tests now throw segmentation fault everytime.

PR #806, opened 2016-05-21
Prevents duplication of frequency for instructors with both DC and SWC badges.

### LINKED ISSUE #787: Frequency of teaching report is wrong
https://amy.software-carpentry.org/workshops/reports/instructor_num_taught reports that Jonah Duckles has taught 26 times - I don't believe that's correct given our current database.

narayanaditya95:
Fixes #787

To be honest, I have no clue why this minor edit fixed the issue.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -324,7 +324,7 @@ class ReportsViewSet(ViewSet):
     @list_route(methods=['GET'])
     def instructor_num_taught(self, request, format=None):
         badges = Badge.objects.instructor_badges()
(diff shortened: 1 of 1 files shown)


gvwilson:
Then I'm :-1: - if we don't know why it worked, we don't really know that it *did*. (Thus speaks the voice of bitter experience :-) )

narayanaditya95:
Yes, closing this PR.
Seems like an undefined behaviour.

PR #807, opened 2016-05-22
Prevents duplication of frequency for instructors with both DC and SWC badges.

### LINKED ISSUE #787: Frequency of teaching report is wrong

narayanaditya95:
Fixes #787

This solution makes a change to the list of instructors displayed.
Earlier, the instructors with SWC/DC badges who didn't teach (as an instructor) in any workshop were also listed with their corresponding count being zero.
Now, only instructors with non-zero frequency are listed.

Is that fine?

### DIFF
--- a/api/test/test_reports.py
+++ b/api/test/test_reports.py
@@ -1,0 +1,83 @@
+import datetime
+import json
+
+from django.core.urlresolvers import reverse
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+from api.views import (
+    ReportsViewSet,
+)
+from api.serializers import InstructorNumTaughtSerializer
+from workshops.models import (
+    Badge,
+    Award,
+    Person,
+    Role,
+    Host,
+    Task,
+    Event,
+)
+
+
+class BaseReportingTest(APITestCase):
+
+    def setUp(self):
+        self.login()
+
+    def login(self):
+        self.admin = Person.objects.create_superuser(
+                username="admin", personal="Super", family="User",
+                email="sudo@example.org", password='admin')
+        self.client.login(username='admin', password='admin')
+
+
+class TestReportingInstructorNumTaught(BaseReportingTest):
+    def setUp(self):
+        super().setUp()
+
+        # get instructor badges
+        swc_instructor, _ = Badge.objects.get_or_create(
+            name='swc-instructor'
+        )
+        dc_instructor, _ = Badge.objects.get_or_create(
+            name='dc-instructor'
+        )
+        # set up an event
+        host = Host.objects.create(domain='host.edu', fullname='Host EDU')
+        event = Event.objects.create(slug='event1', host=host)
+        instructor = Person.objects.create(
+            username='harrypotter', personal='Harry', family='Potter',
+            email='user1@name.org',
+        )
+        instructor_role, _ = Role.objects.get_or_create(name='instructor')
+        Task.objects.create(
+            event=event,
+            person=instructor,
+            role=instructor_role
+        )
+        # Award a SWC Badge
+        Award.objects.create(person=instructor, badge=swc_instructor,
+                             awarded=datetime.date.today())
+        # Award a DC Badge
+        Award.objects.create(person=instructor, badge=dc_instructor,
+                             awarded=datetime.date.today())
+
+        # make sure we *do not* get twice the number expected
+        self.expecting = [
+            {
+                'person': 'http://testserver/api/v1/persons/{}/?format=json'.format(
+                    instructor.pk,
+                ),
+                'name': 'Harry Potter',
+                'num_taught': 1,
+            },
+        ]
+
+    def test_view(self):
+        # test only JSON output
+        url = reverse('api:reports-instructor-num-taught')
+        response = self.client.get(url, {'format': 'json'})
+        content = response.content.decode('utf-8')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(json.loads(content), self.expecting)
--- a/api/views.py
+++ b/api/views.py
@@ -1,7 +1,7 @@
 import datetime
 from itertools import accumulate
 
-from django.db.models import Count, Sum, Case, When, Value, IntegerField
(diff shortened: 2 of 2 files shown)


gvwilson:
I like the code, but we want people who've never taught displayed as well, so that they're easy to find and pester.  I think we get this if you take out the filter on task role.  You should also take out the may_contact filter: even if people don't want us emailing them, they should show up in stats.

narayanaditya95:
> I like the code, but we want people who've never taught displayed as well, so that they're easy to find and pester. 

+1

>  I think we get this if you take out the filter on task role.

-1. Taking that filter out will give us counts of all roles assumed by the person (including those of tutor, trainees, etc).

I have come up with this solution, but it involves two different queries:
```python
past_instructors = Person.objects.filter(badges__in=badges) \
                        .filter(task__role__name='instructor') \
                        .annotate(num_taught=Count('task', distinct=True)) \
                        .order_by('-num_taught')
future_instructors = Person.objects.filter(badges__in=badges) \
                        .exclude(task__role__name='instructor') \
                        .annotate(num_taught=Value(0, IntegerField()))
serializer = InstructorNumTaughtSerializer(
    list(past_instructors) + list(future_instructors),
    many=True,
    context=dict(request=request)
)
```

gvwilson:
@pbanaszkiewicz ?

pbanaszkiewicz:
@narayanaditya95 I don't like the way serializer is fed with query. Since `past_instructors` and `future_instructors` are both Query**Sets**, in some cases you can sum them as you'd sum mathematical sets. Please check if it possible here.

narayanaditya95:
`chain(past_instructors, future_instructors)` gives the same output.
`past_instructors | future_instructors` does not give the expected output.

pbanaszkiewicz:
Okay, it took me quite some time but I think I have a solution.

@narayanaditya95 go check out this script, you can run it by `./manage.py shell` and then `import scriptname`:

```python
from django.db.models import Count, Value, IntegerField, Case, When, Sum, F
from workshops.models import Badge

badges = Badge.objects.instructor_badges()
past_instructors = Person.objects.filter(badges__in=badges) \
                                 .filter(task__role__name='instructor') \
                                 .annotate(num_taught=Count('task', distinct=True))

future_instructors = Person.objects.filter(badges__in=badges) \
                                   .exclude(task__role__name='instructor') \
                                   .annotate(num_taught=Value(0, IntegerField())) \
                                   .distinct()

chain = list(past_instructors) + list(future_instructors)

comp = Person.objects.filter(badges__in=badges).annotate(
        num_taught=Count(Case(When(task__role__name='instructor', then=1)), distinct=True)
    )
comp = Person.objects.filter(badges__in=badges).annotate(
        num_taught=Count(Case(When(task__role__name='instructor', then=F('task'))), distinct=True)
    )
comp = Person.objects.filter(badges__in=badges).annotate(
        num_taught=Sum(Case(When(task__role__name='instructor', then=1)), distinct=True)
    )
comp = Person.objects.filter(badges__in=badges).annotate(
        num_taught=Sum(Case(When(task__role__name='instructor', then=1)))
    )

for left in chain:
    right = comp.get(pk=left.pk)
    assert left.num_taught == right.num_taught, (left, left.num_taught, right.num_taught)

print(len(chain))
print(comp.count())
```

One of these `comp` objects actually works and it works in a correct way (assuming your `past_instructors` and `future_instructors` work correctly - but I think they do).

narayanaditya95:
It's the second one.
I am in awe!
Great!

pbanaszkiewicz:
One last thing: please make a regression test case for this bug.

narayanaditya95:
@pbanaszkiewicz Which module should the regression test be written in?

pbanaszkiewicz:
@narayanaditya95 apparently there are no tests for reports (my bad…), so please create `api/test/test_reports.py` and put the test there in a general `TestReporting` class.

narayanaditya95:
@pbanaszkiewicz I have been getting segmentation faults on running tests ever since [54793f3891b042dc67f990c2753e6ca1ed082781](https://github.com/swcarpentry/amy/commit/54793f3891b042dc67f990c2753e6ca1ed082781) was merged.

pbanaszkiewicz:
@gvwilson to my knowledge you have mac. Can you run the tests on `develop` please?

@narayanaditya95 you probably hit some bug in Python, let's see if everything works for Greg and if so then we'll try to narrow down the problem and report upstream.

gvwilson:
I get this with Python 3.5.1 :: Anaconda 2.5.0 (x86_64) on Mac OS X 10.10.5, on `develop` at 8000e14e25c11028c9f1ec9f61c47bda46ec2251:

~~~
make test
python3 manage.py test
/Users/gvwilson/anaconda3/envs/amy/lib/python3.5/site-packages/selectable/forms/fields.py:27: RemovedInDjango110Warning: `BaseAutoCompleteField._has_changed` method should be renamed `has_changed`.
  class BaseAutoCompleteField(forms.Field):

Creating test database for alias 'default'...
............................................................................s.s..........................................................................................................................................................................................x.
Ran 267 tests in 19.600s

OK (skipped=2, expected failures=1)
~~~

Note: the Travis build for this PR fails: https://travis-ci.org/swcarpentry/amy/builds/133854804 - because `BaseExportingTest` can't be imported.

narayanaditya95:
@gvwilson 
What version of SQLite are you running?

Mine says:
```
sqlite3 --version
3.8.10.2 2015-05-20 18:17:19 2ef4f3a5b1d1d0c4338f8243d40a2452cc1f7fe4
```

narayanaditya95:
@pbanaszkiewicz ping. Review please.

gvwilson:
3.9.2 2015-11-02 18:31:45 bda77dda9697c463c3d0704014d51627fceee328

pbanaszkiewicz:
@narayanaditya95  Thanks for the submission!

If your tests still segfault then please add an issue so that we can investigate further.

PR #810, opened 2016-05-23
Allow bulk emails to filtered workshop staff.

### LINKED ISSUE #797: Email instructors from search results
When I search for instructors using https://amy.software-carpentry.org/workshops/instructors I would like to be able to capture all their email addresses so I can send them an email.

narayanaditya95:
Fixes #797 

@maneesha @gvwilson 
The emails are currently listed in the TO field. Do I have to modify it to use the BCC field?

Please review.


workshops/views.py
```diff
@@ -1771,6 +1771,7 @@ def workshop_staff(request):
                 people = people.filter(q, task__role__name='learner') \
                                .exclude(badges__in=instructor_badges)
 
+    emails = people.values_list('email', flat=True)
```
narayanaditya95: Note: The list of emails here is the entire list of staff that confirm to the filter given. They may be listed in the second or the third page when the results are paginated.

@gvwilson @maneesha Do we want to email the list shown on the particular page number or do we want to email every staff in all the page numbers.

workshops/templates/workshops/workshop_staff.html
```diff
@@ -43,6 +43,7 @@
       </tr>
     {% endfor %}
     </table>
+    <p><a href="mailto:{{ emails|join:',' }}" class="btn btn-primary">EMail All</a></p>
```
narayanaditya95: You can suggest a better caption for the button. Couldn't come up with anything better.

pbanaszkiewicz: Contact all/everyone

PR #811, opened 2016-05-23
Add the ability to send bulk emails to filtered workshop staff.

### LINKED ISSUE #797: Email instructors from search results

narayanaditya95:
Fixes #797 

Sorry about the goof-up. I should probably use better branch names.

### DIFF
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -1794,6 +1794,7 @@ def workshop_staff(request):
                 people = people.filter(q, task__role__name='learner') \
                                .exclude(badges__in=instructor_badges)
 
+    emails = people.filter(may_contact=True).values_list('email', flat=True)
     people = get_pagination_items(request, people)
     context = {
         'title': 'Find Workshop Staff',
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
Ref #810.

workshops/views.py
```diff
@@ -1771,6 +1771,7 @@ def workshop_staff(request):
                 people = people.filter(q, task__role__name='learner') \
                                .exclude(badges__in=instructor_badges)
 
+    emails = people.values_list('email', flat=True)
```
pbanaszkiewicz: Does this exclude people with empty emails or people who don't agree to be contacted (`may_contact=False`)?

narayanaditya95: I added a filter for `may_contact` field.
It does filter out empty emails.

PR #815, opened 2016-05-29
Fix event validion: check format for required or non-empty tags

### LINKED ISSUE #813: Website validation gives incorrect answer for empty fields
1.  Go to https://amy.software-carpentry.org/workshops/event/2016-07-11-ttt-scipy.
2.  Click 'validate event'.
3.  Error message reports 'helper' and 'eventbrite' are invalid.

Both are present, but empty.

pbanaszkiewicz:
Changed event validation. It now checks expected format only for tags
that are required or that have a non-empty value.

This fixes #813.

### DIFF
--- a/workshops/test/test_util.py
+++ b/workshops/test/test_util.py
@@ -756,6 +756,27 @@ Other content.
         assert len(errors) == 12
         assert all([error.startswith('Missing') for error in errors])
 
+    def test_validating_empty_metadata(self):
+        metadata = {
+            'slug': '',
+            'language': '',
+            'startdate': '',
+            'enddate': '',
+            'country': '',
+            'venue': '',
+            'address': '',
+            'latlng': '',
+            'instructor': '',
+            'helper': '',
+            'contact': '',
+            'eventbrite': '',
+        }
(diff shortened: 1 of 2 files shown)


gvwilson:
I don't see how this fixes the problem:
*   the `requirements` list in `validate_tags_from_event_website` still has `required == True` for `helper` (which it should, since we want there to be a `helper` field)
*   but this change skips the check if the value is empty (which is OK for `helper`)...
*   ...which means that an empty list of instructors will now pass as well, won't it?  Which shouldn't be allowed.
What am I missing?

pbanaszkiewicz:
>  but this change skips the check if the value is empty (which is OK for helper)...

The check happens if (the tag is required) or if (it's not *and* the value is non-empty).

>  ...which means that an empty list of instructors will now pass as well, won't it? Which shouldn't be allowed. What am I missing?

It won't pass and the test for that was added in this PR.

gvwilson:
Thanks - I shouldn't read code this early in the morning. :+1: to merge 
- thanks for the fix.

PR #818, opened 2016-05-29
Fix IntegrityError when merging people with similar M2M related objects

### LINKED ISSUE #755: IntegrityError when merging persons with tasks in common and "combine" strategy
Example: merging person A and person B, both have similar task "learner" in "XYZ" event.

pbanaszkiewicz:
For example: merging two people with similar task of being a learner at
the event '2016-05-29-copacabana' would yield IntegrityError when trying
to use "combine" strategy since after merging there would be two
identical tasks (both learner, event '2016-05-29-copacabana' and the
same person) - and that violates the uniqueness constraint.

This fixes #755.

### DIFF
--- a/workshops/util.py
+++ b/workshops/util.py
@@ -797,6 +797,11 @@ def merge_objects(object_a, object_b, easy_fields, difficult_fields,
         base_obj = object_b
         merging_obj = object_a
 
+    # used to catch all IntegrityErrors caused by violated database constraints
+    # when adding two similar entries by the manager (see below for more
+    # details)
+    integrity_errors = []
+
     with transaction.atomic():
         for attr in easy_fields:
             value = choices.get(attr)
@@ -851,14 +856,14 @@ def merge_objects(object_a, object_b, easy_fields, difficult_fields,
 
                 # some entries may cause IntegrityError (violation of
                 # uniqueness constraint) because they are duplicates *after*
-                # being added to the database
+                # being added by the manager
                 for element in summed:
                     try:
                         with transaction.atomic():
                             manager.add(element)
-                    except IntegrityError:
-                        pass
+                    except IntegrityError as e:
+                        integrity_errors.append(str(e))
 
         merging_obj.delete()
 
-        return base_obj.save()
+        return base_obj.save(), integrity_errors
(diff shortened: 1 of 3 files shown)


pbanaszkiewicz:
@gvwilson @chrismedrela @narayanaditya95 I'd like to ask anyone of you for a feedback, thanks!

workshops/util.py
```diff
@@ -847,9 +847,17 @@ def merge_objects(object_a, object_b, easy_fields, difficult_fields,
...
+                summed = related_a.all() | related_b.all()
+
+                # some entries may cause IntegrityError (violation of
+                # uniqueness constraint) because they are duplicates *after*
+                # being added to the database
+                for element in summed:
+                    try:
+                        with transaction.atomic():
+                            manager.add(element)
+                    except IntegrityError:
```
gvwilson: I think this is too general, since it hides *all* integrity errors, not just ones resulting from duplicate entries.

chrismedrela:
@pbanaszkiewicz, I think this PR is fine, except for the issue raised by @gvwilson.

workshops/util.py
pbanaszkiewicz: Yup, and when uniqueness constraint is violated it generates integrity errors. Since they happen when we try to add something to the DB, we know it's going to violate some DB constraint.

gvwilson: Right: my question is, will catching and discarding `IntegrityError` 
mask errors that we actually want reported?  Is there a way to restrict 
this so that it only discards errors that arise when Task objects are 
duplicates?  Or am I being paranoid?

pbanaszkiewicz: I think you are being paranoid but for a good cause.

We can't catch unique constraints violations only.

We could avoid integrity errors if we explicitly singled out overlapping tasks - but that's not easy / feasible in general way.

gvwilson: All right - can you catch them all, then report on the page after the 
merge is complete what they were?  If so, let's deploy that.

narayanaditya95:
@all +1 for merge.

pbanaszkiewicz:
@gvwilson updated, please let me know if this suffices.

gvwilson:
:+1:

PR #819, opened 2016-05-30
Make all Charfield non-nullable.

### LINKED ISSUE #790: Should we remove 'null=True' from all CharFields?
https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.CharField suggests that CharField shouldn't use null=True, but instead should use blank=True.

### LINKED ISSUE #812: Slug should be a required field

narayanaditya95:
Migrate CharFields and populate slugs on Events.
Fix #790
Fix #812


gvwilson:
:+1:

workshops/migrations/0088_auto_20160530_1029.py
```diff
@@ -0,0 +1,60 @@
...
+from django.template.defaultfilters import slugify
+
+
+class Migration(migrations.Migration):
+
+    def slugify(apps, schema_editor):
+        """Generate slug from date and venue."""
+        Event = apps.get_model('workshops', 'Event')
+        for event in Event.objects.filter(slug__isnull=True):
+            event.slug = slugify(str(event.start) + '-' + event.venue)
```
narayanaditya95: Loop until no Integrity Errors? Opinions.

narayanaditya95:
@pbanaszkiewicz 
A non-closing <input> tag is causing XML parsing of response to fail. How do I deal with this?

pbanaszkiewicz:
@narayanaditya95 close the input like so: `<input name="…" tag="value" />`

narayanaditya95:
@pbanaszkiewicz 
What exactly is the purpose of this test?
`test/test_event.py` L321

```
    def test_add_two_minimal_events(self):
        host = Host.objects.get(fullname='Test Host')
        url = reverse('event_add')
        data = {
                'slug': '2016-test-event-random',
                'host': host.id,
                'tags': [self.test_tag.id],
                'administrator': host.id,
                'invoice_status': 'unknown',
            }
        response = self.client.post(url, data)
        assert response.status_code == 302, (
            'expected 302 redirect after first post, got {}'.format(
            response.status_code))
        response = self.client.post(url, data)
        if response.status_code != 302:
            self._check_status_code_and_parse(response, 200)
            assert response.status_code == 302, (
                'expected 302 redirect after second post, got {}'.format(
                    response.status_code))
```

pbanaszkiewicz:
@narayanaditya95 I agree that test should be thrown out / rewritten. I don't know its purpose, it's lacking the docstring, uses normal asserts (I have nothing against them if using py.test, but we don't so…) and the part after second submission is total unknown to me.

narayanaditya95:
Removing that out. We already have a test to ensure non-similar slug.

narayanaditya95:
@pbanaszkiewicz @gvwilson 
Ready to be merged.

pbanaszkiewicz:
We'll merge this in for v1.6.1 or v1.7, whichever is sooner.

pbanaszkiewicz:
@narayanaditya95 do you think you could rebase this PR and rework the migrations so that they apply on top of the current database schema?

workshops/migrations/0088_auto_20160530_1029.py
pbanaszkiewicz: Have you run into troubles with the current approach? If yes, then please add such a loop.

narayanaditya95:
Sure.

narayanaditya95:
@pbanaszkiewicz ping.

PR #821, opened 2016-06-01
host_details: fix 500 instead of 404 when no matching host

pbanaszkiewicz:
The issue was: we should have used `get_object_or_404` instead of
`Host.objects.get` because the latter throws error 500 when specific
host cannot be found.

Additionally changed one test's name so that it's discovered by the test
runner.

### DIFF
--- a/workshops/test/test_views_404.py
+++ b/workshops/test/test_views_404.py
@@ -54,7 +54,12 @@ class TestViewsFor404ing(TestBase):
         rv = self.client.get(url)
         self.assertEqual(rv.status_code, 404)
 
-    def todos_add(self):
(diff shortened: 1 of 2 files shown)


gvwilson:
:+1: to merge, but how did we wind up in this situation?  Aren't hosts chosen from a drop-down (which should be guaranteed to be in the DB)?

pbanaszkiewicz:
@gvwilson probably someone accessed some old link or something.

gvwilson:
Hm - can you please have a look at the server logs and see who was on the system and what they might have tried? I doubt anyone would have bookmarked a particular organization's link, and typing in the name manually seems unlikely too.

PR #823, opened 2016-06-05
WIP: Bulk import speakers from  PyData site

narayanaditya95:
@gvwilson @pbanaszkiewicz 
This is a test implementation of the bulk import of speaker profiles from a PyData conference site.
See the implementation of the API at [#115](https://github.com/pydata/conf_site/pull/115).

I have reused most of the bulk add person machinery, since we are interested in importing all speakers using the API.

There are few issues that need to be discussed.

- The `Speaker` model at the PyData site uses a single `name` field while AMY uses `personal`(req), `middle`(optional) and `family`(required) name fields. How do I populate the two fields at AMY?
```py
class Speaker(models.Model):

    SESSION_COUNT_CHOICES = [
        (1, "One"),
        (2, "Two")
    ]

    user = models.OneToOneField(User, null=True, related_name="speaker_profile", verbose_name=_("User"))
    name = models.CharField(verbose_name=_("Name"), max_length=100,
                            help_text=_("As you would like it to appear in the"
                                        " conference program."))
    biography = models.TextField(blank=True, help_text=_("A little bit about you.  Edit using "
                                                         "<a href='http://warpedvisions.org/projects/"
                                                         "markdown-cheat-sheet/target='_blank'>"
                                                         "Markdown</a>."), verbose_name=_("Biography"))
    biography_html = models.TextField(blank=True)
    photo = models.ImageField(upload_to="speaker_photos", blank=True, verbose_name=_("Photo"))
    annotation = models.TextField(verbose_name=_("Annotation"))  # staff only
    invite_email = models.CharField(max_length=200, unique=True, null=True, db_index=True, verbose_name=_("Invite_email"))
    invite_token = models.CharField(max_length=40, db_index=True, verbose_name=_("Invite token"))
    created = models.DateTimeField(
        default=datetime.datetime.now,
        editable=False,
        verbose_name=_("Created")
    )
```

- Should the API at Pydata site be rewritten to confirm with AMY's database schema. See the [commit](https://github.com/pydata/conf_site/pull/115/commits/6b62b256ced84ba0189123015cb4093be6f52a09) that renames the `name` field of the API to `personal` to make it easier at AMY's end to populate the `Person` table.

- Authentication!: The API at PyData only accepts requests from admin users. I have currently disabled the authentication while testing this PR.
The token authentication methods listed in the [DRF documentation](http://www.django-rest-framework.org/api-guide/authentication/#api-reference) requires the API to be served over HTTPS. This is not true for all PyData conference sites. See <http://pydata.org/paris2016/>.
I am not sure about the Session authentication method.


workshops/templates/workshops/person_bulk_add_form.html
```diff
@@ -1,13 +1,10 @@
 {% extends "base_nav_fixed.html" %}
 
 {% block content %}
+{% load crispy_forms_tags %}
 {% load staticfiles %}
+
 <p><a href="{% url 'person_bulk_add_template' %}">BulkPersonAddTemplate.csv</a> is a blank template of a well-formed .CSV file.</p>
-<form action="" method="post" enctype="multipart/form-data"
```
narayanaditya95: Do we have to show that the "import using the URL field" is a functionality exclusive to PyData events?
But since PyData will be hosting a separate instance of AMY, this shouldn't be a problem.

workshops/views.py
```diff
@@ -554,13 +554,33 @@ def person_bulk_add(request):
                     # the data
                     request.session['bulk-add-people'] = persons_tasks
                     return redirect('person_bulk_add_confirmation')
+        elif form.is_valid() and form.cleaned_data['event']:
+            event = form.cleaned_data['event']
+            try:
+                r = requests.get(event.url + '/api/speaker/')
```
narayanaditya95: This assumes that the url is saved with no trailing '/'.
What is the accepted practice here?
Do I override the `save()` method of the `Person` model to ensure such a constraint?

workshops/views.py
```diff
@@ -554,13 +554,33 @@ def person_bulk_add(request):
...
+                    'Request for "{0}" returned status code {1}'
+                    .format(event.url, e.response.status_code)
+                )
+            except (requests.exceptions.ConnectionError,
+                    requests.exceptions.Timeout):
+                error_messages.append("Network connection error.")
+            else:
+                for person in persons_tasks:
+                    person['event'] = event.slug
+                    person['role'] = 'instructor'
```
narayanaditya95: The role should probably be a field in the API response. The PyData site instance knows best "who did what".
@pbanaszkiewicz @gvwilson ?

gvwilson:
Hi Aditya,

I would split the "name" field on spaces - if there are two values, 
assume they're first and last, and if it's anything else, display it as 
plain text and have the user edit the "personal" and "family" columns.  
The problem with modifying the PyData data model is all the old 
instances that are out there that we don't want to have to update...

Cheers,

Greg


-- 
Dr Greg Wilson
Director of Instructor Training
Software Carpentry Foundation

workshops/templates/workshops/person_bulk_add_form.html
gvwilson: Won't be a problem.

workshops/views.py
gvwilson: @pbanaszkiewicz ?

workshops/views.py
gvwilson: Yes, I think the role should be a field in the API response.  If AMY 
doesn't recognize the value, it should flag it as an error.

narayanaditya95:
Closing this, due to a schema design change.

PR #825, opened 2016-06-07
Github auth + update profile form

chrismedrela:
This is still in progress, not ready for review. I'll split it into multiple commits at the end, after review. 

If you want to run this on your local machine, you need to:

1. Install localtunnel: `npm install -g localtunnel`

2. Expose your 8000 port to the world: `lt --port 8000 --subdomain amytest`. Amy will be available at https://amytest.localtunnel.me. It's important that you use `amytest` subdomain. This tool often exits abnormally, so you just need to rerun it from time to time. 

3. Set `SOCIAL_AUTH_GITHUB_KEY` and `SOCIAL_AUTH_GITHUB_SECRET` environment variables. Both values can be generated in Github > Settings > OAuth applications > Developer Applications. You need to set Homepage URL to `https://amytest.localtunnel.me` and Autorization Callback URL to `http://amytest.localtunnel.me/complete/github/`. Or you can use values provided by me. 

4. Run server: `python3 manage.py runserver 0.0.0.0:8000`.

5. Open in browser: https://amytest.localtunnel.me/account/login.

Note that you can login using github account only if your github account is associated with your Person record. This is represented in `UserSocialAuth` table. So if you want to login successfully you need to execute:

```python
from workshops.models import *
from social.apps.django_app.default.models import UserSocialAuth

UserSocialAuth.objects.create(provider='github', uid='<your github uid>', 
                              user=Person.objects.get(username='<your username>'))
```

You can find out your github uid here: http://caius.github.io/github_id/

@pbanaszkiewicz, deadline for version 1.6 is June 10. I think it's possible to get this branch ready (and implement update profile form) for review til June 10, but I don't think we will merge this before the deadline. Is it hard deadline?


gvwilson:
We really want to invite people to register for future instructor 
training events starting next week, and would like to have 'update 
profile' working at the same time, which means getting GitHub login 
working.  Better to delay than deploy something that's buggy, but best 
of all to deploy something that works.  @pbanaszkiewicz, does it help if 
we push the cut date to June 13, and you guys take some time off next 
week in swap?

pbanaszkiewicz:
@chrismedrela @gvwilson postponing the deadline definitely helps, but we (me and Chris) have a range of exams coming up (control theory this Friday, optimal control [which is *really* hard, sigh…] on Wednesday) so we're kind of limited on time here. Not sure what to do…

gvwilson:
Let's push back a week - will that help?

pbanaszkiewicz:
@gvwilson it does help me, @chrismedrela?

chrismedrela:
It does help me too.

chrismedrela:
@gvwilson, @pbanaszkiewicz, This branch is ready for first review. It lacks tests. The code is not as good as I want it to be (especially naming and wording), but I'd like to get your feedback. I'll split this commit into smaller ones after the review.

amy/urls.py
```diff
@@ -40,4 +42,10 @@
 
     # REST API v1
     url(r'^api/v1/', include('api.urls')),
+
+    # Login with github credentials
+    url('', include('social.apps.django_app.urls', namespace='social')),
+    url(r'^workshops/no_github_user/$',
```
gvwilson: Is this a URL a human being would ever see or bookmark?

api/test/test_export.py
```diff
@@ -46,10 +46,12 @@ def setUp(self):
...
-        self.user1 = Person.objects.create_user('user1', 'User1', 'Name',
-                                                'user1@name.org')
-        self.user2 = Person.objects.create_user('user2', 'User2', 'Name',
-                                                'user2@name.org')
+        self.user1 = Person.objects.create_user(
+            username='user1', email='user1@name.org',
+            personal='User1', family='Name')
+        self.user2 = Person.objects.create_user(
+            username='user2', email='user2@name.org',
+            personal='User2', family='Name')
```
gvwilson: why this change?

workshops/forms.py
```diff
@@ -917,3 +917,39 @@ class Meta:
             'teaching_frequency_expectation': forms.RadioSelect(),
             'max_travelling_frequency': forms.RadioSelect(),
         }
+
+
+class AutoUpdateProfileForm(forms.ModelForm):
+    username = forms.CharField(disabled=True)
+
+    ERROR_MSG = ("This github account is already associated with different "
+                 "Amy account. Log out and log in using github account.")
```
gvwilson: AMY (all caps) and GitHub (CamelCase)

workshops/forms.py
```diff
@@ -917,3 +917,39 @@ class Meta:
...
+            'gender',
+            'may_contact',
+            'airport',
+            'github',
+            'twitter',
+            'url',
+            'username',
+            'affiliation',
+            'domains',
+            'lessons',
```
gvwilson: This list seems to show up elsewhere (e.g., look for the comment "non-M2M-relationships" in views.py). If the lists actually are the same, can we stuff them into a constant somewhere so that everything stays in sync?

workshops/models.py
```diff
@@ -165,7 +165,7 @@ class PersonManager(BaseUserManager):
       $ python manage.py createsuperuser
     """
 
-    def create_user(self, username, personal, family, email, password=None):
+    def create_user(self, username, email, personal=None, family=None, password=None):
```
gvwilson: Why defaults for `personal` and `family`?

workshops/models.py
```diff
@@ -179,7 +179,7 @@ def create_user(self, username, personal, family, email, password=None):
         user.save(using=self._db)
         return user
 
-    def create_superuser(self, username, personal, family, email, password):
+    def create_superuser(self, username, email, password, personal=None, family=None):
```
gvwilson: Why defaults for `personal` and `family`?

workshops/models.py
```diff
@@ -271,26 +288,26 @@ class Person(AbstractBaseUser, PermissionsMixin):
...
     def get_full_name(self):
         middle = ''
         if self.middle is not None:
             middle = ' {0}'.format(self.middle)
-        return '{0}{1} {2}'.format(self.personal, middle, self.family)
+        return '{0}{1} {2}'.format(self.personal or '', middle, self.family or '')
 
     def get_short_name(self):
-        return self.personal
+        return self.personal or self.username
```
gvwilson: Um... why this change? And why is it this PR?

workshops/models.py
```diff
@@ -271,26 +288,26 @@ class Person(AbstractBaseUser, PermissionsMixin):
...
 
     def __str__(self):
         result = self.get_full_name()
         if self.email is not None:
             result += ' <' + self.email + '>'
+        if self.username is not None:
+            result += ' (' + self.username + ')'
```
gvwilson: This is going to make name display pretty ugly...

workshops/models.py
```diff
@@ -312,8 +329,8 @@ def clean(self):
     def save(self, *args, **kwargs):
         # save empty string as NULL to the database - otherwise there are
         # issues with UNIQUE constraint failing
-        self.personal = self.personal.strip()
-        self.family = self.family.strip()
+        self.personal = self.personal.strip() if self.personal else None
+        self.family = self.family.strip() if self.family else None
```
gvwilson: Since when do we allow `None` for *both*?

workshops/socialauth.py
```diff
@@ -0,0 +1,76 @@
+from django.http import HttpResponseForbidden
+
```
gvwilson: why the blank lines?

workshops/socialauth.py
```diff
@@ -0,0 +1,76 @@
...
+            return redirect(reverse('no_github_user'))
+
+
+def find_user_or_abort(details=None, **kwargs):
+    username = details['username']
+    try:
+        user = Person.objects.get(github=username)
+        return {'user': user}
+    except Person.DoesNotExist:
+        raise NoUserAssociatedWithGithubAccount
```
gvwilson: Does this need `()` ?

workshops/socialauth.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def debug(*args, **kwargs):
+    print('---')
+    for k, v in kwargs.items():
+        print('{} -> {}'.format(k, v))
```
gvwilson: remove

workshops/socialauth.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def is_admin(user):
+    if user:
+        return user.groups.filter(name='Admins').exists() or user.is_staff or user.is_superuser
```
gvwilson: long line

workshops/socialauth.py
```diff
@@ -0,0 +1,76 @@
...
+    return f
+
+
+@results_decorator
+def selectable_only_for_admins(request):
+    user = getattr(request, 'user', None)
+    if user is None or not user.is_authenticated():
+        return HttpResponse(status=401)  # Unauthorized
+    elif not is_admin(user):
+        return HttpResponseForbidden()
```
gvwilson: Is an `else` required, or does `results_decorator` handle that?

workshops/views.py
```diff
@@ -208,9 +207,17 @@ def form_valid(self, form):
 #------------------------------------------------------------
 
 
-@login_required
+def dispatch(request):
+    if request.user and is_admin(request.user):
+        return redirect(reverse('dashboard'))
+    else:
+        return redirect(reverse('trainee-dashboard'))
```
gvwilson: I generally dislike "significant omissions": if one is called `trainee-dashboard`, the other should be called `admin-dashbord`, not just `dashboard`.

workshops/views.py
```diff
@@ -2982,3 +2989,44 @@ def get_context_data(self, **kwargs):
...
+def trainee_dashboard(request):
+    context = {
+        'title': 'Your profile',
+        'user': request.user,
+    }
+    return render(request, 'workshops/trainee_dashboard.html', context)
+
+
+@login_required
+def autoupdate_profile(request):
```
gvwilson: Why `autoupdate` ?

gvwilson:
Reviewed what I could.  Would very much like tests.  And yes, please 
break into pieces.

workshops/models.py
```diff
@@ -221,34 +221,51 @@ class Person(AbstractBaseUser, PermissionsMixin):
...
-    airport     = models.ForeignKey(Airport, null=True, blank=True, on_delete=models.PROTECT)
-    github      = models.CharField(max_length=STR_MED, unique=True, null=True, blank=True)
-    twitter     = models.CharField(max_length=STR_MED, unique=True, null=True, blank=True)
-    url         = models.CharField(max_length=STR_LONG, null=True, blank=True)
+    airport     = models.ForeignKey(Airport, null=True, blank=True, on_delete=models.PROTECT,
+                                    verbose_name='Nearest major airport')
+    github      = models.CharField(
+        max_length=STR_MED, unique=True, null=True, blank=True,
+        verbose_name='GitHub username',
+        help_text='Please provide your username, not a numeric user ID.')
```
pbanaszkiewicz: Why would anyone provide a numeric user ID?

workshops/models.py
```diff
@@ -221,34 +221,51 @@ class Person(AbstractBaseUser, PermissionsMixin):
...
         through="Qualification",
+        verbose_name='Topic and lessons you\'re comfortable teaching',
+        help_text='Please mark ALL that apply.',
         blank=True,
     )
     domains = models.ManyToManyField(
         "KnowledgeDomain",
         limit_choices_to=~Q(name__startswith='Don\'t know yet'),
+        verbose_name='Areas of expertise',
+        help_text='Please check all that apply.',
```
pbanaszkiewicz: So "check" or…

workshops/models.py
```diff
@@ -221,34 +221,51 @@ class Person(AbstractBaseUser, PermissionsMixin):
...
+                  'you affiliated with (if any)?')
 
     badges = models.ManyToManyField(
         "Badge", through="Award",
         through_fields=('person', 'badge'))
     lessons = models.ManyToManyField(
         "Lesson",
         through="Qualification",
+        verbose_name='Topic and lessons you\'re comfortable teaching',
+        help_text='Please mark ALL that apply.',
```
pbanaszkiewicz: …or "mark"?

workshops/socialauth.py
pbanaszkiewicz: Please conform to our styling:

1. sorted alphabetically
2. grouped by dependency, ie. system modules in the first group, related dependencies (e.g. Django, or selectable) in the second group, and project imports in the third group.

workshops/socialauth.py
```diff
@@ -0,0 +1,76 @@
...
+
+def abort_if_no_user(user, *args, **kwargs):
+    if user is None:
+        raise NoUserAssociatedWithGithubAccount
+
+
+class NoUserAssociatedWithGithubAccountMiddleware(object):
+    def process_exception(self, request, exception):
+        if isinstance(exception, NoUserAssociatedWithGithubAccount):
+            return redirect(reverse('no_github_user'))
```
pbanaszkiewicz: I have no idea what this does. Add comments maybe?

workshops/socialauth.py
pbanaszkiewicz: Can this be in Person manager? Or new/custom `PersonQuerySet`?

workshops/templates/account/no_github_user.html
```diff
@@ -0,0 +1,12 @@
+{% extends "account/base.html" %}
+
+{% block content %}
+<div class="row">
+  <div class="col-md-12">
+    <div class="alert alert-danger" role="alert"><strong>Error:</strong>
+    No account is associated with this github account.</div>
```
pbanaszkiewicz: Spelling

amy/settings.py
```diff
@@ -15,6 +15,7 @@
 import sys
 from django.utils.translation import ugettext_lazy as _
 
+
```
pbanaszkiewicz: newline driven change

amy/settings.py
```diff
@@ -88,7 +89,7 @@
         'OPTIONS': {
             'debug': DEBUG,
 
-            # default processors + a request processor + amy-version
+            # default processors + a request processor + amy-version + python-social-auth
```
pbanaszkiewicz: isn't this line too long?

amy/settings.py
```diff
@@ -195,6 +200,36 @@
...
+
+
+SOCIAL_AUTH_PIPELINE = (
+    'workshops.socialauth.debug',
+    'social.pipeline.social_auth.social_details',
+    'social.pipeline.social_auth.social_uid',
+    'social.pipeline.social_auth.auth_allowed',
+    'workshops.socialauth.find_user_or_abort',
+    # get user from
+    # 'social.pipeline.social_auth.social_user',  # set kwargs['social'] and kwargs['user'] if UserSocialAuth record exists
```
pbanaszkiewicz: multiple comments in line? Line too long?

amy/settings.py
```diff
@@ -195,6 +200,36 @@
...
+SOCIAL_AUTH_GITHUB_KEY = os.environ.get('SOCIAL_AUTH_GITHUB_KEY', '').strip()
+SOCIAL_AUTH_GITHUB_SECRET = os.environ.get('SOCIAL_AUTH_GITHUB_SECRET', '').strip()
+if not SOCIAL_AUTH_GITHUB_KEY or not SOCIAL_AUTH_GITHUB_SECRET:
+    print('Logging using github account will *not* work, '
+          'because you didn\'t set SOCIAL_AUTH_GITHUB_KEY and/or '
+          'SOCIAL_AUTH_GITHUB_SECRET environment variables.')
+
+
+SOCIAL_AUTH_PIPELINE = (
+    'workshops.socialauth.debug',
```
pbanaszkiewicz: is this `debug` necessary? What does it do?

workshops/templates/forms/autoupdateprofile.html
```diff
@@ -0,0 +1,12 @@
...
+{% load crispy_forms_tags %}
+{% load selectable_tags %}
+
+{% block content %}
+{% crispy form form_helper %}
+{% endblock %}
+
+{% block extrajs %}
+{{ form.media }}
+{% endblock extrajs %}
```
pbanaszkiewicz: I believe we have a generic form template somewhere. Perhaps you could reuse it.

workshops/views.py
```diff
@@ -208,9 +207,17 @@ def form_valid(self, form):
 #------------------------------------------------------------
 
 
-@login_required
+def dispatch(request):
```
pbanaszkiewicz: No docstring, no comments… And the `dashboard` view stays the "Home page"?

pbanaszkiewicz:
@chrismedrela @gvwilson 
What I see in this PR is that this feature requires us to change `Person` model (make `personal` and `family` nullable), but we can only allow specific people to log in… Can we somehow work around this constraint?

And why is this change required?

Other thing is that we don't have a mechanism for adding people to the "github-login-enabled" list, other than making entries in the `UserSocialAuth` model by hand.

Can people log in even though they're marked as "inactive" (`.is_active = False`)?

I don't trust this code -- this is opening AMY for other people, something that is not tested thoroughly.

@chrismedrela Finally, there's a matter of communication. In most places I'm guessing why you made a specific change, because you're not communicating why. The code lacks comments, lacks docstrings, and even lacks commits -- go through @wking's commits to see how he communicates rationale in his commits.

amy/urls.py
chrismedrela: I've deleted that. Now, user is redirected to `login` page with error message.

api/test/test_export.py
chrismedrela: It's related to neither GitHub auth nor update profile form. I think it's clear what is the meaning of all arguments now.

workshops/forms.py
chrismedrela: This list is different. For example, it contains M2M fields (like `domains` and `lessons`) and doesn't contain all non-relational fields (for example `notes`).

workshops/models.py
chrismedrela: Oops, I left mess in code. I've reverted back to required ones.

workshops/models.py
chrismedrela: Again, I've reverted this back. At some point I thought that `personal` should be optional, but it's not the case.

workshops/models.py
chrismedrela: OK, reverted back.

workshops/socialauth.py
chrismedrela: No, it's equivalent to `raise NoUserAssociatedWithGithubAccount()`. See http://stackoverflow.com/questions/16706956/is-there-a-difference-between-raise-exception-and-raise-exception-without

workshops/socialauth.py
chrismedrela: OK, I've made this a method in `Person`.

workshops/socialauth.py
chrismedrela: When `None` is returned, the response is what the view returns. I've added explicit `else: return None`.

workshops/models.py
chrismedrela: Reverted back.

workshops/forms.py
chrismedrela: Done.

workshops/models.py
chrismedrela: Reverted back.

workshops/socialauth.py
chrismedrela: Removed.

workshops/views.py
chrismedrela: OK, renamed `dashboard` to `admin-dashboard` (and related templates).

workshops/models.py
chrismedrela: I've deleted this `help_text`.

workshops/models.py
chrismedrela: Changed to `Please check all that apply.`

workshops/templates/account/no_github_user.html
chrismedrela: Do you mean `GitHub` instead of `github`?

amy/settings.py
chrismedrela: No, it's sth that I should remove before pushing code.

workshops/templates/account/no_github_user.html
pbanaszkiewicz: Yes, I meant github→GitHub. Sorry for not being explicit.

workshops/templates/account/login.html
```diff
@@ -27,6 +27,15 @@
           <input type="submit" class="btn btn-default" value="Log in">
         </div>
       </div>
+
+      <div class="form-group">
+        <div class="col-sm-offset-2 col-sm-10">
+          <a class="btn btn-default" id="github-button" name="github" href="/login/github/">
+            Log in with Github account
```
pbanaszkiewicz: Github → GitHub

workshops/templates/forms/autoupdateprofile.html
chrismedrela: We have, but it inherits from `base_nav_fixed.html`. I've renamed this template to `generic_form_nonav.html`.

workshops/socialauth.py
chrismedrela: Done.

workshops/templates/account/login.html
chrismedrela: Changed to `Log in with your Github account`.

pbanaszkiewicz: You forgot about "Git**H**ub"

chrismedrela:
I've split this PR into smaller commits, cleaned the code a bit and made changes that you all proposed. I didn't manage to write tests today. This is the next thing that I'll do.

1. `Person.personal` and `Person.family` fields are still required. I left mess in code, because I had made those fields optional, and then reverted back to required ones, but I forget to be consistent in all places. I've removed that mess.

2. I named the form `AutoUpdateProfileForm`, because I couldn't get with a better name. I don't want to name it `UpdateProfileForm`, because it's very close to `ProfileUpdateRequest`. Can you think of a better name?

3. I've created a new "Admins" group that includes Admins, Mentors and Examiners. Access to all views (except for trainee dashboard and autoupdate request form) is restricted to users who are in this group. Unfortunately, we already have a group named "administrators" (only for real Admins, *not* for Mentors or Examiners). Can you think of a better name for this group, so we avoid name collision?

4. GitHub Authentication works in the following way:

  - Functions from `SOCIAL_AUTH_PIPELINE` are executed in order, passing `**kwargs` to each one. This dictionary is updated with the result of each function and then passed to the next function in pipeline. At some point you need to set `kwargs['user']` to Person record.

  - `find_user_or_abort` searches for an active user with given GitHub nick. If such user exists, we`kwargs['user']` is set. If there is no such user, we raise `NoPersonAssociatedWithGithubAccount` exception that is caught by `GithubAuthMiddleware`. This middleware adds "No account is associated with your GitHub account." message and then, redirects to login page, where the error message is displayed.

  - We ignore records from `UserSocialAuth` table.

5. If we want to add a person to "github-login-enabled" list, it's enough to set their `Person.github`.

6. I've changed the code, so that inactive people cannot log in.

workshops/templates/account/login.html
chrismedrela: Done. :-)

amy/settings.py
```diff
@@ -196,6 +201,40 @@
...
+    'social.backends.github.GithubOAuth2',
+    'django.contrib.auth.backends.ModelBackend',
+)
+SOCIAL_AUTH_ADMIN_USER_SEARCH_FIELDS = ['github']
+SOCIAL_AUTH_GITHUB_KEY = os.environ.get('SOCIAL_AUTH_GITHUB_KEY', '').strip()
+SOCIAL_AUTH_GITHUB_SECRET = os.environ.get('SOCIAL_AUTH_GITHUB_SECRET', '').strip()
+if not SOCIAL_AUTH_GITHUB_KEY or not SOCIAL_AUTH_GITHUB_SECRET:
+    print('Logging using github account will *not* work, '
+          'because you didn\'t set SOCIAL_AUTH_GITHUB_KEY and/or '
+          'SOCIAL_AUTH_GITHUB_SECRET environment variables.')
```
pbanaszkiewicz: `SOCIAL_AUTH_GITHUB_KEY` and `SOCIAL_AUTH_GITHUB_SECRET` should be required if `not DEBUG`, and this message should go to `stderr` because chances are only that output is captured in logs.

requirements.txt
```diff
@@ -16,3 +16,5 @@ fake-factory>=0.5.3,<0.6
 drf-nested-routers>=0.11.1,<0.12
 django_compressor>=2.0,<2.1
 pygithub>=1.26,<=1.27
+python-social-auth
+Requests-OAuthlib
```
pbanaszkiewicz: You should pin the versions

pbanaszkiewicz:
@chrismedrela:

Ad 2: maybe `SelfUpdateProfileForm` or `UpdateSelfProfileForm`?
Ad 3: maybe `managers`? What's the point of this group anyway?
Ad 4: can we add entries to `UserSocialAuth` on the fly when we find someone in `Person` table with matching github username? This is just an alternative approach giving us lightly more control.

Ad rest: excellent, thanks!

@chrismedrela as you can see, some of recently merged PRs caused some conflicts and probably need a merging migration too.

@gvwilson the deadline was yesterday, but there's still some things to do in this PR. How soon should we release v1.6?

@chrismedrela can you please provide me with `SOCIAL_AUTH_GITHUB_KEY` and `SOCIAL_AUTH_GITHUB_SECRET` values for `amy-dev.software-carpentry.org` domain?

@gvwilson, @narayanaditya95 looking forward to other ideas for naming (see "Ad 2", "Ad 3" above).

chrismedrela:
@pbanaszkiewicz, we have one group of AMY users: Admins, Mentors and Examiners. We'll have another one: Trainees. We need to restrict most views to the first group, that is only people who are in "Admins" group can view these views. That's the point of this group.

You can create your own pair of `SOCIAL_AUTH_GITHUB_{KEY, SECRET}`. Go to Github settings > OAuth applications > Developer Applications > Register a new application. Set homepage URL to `https://amy-dev.software-carpentry.org` and authorization callback URL to `https://amy-dev.software-carpentry.org/complete/github/`.

chrismedrela:
@pbanaszkiewicz, ready to be merged.

pbanaszkiewicz:
@chrismedrela I run into errors when migrating a fake database with your migrations. Can you confirm such error?

When applying 0097_add_admins_group.py:
```
TypeError: int() argument must be a string, a bytes-like object or a number, not 'Group'
```

pbanaszkiewicz:
I confirm this error on a separate machine.

Here's what I did:

```
make dev_database (which does rm db.sqlite3, migrate, fake_database)
checkout && pull your branch
make upgrade
./manage.py migrate
```

For some very strange reason Django migrates some old migrations first:

```
Operations to perform:
  Apply all migrations: reversion, contenttypes, workshops, auth, admin, sessions, default
Running migrations:
  Rendering model states... DONE
  Applying default.0001_initial... OK
  Applying default.0002_add_related_name... OK
  Applying default.0003_alter_email_max_length... OK
  Applying default.0004_auto_20160423_0400... OK
  Applying workshops.0097_add_admins_group...Traceback (most recent call last):
  File "./manage.py", line 10, in <module>
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/core/management/__init__.py", line 345, in execute
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/core/management/base.py", line 399, in execute
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/core/management/commands/migrate.py", line 200, in handle
    executor.migrate(targets, plan, fake=fake, fake_initial=fake_initial)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/migrations/executor.py", line 92, in migrate
    self._migrate_all_forwards(plan, full_plan, fake=fake, fake_initial=fake_initial)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/migrations/executor.py", line 121, in _migrate_all_forwards
    state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/migrations/executor.py", line 198, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/migrations/migration.py", line 123, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/migrations/operations/special.py", line 183, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/home/piotr/workspace/SwC/amy/workshops/migrations/0097_add_admins_group.py", line 14, in forward
    p.groups.add(group)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/fields/related_descriptors.py", line 843, in add
    self._add_items(self.source_field_name, self.target_field_name, *objs)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/fields/related_descriptors.py", line 986, in _add_items
    '%s__in' % target_field_name: new_ids,
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/query.py", line 790, in filter
    return self._filter_or_exclude(False, *args, **kwargs)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/query.py", line 808, in _filter_or_exclude
    clone.query.add_q(Q(*args, **kwargs))
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/sql/query.py", line 1243, in add_q
    clause, _ = self._add_q(q_object, self.used_aliases)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/sql/query.py", line 1269, in _add_q
    allow_joins=allow_joins, split_subq=split_subq,
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/sql/query.py", line 1199, in build_filter
    condition = lookup_class(lhs, value)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/lookups.py", line 19, in __init__
    self.rhs = self.get_prep_lookup()
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/fields/related_lookups.py", line 56, in get_prep_lookup
    self.lookup_name, self.rhs)
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/fields/__init__.py", line 746, in get_prep_lookup
    return [self.get_prep_value(v) for v in value]
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/fields/__init__.py", line 746, in <listcomp>
    return [self.get_prep_value(v) for v in value]
  File "/home/piotr/workspace/anaconda/envs/amy/lib/python3.5/site-packages/django/db/models/fields/__init__.py", line 976, in get_prep_value
    return int(value)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'Group'
```

These old migrations were used by `make dev_database` call:
```
Operations to perform:
  Apply all migrations: contenttypes, reversion, sessions, auth, workshops, admin
Running migrations:
  Rendering model states... DONE
  Applying contenttypes.0001_initial... OK
  Applying auth.0001_initial... OK
  Applying workshops.0001_initial... OK
  Applying admin.0001_initial... OK
  Applying admin.0002_logentry_remove_auto_add... OK
  Applying contenttypes.0002_remove_content_type_name... OK
  Applying auth.0002_alter_permission_name_max_length... OK
  Applying auth.0003_alter_user_email_max_length... OK
  Applying auth.0004_alter_user_username_opts... OK
  Applying auth.0005_alter_user_last_login_null... OK
  Applying auth.0006_require_contenttypes_0002... OK
  Applying auth.0007_alter_validators_add_error_messages... OK
  Applying reversion.0001_initial... OK
  Applying reversion.0002_auto_20141216_1509... OK
  Applying sessions.0001_initial... OK
  Applying workshops.0002_auto_20150219_1305... OK
  Applying workshops.0003_remove_person_slug... OK
  Applying workshops.0003_auto_20150227_1245... OK
  Applying workshops.0004_merge... OK
```

chrismedrela:
@pbanaszkewicz, which branch are you on when executing `make dev_database`?

pbanaszkiewicz:
@chrismedrela `develop`

chrismedrela:
I reproduced this on my machine.

chrismedrela:
Those `default.000*_*` migrations are not old migrations. They are introduced by `python-social-auth`.

chrismedrela:
OK, this is actually bug in my migration code. What is really strange is that this bug didn't appear when you make dev_database on `github-login` branch.

pbanaszkiewicz:
I'm glad that we have the amy-dev instance. What a surprise it would be if I upgraded the production in this state.

chrismedrela:
I've added auth security tests and fixed the bug in migrations. This version should work on `amy-dev` (no issues with migrations on backup of production data on my machine). It's still missing tests for the update profile form, but that should be much easier and I'll write them soon. In the meantime, you can make final review.

I tried to modify according commits instead of adding new one, but I gave up after lots of rebases and reflogs.

workshops/models.py
```diff
@@ -343,11 +343,6 @@ def is_staff(self):
         """
         return self.is_superuser
 
-    @property
-    def is_admin(self):
-        is_admin = self.groups.filter(name='Admins').exists()
-        return is_admin or self.is_staff or self.is_superuser
-
```
pbanaszkiewicz: Why is this no longer a property?

workshops/util.py
```diff
@@ -877,7 +874,25 @@ def merge_objects(object_a, object_b, easy_fields, difficult_fields,
...
+        'You already used some access control decorator on this view.'
+    f = login_required(f)
+    f.access = 'only_for_logged_in'
+    return f
+
+
+def login_not_required(f):
+    assert not hasattr(f, 'access'), \
+        'You already used some access control decorator on this view.'
+    f.access = 'login_not_required'
```
pbanaszkiewicz: What does this magic do in these 3 functions?

workshops/models.py
```diff
@@ -369,6 +364,14 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
 
+def is_admin(user):
+    if user is None or user.is_anonymous():
+        return False
+    else:
+        is_admin = user.groups.filter(name='administrators').exists()
+        return is_admin or user.is_staff or user.is_superuser
```
pbanaszkiewicz: 1. I'm unsure the use of `user.is_staff`. What if we change its meaning later on so that it will break logic here? Dropping `or user.is_staff` condition here would remove the only use of `user.is_staff` we have right now -- so this field is only used by the Django auth.

2. I'd rename `is_admin` variable from inside `is_admin` function to something like `has_admin_group`.

workshops/test/test_security.py
```diff
@@ -0,0 +1,254 @@
...
+
+                # v.callback is always a function, even when the view is
+                # class based. We try to find the class implementing the view.
+                view_class = getattr(v.callback, 'view_class', None)
+                cls = getattr(v.callback, 'cls', None)
+                class_ = view_class or cls or None
+
+                is_function_based_view = (model_admin is None and
+                                          admin_site is None and
+                                          class_ is None)
```
pbanaszkiewicz: Can you just test if `v` is an instance(/subclass?) of a `View` (probably `django.views.generic.View`)?

workshops/test/test_security.py
```diff
@@ -0,0 +1,254 @@
...
+
+                if is_function_based_view:
+                    self.assertTrue(access is not None,
+                                    "You have a function based view with"
+                                    "no access control decorator. This view is"
+                                    "probably accessible to every user. If "
+                                    "this is what you want, use "
+                                    "@login_not_required decorator.")
```
pbanaszkiewicz: You're dropping spaces when concatenating string in parentheses, for example
```
                                    "You have a function based view with"
                                    "no access control decorator. This view is"
```

becomes `"You have a function based view withno access control decorator. This view is"`.

pbanaszkiewicz: Additionally I can see inconsistency with using `'` vs `"` for strings.

AMY's style evolved over time and what I now do is:
* `'` for strings everywhere
* except for docstrings which are `"""` in most cases, except if I can manage to fit in one line by dropping `""`, then I'm using `"`

So:
```python
def function():
    """Do this and that."""
    v = 'Test string'
    return v
```

workshops/test/test_security.py
```diff
@@ -0,0 +1,254 @@
...
+                    self.assertIn(access, ['only_for_admins',
+                                           'only_for_logged_in',
+                                           'login_not_required'])
```
pbanaszkiewicz: Okay, I can see why you needed that magic over there.

Have you thought about saving/working with explicit permissions list? If so, what are your thoughts?

workshops/test/test_security.py
```diff
@@ -0,0 +1,254 @@
...
+
+                        self.assertNotEqual(
+                            len(found), 0,
+                            "This view lacks access control mixin and is "
+                            "probably accessible to every user. If this is "
+                            "what you want, use LoginNotRequiredMixin.")
+                        self.assertEqual(
+                            len(found), 1,
+                            "You have more than one access control mixin "
+                            "defined in this view.")
```
pbanaszkiewicz: You have a test for no permission mixin, and you have a test for two excluding permission mixins. Why f-b-views don't have the latter test? Because you *can* chain decorators if you want, so…

workshops/util.py
pbanaszkiewicz: Okay, I understand why you're setting `f.access` here.
1. docstrings!!!
2. functions and classes can set their own parameters, and you can overwrite them. I'd use `._access_control` or some other "hidden" name instead of `.access`.

What's more is that this kind of decorators basically hide "permissions by inheritance" model @gvwilson was talking about yesterday.

Writing `@perms_required(['workshops.do_a', 'workshops.do_b'])` gives us more control over the perms (and I know at least one place where we have to use this approach to give only specific parts accessible to some user groups), while writing `@groups_required(['administrators'])` is cozy and can make me lazy (`@only_for_admins` and `@only_for_logged_in` are basically other names for decorators that already exist) -- I can forget to update `administrators` permissions list, for example.

I kinda feel mixing different decorators is not good.

Finally there's some naming convention I feel would be better here. I like `_required` approach, so maybe:

* `admin_required` instead of `only_for_admins`
* Django's `login_required` imported as `django_login_required` and `only_for_logged_in` renamed to `login_required`
* `login_not_required` is fine

@gvwilson any additional comments? Especially regarding more control (explicit permissions list) vs less work (explicit/implicit group-based permissions).

pbanaszkiewicz:
@chrismedrela excellent work. I especially like the part of tests that checks if all views have some sort of access control decorator/mixin. I'll try to run this on `amy-dev` for everyone to test.

workshops/github_auth.py
```diff
@@ -0,0 +1,28 @@
...
+
+from workshops.models import Person
+
+
+class NoPersonAssociatedWithGithubAccount(SocialAuthBaseException):
+    pass
+
+
+def find_user_or_abort(request, details=None, **kwargs):
+    username = details['username']
```
pbanaszkiewicz: What other details we can get here?

workshops/templates/workshops/trainee_dashboard.html
```diff
@@ -0,0 +1,58 @@
...
+  </td></tr>
+  <tr><th>your badges</th><td>
+      <ul>
+        {% for badge in user.badges.all %}
+          <li>{{ badge }}</li>
+        {% empty %}
+          <li>No badges awarded.</li>
+        {% endfor %}
+      </ul>
+      <p>If you think that you're missing a badge, please email us at <a href="mailto:TODO">TODO@TODO.com</a></p>
```
pbanaszkiewicz: TODO???

gvwilson: Yeah - that should *definitely* be `FIXME`. :-)

workshops/github_auth.py
```diff
@@ -0,0 +1,28 @@
...
+    try:
+        user = Person.objects.get(github=username, is_active=True)
+        return {'user': user}
+    except Person.DoesNotExist:
+        # This exception is caught by GithubAuthMiddleware
+        raise NoPersonAssociatedWithGithubAccount
+
+
+class GithubAuthMiddleware(object):
```
pbanaszkiewicz: Is this compatible with the [new style middleware](https://docs.djangoproject.com/en/dev/releases/1.10/#new-style-middleware)? Also, we're Python 3.5+ only, so no need to inherit `object`.

chrismedrela: It looks like it's enough to inherit from `MiddlewareMixin`, but this class is available starting from Django 1.10, so first we need to upgrade Django.

workshops/templates/workshops/trainee_dashboard.html
chrismedrela: @gvwilson, what email address should I use here?

gvwilson: admin@software-carpentry.org

workshops/github_auth.py
chrismedrela: In my case it is:
```python
{
    'username': 'chrismedrela', 
    'email': '', 
    'last_name': '', 
    'fullname': '', 
    'first_name': ''
}
```

workshops/test/test_security.py
chrismedrela: `v` is an instance of `RegexURLPattern` (yes, it's confusing, so I'll rename `all_views` -> `all_urls` and `v` -> `url`)

workshops/models.py
chrismedrela: 1. OK, I've changed that line to `return has_admin_group or user.is_superuser`. 
2. :+1:

workshops/models.py
chrismedrela: `user` can be `None`, `AnonymousUser` or `Person`. You can cover all three cases if you use a function instead of a method.

workshops/test/test_security.py
chrismedrela: See asserts inside `only_for_logged_in` or `only_for_admins`. But I'll change `access` to a list.

chrismedrela:
@pbanaszkiewicz, @gvwilson, I've added tests for trainee dashboard and autoupdate profile form. If you think this PR is good enough, then it's ready to be merged.

chrismedrela:
@pbanaszkiewicz, I agree that mixing two different ways of authorization (group-based `@admin_required` and permission-based `@permission_required`) doesn't feel good. I'd stick with the simplest approach (that is, group-based) for all views. When it's necessary, I'd add `@permission_required`. By the way, what is the use case where you have to use permissions?

pbanaszkiewicz:
Hello Chris,

> I'd stick with the simplest approach (that is, group-based) for all views. When it's necessary, I'd add @permission_required

I agree with that.

> By the way, what is the use case where you have to use permissions?

Probably every page with more than one form (except search form) should have more granular control, for example `event_details`.

Alright, so I'm going to update the amy-dev instance and bother you with one last thing after that.

Cheers,
Piotr

workshops/forms.py
pbanaszkiewicz: @chrismedrela can you please add `languages` to this form?

pbanaszkiewicz:
@chrismedrela @gvwilson I'm sorry but there's a security flaw that I'm very concerned about. With the current approach we allow people to log into existing user's account in AMY if and only if the user's account GitHub login and the GitHub login from logging data are the same: `if Person.objects.get(username='pbanaszkiewicz').github == loggin_data['github_login']: log_them_in()`.

If someone managed (with, I don't know, an overlooked API endpoint) to change Greg's GH login into that intruder's login, then the intruder could log into Greg's account in AMY.

What I think we should do is go with the social-auth workflow (login with github, create a new user and associate that user with this github login), and allow responsible people (==admins) to merge duplicate users.

This is however not a small change and we are already running so behind the schedule…

I need your opinions on this matter, @gvwilson, @chrismedrela and @narayanaditya95.

gvwilson:
Hi Piotr,

If I understand correctly, you're worried that someone could use the API 
to change data in the database in order to give themselves access.  So 
here's another scenario: they use an overlooked API endpoint to give 
their own account superuser access.  Or yet another overlooked API 
endpoint to download all of the email addresses, or just about anything 
else - if we're going to posit overlooked API endpoints, then anything 
is possible and there's nothing we can do about it, no?

Thanks,

Greg

pbanaszkiewicz:
Hi Greg,

I knew you were going to use the overlooked API card.

The overlooked API was just an example. It can be any other error, even 
XSS in person's notes field can be very harmful in this case.

I also want to add that if the our current workflow was okay, other 
websites would use it.

Here's an similar situation described by social-auth devs:
http://psa.matiasaguirre.net/docs/use_cases.html#associate-users-by-email

We should switch to using social-auth table (preferred) or additionally 
validate users with their emails (also has some disadvantages, e.g. 
@chrismedrela doesn't share his email when logging in).

Cheers,
Piotr

gvwilson:
I will defer to your judgment on this, but as you said, we're already 
behind schedule - I need a firm date for deploying user login.

chrismedrela:
@pbanaszkiewicz, ready for review.

chrismedrela:
@pbanaszkiewicz, ready for review.

.gitignore
```diff
@@ -1,5 +1,6 @@
 htmlerror
 *~
+.idea/*
```
pbanaszkiewicz: For future reference: you can gitignore your editor's files globally with core setting `excludesfile`.

pbanaszkiewicz:
@chrismedrela I merged this PR in with a couple of my commits, summary:

1. `AutoProfileForm` was not working at all for some fields (or for all fields)
2. `Person` methods are coupled too tightly to GitHub provider and we'll be working on that later
3. I simplified and user-friendlied the "sync social_auth table" behavior on person's details view.

PR #832, opened 2016-06-13
Fixing unparenthesized method 'call'.

gvwilson:
The use of `lower` instead of `lower()` was preventing data download from working (bug detected 2016-06-13).

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -264,7 +264,7 @@ class ReportsViewSet(ViewSet):
         This function conditionally turns iterables into lists based on the
         format requested by browser."""
         # choose either '?format=...' or '/url.format' or None
-        format_ = (request.query_params.get('format') or format or '').lower
+        format_ = (request.query_params.get('format') or format or '').lower()
 
         if format_ in self.formats_requiring_lists:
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
@gvwilson can you add a regression test?

gvwilson:
Sure - which of the api/tests/*.py files should it go in?

pbanaszkiewicz:
In `test_reports.py` as a standalone subclass of `BaseReportingTest` with only 1 test for the `listify` method of `ReportsViewSet`. Thanks!

pbanaszkiewicz:
@gvwilson any progress here?

gvwilson:
Sorry Piotr - I'm looking after my daughter this week.  Can you please 
file a ticket for me to add a test, and deploy as-is?  I know it's bad 
practice, but we're both confident in the change.

pbanaszkiewicz:
@gvwilson I'll add the tests on your behalf, I hope you won't mind :-)

gvwilson:
No no no, that's fine, that's fine, please...

PR #833, opened 2016-06-14
Add ability to import a PyData host from a conference site.

### LINKED ISSUE #829: Import entire PyData conference as an Event
A PyData conference (say PyCon US '16) should be represented as an organization that hosts a series of events (here talks at the PyData conference).

The FK from the `Event` model to the `Organization` named `host` still performs the same functionality.

narayanaditya95:
Since each PyData conference would exist as a Host, we would like to import details about the PyData conference into the `Host` model.

The API is can be found at  [#116](https://github.com/pydata/conf_site/pull/116).
```
{
    "title": "PyData",
    "start_date": "2040-01-01",
    "end_date": "2040-01-04"
}
```
The `country` field is not populated as it not saved as any field in the PyData instance.

@gvwilson Please review.

Fixes #829

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -610,6 +610,11 @@ class HostForm(forms.ModelForm):
         model = Host
         fields = ['domain', 'fullname', 'country', 'notes']
 
+    class Media:
+        js = (
+            'import_host_from_url.js',
+        )
+
 
 class MembershipForm(forms.ModelForm):
     class Meta:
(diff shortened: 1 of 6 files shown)


gvwilson:
Looks OK to me, but I don't speak Javascript, so I can't review the core of it.

narayanaditya95:
There is a problem here though.
Not all PyData conferences are hosted at the root domain. PyData Paris is hosted at `http://pydata.org/paris2016/`.

We would have to migrate the `domain` field to a general `url` field.

gvwilson:
:+1: if @pbanaszkiewicz and @maneesha agree.

narayanaditya95:
Closing this PR as PyData conferences will not exist as Events.

Issue #761, opened 2016-05-04
Self-organized DC workshop request form


tracykteal:
Data Carpentry now has guidelines for self-organized workshops, and we'd like to have a separate self-organized workshop registration form for instructors to register a self-organized workshop and confirm that instructors will meet the self-organized workshop criteria
http://www.datacarpentry.org/self-organized-workshops/

I drafted a form that has the necessary fields. For fields that are the same as in the existing form, that same field is fine. I added a few fields that are just relevant for self-organized workshops and took out some that weren't relevant. 

https://docs.google.com/forms/d/1SHctDIQYLEDy139NNTKkgi0ctbMeh2zdE927hsWh1Ew/viewform

Can we make this a form in Amy, so the information goes in to the Amy system? I imagine some discussion will be needed to make sure the questions can match with the database structure. Let me know if a call would be helpful.

pbanaszkiewicz:
@tracykteal the form you proposed differs from the current similar form for SWC (https://amy.software-carpentry.org/workshops/submit/). The existing form is rather intended for people to tell AMY that they've been running self-organized workshops, however I don't think anyone used that form ever :cry:

@maneesha please decide if you want to stick with the existing form or implement something similar to what @tracykteal proposed for DC.

tracykteal:
Yes, this new form has a slightly different motivation, in that we want to make sure people are prepared to run self-organized workshops and that we get them registered. This is partly in response to the challenges @maneesha has seen with self-organized workshops. It also matches the process approved by the DC Steering Committee for self-organized workshops http://www.datacarpentry.org/self-organized-workshops/

tracykteal:
Any update on this? Anything we can do to help?

pbanaszkiewicz:
@tracykteal just to make sure: this new form is meant to replace https://amy.software-carpentry.org/workshops/submit/, correct?

tracykteal:
No, it would be an additional form. We want to have one for hosts to request workshops and one for self-organized workshops, because they have different sets of questions.

pbanaszkiewicz:
> We want to have one for hosts to request workshops and one for self-organized workshops, because they have different sets of questions.

We currently have 2 different forms for requesting a workshop:
* https://amy.software-carpentry.org/workshops/dc/request/
* https://amy.software-carpentry.org/workshops/swc/request/

We also have 1 form for informing admins about a possibly self-organized workshop that already has a webpage: https://amy.software-carpentry.org/workshops/submit/

[...]

PR #836, opened 2016-06-18
DC self-organized workshop requests

pbanaszkiewicz:
This PR solves #761 by adding a DC-specific form available to end users, a related model (with 2 additional simple models) and admin interface.

Additional changes outside of scope of the #761:

* links to lists of requests (ie. training requests, workshop submission, invoice requests) were moved over to a separate subgroup in top navigation bar
* abstract mixins for models were added: one provides `active` boolean field, and the other one provides `created_at` and `last_updated_at` date time fields with correct `auto_now`/`auto_now_add` values


workshops/models.py
```diff
@@ -18,6 +18,7 @@
 STR_SHORT   =  10         # length of short strings
 STR_MED     =  40         # length of medium strings
 STR_LONG    = 100         # length of long strings
+STR_LONGEST = 255  # length of the longest strings
```
gvwilson: why not increase STR_LONG and use that instead of STR_LONGEST ?

workshops/models.py
```diff
@@ -32,6 +33,26 @@ class Meta:
         abstract = True
 
 
+class ActiveMixin(models.Model):
```
gvwilson: We use tags on events for this, don't we?

gvwilson:
:+1:

workshops/models.py
pbanaszkiewicz: Increasing `STR_LONG` would trigger a migration.

workshops/models.py
pbanaszkiewicz: No. Similar field on Event is `.completed`, but it has inverted meaning.

workshops/models.py
gvwilson: There are already several migrations in this PR, no?

workshops/models.py
gvwilson: Then would it make sense to use `.completed` here (same name, same sense)?

workshops/models.py
pbanaszkiewicz: Yes, but my intention was to add a variable substituting `max_length=255`, which was used in many places :)

workshops/models.py
pbanaszkiewicz: `.active` was used before I factored out the Mixin. `Event.completed` is actually the only place where we have `.completed`. If anything it would be `Event` that should have `.active`.

pbanaszkiewicz:
I'm going to rebase against develop and add a merging migration.

PR #837, opened 2016-06-18
Import PyData conf as Events

narayanaditya95:
I created a new package within the `workshops` package.
This would allow us to separate code specific to PyData in a separate package that is pluggable. (No more ugly `if` conditionals)
The URLs from the `workshops.pydata` package are inserted before the URLs from the `workshops` package.

Added ability to import events.
Currently, it only populates the `slug`, `start`, `end` and `tags` fields.


workshops/pydata/utils.py
```diff
@@ -0,0 +1,7 @@
+
+def parse_event(conf):
```
narayanaditya95: Mapping from fields from conf. site to fields in AMY.

narayanaditya95:
See #839.

PR #838, opened 2016-06-18
Add title and url to tasks

narayanaditya95:
This would allow us to record presentations that each presenter gave at a PyData conference.


narayanaditya95:
Up for review.

gvwilson:
Looks good to me if this is the direction we're going.

narayanaditya95:
@gvwilson Do you have apprehensions regarding this schema change?

narayanaditya95:
@gvwilson @pbanaszkiewicz I have added `title` columns to all forms and tables that list tasks (I hope I didn't miss anything). Although we do not aim to populate these `title` fields for SWC/DC workshops, the column would still exist.
Is that fine or should we have some conditional templating based on [`PYDATA_ENABLED`](https://github.com/swcarpentry/amy/pull/837/files#diff-5be74688ddb29fccfa52e37f1635073bR3) variable.

gvwilson:
Nope - looks good to me.

gvwilson:
I'm :+1: on displaying the field in all AMY instances rather than 
conditional layout, but it's @pbanaszkiewicz's call.

narayanaditya95:
@gvwilson I realized we aren't storing the URL for a PyData presentation. An admin would like to visit the presentation page at conf. site from the task page at AMY.
Should we add another `URLField`?

narayanaditya95:
@gvwilson Ping.

gvwilson:
Yup - gave it a thumbs-up (as an emoji on the commit comment), so :+1: 
to merge.

narayanaditya95:
> @gvwilson I realized we aren't storing the URL for a PyData presentation. An admin would like to visit the presentation page at conf. site from the task page at AMY.
Should we add another URLField?

What are your views regarding this?

gvwilson:
I don't want to introduce too many schema changes until we have feedback 
from Leah that this is a good direction for the interface, so please 
*don't* add the URL field for now - if we need to later, it'll be easy 
to do.

pbanaszkiewicz:
@gvwilson please refresh my memory, when can we expect feedback from Leah? And do you want to have some dev instance prepared for that occasion?

gvwilson:
I'm still trying to get feedback from Leah, and yes, we'll want a dev 
instance.

pbanaszkiewicz:
@narayanaditya95 can we close this PR because it's contained within #839?

narayanaditya95:
Let's keep this PR open as this PR isn't exclusive to just PyData. I would be better to discuss the schema change here and talk about PyData-specific code changes at #839.

#839 Does include it, but it should be merged independently and after this PR #838.

pbanaszkiewicz:
@narayanaditya95 so do we need `Task.url` or not?

narayanaditya95:
@pbanaszkiewicz Although @gvwilson said "not now", I think recording the url is important as it also records the `pk` of the presentation object at PyData. That would allow us to refetch and update the task object.
I will be adding the url field today.

narayanaditya95:
@pbanaszkiewicz Rebased and added the URL field. Review please.

narayanaditya95:
Moving data migrations pertaining exclusively to `pydata` package to its `fixtures` directory.

narayanaditya95:
@pbanaszkiewicz please review.

workshops/test/test_tasks.py
```diff
@@ -70,6 +71,12 @@ def test_task_detail_view_reachable_from_event_person_and_role_of_task(self):
         response = self.client.get(reverse('task_details', args=[str(correct_task.id)]))
         assert response.context['task'].pk == correct_task.pk
 
+    def test_task_wrong_url_not_accepted(self):
+        task = self.fixtures['test_task_1']
```
pbanaszkiewicz: Where does `self.fixtures` come from?

pbanaszkiewicz:
@narayanaditya95 looks OK, but I'm not going to merge until #873 is fixed :<

workshops/test/test_tasks.py
narayanaditya95: It comes from the `setUp` method on the class.

workshops/test/test_tasks.py
```diff
@@ -70,6 +71,12 @@ def test_task_detail_view_reachable_from_event_person_and_role_of_task(self):
...
+        task.url = 'htp://www.example.com'
+        task_form = TaskForm(initial={'task':task})
+        self.assertFalse(task_form.is_valid())
```
narayanaditya95: "Always ensure that your tests fail before the fix".
Subtle. :)

pbanaszkiewicz: Yeah, so where's that test I was talking about? :stuck_out_tongue:

narayanaditya95: Done.

pbanaszkiewicz: @narayanaditya95 thanks!

1. It's a good habit to write docstrings to your tests because they're showed by the TestRunner when tests fail.

2. I'll merge this when we open v1.7 for merging.

narayanaditya95: Added docstrings and one additional assertion. :)

PR #839, opened 2016-06-21
Import event person task

narayanaditya95:
Since all imports build incrementally on each other, I decided to club them together in one PR. (I get lost easily otherwise too)

I will rewrite the git history after the final approval. Rewriting git history after every suggestion hides the previous conversation, something that I hate.

Since I haven't yet hosted a PyData conf. site with an API over it, I am bundling the API responses.
```
GET /api/
{
    "title": "PyData",
    "start_date": "2040-01-01",
    "end_date": "2040-01-04"
}
```
```
GET /api/speaker/1/
{
    "username": "admin",
    "name": "admin",
    "email": "admin@pydata.org"
}
```
```
GET /api/presentation/1/
{
    "slot": {
        "day": "2016-06-14",
        "kind": "Test",
        "name": "2016-06-14 Test (09:55:44 - 00:00:00) ",
        "start": "09:55:44",
        "end": null
    },
    "section": "PyData General Sessions",
    "speaker": {
        "username": "admin",
        "name": "admin",
        "email": "admin@pydata.org"
    },
    "title": "Big Data",
    "description": "Test",
    "abstract": "Test",
    "cancelled": true,
    "additional_speakers": []
}
```

@gvwilson @pbanaszkiewicz 
This is up for the first review.

### DIFF
--- a/pydata/apps.py
+++ b/pydata/apps.py
@@ -1,0 +1,7 @@
+from django.apps import apps, AppConfig
+
+
+class PyDataConfig(AppConfig):
+    name = 'pydata'
+    label = 'PyData'
+    verbose_name = 'AMY for PyData conferences'
--- a/pydata/apps.py
+++ b/pydata/apps.py
@@ -5,3 +5,6 @@ class PyDataConfig(AppConfig):
     name = 'pydata'
     label = 'PyData'
     verbose_name = 'AMY for PyData conferences'
+
+    def ready(self):
+        from . import checks
--- a/pydata/apps.py
+++ b/pydata/apps.py
@@ -1,3 +1,5 @@
+import re
+
 from django.apps import apps, AppConfig
 from django.utils.functional import curry
 
@@ -33,3 +35,51 @@ class PyDataConfig(AppConfig):
             'level',
             curry(Sponsorship._get_FIELD_display, field=amount_field),
         )
+
+        # Add "Import from URL" button to SponsorshipForm
+        from workshops.forms import SponsorshipForm
+
+        class Media:
+            js = ('import_sponsor.js', )
+
+        SponsorshipForm.Media = Media
+
+        setattr(
+            Sponsorship,
+            'PROFILE_REGEX',
+            re.compile(r'^(?P<url>.+?(?=/sponsors))/sponsors/(?P<id>[^/]+)/?'),
+        )
+
+        Task = apps.get_model('workshops', 'Task')
+        setattr(
+            Task,
+            'PRESENTATION_REGEX',
+            re.compile(r'^(?P<url>.+?(?=/schedule))/schedule/presentation/(?P<id>[^/]+)/?'),
+        )
+
+        Person = apps.get_model('workshops', 'Person')
+        setattr(
+            Person,
+            'PROFILE_REGEX',
+            re.compile(r'^(?P<url>.+?(?=/speaker))/speaker/profile/(?P<id>[^/]+)/?'),
+        )
+
+        # Add "Import from URL" button to PersonForm on PersonCreate view
+        from workshops.forms import PersonForm
+        from workshops.views import PersonCreate
+
+        class Media:
+            js = ('import_person.js', )
+
+        PersonForm.Media = Media
+        PersonCreate.template_name = 'pydata/person_create_form.html'
+
+        # Add "Import from URL" button to TaskForm on TaskCreate view
+        from workshops.forms import TaskForm
+        from workshops.views import TaskCreate
+
+        class Media:
+            js = ('import_task.js', )
+
+        TaskForm.Media = Media
+        TaskCreate.template_name = 'pydata/task_create_form.html'
--- a/pydata/apps.py
+++ b/pydata/apps.py
@@ -1,4 +1,5 @@
 from django.apps import apps, AppConfig
+from django.utils.functional import curry
 
 
 class PyDataConfig(AppConfig):
@@ -8,3 +9,27 @@ class PyDataConfig(AppConfig):
 
     def ready(self):
         from . import checks
+
+        Sponsorship = apps.get_model('workshops', 'Sponsorship')
+        # Add choices to the `amount` field
+        Sponsorship.LEVELS = (
+            (0, 'Founding'),
+            (15000, 'Diamond'),
+            (8000, 'Platinum'),
+            (5000, 'Gold'),
+            (3000, 'Silver'),
+            (1500, 'Supporting'),
+            (1, 'Community'),
+        )
+        amount_field = Sponsorship._meta.get_field('amount')
+        amount_field.choices = Sponsorship.LEVELS
+        amount_field.verbose_name = 'Sponsorship level'
+
+        # Add an attribute to sponsorship object to return the level text
+        # Mimics `get_amount_choice` method
+        # See https://github.com/django/django/blob/49b4596cb4744e4b68d56e6a540a3e15c1582963/django/db/models/fields/__init__.py#L702
+        setattr(
+            Sponsorship,
+            'level',
+            curry(Sponsorship._get_FIELD_display, field=amount_field),
+        )
--- a/pydata/static/import_person.js
+++ b/pydata/static/import_person.js
@@ -1,0 +1,38 @@
+function import_from_url(url) {
+  return $.get("/workshops/person/import/", {'url': url}, function(data) {
+    $("#event_import_url").parent().removeClass('has-error');
+    $('#import_url_modal').modal('hide');
+    $('#error_message').addClass('hidden');
+
+    $("#id_personal").val(data.personal);
+    $("#id_family").val(data.family);
+    $("#id_email").val(data.email);
+    $("#id_url").val(url);
+  });
+}
+
+$('#import_url_button').click(function() {
+  $('#import_url_modal').modal();
+});
+$('#import_url_form').submit(function(e) {
+  e.preventDefault();
+
+  // indicate loading data
+  var btn = $(this).find('button[type=submit]');
+  btn.attr('disabled', true);
+
+  // load data from URL
+  import_from_url(
+    $(this).find(':input[name=url]').val()
+  )
+  .fail(function(data) {
+    // something went wrong, let's indicate it
+    $("#event_import_url").parent().addClass('has-error');
+    $('#error_message').text(data.responseText);
+    $('#error_message').removeClass('hidden');
+  })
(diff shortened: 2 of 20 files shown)


pydata/models.py
```diff
@@ -0,0 +1,22 @@
...
+from workshops.models import Person as BasePerson
+from workshops.models import Task as BaseTask
+
+
+class Person(BasePerson):
+    """ Represents a PyData speaker."""
+
+    PROFILE_REGEX = re.compile(r'^(?P<url>.+?(?=/speaker))/speaker/profile/(?P<id>[^/]+)/?')
+
+    class Meta:
```
narayanaditya95: Since these models are just used to override behaviour (if any), they should not exist in the database.
Without `abstract=True`, the django test framework complains.

pydata/templates/pydata/generic_import_form.html
```diff
@@ -0,0 +1,17 @@
+{% extends "base_nav_fixed.html" %}
```
narayanaditya95: @pbanaszkiewicz This template can be moved to `workshops/templates` if you want it to.

workshops/static/import_from_url.js
```diff
@@ -8,6 +8,7 @@ function import_from_url(url) {
     $("#id_start").val(data.start);
     $("#id_end").val(data.end);
     $("#id_reg_key").val(data.reg_key);
+    $("#id_tags").val(data.tags);
```
narayanaditya95: On the import of a PyData conference, the `PyData` tag is selected.
@pbanaszkiewicz You may want to replicate it for SWC and DC event imports.

narayanaditya95:
@gvwilson @pbanaszkiewicz Ping.

gvwilson:
I apologize, but this one's beyond my understanding - if @pbanaszkiewicz likes it, :+1: from me.

pbanaszkiewicz:
@narayanaditya95 I'll review the changes on Friday evening or Saturday morning of my time.

pydata/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    family = speaker['name'].rsplit(' ', 1)[-1]
+    return {
+        'personal': personal,
+        'family': family,
+        'email': speaker['email'],
+    }
+
+
+def parse_task(presentation, conf_url):
+    event = Event.objects.get(url__contains=conf_url)
```
narayanaditya95: If we allow `EventLookup` to search against the `url` field, we wouldn't have to fetch the Event from database and pass it's `pk` within the response.

workshops/models.py
```diff
@@ -1338,6 +1338,10 @@ class Task(models.Model):
     event      = models.ForeignKey(Event)
     person     = models.ForeignKey(Person)
     role       = models.ForeignKey(Role)
+    title      = models.CharField(
+        max_length=STR_LONG,
+        blank=True,
```
pbanaszkiewicz: Shouldn't you add a `default=""` or `null=True`?

pydata/models.py
pbanaszkiewicz: And why do you need these abstract models?

pydata/templates/pydata/generic_import_form.html
pbanaszkiewicz: Nah, it's fine the way it is. By the way, if we had decided to move this template, it couldn't have "import" -- because we also have "update from URL" function.

pydata/templates/pydata/person_create_form.html
```diff
@@ -0,0 +1,5 @@
+{% extends "pydata/generic_import_form.html" %}
+
+{% block modal %}
+{% include "pydata/person_import_from_url.html" %}
```
pbanaszkiewicz: I'd call this template "person_import_modal" -- had I been wiser I'd call the similar template in `workshops` this way too.

workshops/static/import_from_url.js
pbanaszkiewicz: If I understand correctly, `data.tags` contains a list of tags like 'DC', 'TTT', 'SWC', or 'PyData', right?

I don't think I'm able to discover if a workshop is DC or SWC type. `data.tags` should be empty in this case.

pydata/urls.py
```diff
@@ -0,0 +1,14 @@
...
+
+urlpatterns = [
+    url(r'persons/add/?$', views.PersonCreate.as_view(), name='person_add'),
+    url(r'persons/import/?$', views.PersonImport.as_view(), name='person_import'),
+
+    url(r'^events/import/?$', views.ConferenceImport.as_view(), name='event_import'),
+
+    url(r'^tasks/add/?$', views.TaskCreate.as_view(), name='task_add'),
+    url(r'^tasks/import/?$', views.TaskImport.as_view(), name='task_import'),
+]
```
pbanaszkiewicz: I was just thinking if you really need to plug into existing views. Maybe you should have one huge view that imports everything from PyData instead?

narayanaditya95: We are over-riding the existing URLs.
I do not get how having a single view would allow us to individually tweak and override relevant URLs.
Sorry, I couldn't follow.

pydata/models.py
narayanaditya95: To encapsulate the `PROFILE_REGEX` logic within the `Person` model.
Just as `Event` model encapsulates few regexes related to the workshop URL.

I would not want them to be littered across various views.

pydata/templates/pydata/generic_import_form.html
narayanaditya95: This template would be used at the `/model/add/` view that is only concerned with the import of a model instance.
A similar template would exist for `/model/<slug>/edit` view that would allow both imports and edits.

I haven't added the functionality to edit the existing PyData events, persons and tasks before we get a response from Leah or Andy.

workshops/static/import_from_url.js
narayanaditya95: > If I understand correctly, data.tags contains a list of tags like 'DC', 'TTT', 'SWC', or 'PyData', right?

Yes

> I don't think I'm able to discover if a workshop is DC or SWC type. data.tags should be empty in this case.

Okay. If there existed a way to distinguish between the two, `data.tags` can be a part of the AJAX response.

pydata/urls.py
pbanaszkiewicz: I meant: **instead of** "individually tweaking and overriding URLs". Like: instead of replacing AMY's functionality, add to it.

workshops/static/import_from_url.js
pbanaszkiewicz: Please wrap it into a conditional then:
```javascript
if (data.tags != undefined) {
    $("#id_tags").val(data.tags);
}
```

narayanaditya95:
A dev instance of PyData conf site is running at http://162.242.208.88/.

workshops/static/import_from_url.js
narayanaditya95: Without a conditional, `$("#id_tags").val(data.tags);` does not throw any TypeError. If only resets the selection within the `tags` fields.
Are we interested in clearing the selection after an `import` command?

pydata/urls.py
narayanaditya95: If I understand right, you suggest we make changes within the `workshops` package?
Wouldn't that involve a lot of conditionals? There is a high probability of making the existing code too complex with a lot of branching.

pydata/utils.py
narayanaditya95: @pbanaszkiewicz 
Can we allow lookups on the `event` field using the URL of the event? It would save us a database query here.
Opinions?

pydata/static/import_person.js
```diff
@@ -0,0 +1,38 @@
+function import_from_url(url) {
```
narayanaditya95: @pbanaszkiewicz 
I wish we could have some sort of inheritance in javascript as well.

narayanaditya95:
@pbanaszkiewicz 
What is your opinion regarding the recent refactor? I tried to minimize code duplication.

workshops/models.py
narayanaditya95: 'null=False, default=''` is implicit here.

pydata/views.py
```diff
@@ -0,0 +1,124 @@
...
+
+from workshops.models import Event, Tag, Role
+from workshops.views import TaskCreate as BaseTaskCreate
+from workshops.views import PersonCreate as BasePersonCreate
+
+from .forms import ImportTaskFullForm, ImportPersonForm
+from .models import Person, Task
+
+
+class BaseImport(LoginRequiredMixin, View):
```
pbanaszkiewicz: I think you need to use some other mixin because we changed the logging permissions.

pbanaszkiewicz:
@narayanaditya95 I don't have many comments. I think we need to see the one-import-to-rule-them-all page and see how you manage to add all the tasks / events / whatever without code duplication (so IMO this is WIP).

pbanaszkiewicz:
@narayanaditya95 again, can you please rebase? Thanks!

gvwilson:
@narayanaditya95 please rebase as soon as you can.

narayanaditya95:
About to push an update today.

pydata/apps.py
```diff
@@ -0,0 +1,85 @@
...
+            (15000, 'Diamond'),
+            (8000, 'Platinum'),
+            (5000, 'Gold'),
+            (3000, 'Silver'),
+            (1500, 'Supporting'),
+            (1, 'Community'),
+        )
+        amount_field = Sponsorship._meta.get_field('amount')
+        amount_field.choices = Sponsorship.LEVELS
+        amount_field.verbose_name = 'Sponsorship level'
```
narayanaditya95: @pbanaszkiewicz Can you help me with monkey patching django models if you have the experience of achieving it before. I can confirm that the above code did work, but it no longer patches the `amount` field to add choices. I haven't been able to debug the erratic behaviour.

pbanaszkiewicz: Whoa, I had merged before I saw this comment. @narayanaditya95 should I revert and work with you on this monkey patching?

narayanaditya95: This wasn't supposed to be merged.
Please revert.

pbanaszkiewicz: @narayanaditya95 I reset the `develop` branch, but unfortunately that didn't reopen this PR. I'm really sorry about this mess, @narayanaditya95, but can you make a new PR with these changes?

narayanaditya95: Okay

Issue #840, opened 2016-06-22
Person without email causes attribute error


pbanaszkiewicz:
In `Person.clean()`, `self.email` can be None.

prerit2010:
@pbanaszkiewicz  Can you  please elaborate how to reproduce this bug ? I created a new person with no email id, but it didn't throw any error.

pbanaszkiewicz:
@prerit2010 I found it when I was changing permissions for a person without email.

pbanaszkiewicz:
Closed via #841.

PR #841, opened 2016-06-23
Fixed attribute error-- on changing the permission for a person without email"

prerit2010:


workshops/models.py
```diff
@@ -332,7 +332,10 @@ def clean(self):
         """This will be called by the ModelForm.is_valid(). No saving to the
         database."""
         # lowercase the email
-        self.email = self.email.lower()
```
narayanaditya95: How about `self.email = self.email.lower() if self.email`.

prerit2010: @narayanaditya95  Then it should be : `self.email = self.email.lower() if self.email else None`

pbanaszkiewicz:
Looks good. @prerit2010 can you add a regression test to `test_person.TestPerson`? Thanks a lot.

pbanaszkiewicz:
@prerit2010 thanks.

Ref #840.

pbanaszkiewicz:
Since we have #844 scheduled for v1.6, let's fix this bug in this release as well. Thanks @prerit2010 for this contribution!

PR #848, opened 2016-06-27
Describing instructor training workflow

gvwilson:
Point-form use cases of instructor training workflow.


pbanaszkiewicz:
@gvwilson can you please move to `.md` so that GitHub renders it? Unless you specifically wanted it to not render.

gvwilson:
I'll work on it - can you comment on the .txt?

pbanaszkiewicz:
Oh, I see -- by using {[< brackets you actually use Markdown syntax, so you have to use `.txt`. Additionally it's easy to comment on specific lines this way.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+
+* [user registers self][a]: user match by email address or GitHub ID.
+  - User fills in {single user application form} on AMY with name, email address, GitHub ID, etc.
+  - AMY matches to pre-existing <Person> using email or GitHub ID.
+  - <Training Application> created, connected to pre-existing <Person>, and added to <Training Application List>.
+
+* [user registers self][b]: no user match.
+  - User fills in form on AMY with name, email address, GitHub ID, etc.
+  - AMY does *not* match email or GitHub ID.
+  - AMY creates new <Person>.
```
pbanaszkiewicz: But not automatically -- only after human intervention, right?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+  - Administrator creates <Person>.
+  - Administrator => [admin registers user][a].
+
+## Group application for training.
+
+* [user registers group]: individual user registers a group of friends or colleagues.
+  - User fills in {group application form} with:
+    - Their own name, email address, institution.
+    - Names and email addresses of group members
+  - User and others are sent unique links to fill in full application form.
```
pbanaszkiewicz: Potential spam-generator?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+    - Creates one new <Training Application> record for each <Person>.
+    - Connects the <Training Application> to the <Person>.
+    - Connects the <Training Application> to the <Group Training Application>.
+    - Adds the <Training Application> to <Training Application List>.
+
+## Reviewing training queue.
+
+* [admin reviews training queue]: to see who's waiting and how long.
+  - Admin goes to {training queue display}.
+  - Admin selects filtering and sorting criteria (if any).
```
pbanaszkiewicz: What can you sort by? Something like registration-details, e.g. institution?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+  - Admin pages through display showing both <Training Application>s and <Group Training Application>s.
```
pbanaszkiewicz: Does the table show <Group Training Application> as a single entry or as each entry belonging to the that <Group Training Application>?

PS It's hard to mix objects of different types (models) in a table.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+
+## Setting up training course.
+
+* [admin creates training event]: and assigns people to it.
+  - Admin creates new <Event>.
+  - Admin goes to {training queue display}.
+  - Admin selects <Event>.
+  - Admin selects some mix of <Training Application> and <Group Training Application>.
```
pbanaszkiewicz: <Group Training Application> is allowed as a whole to the <Event>, correct?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+
+* [admin removes people from training event][a]: and discards applications
+  - Admin goes to {assigned training application display}.
+  - Admin selects <Training Application>s and <Group Training Application>s.
+  - Admin removes them from training event.
+  - AMY deletes associated <Task>s.
+
+* [admin removes people from training event][b]: but keeps applications
+  - => [admin removes people from training event][a]
+  - AMY moves <Training Application>s and <Group Training Application>s back to <Training Application List>.
```
pbanaszkiewicz: So you have to explicitly write "AMY deletes `<Training Application>s` and `<Group Training Application>s`" in the `[admin removes people from training event][a]` point.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+
+## Reviewing training progress.
+
+* [admin views training progress]: both past and present.
```
pbanaszkiewicz: "past and present" what? Past and present progress?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+  - Admin views {training progress page}.
+  - Admin selects filtering and sorting criteria (if any).
```
pbanaszkiewicz: Any possible actions to be taken as part of reviewing training progress? Or is it in the following headers?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+
+## Tracking submission of written work.
+
+* [trainee reports submission of written work]: such as pull request or comment on lesson.
+  - Trainee views their own {training progress}.
+  - Trainee enters URL of written work.
+  - AMY updates <Training Task> associated with <Person> and <Event>.
+  - AMY adds <Training Progress> to <Training Progress Review List>.
```
pbanaszkiewicz: Should we add some feedback loop? Should we allow trainees to update the `<Training Progress>` and send a notification to the trainer or other person responsible for reviewing trainee's progress?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+
+* [trainer reports submission of written work]: on behalf of trainee.
+  - Trainer looks up <Training Progress> associated with trainee.
+  - Trainer enters URL of written work in <Training Task>.
+  - AMY adds <Training Progress> to <Training Progress Review List>.
+
+* [trainer reports completion of written work]: on behalf of trainee.
+  - Trainer looks up <Training Progress> associated with trainee.
+  - Trainer enters URL of written work in <Training Task>.
+  - Trainer marks <Training Task> as complete.
```
pbanaszkiewicz: So the trainer can mark some `<Training Progress>` or `<Training Task>` as complete only after they first enter the URL to the written work?

I thought the trainer uses existing `<Training Progress>` URL to that was successfully approved after the review.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+
+* [trainer reports successful completion of interactive work]: such as discussion session or lesson demo.
+  - Trainer views {training progress} for trainee.
+  - Trainer selects <Training Task>.
+  - Trainer:
+    - Enters date of interactive event.
+    - Adds comments (if any).
+  - Trainer indicates whether <Training Task> is now complete.
+
+Note: each <Training Task> may accumulate multiple date+comment pairs, one for each attempt to pass that task.
```
pbanaszkiewicz: This means that `<Training Task>` is associated with a feedback loop only if it corresponds to an "interactive work such as discussion session or lesson demo", and not the written work (see above).

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+    - Do not yet have a PDF certificate.
+  - Admin adjusts information in form.
+    - Changes name of instructor(s) awarding certificates.
+    - Changes date of award.
+  - AMY generates PDF certificates.
+  - User views/downloads certificate on {certificate display}.
+
+## Questions
+
+- How to display a mix of both <Training Application>s and <Group Training Application>s?
```
pbanaszkiewicz: I'd say that `<Group Training Applications>` should have a list of `<Training Application>s` so that we can display objects of the same type.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,174 @@
...
+- How to select from a mix of both <Training Application>s and <Group Training Application>s?
```
pbanaszkiewicz: That's UI/UX concern, isn't it?

pbanaszkiewicz:
I hope I understood you correctly, just to make sure here's a stripped down workflow (without UI/UX/database/AMY interactions)
untitled diagram

docs/instructor-training-workflow.txt
gvwilson: Good catch - modified.

docs/instructor-training-workflow.txt
gvwilson: Good catch - fixed.

docs/instructor-training-workflow.txt
gvwilson: Yes - we can worry about that during implementation.

docs/instructor-training-workflow.txt
gvwilson: I'll take the Group Application stuff out of the workflow for now - we need to get individual applications in first.

docs/instructor-training-workflow.txt
gvwilson: See above.

docs/instructor-training-workflow.txt
gvwilson: Good catch.

docs/instructor-training-workflow.txt
gvwilson: In the following.

docs/instructor-training-workflow.txt
gvwilson: Yeah, that sentence was useless. Fixed.

docs/instructor-training-workflow.txt
gvwilson: Not yet.

docs/instructor-training-workflow.txt
gvwilson: Good point - I've added to the Questions section.

docs/instructor-training-workflow.txt
gvwilson: Taking Groups out for now.

docs/instructor-training-workflow.txt
gvwilson: It was a database question - as you pointed out, it's hard to have heterogeneous objects.

docs/instructor-training-workflow.txt
pbanaszkiewicz: If "Not yet" is for the notification -- I was thinking about a notification system for a loooooong time and there are some excellent plugins to Django that provide GitHub-like notifications. We may be just a step away from adding them to AMY.

docs/instructor-training-workflow.txt
pbanaszkiewicz: It may be me misunderstanding. Why does the trainer have to report submission of trainee's written work on the trainee's behalf? This should be submitted by the trainee themselves, right?

docs/instructor-training-workflow.txt
gvwilson: I understand, but this is a "want to have", and other things are "must 
have".
Cheers,
Greg

docs/instructor-training-workflow.txt
gvwilson: Nope, you're right: there needs to be another workflow "trainee submits written work".

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+
+- [use cases] are in square brackets
+- {web pages} are in curly braces
+- <database objects> are in angle brackets
+
+## Individual application for training.
+
+* [user registers self][a]: user match by email address or GitHub ID.
+  - User fills in {single user application form} on AMY with name, email address, GitHub ID, etc.
+  - AMY matches to pre-existing <Person> using email or GitHub ID.
```
chrismedrela: When you say "GitHub ID" do you mean GitHub username? The applicant doesn't tell us its ID in the form.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - <Training Application> created, connected to pre-existing <Person>, marked "Pending", and added to <Training Application List>.
```
chrismedrela: `Training Application List` <- do you mean a view?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+
+* [user registers self][b]: no user match.
+  - User fills in form on AMY with name, email address, GitHub ID, etc.
+  - AMY does *not* match email or GitHub ID.
+  - <Training Application> created, marked "Pending", and added to <Training Application List> *without* <Person> object.
```
chrismedrela: Do I understand correctly that you want `<TrainingApplication>` to have nullable ForeignKey to `<Person>`?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - Admin selects <Training Application> marked "Pending" that has been matched with <Person>.
+  - Admin approves <Training Application>.
+  - AMY clears <Training Application>'s "Pending" flag.
+
+* [admin approves training][b]: no user record in database.
+  - Admin views <Training Application List>.
+  - Admin selects <Training Application>.
+  - Admin approves <Training Application>.
+  - AMY clears <Training Application>'s "Pending" flag.
+  - AMY creates <Person> record and associates with <Training Application>.
```
chrismedrela: 1) Do you want `may_contact` set to True? If yes, should we add a note to the `TrainingRequestForm` that we may contact them?
2) What is the password for the new Person record?
3) Is there any reason for the "Pending" state except that we need to match application to a Person?
4) What if somebody makes two requests? The admin probably needs something like "discard this application".

Do I understand correctly that:

- We use `create_username` and ignore GitHub login when choosing applicant username. 
- We set `airport` to null, because we don't have such field in the form.
- In the new Person record, we leave `twitter`, `url`, `notes`, `badges`, `lessons`, `domains`, `languages` and `orcid` empty.
- After approving the request, the applicant still cannot login using GitHub and `Person.is_active == False`.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+* [admin approves training][a]: user record already in database.
+  - Admin views <Training Application List>.
+  - Admin selects <Training Application> marked "Pending" that has been matched with <Person>.
+  - Admin approves <Training Application>.
+  - AMY clears <Training Application>'s "Pending" flag.
+
+* [admin approves training][b]: no user record in database.
+  - Admin views <Training Application List>.
+  - Admin selects <Training Application>.
+  - Admin approves <Training Application>.
```
chrismedrela: I think at this point Admin should be able to (optionally) manually match Person. In that case, we don't create Person record.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - User fills in form on AMY with name, email address, GitHub ID, etc.
+  - AMY does *not* match email or GitHub ID.
+  - <Training Application> created, marked "Pending", and added to <Training Application List> *without* <Person> object.
+
+## Administrator approves individual for training.
+
+* [admin approves training][a]: user record already in database.
+  - Admin views <Training Application List>.
+  - Admin selects <Training Application> marked "Pending" that has been matched with <Person>.
+  - Admin approves <Training Application>.
```
chrismedrela: I think that at this point Admin should be able to view what Person was matched by AMY and change it to other Person if AMY mismatched.

docs/instructor-training-workflow.txt
chrismedrela: OK, to sum up, I think this step can be simplified:
```
[admin approves training]
- Admin views {Training Application List}.
- Admin selects <Training Application>.
- Admin approves <Training Application> and chooses one from two options:
    a) choose a <Person> that should be associated with the <Training Application> using autocomplete field;
       if AMY can match a <Person>, then the default value is that <Person>
    b) create a new <Person> (default option if AMY didn't propose any Person)
- AMY clears <Training Application>'s "Pending" flag.
- If Admin chose "create new Person", AMY creates <Person> record.
- AMY associates the selected <Person> or the new <Person> with <Training Application>.
```

docs/instructor-training-workflow.txt
chrismedrela: 3) OK, now I think the matching should be done when an Admin approves the request. This has the advantage of simplifying this use case.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - AMY clears <Training Application>'s "Pending" flag.
+
+* [admin approves training][b]: no user record in database.
+  - Admin views <Training Application List>.
+  - Admin selects <Training Application>.
```
chrismedrela: Does it mean that the "Approve" button is displayed in {detailed Training Application view} rather than in {Training Application list view}?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - Admin approves <Training Application>.
+  - AMY clears <Training Application>'s "Pending" flag.
+  - AMY creates <Person> record and associates with <Training Application>.
+
+## Administrator registers individual for training.
+
+* [admin registers user][a]: user is already in database.
+  - Administrator looks up email address or GitHub ID.
+  - Administrator => [user registers self][a].
```
chrismedrela: Does it mean that there is a button in detailed `<Person>` view redirecting to {Training Application form} with prefilled fields like `personal`, `email` and so on?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+* [admin registers user][b]: user is *not* already in database.
+  - Administrator creates <Person>.
+  - Administrator => [admin registers user][a].
+
+## Reviewing training queue.
+
+* [admin reviews training queue]: to see who's waiting and how long.
+  - Admin goes to {training queue display}.
+  - Admin selects filtering and sorting criteria (if any).
+  - Admin pages through display showing both <Training Application>s.
```
chrismedrela: I don't understand this. Do you mean both individual and group applications?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+## Setting up training course.
+
+* [admin creates training event]: and assigns people to it.
+  - Admin creates new <Event>.
+  - Admin goes to {training queue display}.
+  - Admin selects <Event>.
+  - Admin selects some <Training Application>s.
+  - AMY creates <Task>s to add all the selected <Person>s to the <Event>.
+  - AMY connects the <Training Application>s to the <Event>.
+  - AMY moves <Training Application>s to <Assigned Application List>.
```
chrismedrela: How is this list represented in the database? Is it that `<Training Application>` have a nullable ForeignKey to `<Event>` and if this field is not null, then we consider that application to be in `<Assigned Application List>`?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - AMY provides single-click "send to all" email link to admin to inform trainees.
```
chrismedrela: Do you mean something like `<a href="mailto:one@person, second@person">send to all</a>` or do you want to send the email through AMY?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - AMY creates <Training Progress> record for each <Person>.
```
chrismedrela: What this record represents? Is it a model with fields like `passed (bool)`, `attended (bool)`, `created_by (Person)`, `date`, `requirement` (set to the "Training" Requirement)?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+
+## Administering training course applications.
+
+* [admin removes people from training event][a]: and discards applications
+  - Admin goes to {assigned training application display}.
```
chrismedrela: Do {assigned training applications display} and {unassigned training applications display} differ (except for the set of records that is displayed)(for example, different columns are shown)?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - Admin removes them from training event.
+  - AMY deletes associated <Task>s.
+  - AMY deletes <Training Application>s.
+
+* [admin removes people from training event][b]: but keeps applications
+  - Admin goes to {assigned training application display}.
+  - Admin selects <Training Application>s.
+  - Admin removes them from training event.
+  - AMY deletes associated <Task>s.
+  - AMY moves <Training Application>s back to <Training Application List>.
```
chrismedrela: `- AMY disassociates <Training Application> with the <Event>.`

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+## Reviewing training progress.
+
+* [admin views training progress]: to see who's done what.
+  - Admin views {training progress page}.
+  - Admin selects filtering and sorting criteria (if any).
+
+## Tracking submission of written work.
+
+* [trainee reports submission of written work]: such as pull request or comment on lesson.
+  - Trainee views their own {training progress}.
```
chrismedrela: This view is different from the one on line 83, is that right?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - Trainee enters URL of written work.
+  - AMY updates <Training Task> associated with <Person> and <Event>.
```
chrismedrela: What is `<Training Task>`?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+
+* [admin reports submission of written work]: on behalf of trainee.
+  - Admin looks up <Training Progress> associated with trainee.
+  - Admin enters URL of written work in <Training Task>.
+
+* [admin reports submission of written work]: on behalf of trainee.
+  - Admin looks up <Training Progress> associated with trainee.
+  - Admin enters URL of written work in <Training Task>.
+  - AMY adds <Training Progress> to <Training Progress Review List>.
+
```
chrismedrela: Lines 98-100 are duplication of 94-96.

chrismedrela: `<Training Progress Review List>` <- is this a view? If not, how will we represent this list in database?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+## Tracking participation in interactive work.
+
+* [admin reports successful completion of interactive work]: such as discussion session or lesson demo.
+  - Admin views {training progress} for trainee.
+  - Admin selects <Training Task>.
+  - Admin:
+    - Enters date of interactive event.
+    - Adds comments (if any).
+  - Admin indicates whether <Training Task> is now complete.
```
chrismedrela: This requires Mentors and Examiners to go to the {training progress} view *for each trainee*. If I leaded a session, I'd like to have a view of all trainees, select trainees that attended my session, and then click "bulk pass".

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - Admin enters URL of written work in <Training Task>.
+
+* [admin reports submission of written work]: on behalf of trainee.
+  - Admin looks up <Training Progress> associated with trainee.
+  - Admin enters URL of written work in <Training Task>.
+  - AMY adds <Training Progress> to <Training Progress Review List>.
+
+* [admin reports completion of written work]: on behalf of trainee.
+  - Admin looks up <Training Progress> associated with trainee.
+  - Admin enters URL of written work in <Training Task>.
```
chrismedrela: I don't get it. What are `<Training Progress>` and `<Training Task>`?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - Admin selects <Training Progress>.
+  - Admin => [admin changes completion status of work for trainee].
+
+## Completing training.
+
+* [training automatically completed]: triggered by changes in <Training Task>s.
+  - AMY records change to <Training Task> and all <Training Task>s for <Person> now complete.
+  - AMY displays {badge creation form}.
+  - Admin confirms creation of badge.
+  - AMY moves <Training Progress> to <Past Training Progress List>.
```
chrismedrela: `- AMY creates certificates`?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - Admin selects some <Training Application>s.
+  - AMY creates <Task>s to add all the selected <Person>s to the <Event>.
+  - AMY connects the <Training Application>s to the <Event>.
+  - AMY moves <Training Application>s to <Assigned Application List>.
+  - AMY provides single-click "send to all" email link to admin to inform trainees.
+  - AMY creates <Training Progress> record for each <Person>.
+
+## Administering training course applications.
+
+* [admin removes people from training event][a]: and discards applications
```
chrismedrela: What if an Admin remove people from training event after the training took place? In that case, we need to mark `<Training Progress>` as `removed from training` or something like that.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+    - Do not yet have a PDF certificate.
+  - Admin adjusts information in form.
+    - Changes name of instructor(s) awarding certificates.
+    - Changes date of award.
+  - AMY generates PDF certificates.
+  - User views/downloads certificate on {certificate display}.
+
+## Questions
+
+- What to do if Admin deletes <Task> from training <Event> directly rather than via [admin removes people from training event]?
```
chrismedrela: Then nothing bad happens. The trainee application is just moved to `<Training Application List>`, assuming that this list is computed dynamically, i.e. it's a queryset.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+- What workflow *should* we use for marking training events as closed/stalled?
+- Do we only allow feedback loop/multiple tries for interactive work (discussions and demos), but *not* for written work?  That's how the workflows are written now...
+- Can we support group applications by providing one extra free-text field in the application called "Group Name", and asking people who want to co-apply to fill it in with the same value?  We can then sort/filter/group by that value to find people in the same group.
```
chrismedrela: That's a great idea! The only question is if a group of people will want to apply if every person in the group have to fill the form, instead of delegating one person to fill one group application. And one minor issue: in that case we don't want `reason` field in every application, only one for one group.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+    - Changes name of instructor(s) awarding certificates.
+    - Changes date of award.
+  - AMY generates PDF certificates.
+  - User views/downloads certificate on {certificate display}.
+
+## Questions
+
+- What to do if Admin deletes <Task> from training <Event> directly rather than via [admin removes people from training event]?
+- What workflow *should* we use for marking training events as closed/stalled?
+- Do we only allow feedback loop/multiple tries for interactive work (discussions and demos), but *not* for written work?  That's how the workflows are written now...
```
chrismedrela: Probably yes, assuming that a Trainee may fail Discussion/Demo Session and we want to give them another chance.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,211 @@
...
+  - Admin adjusts information in form.
+    - Changes name of instructor(s) awarding certificates.
+    - Changes date of award.
+  - AMY generates PDF certificates.
+  - User views/downloads certificate on {certificate display}.
+
+## Questions
+
+- What to do if Admin deletes <Task> from training <Event> directly rather than via [admin removes people from training event]?
+- What workflow *should* we use for marking training events as closed/stalled?
```
chrismedrela: Does it differ from `[admin ends training course]`?

chrismedrela:
@gvwilson, I reviewed what I could. My main impression is that the very first thing that we should agree on is not the workflow, but the schema and how will be represent data about progress in database.

docs/instructor-training-workflow.txt
gvwilson: Good idea - change made.

docs/instructor-training-workflow.txt
gvwilson: This is in the data model - I expect it will be the entire set of training applications.  I'll update to make this clearer.

docs/instructor-training-workflow.txt
gvwilson: I *think* so - we eventually need a Person, but we don't want to create those for people whose applications we reject.

docs/instructor-training-workflow.txt
gvwilson: Yup.

docs/instructor-training-workflow.txt
gvwilson: I think we need to approve applications one by one, because some will have Person objects and some won't.  If this works, we can go back later and add something to the application list display to allow bulk approval.

docs/instructor-training-workflow.txt
gvwilson: Agreed.

docs/instructor-training-workflow.txt
gvwilson: Agreed, but I usually write use cases without conditionals or loops, because control flow quickly gets out of hand and makes paths hard to read.  Of course, duplication makes paths hard to read too.  Life itself is hard sometimes...

docs/instructor-training-workflow.txt
gvwilson: I think so.

docs/instructor-training-workflow.txt
gvwilson: Yes, "both" was leftover from when there were both - I've taken it out.

docs/instructor-training-workflow.txt
gvwilson: Yes - I've removed mention of the <Assigned Application List>, since we can just filter by nullity of the foreign key.

docs/instructor-training-workflow.txt
gvwilson: the href (same as we're doing for workshop attendees)

docs/instructor-training-workflow.txt
gvwilson: The Training Progress record points at the specific tasks that need to be accomplished.

docs/instructor-training-workflow.txt
gvwilson: Agreed - can @chrismedrela and @pbanaszkiewicz decide whether to delete objects entirely, or mark them as deleted?  We had this discussion once before around Events and Persons.

docs/instructor-training-workflow.txt
gvwilson: It's a filter - wording clarified.

docs/instructor-training-workflow.txt
gvwilson: I've written on the assumption that `<Training Progress>` is one-to-many with `<Training Task>`, but you may decide to keep everything in `<Training Progress>` - that would be simpler...

docs/instructor-training-workflow.txt
gvwilson: Thanks for catching this.

docs/instructor-training-workflow.txt
gvwilson: Agreed - added.

docs/instructor-training-workflow.txt
gvwilson: We could.

docs/instructor-training-workflow.txt
gvwilson: But there will still be a reference from the Training Application to the Event, won't there?

docs/instructor-training-workflow.txt
gvwilson: Nope - good catch.

docs/instructor-training-workflow.txt
gvwilson: Agreed - let's get individual applications working, then add this.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,249 @@
...
+  - AMY marks <Training Progress> as "Completed".
+
+## Questions
+
+- What to do if Admin deletes <Task> from training <Event> directly rather than via [admin removes people from training event]?
+- Do we only allow feedback loop/multiple tries for interactive work (discussions and demos), but *not* for written work?  That's how the workflows are written now...
+- Can we support group applications by providing one extra free-text field in the application called "Group Name", and asking people who want to co-apply to fill it in with the same value?  We can then sort/filter/group by that value to find people in the same group.
+- Do we use one <Training Progress> with multiple fields, or have <Training Progress> point at multiple (atomic) <Training Task>s?  This doc assumes the latter.
+- Should each <Training Task> accumulate multiple date+comment pairs (one for each attempt to pass that task), or should people have multiple <Training Task> objects of a particular type (e.g., "demo"), N-1 of which are "fail", the last of which is "pass"?
+- Should we actually delete <Training Application>s that we don't want, or mark as "discarded"?
```
chrismedrela: I always prefer to mark records as "discarded" or "deleted" rather than remove them entirely from db, because you can revert this operation.

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,249 @@
...
+
+* [admin ends training course]: because it has timed out.
+  - Admin selects training <Event>.
+  - Admin marks <Event> as "Stalled".
+  - AMY marks all associated <Training Progress>es as "Unsuccessful".
+  - AMY marks <Training Progress> as "Completed".
+
+## Questions
+
+- What to do if Admin deletes <Task> from training <Event> directly rather than via [admin removes people from training event]?
```
chrismedrela: OK, so this looks like redundancy - you have Tasks indicating that a Person was matched with an Event, and at the same time, you can match Application to an Event. What if we drop `TrainingApplication.event` and rely only on Tasks records?

chrismedrela:
@gvwilson, I'd like to propose one major change to the schema you proposed. You propose to have:

- `Training Progress` associated with an `Event` and `Person`
- many `Training Task`s to one `Training Progress`.

I propose to drop `Training Progress` and associate `Training Task` directly to a `Person` (and don't associate it with any `Event`). If I submit a homework, it doesn't matter which Instructor Training I finished. If we want to add a new requirement that must be passed before an Instructor Training, how do you represent that in database?

docs/instructor-training-workflow.txt
```diff
@@ -0,0 +1,249 @@
...
+- Do we only allow feedback loop/multiple tries for interactive work (discussions and demos), but *not* for written work?  That's how the workflows are written now...
+- Can we support group applications by providing one extra free-text field in the application called "Group Name", and asking people who want to co-apply to fill it in with the same value?  We can then sort/filter/group by that value to find people in the same group.
+- Do we use one <Training Progress> with multiple fields, or have <Training Progress> point at multiple (atomic) <Training Task>s?  This doc assumes the latter.
+- Should each <Training Task> accumulate multiple date+comment pairs (one for each attempt to pass that task), or should people have multiple <Training Task> objects of a particular type (e.g., "demo"), N-1 of which are "fail", the last of which is "pass"?
```
chrismedrela: I prefer the latter. The former requires to store a list of pairs which would result in a new model.

PR #851, opened 2016-06-28
Change default invoice status to 'not-invoiced'

### LINKED ISSUE #850: Default invoice status
Default invoice status when creating a new event should be "Not Invoiced" -- currently it is "Unknown"

narayanaditya95:
@maneesha requested a change in the default invoice status.
Fixes #850.


narayanaditya95:
Ready to be merged.

workshops/views.py
```diff
@@ -1112,6 +1112,9 @@ class EventCreate(LoginRequiredMixin, PermissionRequiredMixin,
     form_class = EventForm
     template_name = 'workshops/event_create_form.html'
 
+    def get_initial(self):
+        return {'invoice_status': 'not-invoiced'}
+
```
pbanaszkiewicz: Why not change the default value in model? (Except that it triggers a migration.)

narayanaditya95: I tried that before, it ends up breaking a test. 
`tesr/test_event.py L:50`
```
    def test_get_uninvoiced_events(self):
        """Test that the events manager can find events that owe money"""

        uninvoiced_events = Event.objects.uninvoiced_events()

        # There should be as many as there are strictly future events.
        assert len(uninvoiced_events) == self.num_uninvoiced_events

        # Check that events with a fee of zero or None are still on this list
        assert any([x for x in uninvoiced_events if not x.admin_fee])
```
It asserts that the number of uninvoiced events are the same as the number of future event, which breaks after changing the default.
Moreover, `unknown` seems to be the right default for the `invoice_status` rather than `not-invoiced` when the field `invoice_status` is not specified.

pbanaszkiewicz: So you're saying that we shouldn't fix bugs because doing so breaks our tests? :-)

Can you fix the test and change the default value in the model?

narayanaditya95:
@pbanaszkiewicz Ping.

workshops/test/base.py
```diff
@@ -299,7 +299,8 @@ def _setUpEvents(self):
         self.num_uninvoiced_events = 0
         self.num_upcoming = 0
         for e in Event.objects.all():
-            if e.invoice_status == 'not-invoiced' and e.start < today:
+            e.is_past_event = e.start < today and (e.end is None or e.end < today)
```
narayanaditya95: There was a difference in logic between the statistics we generate here and `Event.objects.uninvoiced_events().count()`. The method `uninvoiced_events()` also filters out events which have a `end` date in the future or nulled. I replicated here and `self.num_uninvoiced_events` now returns the correct number of events which haven't been invoiced.

workshops/models.py
```diff
@@ -685,7 +685,7 @@ class Event(AssignmentMixin, models.Model):
         max_length=STR_MED,
         choices=INVOICED_CHOICES,
         verbose_name='Invoice status',
-        default='unknown', blank=False,
+        default='not-invoiced', blank=False,
```
pbanaszkiewicz: Doesn't that trigger a migration?

narayanaditya95: Done

pbanaszkiewicz:
@narayanaditya95 sorry for being so nitpicky, but you must be working off of an old commit because that migration should have 0104 number. Please rebase against `develop` and re-create the migration.

narayanaditya95:
I am getting used to "pull-first push-later" workflow. :)

pbanaszkiewicz:
Thanks for this fix, @narayanaditya95!

PR #863, opened 2016-06-29
Make links in help texts in training request form clickable

### LINKED ISSUE #845: Training request form should use <a>s in help texts
Fields at the bottom of the form use `help_text`s but don't replace links with <a>links</a>.

chrismedrela:
Fixes #845.


narayanaditya95:
This change may generate a migration.

chrismedrela:
Good catch -- I've not check that. Now I checked it and it does *not* generate any migration. This PR changes help_text in a form, not a model.

narayanaditya95:
Sorry, I overlooked that it is a modelform and not a model.

PR #870, opened 2016-07-01
Add group name and comment fields to Training Request Form

pbanaszkiewicz:
Group name field can be used to allow group registration.

Comment field can be used for any other notes.


pbanaszkiewicz:
I didn't say why we need these fields: this way we solve group training applications in a short-term.

PR #871, opened 2016-07-01
Instructor training workflow

chrismedrela:
@pbanaszkiewicz , here is the schema proposal, it turns out we need only two models. Please review. And don't merge. :)


chrismedrela:
Ready for review.

Issues:

1. I've left some questions in the code (see FIXME comments).

2. Handling of multiple forms on one webpage. For example, have a look at trainees view. There are two functionalities: bulk discarding and bulk add progresses. They have to live in the same `<form>` tag, because they share checkboxes that let you select trainees. This results in a lot of ugly workarounds in the code. 

3. Better way to display discarded progresses. They're striked and there is cross icon now. The strike is not visible and the cross can be misinterpreted as delete button.

4. I've put everything into one commit, because I don't want to have history going like this: "fix A", "implement B", "introduce C feature", "bunch of changes to A, B and C", "fix bug in A and C", "another bunch of changes to A, B and C". So I'd like to split it into smaller commits after final review, just before merging. Look at this commit message to see how the messages of smaller commits will look like.

gvwilson:
@chrismedrela can you please re-base? @pbanaszkiewicz can you please do a technical review (esp. look for the FIXME lines)?

workshops/filters.py
```diff
@@ -207,6 +211,139 @@ def get_order_by(self, order_value):
         return super().get_order_by(order_value)
 
 
+def filter_all_persons(queryset, all_persons):
+    if all_persons:
```
pbanaszkiewicz: Please add a docstring -- I understood the purpose of this function after reading `all_persons` label.

workshops/filters.py
```diff
@@ -207,6 +211,139 @@ def get_order_by(self, order_value):
...
+        return queryset
+    else:
+        return queryset.filter(
+            task__role__name='learner',
+            task__event__tags__name='TTT').distinct()
+
+
+def filter_trainee_name(queryset, name):
+    return queryset.filter(Q(personal__icontains=name) |
```
pbanaszkiewicz: This is tricky to get right - can you find Greg if you search for "Greg Wilson"? Take a look at `search` view, where we have tokenization.

workshops/filters.py
```diff
@@ -207,6 +211,139 @@ def get_order_by(self, order_value):
...
+    else:  # return only trainees without unevaluated homework
+        return queryset.exclude(trainingprogress__state='n') \
+                       .distinct()
+
+
+def filter_training(queryset, training):
+    if training is None:
+        return queryset
+    else:
+        print(training)
```
pbanaszkiewicz: Debugging remains?

workshops/filters.py
```diff
@@ -207,6 +211,139 @@ def get_order_by(self, order_value):
...
+        name='is_swc_instructor',
+    )
+
+    is_dc_instructor = django_filters.BooleanFilter(
+        label='Is DC Instructor?',
+        name='is_dc_instructor',
+    )
+
+    non_evaluated_homework = django_filters.BooleanFilter(
+        label='Unevaluated homework?',
```
pbanaszkiewicz: Maybe rather 'Has unevaluated homework' instead of a question.

workshops/filters.py
```diff
@@ -207,6 +211,139 @@ def get_order_by(self, order_value):
...
+        return queryset
+    elif choice == 'u':  # unmatched
+        return queryset.filter(person=None)
+    elif choice == 'p':  # matched person, unmatched training
+        return queryset.filter(person__isnull=False)\
+                       .exclude(person__task__role__name='learner',
+                                person__task__event__tags__name='TTT')\
+                       .distinct()
+    else:
+        assert choice == 't'  # matched person and training
```
pbanaszkiewicz: Why asserts in production code?

workshops/forms.py
```diff
@@ -1,15 +1,16 @@
 import re
+from django.core.exceptions import ValidationError
 
 from django import forms
 from django.core.validators import RegexValidator
 from django.forms import (
     HiddenInput, CheckboxSelectMultiple, TextInput, modelformset_factory,
     RadioSelect,
-)
+    URLField)
```
pbanaszkiewicz: We keep `)` on a separate line to not trigger line changes when new imports are added. Please use this format.

workshops/forms.py
```diff
@@ -1,15 +1,16 @@
 import re
+from django.core.exceptions import ValidationError
```
pbanaszkiewicz: This import should be grouped with other `from django*` imports.

workshops/forms.py
```diff
@@ -21,7 +22,7 @@
     InvoiceRequest, EventSubmission, Language,
     TrainingRequest,
     DCSelfOrganizedEventRequest,
-)
+    TrainingProgress, TrainingRequirement)
```
pbanaszkiewicz: Please keep `)` on a separate line.

workshops/forms.py
```diff
@@ -1036,6 +1037,32 @@ class Meta:
...
+        lookup_class=lookups.PersonLookup,
+        label='Matched Person',
+        required=False,
+        help_text=AUTOCOMPLETE_HELP_TEXT,
+        widget=selectable.AutoComboboxSelectWidget,
+    )
+
+    class Meta:
+        model = TrainingRequest
+        exclude = ()
```
pbanaszkiewicz: Recently you declared to start using `fields` instead of `exclude` to avoid some mistakes :-)

workshops/forms.py
```diff
@@ -1080,3 +1107,244 @@ class Meta:
...
+        lookup_class=lookups.TTTEventLookup,
+        label='Event',
+        required=False,
+        help_text=AUTOCOMPLETE_HELP_TEXT,
+        widget=selectable.AutoComboboxSelectWidget,
+    )
+
+    # In the form, "----" option is displayed for requirement even though
+    # it's required non-blank field. This is a new bug in Django. See
+    # https://code.djangoproject.com/ticket/26813.
```
pbanaszkiewicz: Yeah, controlling some of Django form elements (combinations of null/blank/required and selected by default or not selected by default) is just shitty. This is uneven fight.

workshops/forms.py
```diff
@@ -1080,3 +1107,244 @@ class Meta:
...
+    # When set to True, the form is valid only if every request is matched to
+    # one person. Set to True when 'unmatch' button is clicked, because
+    # unmatching makes sense only if each selected TrainingRequest is matched
+    # with one person.
+    check_person_matched = False
+
+    def clean(self):
+        super().clean()
+        if (self.check_person_matched and
+            any(r.person is None for r in self.cleaned_data.get('requests', []))):
```
pbanaszkiewicz: Assign the `any()` part to a variable and simplify writing of this conditional?

Also, is there a test that checks all of these conditions here?

workshops/management/commands/fake_database.py
```diff
@@ -17,7 +17,7 @@
     Qualification,
     Host,
     Event,
-    Task,
+    Task
```
pbanaszkiewicz: Don't drop that comma.

workshops/models.py
```diff
@@ -1,5 +1,6 @@
 import datetime
 import re
+from django.core.exceptions import ValidationError
 from social.apps.django_app.default.models import UserSocialAuth
```
pbanaszkiewicz: These two imports should be grouped in the third-party imports group.

workshops/models.py
```diff
@@ -397,6 +400,42 @@ def is_staff(self):
...
+
+        requirements = {training, homework, discussion, demo}
+        progresses = self.trainingprogress_set.filter(discarded=False, state='p')
+        passed_requirements = {p.requirement for p in progresses}
+        return requirements - passed_requirements
+
+    @property
+    def missing_dc_instructor_requirements(self):
+        """ Returns set of requirements that are not passed yet by the
+        trainee and are mandatory to become DC Instructor.. """
```
pbanaszkiewicz: Double dot here and above in the docstring.

workshops/models.py
```diff
@@ -397,6 +400,42 @@ def is_staff(self):
...
+        passed_requirements = {p.requirement for p in progresses}
+        return requirements - passed_requirements
+
+    @property
+    def training_tasks(self):
+        """ Returns Tasks related to Instuctor Training events at which this
+        person was trained."""
+        return Task.objects.filter(person=self,
+                                   role__name='learner',
+                                   event__tags__name='TTT')
```
pbanaszkiewicz: I think you took these functions a little bit too far. They don't really seem to be doing much with `Person` model itself, they rather seem like auxiliary functions / utilities. I'd move them to `util.py`.

workshops/models.py
```diff
@@ -397,6 +400,42 @@ def is_staff(self):
...
+        trainee and are mandatory to become SWC Instructor.. """
+        training = TrainingRequirement.objects.get(name='Training')
+        homework = TrainingRequirement.objects.get(name='SWC Homework')
+        discussion = TrainingRequirement.objects.get(name='Discussion')
+        demo = TrainingRequirement.objects.get(name='SWC Demo')
+
+        requirements = {training, homework, discussion, demo}
+        progresses = self.trainingprogress_set.filter(discarded=False, state='p')
+        passed_requirements = {p.requirement for p in progresses}
+        return requirements - passed_requirements
```
pbanaszkiewicz: This little function makes 5 queries and then loops through results while filtering. I really hope:

1. you can make it more efficient
2. it's not being used in one view for >1 person.

workshops/models.py
```diff
@@ -1950,3 +1993,152 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
         default='', null=False, blank=True,
         help_text='What else do you want us to know?',
         verbose_name='Anything else?')
+
+    STATES = (
+        ('p', 'Pending'),  # initial state
+        ('a', 'Accepted'),  # state after matching a Person record
+        ('d', 'Discarded'),  # FIXME or deleted?
```
pbanaszkiewicz: Seems reasonable that we'd discard a training request. Thoughts, @gvwilson?

workshops/models.py
```diff
@@ -1950,3 +1993,152 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
...
+
+        if self.state == 'a' and self.person is None:
+            raise ValidationError({'person': 'Accepted training request must '
+                                             'be matched to a person.'})
+
+    def get_absolute_url(self):
+        return reverse('trainingrequest_details', args=[self.pk])
+
+
+class TrainingRequirement(models.Model):
```
pbanaszkiewicz: Can you please make it tracked by reversion as well?

workshops/models.py
```diff
@@ -1950,3 +1993,152 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
...
+    # FIXME Maybe let's rename it as evaluated_by?
+    awarded_by = models.ForeignKey(Person,
+                                   on_delete=models.PROTECT,
+                                   null=True, blank=True,
+                                   related_name='+')
+    requirement = models.ForeignKey(TrainingRequirement,
+                                    on_delete=models.PROTECT)
+
+    STATES = (
+        ('n', 'Not-evaluated yet'),
```
pbanaszkiewicz: Dash not required here: 'Not evaluated'

workshops/models.py
```diff
@@ -1950,3 +1993,152 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
...
+
+
+@reversion.register
+class TrainingProgress(CreatedUpdatedMixin, models.Model):
+    trainee = models.ForeignKey(Person, on_delete=models.PROTECT)
+
+    # Mentor/examiner who evaluates homework / session. May be null when a
+    # trainee submits their homework.
+
+    # FIXME Maybe let's rename it as evaluated_by?
```
pbanaszkiewicz: I'd go with 'evaluated_by'.

workshops/models.py
```diff
@@ -1950,3 +1993,152 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
...
+    state = models.CharField(choices=STATES, default='p', max_length=1)
+
+    # When we end training and trainee has gone silent, or passed their
+    # deadline, we set this field to True.
+    discarded = models.BooleanField(
+        default=False,
+        verbose_name='Discarded -- check if the trainee has gone silent '
+                     'or passed their training deadline.')
+
+    event = models.ForeignKey(Event, null=True, blank=True)
```
pbanaszkiewicz: Just a thought: you can narrow choices here to only TTT events with `limit_choices_to`.

workshops/models.py
```diff
@@ -1950,3 +1993,152 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
...
+    url = models.URLField(null=True, blank=True)
+    notes = models.TextField(blank=True)
+
+    def get_absolute_url(self):
+        return reverse('trainingprogress_edit', args=[str(self.id)])
```
pbanaszkiewicz: `get_absolute_url`s by general go to details pages - is one available for `TrainingProgress`?

workshops/models.py
```diff
@@ -1950,3 +1993,152 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
...
+            raise ValidationError({'url': msg})
+        elif not self.requirement.url_required and self.url:
+            msg = 'In the case of {}, this field must be left empty.'.format(self.requirement)
+            raise ValidationError({'url': msg})
+
+        if self.requirement.event_required and not self.event:
+            msg = 'In the case of {}, this field is required.'.format(self.requirement)
+            raise ValidationError({'event': msg})
+        elif not self.requirement.event_required and self.event:
+            msg = 'In the case of {}, this field must be left empty.'.format(self.requirement)
```
pbanaszkiewicz: These validation errors are general to the form, so they won't show up near the `self.url` or `self.event` field -- maybe you should use `clean_url` and `clean_event` methods instead?

workshops/models.py
```diff
@@ -1950,3 +1993,152 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
...
+        if self.state != 'n' and self.awarded_by is None:
+            msg = 'State must be "not-evaluated yet" when there is no ' \
+                  'mentor/examiner matched.'
+            raise ValidationError({'state': msg})
+
+        super().clean()
+
+    # FIXME Should this live in template? It's much easier to write this in
+    # Python rather than as a stack of if-elif-elif statements in Django
+    # Template.
```
pbanaszkiewicz: I solved this kind of enumerate problem with a template tag.

workshops/models.py
```diff
@@ -1950,3 +1993,152 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
...
+    # FIXME I have no idea how to move this into template. The logic is too
+    # complicated.
+    @property
+    def long_description(self):
+        """Used in templates in tooltip."""
+
+        state2human_readable_name = {
+            'n': 'non-evaluated',
+            'f': 'failed',
+            'p': 'passed',
```
pbanaszkiewicz: But you have access to `TrainingProgress.STATES`.

workshops/models.py
```diff
@@ -1950,3 +1993,152 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
...
+
+        switch = {
+            'n': 'warning',
+            'f': 'danger',
+            'p': 'success',
+        }
+        return switch[self.state]
+
+    # FIXME I have no idea how to move this into template. The logic is too
+    # complicated.
```
pbanaszkiewicz: Definitely make this a template tag.

workshops/static/bulk_email.js
```diff
@@ -0,0 +1,27 @@
+/*
+FIXME I'm not JS guru and I've no idea what is The Right Way to export
+a function. I guess not the way I did it, because this way pollutes global
+space.
```
pbanaszkiewicz: Looks good to me -- but I'm not JS expert either (never heard anyone developing AMY call themselves a JS expert :-) ).

workshops/views.py
```diff
@@ -2,7 +2,7 @@
 import datetime
 import io
 import re
-from django.views.decorators.http import require_POST
+from crispy_forms.layout import Button, Submit
```
pbanaszkiewicz: Please move this import to the third-party imports group.

workshops/views.py
```diff
@@ -120,7 +123,7 @@ def get_context_data(self, **kwargs):
         else:
             context['title'] = 'New object'
 
-        context['form_helper'] = bootstrap_helper
+        context['form_helper'] = BootstrapHelper()
```
pbanaszkiewicz: Why?

workshops/views.py
```diff
@@ -3103,10 +3106,131 @@ def duplicates(request):
...
+
+
+@login_required
+def trainee_dashboard(request):
+    swc_form = SendHomeworkForm(submit_name='swc-submit')
+    dc_form = SendHomeworkForm(submit_name='dc-submit')
+
+    # FIXME We hardcode requirement set. What's more, what we do is a bit
+    # different for each Requirement. This will results in problems when we
+    # want to switch to dynamic RequirementSet. Any ideas how to fix that?
```
pbanaszkiewicz: Hmmm, maybe say implement a real requirement set?

workshops/views.py
```diff
@@ -3103,10 +3106,131 @@ def duplicates(request):
...
+        badge__name='swc-instructor').exists()
+    request.user.is_dc_instructor = request.user.award_set.filter(
+        badge__name='dc-instructor').exists()
+
+    if request.method == 'POST':
+        # determine which form (SWC or DC homework submission) was sent
+        if 'swc-submit' in request.POST:
+            swc = True
+        else:
+            assert 'dc-submit' in request.POST
```
pbanaszkiewicz: Again, asserts in code?

workshops/views.py
```diff
@@ -3103,10 +3106,131 @@ def duplicates(request):
...
+            swc = False
+
+        if swc:
+            requirement = TrainingRequirement.objects.get(name='SWC Homework')
+        else:
+            requirement = TrainingRequirement.objects.get(name='DC Homework')
```
pbanaszkiewicz: You can merge this conditions with above conditions.

workshops/views.py
pbanaszkiewicz: Also, please explain how does this currently work.

workshops/views.py
```diff
@@ -3136,67 +3260,302 @@ def trainingrequest_create(request):
...
+                username=create_username(training_request.personal,
+                                         training_request.family),
+                personal=training_request.personal,
+                family=training_request.family,
+                email=training_request.email,
+            )
+
+    except IntegrityError as e:
+        # FIXME Any idea how to narrow scope of this except clause only to
+        # IntegrityErrors related to email address?
```
pbanaszkiewicz: Unfortunately this does not seem to be possible, or at least I wasn't successful finding a solution when I was fixing the very same problem in some management command.

workshops/views.py
```diff
@@ -3136,67 +3260,302 @@ def trainingrequest_create(request):
...
+
+        # email address is not unique
+        messages.error(request, 'Could not create a new person -- '
+                                'there already exists a person with '
+                                'exact email address.')
+        return None
```
pbanaszkiewicz: Why does this function return `None` or a redirection?

pbanaszkiewicz:
@chrismedrela I'm sleepy and a little sick, and my review is definitely incomplete, but I want to run this branch before I come back to templates and views. Can you please rebase?

chrismedrela:
@pbanaszkiewicz, yes, but let me do this this morning, it's too late.

chrismedrela:
@pbanaszkiewicz, rebased.

workshops/filters.py
chrismedrela: I've replaced it with something much simpler:
```python
def filter_trainee_name(queryset, name):
    tokens = re.split('\s+', name)  # 'Greg Wilson' -> ['Greg', 'Wilson']
    # Each token must match email address or github username or personal or
    # family name.
    for token in tokens:
        queryset = queryset.filter(Q(personal__icontains=token) |
                                   Q(family__icontains=token) |
                                   Q(email__icontains=token) |
                                   Q(github__icontains=token))
    return queryset
```
What do you think about that?

workshops/filters.py
chrismedrela: Part of contract of this function is that `choice` must be `''`, `'u'`, `'p'` or `'t'`. We want to fail loud when `choice` is something different.

workshops/forms.py
chrismedrela: This is a form for admins, and admins should be able to edit all fields. It's much different situation than `TrainingRequestForm` which is accessible to trainees and where inclusion of unwanted fields is risky.

workshops/forms.py
chrismedrela: I rethought it and I think that in this particular case this is a feature -- mentors and examiners must explicitly choose requirement type (like Homework or SWC Discussion). If there was default type, they could forget about selecting a different type than the default one.

workshops/forms.py
chrismedrela: What exactly do you want to test here?

workshops/models.py
chrismedrela: BTW, PEP257 doesn't say anything about putting space after starting `"""` and before ending `"""`.

workshops/models.py
chrismedrela: This property is used in templates: 

- https://github.com/swcarpentry/amy/pull/871/files#diff-f4cb88d18f36cf5a6e5fc2aeb1a32602R68 
- https://github.com/swcarpentry/amy/pull/871/files#diff-e4b786fea3b24ff03fcc3d341a1c213bR53

workshops/models.py
chrismedrela: Sth like this does only one query per person:
```python
    @property
    def missing_swc_instructor_requirements(self):
        """Returns set of requirements' names (list of strings) that are not
        passed yet by the trainee and are mandatory to become SWC Instructor.
        """
        requirements = [
            'Training',
            'SWC Homework',
            'Discussion',
            'SWC Demo'
        ]
        passed = self.trainingprogress_set\
                     .filter(discarded=False, state='p',
                             requirement__name__in=requirements) \
                     .values_list('requirement__name', flat=True)
        return set(requirements) - set(passed)
```
Unfortunately, this is used in a view for >1 person and I have no idea how to make it faster.

workshops/models.py
chrismedrela: We don't track models like `Role` or `Tag`. Should we also start tracking them?

workshops/models.py
chrismedrela: No, we don't have detail/read page for `TrainingProgress`. Users can see details about progress record in a popover in in trainees list view, when they hover mouse over one of progress records.

workshops/models.py
chrismedrela: This validation requires that `requirement` is already cleaned, so it cannot go to `clean_FIELD` methods.

workshops/models.py
gvwilson: We'll eventually want to add such a page, but that can wait...

workshops/filters.py
pbanaszkiewicz: Wouldn't searching for "Wilson Banaszkiewicz" return empty record with this code? Other thing is that I don't know if I want such query to return anything…

So: I'm not 100% certain this query will do.

workshops/filters.py
pbanaszkiewicz: Do that in tests, not in production code. The reason why you shouldn't rely on `assert`s is that they can be disabled with specific Python flag.

workshops/forms.py
pbanaszkiewicz: But since this is Django's bug, we can expect this "bug/feature" behavior to change. I'd not rely on it.

workshops/forms.py
pbanaszkiewicz: Tests aren't included in this PR (unless I missed them somehow), but they should check `clean()` behavior for different combinations of `self.check_person_matched` and the-long-condition.

workshops/models.py
pbanaszkiewicz: OK. Doesn't really matter to me, anyway.

workshops/models.py
pbanaszkiewicz: Then you should not make a property out of this helper function.

I talked to you and Greg about the root issue here and the response was "don't try to pre-optimize". You see, if when person is added to the training event, they should be assigned "empty" requirements upfront. Then you'd have no troubles checking which requirements were fulfilled and which were not. (So this is like copying empty requirements when people are being allowed/invited to the training.)

workshops/models.py
pbanaszkiewicz: Yeah, we could, but it seems out of the scope of this PR.

workshops/models.py
pbanaszkiewicz: OK.

workshops/models.py
pbanaszkiewicz: Okay, but I'll test this out.

pbanaszkiewicz:
There's still some work, mostly with writing tests and template tags, and responding to the rest of the comments.

Should we schedule this PR for v1.7 (overdue) or v1.8? I'm thinking about v1.8, but this depends on @chrismedrela / @gvwilson  deadlines.

chrismedrela:
@pbanaszkiewicz, I got feedback from Christina and we have three other people that will do walkthroughs soon (last one on Friday). I aim for v1.8 in order not to delay v1.7 release.

workshops/filters.py
chrismedrela: Yes, it'd return empty queryset. Examples of searches that will return Greg:

- Greg
- Greg Wilson
- Wilson Greg
- Greg Wilson gvwilson
- gvwilson Greg
- Wilson gvwilson@software-carpentry.org
- Greg greg
- reg ilso
- software-carpentry
- greg carpentry

workshops/filters.py
chrismedrela: What exactly should I do in tests?

And what about adding something like this at the beginning of the function:
```python
if choices not in ('', 'u', 'p', 't'):
    raise ValueError('Invalid choice value')
```

workshops/forms.py
chrismedrela: I've just noticed that this Django bug is actually about RadioSelect, not select widget:

> Unlike the select widget, where a blank option is idiomatic even for required fields, ...

So we can actually rely on this behaviour.

workshops/models.py
chrismedrela: I don't get how assigning "empty" requirements upfront helps. You still need to execute one query for each person.

workshops/views.py
chrismedrela: The helper can be edited by views (i.e. by `TrainingProgressCreate`). We should have one new helper for each view, not one shared for all views.

workshops/views.py
chrismedrela: Implementing real requirement set doesn't help, because we have different business logic for each requirement.

workshops/views.py
chrismedrela: Control flow. We want to exit `trainingrequest_details` view with returned redirection. If it's `None`, we continue execution of this view.

chrismedrela:
Close #921 when this get merged.

chrismedrela:
Tomorrow, I'll finish writing tests and this branch will be ready for final review. After final review, I'll break this branch into multiple commits, in the way that I've described in the commit message.

workshops/views.py
pbanaszkiewicz: And what's that logic behind them? Can you summarize here?

pbanaszkiewicz:
@gvwilson @chrismedrela I want to close and release v1.7 soon (in next 4 days, let's say). Do you want this PR to be in v1.7?

gvwilson:
Yes please.

chrismedrela:
@gvwilson, I've written tests and implemented "download all training requests as CSV" feature. 

@pbanaszkiewicz This branch is ready for final review. Note, that I did a lot of changes after your last review. After the review, I'll split it into multiple commits (one day of work). There is one more thing that I need to do: write a one-off tool to import data from spreadsheets.

And yes, let's schedule it for v1.7. Let's merge this huge PR this weekend.

gvwilson:
:+1: to merging this on the w'end.
Thx,
G

workshops/filters.py
```diff
@@ -1,8 +1,12 @@
-import django.forms.widgets
+import re
 
+import django.forms.widgets
 import django_filters
+from django.db.models import Q
+from django.forms import widgets
```
pbanaszkiewicz: if you import `widgets` from `django.forms`, why import `django.forms.widgets` separately?

workshops/filters.py
```diff
@@ -70,8 +75,15 @@ def filter(self, qs, value):
             return qs
 
 
-class FilterSetWithoutHelpText(django_filters.FilterSet):
-    """Because of some stupidity this got merged to django-filters:
+class AMYFilterSet(django_filters.FilterSet):
+    """
+    This base class server two roles:
```
pbanaszkiewicz: server?

workshops/filters.py
```diff
@@ -207,7 +222,218 @@ def get_order_by(self, order_value):
...
+    if all_persons:
+        return queryset
+    else:
+        return queryset.filter(
+            task__role__name='learner',
+            task__event__tags__name='TTT').distinct()
+
+
+def filter_trainee_name(queryset, name):
+    if name == '':
```
pbanaszkiewicz: missing docstring and why not more pythonic `if not name`?

workshops/filters.py
```diff
@@ -207,7 +222,218 @@ def get_order_by(self, order_value):
...
+
+def filter_unevaluated_homework(queryset, flag):
+    if flag:  # return only trainees with an unevaluated homework
+        return queryset.filter(trainingprogress__state='n').distinct()
+    else:
+        return queryset
+
+
+def filter_training_request(queryset, flag):
+    if flag is None:
```
pbanaszkiewicz: this one should have logic explained either with docstring or with comments

workshops/filters.py
```diff
@@ -207,7 +222,218 @@ def get_order_by(self, order_value):
...
+        return queryset
+    elif choice == 'u':  # unmatched
+        return queryset.filter(person=None)
+    elif choice == 'p':  # matched trainee, unmatched training
+        return queryset.filter(person__isnull=False)\
+                       .exclude(person__task__role__name='learner',
+                                person__task__event__tags__name='TTT')\
+                       .distinct()
+    else:
+        assert choice == 't'  # matched trainee and training
```
pbanaszkiewicz: You still have asserts in the code.

workshops/forms.py
```diff
@@ -33,35 +34,89 @@
...
+        if add_submit_button:
+            self.add_input(Submit(submit_name, submit_label))
+
+        if add_delete_button:
+            self.add_input(Submit(
+                'delete', 'Delete',
+                onclick='return confirm("Are you sure you want to delete it?");',
+                form='delete-form',
+                css_class='btn-danger',
+                style='float: right;'))
```
pbanaszkiewicz: While in most I'm okay with refactoring of bootstrap helper classes, this deletion button seems wrong, because we need more customized messages in `onclick` (e.g. "Are you sure you want to remove 2016-07-30-Krakow event?"). It also requires (according to the docstring) a separate POST form while in the previous approach a link was enough.

workshops/forms.py
```diff
@@ -70,26 +125,11 @@ def __init__(self, form=None):
...
     """For use in inline formsets."""
     template = 'bootstrap/table_inline_formset.html'
 
 
-bootstrap_helper = BootstrapHelper()
-bootstrap_helper_get = BootstrapHelperGet()
-bootstrap_helper_with_add = BootstrapHelperWithAdd()
-bootstrap_helper_filter = BootstrapHelperFilter()
-bootstrap_helper_wider_labels = BootstrapHelperWiderLabels()
-bootstrap_helper_inline_formsets = BootstrapHelperFormsetInline()
```
pbanaszkiewicz: Isn't removal of these pre-created instances against DRY?

workshops/forms.py
```diff
@@ -33,35 +34,89 @@
 
 class BootstrapHelper(FormHelper):
     """Layout and behavior for crispy-displayed forms."""
-    form_class = 'form-horizontal'
-    label_class = 'col-lg-2'
-    field_class = 'col-lg-8'
     html5_required = True
 
-    def __init__(self, form=None):
+    def __init__(self, form=None,
```
pbanaszkiewicz: Will this be compatible with #910?

workshops/forms.py
```diff
@@ -1094,3 +1206,283 @@ class Meta:
...
+        # the template where this form is used
+
+        # We use formnovalidate on submit buttons to disable browser
+        # validation. This is necessary because this form is used along with
+        # BulkMatchTrainingRequestForm, which have required fields. Both
+        # forms live inside the same <form> tag. Without this attribute,
+        # when you click one of the following submit buttons, the browser
+        # reports missing values in required fields in
+        # BulkMatchTrainingRequestForm.
+        FormActions(
```
pbanaszkiewicz: Isn't form actions a thing from bootstrap v2?

workshops/models.py
```diff
@@ -1,18 +1,19 @@
 import datetime
 import re
-from social.apps.django_app.default.models import UserSocialAuth
 from urllib.parse import urlencode
 
 from django.contrib.auth.models import (
-    AbstractBaseUser, BaseUserManager, PermissionsMixin)
+    AbstractBaseUser, BaseUserManager, PermissionsMixin
```
pbanaszkiewicz: add trailing comma

workshops/static/amy_utils.js
```diff
@@ -0,0 +1,134 @@
+/* A file with functionality that is enabled on each page. */
+
+function bulk_email() {
+    /*
+    Look at all selected <input type="checkbox" email="example@example.org">
+    tags on the webpage and join email addresses using comma as a separator
+    (i.e. "first@example.org,second@example.org").
```
pbanaszkiewicz: I think we need to look for more general solutions (ref #934) -- but maybe later

workshops/templates/workshops/all_airports.html
```diff
@@ -3,11 +3,6 @@
 {% load crispy_forms_tags %}
 {% load pagination %}
 
-{% block sidebar %}
-    <h3>Filter</h3>
-    {% crispy filter.form form_helper %}
-{% endblock %}
```
pbanaszkiewicz: Why do you remove the sidebar from that page?

pbanaszkiewicz: Okay, I can see why (because the sidebar's form is included one level higher in templates hierarchy https://github.com/swcarpentry/amy/pull/871/files#diff-ee7eb5718e08d7967a72f5c6dbc4860bR12 )

workshops/templates/workshops/all_trainees.html
```diff
@@ -0,0 +1,154 @@
...
+{% block content %}
+
+<form role="form" class="form-horizontal training-progress" method="post">
+{% if form.errors.trainees or discard_form.errors.trainees %}
+  <div class="alert alert-danger" role="alert">Select at least one trainee.</div>
+{% elif form.errors or discard_form.errors %}
+  <div class="alert alert-danger" role="alert">Fix errors below and try again.</div>
+{% endif %}
+
+<p>Hint: This list is ordered by last login time by default. You can ask trainees participating in your discussion/demo session to login into AMY and, then, they will appear at the top of this list.</p>
```
pbanaszkiewicz: You can use [alerts](http://getbootstrap.com/components/#alerts) for such hints

workshops/templates/workshops/event.html
```diff
@@ -248,8 +248,6 @@ <h4 id="newtodo">New TODO</h4>
 <script type="text/javascript">
   var gmap = null;
 
-  $('#error_event_url').popover({placement: "left", html: true});
-  $('#error_closest_instructors').popover({placement: "left", html: true});
```
pbanaszkiewicz: Why this change?

workshops/templates/workshops/event_details_table.html
```diff
@@ -8,7 +8,7 @@
...
+  <tr class="{% if not event.url %}bg-danger{% endif %}"><td>Website URL:</td><td colspan="2">{{ event.website_url|default:"—"|urlize_newtab }} {% if event.url %}<a href="{% url 'validate_event' event.slug %}" class="btn btn-primary btn-xs pull-right" id="validate_event">validate event</a>{% else %}<a class="btn btn-danger btn-xs pull-right" id="error_event_url" href="#" data-toggle="popover" title="Validation error" data-content="Cannot validate an event without URL pointing to the GitHub repository, e.g.: <code>https://github.com/swcarpentry/2015-05-24-training</code>" data-html="true">Error</a>{% endif %}</td></tr>
```
pbanaszkiewicz: Changes: removed `data-trigger="focus"` attr, added `data-html="true"` attr. Why?

workshops/templates/workshops/trainingrequest.html
```diff
@@ -1,11 +1,45 @@
...
+{% if form.errors %}
+<div class="alert alert-danger" role="alert">Fix errors below and try again.</div>
+{% endif %}
+
 {% last_modified req %}
 
+{% if req.state == 'p' %}
+  <h2>Accept Request</h2>
+  {% if form.initial.person %}
+    <p>It looks like the submitter has already an account. Click "accept & match to selected trainee account" unless the following account doesn't match submitter.</p>
```
pbanaszkiewicz: `&` → `&amp;`

workshops/test/test_landing_page.py
```diff
@@ -102,3 +97,5 @@ def test_trainee_logs_in(self):
                               follow=True)
 
         self.assertEqual(rv.resolver_match.view_name, 'trainee-dashboard')
+
+
```
pbanaszkiewicz: Too many blank spaces before EOF.

workshops/views.py
```diff
@@ -2,40 +2,65 @@
 import datetime
 import io
 import re
-from django.views.decorators.http import require_POST
+from crispy_forms.helper import FormHelper
```
pbanaszkiewicz: If you fix imports in other files, why don't you fix import in here?

workshops/views.py
```diff
@@ -121,7 +129,10 @@ def get_context_data(self, **kwargs):
         else:
             context['title'] = 'New object'
 
-        context['form_helper'] = bootstrap_helper
+        form = context['form']
+        if not hasattr(form, 'helper'):
+            form.helper = BootstrapHelper(submit_label='Add')
```
pbanaszkiewicz: This is a default helper if no other is available, correct? You should add a comment about this.

workshops/views.py
```diff
@@ -161,6 +177,10 @@ class DeleteViewContext(DeleteView):
     """
     Class-based view for deleting objects that extends default template context
     by adding proper page title.
+
+    GET requests are redirected to appropriate update views.
```
pbanaszkiewicz: Why is it so? What is it for?

workshops/views.py
```diff
@@ -3154,10 +3147,132 @@ def duplicates(request):
 
     return render(request, 'workshops/duplicates.html', context)
 
+# ------------------------------------------------------------
+# Views for trainees
+
+def _homework2state(homework):
```
pbanaszkiewicz: move to util.py? Or to models.py - depending on the function's purpose (missing docstring)

pbanaszkiewicz:
@chrismedrela this has become a very huge PR. Since you already have outlined changes you want to put into commits after you split, I suggest you to provide split PRs for future reviews - it's easier to follow your thought stream and, thanks to GH UI, it's easier to comment on specific logical changes (e.g. you change sidebar filter form and I can see all templates where the sidebar is removed), and (this may be specific only to Firefox) huge commit means LOOOOONG page loading time (like 1min for this PR's single commit's files changed tab).

Please give me a parent commit ID for your commit - I need it to not have any conflicts when trying this branch out (GH doesn't show that parent hash anymore - or I cannot find it).

workshops/views.py
chrismedrela: - In the case of Training, Discussion Session, SWC and DC Demo Session, we display whether they passed these requirements (we look for passed, not-discarded TrainingProgress record).

- In the case SWC/DC homework, we look at the state of latest not-discarded TrainingProgress. If it's "failed" or there is no such record, we ask trainee to submit homework. Otherwise, we display "Homework not evaluated yet" or "Homework accepted" depending on the state.

I'm not sure if this answers your question, because I didn't understand what exactly you asked for.

workshops/filters.py
chrismedrela: What if I add the following snippet of code at the beginning of this function?

```python
if choices not in ('', 'u', 'p', 't'):
    raise ValueError('Invalid choice value')
```

workshops/forms.py
chrismedrela: The reason for this change is to generate POST requests instead of GET ones, which is safer.

I see that the approach I proposed don't let you customize alert message. We could make the message dynamic:
```python
self.add_input(HTML('<input type="submit" name="delete" value="Delete" class="btn btn-primary btn-danger" id="submit-id-delete" style="float: right;" form="delete-form" onclick="return confirm(&quot;{{ form.delete_message }}&quot;);">'))
```
But unfortunately it's not rendered properly. I'll investigate it.

workshops/forms.py
chrismedrela: No, why? DRY is about duplicating code, not about duplicating object instances. I've made this change, because I wanted to alter helper in one place, and suddenly, the change appeared in every form because of shared state. So it's better to create a new helper for each form.

workshops/forms.py
chrismedrela: This will probably result in merge conflict, but it's easy to resolve.

workshops/forms.py
chrismedrela: That's right, `.form-actions` class was removed in v3, but `FormActions` does its job: intent buttons as expected.

workshops/templates/workshops/all_trainees.html
chrismedrela: IMO alerts are for one-off messages, that disappear after reload. This hint is displayed all the time. Are you sure you want to display it as alert?

workshops/templates/workshops/event.html
chrismedrela: Because of https://github.com/swcarpentry/amy/pull/871/files#diff-f460535dbd316f187c28498e625394abR75:
```javascript
/*
  Enable Bootstrap Popovers by default.

  Example usage:

  <span class="btn btn-primary"
        data-toggle="popover"
        data-content="Content of a popup">Hover, focus or click me!</span>
  */
  $('[data-toggle="popover"]').popover({placement: "auto", trigger: "click hover focus"});
```

This piece of JS is loaded on all pages.

workshops/templates/workshops/event_details_table.html
chrismedrela: See above.

workshops/views.py
chrismedrela: We need to somehow handle GET requests. IMO The best way is to redirect to the view with "delete" button.

workshops/views.py
chrismedrela: This is utility function that is used only in `trainee_dashboard` view. It won't be used anywhere else, so why do you want to move it?

chrismedrela:
@pbanaszkiewicz, I've implemented changes that you proposed and rebased against the newest `develop`. The parent commit ID is https://github.com/swcarpentry/amy/commit/b65107dff482db3dfa55754c46391c15b562d714

>  I suggest you to provide split PRs for future reviews

Do you mean multiple PRs or one PRs with multiple commits? The former is not possible, because a lot of changes depends on other changes.

Please ignore `import_training_progress.py` file for a moment, I'm working on it.

workshops/filters.py
pbanaszkiewicz: No. If you provide `choices` to the field, django-filter should say "this is not a valid choice" on its own when you use a wrong choice. At least it did when I tested similar scenario a few months ago.

workshops/forms.py
pbanaszkiewicz: It's not like people type `/delete` into the browser's address bar by mistake :-)

workshops/forms.py
pbanaszkiewicz: Well yes, but it's easier to have one class/instance `FormHelperFilter` that you know you can use when you're writing a new filter form class -- you're then safer, because you're using a predefined "styles" (or arguments to the `BootstrapHelper`), the very same that you're using in every other filter form.

workshops/views.py
pbanaszkiewicz: "GET to /workshops/object/delete → redirect to /workshops/object/", is that what this is doing?

Why not throw 405 Method Not Allowed…? It's safer than redirection + error message "sorry but you must have mistakenly entered that other url, well hell no, you should click this red button on the bottom instead".

workshops/views.py
pbanaszkiewicz: > This is utility function

Well guess what `util.py` is for :P we used to store some utility functions used by single views over there as well, but we got rid of them (they were too slow & complicated, used in search).

pbanaszkiewicz:
> > I suggest you to provide split PRs for future reviews
>
> Do you mean multiple PRs or one PRs with multiple commits? The former is not possible, because a lot of changes depends on other changes.

I mean turn single-commit-PRs into multiple-commit-PRs. It was whole lot more convenient for me to read through commits in #905 than reading through this never ending stream of code lines.

Anyway, this PR looks good enough for me, please merge when you think you're ready.

chrismedrela:
@gvwilson, I need URL to Etherpad where you can register for Demo Session. This URL will be displayed in trainee's dashboard.

workshops/views.py
chrismedrela: You're right, 405 is a better idea. I've changed it.

gvwilson:
Excellent work - thank you.

pbanaszkiewicz:
Thanks @chrismedrela!

Issue #873, opened 2016-07-02
Feedback on the training request form


pbanaszkiewicz:
Feedback:

- [x] @maneesha: show that mandatory checkboxes are mandatory
- [x] @maneesha: fix wording on the confirmation page ("thank you for applying")
- [x] @jduckles: drop "institutional" from "institutional affiliation"
- [ ] @jduckles: fix intro (manage expectations)
- [x] @jduckles, @tracykteal: decide what this form is for (open training / partner training / group training)
- [x] @tracykteal, @ErinBecker: "personal name" → "first name"? "family name" → "last name"? Or "Personal (given) name", "Family name / surname" (**COMMENT** @pbanaszkiewicz: we may want to change this on all our forms / templates)
- [x] @tracykteal: add field for explanations to the "previous training in teaching", "previous experience in teaching"
- [x] @ErinBecker: fix time estimate in the intro ("2h" → "8-10h")
- [x] @ErinBecker: fix wording for location: "please give city, province or state" → "please give city, and
province or state if applicable"
- [x] @ErinBecker: add alphabetical sorting to the areas of expertise (this will affect other forms as well)
- [x] @ErinBecker: "previous training in teaching" should have additional option between "a few days" and "a full degree", for example "a certification or short course"
- [x] @ErinBecker: change options for "How frequently do you program", eg. "hourly/daily/weekly/monthly/not consistently (less than monthly)/never or almost never"  (**COMMENT** @pbanaszkiewicz: I'd go with "every day / a few times a week / a few times a month / a few times a year / never or almost never")
- [x] @ErinBecker: drop "classes" from  "how often would you expect to teach classes . . . or workshops"
- [x] @ErinBecker: add Data Carpentry to the "I agree to abide by" checkbox
- [x] @ErinBecker: help text in the "I agree to complete this training within three months of the Training Course" checkbox uses wrong estimate of 2h again (should be 8-10h)
- [x] @chrismedrela: add GitHub field

pbanaszkiewicz:
@chrismedrela any progress on this issue? The milestone was supposed to be closed on Monday/Tuesday.

chrismedrela:
@pbanaszkiewicz Oops, I didn't notice the milestone deadline and I focused on other stuff. I'll try to do this today. One question: What exactly do I need to change in the intro?

pbanaszkiewicz:
Leave the intro as it is until @jduckles says if he can improve it.

chrismedrela:
Questions:

[...]

PR #849, opened 2016-06-28
Migrate Host to Organization

narayanaditya95:
Fixes #828 

Do not merge yet. I have to make changes to the templates.

PR #875, opened 2016-07-03
Added the Sponsorship model

narayanaditya95:
@gvwilson @pbanaszkiewicz 
Writing tests. Ready for review.


workshops/models.py
```diff
@@ -159,6 +159,35 @@ def self_organized_workshops_per_year_remaining(self):
 #------------------------------------------------------------
 
 
+class Sponsorship(models.Model):
+    '''Represent sponsorship from a host for an event.'''
+
+    SPONSORSHIP_SLABS = (
```
narayanaditya95: @gvwilson @pbanaszkiewicz I would like to know whether a choice field would be better suited for SWC/DC workshops or a plain simple `IntegerField`?
If we go with choices, what are the slabs in which SWC/DC accept sponsorship.

workshops/static/css/amy.css
```diff
@@ -93,10 +93,9 @@ body {
 
 /* when using nav-tab (http://getbootstrap.com/javascript/#tabs) move the
    crispy form down a few notches */
-div[role="tabpanel"] form {
+.tab-content {
```
narayanaditya95: This CSS selector is more specific.
`div[role="tabpanel"] form` also added padding above the `Delete` buttons in the sponsorship table.

workshops/models.py
pbanaszkiewicz: For me slab is [this](http://peppardbuildingsupplies.com/sites/default/files/SWATCH_BS_Grey_0.jpg) :-)

Anyway, what does `Sponsorship.slab` represent here?

workshops/models.py
```diff
@@ -159,6 +159,35 @@ def self_organized_workshops_per_year_remaining(self):
...
+        ('slab_a', 'Slab A'),
+        ('slab_b', 'Slab B'),
+    )
+
+    organization = models.ForeignKey(
+        'Host',
```
pbanaszkiewicz: Switch to `Organization` after #849 is merged.

workshops/models.py
narayanaditya95: I am not sure how I ended up with such a field name.
Anyway, a `slab` is intended to show the divisions in which SWC/DC accept sponsorships. Each slab would have predefined sponsorship amount.
PyData uses it and it's sponsors can opt for divisions such as `PyData Founding`, `PyData Gold`, `PyData Silver`... each with decreasing sponsor amount and benefits.

workshops/models.py
```diff
@@ -726,6 +755,10 @@ class Event(AssignmentMixin, models.Model):
         on_delete=models.PROTECT,
         help_text='Organization responsible for administrative work.'
     )
+    sponsors = models.ManyToManyField(
+        Host, related_name='sponsor', blank=True,
```
pbanaszkiewicz: Am I reading this correctly? `Event.sponsors → Hosts` but `Host.sponsor → Event`?

workshops/models.py
narayanaditya95: Yes.

workshops/views.py
```diff
@@ -156,6 +157,22 @@ def get_success_message(self, cleaned_data):
...
+
+    def delete(self, request, *args, **kwargs):
+        '''Workaround for https://code.djangoproject.com/ticket/21926'''
+        messages.success(
+            self.request,
+            self.success_message.format(self.get_object())
+        )
+        return super().delete(request, *args, **kwargs)
+
+
```
pbanaszkiewicz: You sure you need this? We used to redirect right after deletion.

workshops/models.py
narayanaditya95: Yes, just as `host.administrator` gives the list of events that the host administered, `host.sponsor` gives the list of events sponsored.
On a second thought, I think `Event.sponsor_set` sounds better.

workshops/views.py
narayanaditya95: Yes, we currently also redirect right after deletion. There is no intermediate confirmation page (except for the confirm javascript popup).

pbanaszkiewicz:
@narayanaditya95 looks OK.

I wonder what we should use for the sponsorship levels aka. slabs. You can imagine we don't have an established values for them in SWC/DC. Maybe @gvwilson can provide some more feedback here?

workshops/models.py
pbanaszkiewicz: That was just a note for future :-)

workshops/models.py
pbanaszkiewicz: Ugh, both `organization.administrator` and `organization.sponsor` sound bad, and I believe I'm responsible for the former :-)

I'd use `organization.administered_events` or `organization.administered_set` and `organization.sponsored_events` or `organization.sponsored_set`.

workshops/views.py
pbanaszkiewicz: So why do we need this view?

narayanaditya95:
If SWC/DC does not use a predefined levels, we can switch to an `IntegerField`.
PyData package would then monkeypatch the `slab` field to add `choices` to it. Each predefined level at PyData has a sponsorship amount against it.

workshops/views.py
narayanaditya95: Because it makes the task of deleting and adding a success message easier and non-repetitive.

workshops/models.py
```diff
@@ -159,6 +159,35 @@ def self_organized_workshops_per_year_remaining(self):
...
+        help_text='Organization sponsoring the event'
+    )
+    event = models.ForeignKey(
+        'Event',
+        on_delete=models.CASCADE,
+    )
+    level = models.PositiveIntegerField(
+        choices=SPONSORSHIP_LEVELS,
+        default=1000,
+    )
```
narayanaditya95: @gvwilson Does SWC/DC use predefined levels of sponsorship, or should we simply remove the choices?

gvwilson: Our current levels are defined on 
http://software-carpentry.org/scf/join/.  I think a Partnership model 
with names is overkill - just a text field naming the level should be 
enough, and we can add special notes to the Notes field on the 
organization where needed.

narayanaditya95: The levels defined at http://software-carpentry.org/scf/join/ list the kinds of partnership between an organization and SWC/DC. It does not list the kind of sponsorship an organization can provide for an SWC/DC workshop.
What practice does SWC/DC follow while accepting sponsorship for a particular workshop?

gvwilson: My apologies, I misunderstood the question - organizational sponsorship 
for a specific workshop varies so widely that I don't think it's worth 
trying to capture as anything except a free-text field (in Notes).

narayanaditya95:
@gvwilson I added a nullable `DecimalField` that would allow SWC/DC admins to record monetary sponsorship. The `pydata` app would on the other hand, extend the `DecimalField` to add predefined levels.

@pbanaszkiewicz I have added tests and addressed all the comments raised on the PR.
It is up for review and merge.

narayanaditya95:
bump.

workshops/views.py
pbanaszkiewicz: That's a nice change, can you do that for other deletion views if feasible (can be in a different PR)?

workshops/templates/workshops/event_edit_form.html
```diff
@@ -4,32 +4,68 @@
...
-  </table>
-  {% else %}
-  <p>No tasks.</p>
-  {% endif %}
-
+<div>
+  <ul class="nav nav-tabs" role="tablist">
+    <li role="presentation" class="active"><a href="#event" aria-controls="event" role="tab" data-toggle="tab">Event</a></li>
+    <li role="presentation"><a href="#tasks" aria-controls="tasks" role="tab" data-toggle="tab">Tasks</a></li>
+    <li role="presentation"><a href="#sponsors" aria-controls="sponsors" role="tab" data-toggle="tab">Sponsors</a></li>
```
pbanaszkiewicz: See how `person-edit` page behaves after you add a new task to the person -- it switches you to a tab you just were at.

You'll receive 1000 bonus virtual points if you make the tabs switch depending on the anchor in the browser's address bar. :-)

workshops/views.py
```diff
@@ -1194,7 +1214,20 @@ def event_edit(request, event_ident):
...
                 messages.error(request, 'Fix errors below.')
+        if 'sponsor-event' in request.POST:
+            sponsor_form = SponsorshipForm(request.POST, prefix='sponsor')
 
+            if sponsor_form.is_valid():
+                sponsor = sponsor_form.save()
+
+                messages.success(
+                    request,
+                    '{} was added a new sponsor.'.format(sponsor.organization)
```
pbanaszkiewicz: Should there be "as" instead of "a"?

workshops/templates/workshops/event_edit_form.html
pbanaszkiewicz: PS the trick is to use additional tags with the `message.success`. See `person-edit` for details.

workshops/views.py
```diff
@@ -1194,7 +1214,20 @@ def event_edit(request, event_ident):
...
+                )
+                # to reset the form values
+                return redirect(request.path)
```
pbanaszkiewicz: …or `redirect(event)`, right? The latter is less popular but IMO a little bit clearer where you end up after the redirect.

pbanaszkiewicz:
@narayanaditya95 minor issues and one/two challenges, and apart from them +1 (again, not going to merge until after #873 is fixed).

narayanaditya95:
@pbanaszkiewicz Where do I claim my 1000 bonus virtual points?

pbanaszkiewicz:
@narayanaditya95 good job, I'll merge soon into v1.7.

pbanaszkiewicz:
@narayanaditya95 can you please rebase? Thanks!

narayanaditya95:
Done

PR #880, opened 2016-07-06
Allow filling the `assigned_to` field on event_add view

### LINKED ISSUE #847: Assign admin when creating new event
You should be able to assign admin from this page:
https://amy.software-carpentry.org/workshops/events/add/

Right now you can't assign admin until you save and edit.

narayanaditya95:
Fixes #847.

### DIFF
--- a/workshops/test/test_event.py
+++ b/workshops/test/test_event.py
@@ -328,6 +328,24 @@ class TestEventViews(TestBase):
             Event.objects.create(host=self.test_host,
                                  slug='testing-unique-slug')
 
+    def test_assign_to_field_populated(self):
+        """Ensure that we can assign an admin to an event
+        from the `event_add` view."""
+        data = {
+            'slug': '2016-07-09-test',
+            'host': self.test_host.id,
+            'tags': [self.test_tag.id],
+            'assigned_to': self.admin.pk,
+            'invoice_status': 'unknown',
+        }
+        response = self.client.post(reverse('event_add'), data, follow=True)
+        event = Event.objects.get(slug='2016-07-09-test')
+        self.assertRedirects(
+            response,
+            reverse('event_details', kwargs={'event_ident': event.slug}),
+        )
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
That's even easier than I thought!

There should be a test for adding a new event, so please add check for the `assigned_to` in there. If there isn't, can you please create it? Thanks!

narayanaditya95:
@pbanaszkiewicz Done. Also verified that the test fails without the fix.

workshops/test/test_event.py
```diff
@@ -328,6 +328,24 @@ def test_unique_slug(self):
...
+            'assigned_to': self.admin.pk,
+            'invoice_status': 'unknown',
+        }
+        response = self.client.post(reverse('event_add'), data, follow=True)
+        event = Event.objects.get(slug='2016-07-09-test')
+        self.assertRedirects(
+            response,
+            reverse('event_details', kwargs={'event_ident': event.slug}),
+        )
+        self.assertEqual(response.context['event'].assigned_to, self.admin)
```
pbanaszkiewicz: Please change that to `event.refresh_from_db()` and then assert on that `event.assigned_to`.

narayanaditya95: Why do we have to `event.refresh_from_db()` when we are fetching the event object after a POST?
`self.assertEqual(event.assigned_to, self.admin)` should be enough to test the assertion.

pbanaszkiewicz: @narayanaditya95 you're right, somehow I thought that line 343 was doing a request.

Yes, no need to `refresh_from_db()` in this case, and line 347 should be `self.assertEqual(event.assigned_to, self.admin)`.

narayanaditya95: Why do we have to `event.refresh_from_db()` when we are fetching the event object after a POST?
`self.assertEqual(event.assigned_to, self.admin)` should be enough to test the assertion.

PR #883, opened 2016-07-07
Check for missing migrations in Travis CI

chrismedrela:
Explanation how it works: if you run `makemigrations` with `-e` flag, `1` is returned when there are no changes and `0` when there are missing migrations.

~~Unfortunately, I couldn't test if this PR works.~~ This PR works. This PR have no missing migrations and passed. #884 has missing migrations and failed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ install:
   - pip install -r requirements.txt
   - pip install coveralls
 script:
-  coverage run --source=amy,api,workshops manage.py test
+  - python manage.py makemigrations --dry-run -e; export STATUS_CODE=$?; if [[ "$STATUS_CODE" == "0" ]]; then exit 1; fi
+  - coverage run --source=amy,api,workshops manage.py test
 after_success:
   coveralls


narayanaditya95:
I like the change. +1.

Here are few suggestions from my end:
- Bundle a model change along with the PR without the corresponding migrations. See if it fails.
- How about we make the check for missing migrations as a different check. So that when `check migration` test fails and the tests pass, we see something like "1/2 tests passed". I wish to add another check solely concerned with running the tests within the `pydata` app.

chrismedrela:
@narayanaditya95

> Bundle a model change along with the PR without the corresponding migrations. See if it fails.

Tested in #884.

> How about we make the check for missing migrations as a different check. So that when check migration test fails and the tests pass, we see something like "1/2 tests passed". I wish to add another check solely concerned with running the tests within the pydata app.

I don't know how to configure Travis to do this, feel free to guide me or send PR.

narayanaditya95:
@chrismedrela 
The `matrix` field within the `.travis.yml` allows us to export env variables. Each entry within the `matrix` field corresponding to another check.
The `script` field can then use the env variable to run different test suites.

See a working example at https://github.com/divio/django-cms/blob/release/3.3.x/.travis.yml.

chrismedrela:
> P.S. Although it configures a list of jobs in Travis, Travis still appears as a single check in GitHub.

So why do you want to use `matrix` field?

narayanaditya95:
> So why do you want to use matrix field?

Well, after I realized we anyway have to open the Travis UI to check for failing job, I'm -1 for the suggestion.

.travis.yml
```diff
@@ -6,6 +6,7 @@ install:
   - pip install -r requirements.txt
   - pip install coveralls
 script:
-  coverage run --source=amy,api,workshops manage.py test
+  - coverage run --source=amy,api,workshops manage.py test
+  - python manage.py makemigrations --dry-run -e; export STATUS_CODE=$?; if [[ "$STATUS_CODE" == "0" ]]; then exit 1; fi
```
pbanaszkiewicz: Can you swap the lines? This way we know earlier that we're missing the migrations.

pbanaszkiewicz:
Gentlemen, as you can clearly see by the #885 and discussion in #883 (here), we have still plenty of room for improvement over our tests / upon-PR services.

If you have some ideas I'm open for suggestions. For example we could somehow show the coverage for AMY (the tool running tests already supports `coverage` package), we may add "build: passing", "coverage: XY%" icons to the readme, etc. Feel free to propose any ideas that you have.

@chrismedrela I'm going to merge in v1.7 after you swap the lines.

chrismedrela:
@pbanaszkiewicz, done.

.travis.yml
```diff
@@ -6,6 +6,7 @@ install:
   - pip install -r requirements.txt
   - pip install coveralls
 script:
-  coverage run --source=amy,api,workshops manage.py test
+  - python manage.py makemigrations --dry-run -e; export STATUS_CODE=$?; if [[ "$STATUS_CODE" == "0" ]]; then exit 1; fi
```
pbanaszkiewicz: Question: should we limit `makemigrations` check to only run on our applications, ie. `api`, `pydata` and `workshops`? Because otherwise it runs on all installed packages.

chrismedrela: Well, if there are some missing migrations in other applications, that indicates that something is wrong. Am I right that it should never happen that we have a 3rd party app (or django app) with missing migrations?

narayanaditya95: But there isn't much we can do about when there exist migrations in third-party apps.

chrismedrela: We can narrow the check to our apps when situation like that happen. If we narrowed it now, we wouldn't know when this situation occurred.

pbanaszkiewicz: Leaving as it is, since additional check never hurts (until it does).

PR #885, opened 2016-07-07
Speedup Travis builds by caching pip directory

chrismedrela:
Reduction of build time from 188 seconds to 96 seconds (**2x speedup**).

This works. Compare [the first build](https://travis-ci.org/swcarpentry/amy/builds/143161913) of this PR with [the second one](https://travis-ci.org/swcarpentry/amy/builds/143163940).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: false
 language: python
+cache: pip 
 python:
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
Nice! Will merge into v1.7.

PR #886, opened 2016-07-07
Grant steering committee members access to admin interface

### LINKED ISSUE #876: Steering Committee members aren't allowed into AMY
All access apart from AutoUpdateProfile is allowed only for admins, and people from Steering Committee have "steering committee" group which is not considered admin.

The workaround for now is that I'll add SC members to the "administrators" group.

chrismedrela:
Fix #876

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -422,7 +422,8 @@ def is_admin(user):
         return False
     else:
         return (user.is_superuser or
-                user.groups.filter(name='administrators').exists())
+                user.groups.filter(Q(name='administrators') |
+                                   Q(name='steering committee')).exists())
 
 
 class ProfileUpdateRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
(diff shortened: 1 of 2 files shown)


workshops/models.py
```diff
@@ -422,7 +422,8 @@ def is_admin(user):
         return False
     else:
         return (user.is_superuser or
-                user.groups.filter(name='administrators').exists())
+                user.groups.filter(name='administrators').exists() or
+                user.groups.filter(name='steering committee').exists())
```
pbanaszkiewicz: Make this a one query.

Remember that `is_admin` is run almost on every page, so if you increase number of queries here, AMY's performance may decrease.

pbanaszkiewicz:
Fix needed, then I'll merge.

chrismedrela:
@pbanaszkiewicz , done.

pbanaszkiewicz:
@chrismedrela thanks!

PR #890, opened 2016-07-09
Reveal IntegrityError raised in sync_usersocialauth view

chrismedrela:
The redirect in `finally` clause was executed even when an non-`GithubException` (like `IntegrityError`) was raised. This hid the error. Now it results in loud error.

### DIFF
--- a/workshops/test/test_person.py
+++ b/workshops/test/test_person.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 
 import datetime
-from unittest.mock import patch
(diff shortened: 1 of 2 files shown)


chrismedrela:
This `IntegrityError` is another bug -- I covered it here https://github.com/swcarpentry/amy/issues/891.

pbanaszkiewicz:
Any chance for a regression test?

chrismedrela:
@pbanaszkiewicz, done.

PR #892, opened 2016-07-09
Change widgets in person permissions view

chrismedrela:
Use `AutoCompleteSelectMultipleField` and `CheckboxSelectMultiple` instead of `SelectMultiple`. 

Reason: I have no idea how can I unselect all items in `SelectMultiple` widget.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -433,6 +433,14 @@ class PersonCreateForm(PersonForm):
 
 
 class PersonPermissionsForm(forms.ModelForm):
+    user_permissions = selectable.AutoCompleteSelectMultipleField(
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
You have to use CTRL+click on the entries that are already selected. This is the default behavior of your operating system.

I'm -1 on this idea: `SelectMultiple` allows me to see all permissions and allows me to quickly select/unselect a whole bunch at once (CTRL + mouse button push + move up/down).

I'm undecided on `CheckboxSelectMultiple` for groups, because since we don't have many of them this could be an improvement…

chrismedrela:
CTRL+Click works for me, so I'm closing it.

PR #893, opened 2016-07-09
Fix #891 -- UserSocialAuth cache got out of sync

### LINKED ISSUE #891: Synchronizing UserSocialAuth doesn't work with duplicated GitHub usernames
When we have two people with the same GitHub username and UserSocialAuth is already synchronized for the first person, we cannot synchronize for the another one. This is because one GitHub account can be associated with only one AMY account. This results in `IntegrityError` raised in `sync_usersocialauth` view.

What should we do? What about displaying an error message (I mean `messages.error`) with information who has the same GitHub account?

chrismedrela:
When `Person.github` is changed, all `UserSocialAuth` records associated with that person are now deleted.

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -431,6 +431,14 @@ class Person(AbstractBaseUser, PermissionsMixin):
         self.email = self.email.lower() if self.email else None
 
     def save(self, *args, **kwargs):
+        # If GitHub username has changed, clear UserSocialAuth table for this
+        # person.
+        if self.pk is not None:
+            orig = Person.objects.get(pk=self.pk)
+            github_username_has_changed = orig.github != self.github
+            if github_username_has_changed:
+                UserSocialAuth.objects.filter(user=self).delete()
+
         # save empty string as NULL to the database - otherwise there are
         # issues with UNIQUE constraint failing
         self.personal = self.personal.strip()
--- a/workshops/test/test_person.py
+++ b/workshops/test/test_person.py
@@ -1,7 +1,8 @@
 # coding: utf-8
 
 import datetime
-from unittest.mock import patch, MagicMock
+from social.apps.django_app.default.models import UserSocialAuth
+from unittest.mock import patch
 from urllib.parse import urlencode
 
 from django.contrib.auth import authenticate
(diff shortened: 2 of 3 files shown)


workshops/models.py
```diff
@@ -404,6 +404,14 @@ def clean(self):
         self.email = self.email.lower() if self.email else None
 
     def save(self, *args, **kwargs):
+        # If GitHub username has changed, clear UserSocialAuth table for this
+        # person.
+        if self.pk is not None:
+            orig = Person.objects.get(pk=self.pk)
+            github_username_has_changed = orig.github != self.github
+            if github_username_has_changed:
```
gvwilson: Why not `if orig.github != self.github` ? Why the intermediate variable?

workshops/test/test_person.py
```diff
@@ -119,7 +122,7 @@ def _check_person(self, doc, person):
                     .format(key, value, type(value), node.text)
 
     def _get_field(self, doc, key):
-        '''Get field from person display.'''
+        """ Get field from person display. """
```
gvwilson: Changes from single to double quotes should land in a separate PR.

workshops/test/test_person.py
```diff
@@ -183,7 +187,8 @@ def test_person_award_badge(self):
 
         rv = self.client.post(url, data=values)
         assert rv.status_code == 302, \
-            'After awarding a badge we should be redirected to the same page, got {} instead'.format(rv.status_code)
+            'After awarding a badge we should be redirected to the same ' \
+            'page, got {} instead'.format(rv.status_code)
```
gvwilson: Docstring changes belong in a separate PR.

workshops/test/test_person.py
```diff
@@ -734,9 +739,9 @@ def test_merging_m2m_attributes(self):
             # instead testing awards, let's simply test badges
             'badges': set(Badge.objects.filter(name='swc-instructor')),
             # we're saving/combining qualifications, but it affects lessons
-            'lessons': set([self.sql, self.git]),
-            'domains': set([KnowledgeDomain.objects.first(),
-                            KnowledgeDomain.objects.last()]),
+            'lessons': {self.sql, self.git},
+            'domains': {KnowledgeDomain.objects.first(),
+                        KnowledgeDomain.objects.last()},
```
gvwilson: Are these changes directly related to changing Git username?

workshops/test/test_person.py
chrismedrela: There are two commits, one fixing bug and one fixing PEP8 issues. Is that enough or do you want me to split this into two PRs?

workshops/test/test_person.py
chrismedrela: No, it's only about using set literal `{a, b, c}` instead of `set([a, b, c])`.

workshops/models.py
chrismedrela: The name of this variable explains what we are actually testing. This could be rewritten as:
```python
if orig.github != self.github:  # if github username has changed
```

workshops/test/test_person.py
gvwilson: I'm not fussy if @pbanaszkiewcz isn't.

workshops/test/test_person.py
gvwilson: :+1:

workshops/models.py
gvwilson: I think it's clear enough from context, but it's your call.

workshops/test/test_person.py
pbanaszkiewicz: The rule of thumb is to fix not-PEP8-compliant code when you edit it, and not just fix it because you want to.

And I'm very sorry because from time to time I do not follow this rule :-)

So in this case I'd just leave the additional commit as it is.

Question though: I've seen you use `""" spaces right next to the triple quotes. """`. Is it something dictated by PEP8 / PEP257?

workshops/test/test_person.py
pbanaszkiewicz: Nitpicking: you can, and we do that in some places, surround the `text.format()` part in parentheses and then you don't need backslashes.

pbanaszkiewicz:
@chrismedrela alright, so there are a couple issues here.

Firstly, I want to remind you, that in the long-run I want to have a separate page where people can manage their social-auth integration with different providers (so that we drop the connection between `person.github` to anything login-related).

When we do have that page, we can safely modify `person.github` and the person themselves will be able to even add more GitHub accounts to log in.

Secondly, this PR doesn't inform anyone that changing `person.github` is going to drop the social-auth logging. And there are a few places where we should be displaying the warning (ProfileUpdateRequest, person edit, AutoSelfUpdateProfile ?, maybe more).

Thirdly, the social-auth management page is inevitable. Maybe it's less work to Do This Right™ now instead of patching current approach and then making it right?

chrismedrela:
@pbanaszkiewicz 

1. I don't understand what exactly you want to achieve. How do you want to break connection between `Person.github` and login stuff and, at the same time, have this field in sync with `UserSocialAuth`? If don't want to keep it in sync, what's the point of `Person.github`?

2. What other providers do you consider?

3. I'm not sure if sophisticated social-auth management is worth effort. This is going to take a lot of effort (just look how much time did it take to integrate with GitHub). And what are the benefits? Support for other providers and ability to have multiple GitHub accounts associated with one AMY account. Anything more? @gvwilson, any opinion?

pbanaszkiewicz:
@chrismedrela:

Before GitHub login, `Person.github` had only one role: to inform admins about person's GitHub username. Currently it serves a double role: as a informative field and as a semi-automatic social-auth login mechanism.

So to answer your concerns:

1. I don't want `Person.github` to be in sync with user social auth. I want to have a separate page that can be used to manage social links with user's account (like. CRUD from user social auth table).

2. Google, ORCID.

3. Yes, I want these benefits. Additionally: no more semi-automatic auto-connecting/disconnecting `Person.github`. I'm pretty sure this is not going to take as much effort as the base step to login from GH did.

chrismedrela:
@pbanaszkiewicz, so do you want to remove `Person.github` or leave it and let it be out of sync with what we have in `UserSocialAuth`?

And how the workflow would look like? How do we let people login for the first time? The first GitHub/Google/ORCID account needs to be manually associated by admins, so we cannot get rid of semi-automatic auto-connecting.

pbanaszkiewicz:
> so do you want to remove `Person.github` or leave it and let it be out of sync with what we have in `UserSocialAuth`?

Just out of sync.

> How do we let people login for the first time?

They already do that because of the entries in `UserSocialAuth`.

> The first GitHub/Google/ORCID account needs to be manually associated by admins

So the workflow is no different than what we have now. I only want a management page for all your social links/logins. This bug, for example, wouldn't exist if we had such management page.

chrismedrela:
OK, now it's much clearer for me what you want to do. The only challenge that I still see is that we still need semi-automatic auto-connecting, so that we can set the first social login for each person. And that requires some connection between `Person.github` and `UserSocialAuth`. So even if we implement management page for social login, this bug still exists.

chrismedrela:
@pbanaszkiewicz, ping.

pbanaszkiewicz:
@chrismedrela So we're scheduling the social link management page for v1.8? Also, please rebase.

chrismedrela:
@pbanaszkiewicz, done. Yes, I'd like to focus on instructor training workflow first, and implement social link management page later on.

pbanaszkiewicz:
@chrismedrela ready for merge?

chrismedrela:
Yes.

PR #895, opened 2016-07-12
Fix intro in training request form

chrismedrela:
Ref #873.

Please note that I left the three conditions that the applicant has to meet (Code of Conduct, completion of homework, discussion and demo session and teaching one SWC/DC workshops within 12 months).

### DIFF
--- a/workshops/templates/forms/trainingrequest.html
+++ b/workshops/templates/forms/trainingrequest.html
@@ -10,11 +10,30 @@
 {% endblock %}
 
 {% block content %}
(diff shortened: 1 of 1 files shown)


gvwilson:
Looks good - @jduckles and @tracykteal, should the opener link to Data Carpentry pages as well?

jduckles:
We both get notified by emails to partnerships@software-carpentry.org and the lead-in text mentions that the partnerships are joint. The pages are also identical, so I think two links in this case just causes more confusion than it solves. What do you think @tracykteal?

PR #900, opened 2016-07-18
Speeding up tests with --keepdb and --parallel

chrismedrela:
This PR introduces two improvements:

1. Use in-file database instead of in-memory one while running tests with `--keepdb`. This change lets us preserve database between test launches and, therefore, avoid the overhead of applying migrations during each test launch. Note that a new fresh in-memory database is still used by default, that is when you don't set `--keepdb` flag (i.e. `python3 manage.py test`).

2. `TestCase.subTest` cannot be used when tests are launched in parallel, that is with `--parallel` flag. To bypass this limitation, we use a dummy `subTest` mock when when tests are run in parallel.

The first improvement saves 30 seconds on my machine once the database is created. All test suite takes now 50 seconds instead of 80 seconds.

The second improvement in combination with the first one let you speed up tests twice (25 seconds instead of 50 seconds on my machine, my proc has 4 cores).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,11 @@ htmlerror
 *.pyc
 db.sqlite3
 db.sqlite3.bak
+# in-file test database used when --keepdb flag is set
 test_db.sqlite3
 test_db.sqlite3.bak
+# clones of in-file test database created while launching tests in parallel
+test_db*
 
 # coverage reports
 .coverage
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ htmlerror
 *.pyc
 db.sqlite3
 db.sqlite3.bak
+test_db.sqlite3
+test_db.sqlite3.bak
 
 # coverage reports
 .coverage
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -177,8 +177,15 @@ DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.sqlite3',
         'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
+        'TEST': {},
     }
 }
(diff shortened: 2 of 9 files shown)


chrismedrela:
Unfortunately, squashing migrations doesn't make any difference. It still takes 30 seconds to apply migrations on my machine, even when first 87 migrations are squashed. So that was waste of time. :( See this branch: https://github.com/chrismedrela/amy/tree/squash-migrations

BTW, pro hint: if you get `contenttype.name cannot be NULL` IntegrityError when applying squashed migration, you need to add `('contenttypes', '0002_remove_content_type_name'),` to `Migration.dependencies`.

gvwilson:
Thanks - was the speed of tests a real obstacle to development?

chrismedrela:
@gvwilson, When I wanted to run one test, I had to wait half a minute for results. Now it runs in few seconds. This really slowed down writing tests, especially that I need to run each test a few times before I get it right.

pbanaszkiewicz:
@chrismedrela 

> `TestCase.subTest` cannot be used when tests are launched in parallel, that is with `--parallel` flag. To bypass this limitation, we use a dummy `subTest` mock when when tests are run in parallel.

Does that mean these sub tests are **not** run at all when using `--parallel`?

chrismedrela:
@pbanaszkiewicz, no, they are run, but if any of the subtest fails, other subtests are not checked and we go to the next test.

pbanaszkiewicz:
@chrismedrela please update this PR so that I can merge it

pbanaszkiewicz:
@chrismedrela ping

chrismedrela:
@pbanaszkiewicz pong

PR #904, opened 2016-07-18
Use create_superuser instead of createsuperuser in README.md

chrismedrela:
This is our custom command and it's not documented and hard to discover. Why not using it in Getting Started section?

BTW, do we want a new label for documentation issues?

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,10 +57,10 @@ before starting work on new features.
     $ make dev_database
     ~~~
     
-5.  Create an administrator account:
+5.  Create an administrator account (username: `admin`, password: `admin`):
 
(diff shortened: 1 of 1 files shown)


chrismedrela:
`create_superuser` is ours and it's launched by `make`. The difference is that you don't have to enter personal and family name, email and password with `create_superuser`.

pbanaszkiewicz:
I gave this some thought and decided to dismiss this PR. `create_superuser` (ours) was intended for developers who are too lazy to type in the details; `createsuperuser` (Django's) is what should users be doing when setting up AMY on their servers.

PR #908, opened 2016-07-19
Fix #907 -- leakage of email addresses in "bulk email" buttons

### LINKED ISSUE #907: Possible leakage of email addresses in "mail everyone"
In a few places:
- https://github.com/swcarpentry/amy/blob/d9e9c32fea7cf978cf9c264bff50daeabd9e1823/workshops/templates/workshops/event.html#L129
- https://github.com/swcarpentry/amy/blob/d9e9c32fea7cf978cf9c264bff50daeabd9e1823/workshops/templates/workshops/instructors_by_date.html#L25
- https://github.com/swcarpentry/amy/blob/d9e9c32fea7cf978cf9c264bff50daeabd9e1823/workshops/templates/workshops/workshop_staff.html#L46

we use `mailto:bunch_of_emails`. These email addresses may end up in "To" field instead of "BCC" (correct me if I'm wrong). In that case, recipients can see other recipients' email adresses. However, I'm not sure if this is unwanted behaviour. @gvwilson, any opinion?

Fix: use `mailto:?bcc=` instead of `mailto:`.

chrismedrela:
In a few places, there were links with href=`mailto:bunch,of,emails` instead of `mailto:?bcc=bunch,of,emails`. Those email addresses ended up in "To" field instead of "BCC" (blind copy), resulting in recipients seeing other recipients' email addresses.

### DIFF
--- a/workshops/templates/workshops/event.html
+++ b/workshops/templates/workshops/event.html
@@ -126,7 +126,7 @@
 
 {% if tasks %}
 <h3>Tasks</h3>
-{% if all_emails %}<p><a href="mailto:{{ all_emails|join:',' }}" class="btn btn-primary">Mail everyone</a></p>{% endif %}
+{% if all_emails %}<p><a href="mailto:?bcc={{ all_emails|join:',' }}" class="btn btn-primary">Mail everyone</a></p>{% endif %}
 <table class="table table-striped">
   <tr>
     <th width="25">DC?</th>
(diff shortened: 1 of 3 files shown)


workshops/templates/workshops/event.html
```diff
@@ -126,7 +126,7 @@ <h4 class="modal-title" id="preview_modal_label">Preview on Software-Carpentry w
 
 {% if tasks %}
 <h3>Tasks</h3>
-{% if all_emails %}<p><a href="mailto:{{ all_emails|join:',' }}" class="btn btn-primary">Mail everyone</a></p>{% endif %}
+{% if all_emails %}<p><a href="mailto:?bcc={{ all_emails|join:',' }}" class="btn btn-primary">Mail everyone</a></p>{% endif %}
```
pbanaszkiewicz: One question though: if this is using `?param=value` syntax, shouldn't the content of `value` be urlencoded?

chrismedrela: Probably yes, but after googling a bit it seems that there is no clear consensus which characters should be encoded and, what is more important, email clients often don't interpret encoded characters correctly, so I'd leave it as it is.

PR #910, opened 2016-07-19
Add a cancel button to all forms.

### LINKED ISSUE #881: "Cancel" button
Any place there is a "Submit" button I'd also like to see a "cancel" button so that any changes accidentally made while in editing mode are not saved.

narayanaditya95:
Fixes #881

@pbanaszkiewicz The buttons are stacked together. Are we interested in pushing them to the extreme left and the extreme right? Please review.


pbanaszkiewicz:
@narayanaditya95 can you explain what this does? Does this simply reload the page? Do the form inputs "clear" to original values? Is this what @maneesha wanted?

narayanaditya95:
It force reloads the page. All form fields are researched from the server. 
Yes, the PR addresses the issue @maneesha raised

pbanaszkiewicz:
@narayanaditya95 this adds a button to the search form, and to the external forms. I don't think they should have a "Cancel" button.

narayanaditya95:
@pbanaszkiewicz Looked into an existing HTML5 feature that allows us to reset form fields to their original values.

@chrismedrela Can you describe the utility of the `add_delete_button` argument. I cannot get the rendered delete button to perform its intended functionality.

pbanaszkiewicz:
So we're going with the `<input type=reset>` instead of hard refresh?

pbanaszkiewicz:
The reset button won't clear autocomplete M2M fields, like `Person.languages`. :/

narayanaditya95:
@pbanaszkiewicz 
That was a nice catch. It did work on ForeignKeys.
Force reload looks like the better option now, lest we can fix this irregularity.

pbanaszkiewicz:
Postponing to v1.7.1 because I want to release v1.7 today.

PR #911, opened 2016-07-19
Deprecate `Event.get_by_ident()` and `Event.get_ident`.

### LINKED ISSUE #874: Deprecate `Event.get_by_ident`
With non-nullable and required `slug` field for each event, `Event.get_by_ident` can be deprecated in favour of `Event.objects.get(slug=...)`.

narayanaditya95:
Rename URL params from `event_ident` to `slug`.
Fixes #874


pbanaszkiewicz:
Cleaning! I like this.

PR #915, opened 2016-07-22
Remove help text of autocomplete fields

chrismedrela:


pbanaszkiewicz:
What's rationale behind this removal?

chrismedrela:
After walkthroughs with AMY users, I think that autocomplete fields are intuitive and don't require any explanation. Removing these help texts simplifies UI. The more text on webpages, the more likely it is that an important message is left unread (less is more).

gvwilson:
:+1: - if this is what Chris got from user testing, let's go with it.

pbanaszkiewicz:
This one likely needs to be updated (I think I found 2 more places where `help_text=AUTOCOMPLETE_HELP_TEXT` is used).

pbanaszkiewicz:
@chrismedrela can you please update this PR?

chrismedrela:
@pbanaszkiewicz , done.

PR #922, opened 2016-07-26
Group application workflow

chrismedrela:
Not intended to be merged.

### DIFF
--- a/docs/group-applications-workflow.md
+++ b/docs/group-applications-workflow.md
@@ -1,0 +1,37 @@
+1. Admin creates an event with slug. Admin remembers the slug.
+
+1. Admin bulk uploads people sent by the partner and assigns them to the training. This requires to use CSV file with five columns:
+
+    - personal name,
+    - family name,
+    - email address,
+    - the training slug (all rows with the same value) and
+    - role ('learner' in all rows).
+
+1. Admin goes to "trainees" page and filter by training. That way, Admin can list exactly the same set of people that he uploaded. In the same view, Admin can filter by whether an training request was sent or not.
+
+1. Admin filters only trainees who haven't sent training request yet (and are assigned to the training) and bulk email them. This email contains link to training request form, with group name. Training id is not included in the link, because trainees are already matched with the training.
+
+1. Admin filters only trainees who already have sent training request (and are assigned to the training) and bulk email them. In this email, we ask them to confirm if they're still interested in the training.
+
+1. One of trainees submits training request form.
+
+1. Admin goes to "training requests" page and, then, to details of the requests that was submitted. Admin manually accepts the request and matches it with the right Person record, which is proposed by AMY since we already have that person in database.
+
+1. One of trainees replies to email and confirms that she's still interested in the training.
+
+1. Issue: How do we indicate that fact in AMY? What should Admin do?
+
+1. One of trainees replies to email and resigns from the training.
+
+1. Admin unassignes her from the training and discard her training request.
+
+This workflow requires the following new features in "trainees" view:
+
+- Filter trainees by presence of training request.
+
+- Bulk email selected trainees with two presets of email content.
+
+- Column with list of training requests.
+
+And "filter by training" in "training requests" page.
--- a/docs/group-applications-workflow.md
+++ b/docs/group-applications-workflow.md
@@ -1,3 +1,7 @@
+# New group application workflow
+
+When a partner asks us for a training for a group of people: 
+
 1. Admin creates an event with slug. Admin remembers the slug.
 
 1. Admin bulk uploads people sent by the partner and assigns them to the training. This requires to use CSV file with five columns:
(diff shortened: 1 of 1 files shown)


docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
+1. Admin creates an event with slug. Admin remembers the slug.
```
gvwilson: OK

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. Admin bulk uploads people sent by the partner and assigns them to the training. This requires to use CSV file with five columns:
+
+    - personal name,
+    - family name,
+    - email address,
+    - the training slug (all rows with the same value) and
+    - role ('learner' in all rows).
```
gvwilson: OK.

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. Admin goes to "trainees" page and filter by training. That way, Admin can list exactly the same set of people that he uploaded. In the same view, Admin can filter by whether an training request was sent or not.
```
gvwilson: Is "filter by whether we have a training request in the database" new functionality? Or does it already exist?

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. Admin filters only trainees who haven't sent training request yet (and are assigned to the training) and bulk email them. This email contains link to training request form, with group name. Training id is not included in the link, because trainees are already matched with the training.
```
gvwilson: Does this mean that the "view training requests" page *always* has a "generate list of email addresses" feature?  I.e., if I filter by other criteria, will I always get a "mailto:" link with a bunch of addresses? (That seems useful.)

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. Admin filters only trainees who already have sent training request (and are assigned to the training) and bulk email them. In this email, we ask them to confirm if they're still interested in the training.
```
gvwilson: OK.

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. One of trainees submits training request form.
```
gvwilson: This is the start of a separate workflow.

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. Admin goes to "training requests" page and, then, to details of the requests that was submitted. Admin manually accepts the request and matches it with the right Person record, which is proposed by AMY since we already have that person in database.
```
gvwilson: Can the admin filter training requests by group name to more easily find those submitted in response to a recently-sent email?

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. One of trainees replies to email and confirms that she's still interested in the training.
```
gvwilson: This is the start of a separate workflow (the third) -- presumably this is the case where we already had a training request from someone, and their name was then given to us by a partner organization?

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. Issue: How do we indicate that fact in AMY? What should Admin do?
```
gvwilson: That's a good question: we somehow want to mark them "confirmed", but we don't have anything like that. I suggest that we split this workflow:

1. Person replies and says "yes I'm still interested" => we do nothing.
2. Person replies and says "no I'm not" => we take them out of this event.  If they say they're not interested in training at all any more, we discard their training application as well.  (We don't automatically discard their training application because they might still be interested in training, just unable to attend this particular event.)

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+1. One of trainees replies to email and resigns from the training.
+
+1. Admin unassignes her from the training and discard her training request.
+
+This workflow requires the following new features in "trainees" view:
+
+- Filter trainees by presence of training request.
```
gvwilson: Yes.

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+- Bulk email selected trainees with two presets of email content.
```
gvwilson: Yes.

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+- Column with list of training requests.
```
gvwilson: On what page?

docs/group-applications-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+
+And "filter by training" in "training requests" page.
```
gvwilson: Yes.

gvwilson:
Good job - I think you've thought everything through. :+1: to implement.

docs/group-applications-workflow.md
chrismedrela: New functionality.

docs/group-applications-workflow.md
chrismedrela: Yes, this option can be available all the time. There are three kinds of email that you may want to send (empty one, asking for confirmation and asking for filling the form). So we can have a button like this:
image
When you click it, an empty email is created. When you click the arrow, you can access other options:
image

Or, when you click blue button, what happens could be dynamic and depend on the filters:

- If you filtered by training and filtered only people who have sent a request, "Ask to confirm interest in training" option is used. 
- If you filtered by training and filtered only people who have not sent a request, "Ask to submit group application" option is used.
- Otherwise, an empty email is created.

Any opinions?

docs/group-applications-workflow.md
chrismedrela: Not yet, good catch.

docs/group-applications-workflow.md
chrismedrela: That's right.

docs/group-applications-workflow.md
chrismedrela: `trainees` view.

docs/group-applications-workflow.md
chrismedrela: I've just realized that this feature is actually not necessary.

docs/group-applications-workflow.md
gvwilson: Simplest thing right now is to create an empty email and expect the 
admin to fill it in from some other source.  Please go ahead and 
implement that, and create an issue saying we should create other kinds 
of email messages automatically in future.

chrismedrela:
Implemented in #871.

pbanaszkiewicz:
@chrismedrela close if possible

PR #932, opened 2016-07-30
Fix #834 with a function for YYYY-MM-DD dates formatting

### LINKED ISSUE #834: Date(?) stamp when updating record from url
When updating an event from url, the list of instructors/helpers is recorded in the notes field with a heading that looks like a date but isn't.  For example an update I made today has the heading "UPDATE 2016-5-4:"

What is this heading supposed to be?  Can we make it be the real date of the update (i.e., UPDATE 2016-06-16:")

pbanaszkiewicz:
The JavaScript `Date` object was extended with a prototype function
`yyymmdd` (adapted from http://stackoverflow.com/a/3067896) to display
zeros before month and day number when they consist of single digits.


gvwilson:
:+1:

PR #933, opened 2016-07-30
Fix #846 by adding "cancelled" tag

### LINKED ISSUE #846: Denote cancelled workshops in AMY
Add tag for cancelled workshops.
This should also result in the workshop being unpublished.  
(So published == has date and has location and has workshop page and  tag != "cancelled")

pbanaszkiewicz:
Logic behind calculating (un)published events also changed so that any
"cancelled" event will automatically disappear from published events
list.


workshops/test/base.py
```diff
@@ -298,6 +298,17 @@ def _setUpEvents(self):
...
+        # create a full-blown event that got cancelled
+        e = Event.objects.create(start=event_start, end=event_end,
+                                 slug='starts-today-cancelled',
+                                 url='http://example.org/cancelled-event',
+                                 latitude=-10.0, longitude=10.0,
+                                 country='US', venue='University',
+                                 address='Phenomenal Street',
+                                 host=test_host)
+        tags = Tag.objects.filter(name__in=['SWC', 'cancelled'])
+        e.tags.set(tags)
```
gvwilson: Is there a test somewhere to make sure that we can see cancelled events when we want to?

pbanaszkiewicz: Depends what you understand by "can see". See a listing on all-events page? We don't test that kind of thing, because we assume [django-filter](http://django-filter.readthedocs.io/en/latest/) works correctly (and with these simple filters you actually only need to manually check if they work).

If you meant something like a queryset method (`Event.object.cancelled_events`) then there's no such method and therefore no corresponding test.

If you meant whether we can see cancelled events among unpublished events, then a test for that exists in `workshops.test.test_event.TestEvent.test_unpublished_events`.

chrismedrela:
@pbanaszkiewicz, This PR looks good. 

If you give me PR to review, can I merge it?

pbanaszkiewicz:
@chrismedrela sure, go ahead! (But be prepared to fix things that break after the merge -- this stuff sometimes happens.)

PR #949, opened 2016-08-11
Fix bug when accepting training request

chrismedrela:
It wasn't possible to "accept & match to selected trainee account", because validation on the model was triggered. `TrainingRequest` requires that there is no matched person when the state is "pending", while the form, as submitted, has a trainee chosen.
    
Fixed by disabling model validation.


pbanaszkiewicz:
Can you please provide a regression test?

chrismedrela:
Done.

pbanaszkiewicz:
Thanks @chrismedrela!

PR #951, opened 2016-08-12
Hide survey fields on EventCreate form and EventDetail view

### LINKED ISSUE #944: Hide section for survey results
We've used a couple of different survey tools - this was requested in an earlier iteration of survey tools.  We're not recording these here now so I'd like to hide this from the view until we decide what to do.

narayanaditya95:
Fixes #944

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -3,7 +3,7 @@ import re
 from captcha.fields import ReCaptchaField
 from crispy_forms.bootstrap import FormActions
 from crispy_forms.helper import FormHelper
-from crispy_forms.layout import Layout, Div, HTML, Submit, Reset, Fieldset, Field
+from crispy_forms.layout import Layout, Div, HTML, Submit, Reset, Field
 from crispy_forms.bootstrap import AccordionGroup, Accordion
 from django import forms
 from django.core.exceptions import ValidationError
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -3,7 +3,8 @@ import re
 from captcha.fields import ReCaptchaField
 from crispy_forms.bootstrap import FormActions
 from crispy_forms.helper import FormHelper
-from crispy_forms.layout import Layout, Div, HTML, Submit, Reset
+from crispy_forms.layout import Layout, Div, HTML, Submit, Reset, Fieldset, Field
+from crispy_forms.bootstrap import AccordionGroup, Accordion
 from django import forms
 from django.core.exceptions import ValidationError
 from django.core.validators import RegexValidator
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
Should we perhaps comment out instead of deleting?

narayanaditya95:
I am in favour of re-implementing it as accordions so that they can be revealed on demand.
Crispy form has in-built support for accordions.[1]

[1]: http://django-crispy-forms.readthedocs.io/en/latest/layouts.html

pbanaszkiewicz:
Alright, let's go this way. Please leave a link to accordions somewhere in the code (perhaps in Event model near the field declarations?).

narayanaditya95:
@pbanaszkiewicz 
I rewrote the crispy implementation. It does not make any cosmetic change to the form rendering apart from adding the ability to hide the `Location details` accordian.

workshops/forms.py
```diff
@@ -393,17 +394,7 @@ def clean_end(self):
...
-                  'assigned_to', 'tags', 'url', 'language', 'reg_key',
-                  'admin_fee', 'invoice_status', 'attendance', 'contact',
-                  'notes', 'country', 'venue', 'address', 'latitude',
-                  'longitude', 'learners_pre', 'learners_post',
-                  'instructors_pre', 'instructors_post', 'learners_longterm')
-
-        # WARNING: don't change put any fields between 'country' and
-        #          'longitude' that don't relate to the venue of the event
-
+        fields = '__all__'
```
narayanaditya95: The field order is now determined by `self.helper.layout`.

workshops/forms.py
```diff
@@ -393,17 +391,10 @@ def clean_end(self):
...
+                  'assigned_to', 'tags', 'url', 'language', 'reg_key', 'venue',
                   'admin_fee', 'invoice_status', 'attendance', 'contact',
-                  'notes', 'country', 'venue', 'address', 'latitude',
-                  'longitude', 'learners_pre', 'learners_post',
-                  'instructors_pre', 'instructors_post', 'learners_longterm')
-
-        # WARNING: don't change put any fields between 'country' and
-        #          'longitude' that don't relate to the venue of the event
-
+                  'notes', 'country', 'address', 'latitude', 'longitude', )
```
narayanaditya95: This seems like a violation of DRY. I am repeating the fields I would to display in both `helper.layout` and `Meta.fields`.

workshops/forms.py
```diff
@@ -340,37 +341,34 @@ class EventForm(forms.ModelForm):
...
+        'url',
+        'language',
+        'reg_key',
+        'admin_fee',
+        'invoice_status',
+        'attendance',
+        'contact',
+        'notes',
+        Accordion(
+            AccordionGroup('Location details',
```
pbanaszkiewicz: So previously we had some implementation that's now in the `Accordion`? The output look is the same -- which means less code maintenance for us (big win)!

PR #956, opened 2016-08-12
Grant trainers access to admin dashboard

chrismedrela:


chrismedrela:
Don't merge yet.

pbanaszkiewicz:
Looks good to me (I would not change this so that people with some badge can be considered admins).

PR #959, opened 2016-08-15
Move "trainees" link in navigation bar

chrismedrela:
from drop-down menu to the navigation bar.


pbanaszkiewicz:
Does it fit on narrower pages?

pbanaszkiewicz:
@chrismedrela does it fir on narrow pages?

chrismedrela:
Unfortunately not. So I'm closing it.

PR #960, opened 2016-08-15
Use `event.slug` instead of `event.get_ident`

### LINKED ISSUE #957: API still calls event.get_ident
Which triggers an error for every admin on the dashboard :(

narayanaditya95:
Fixes #957


gvwilson:
:+1:

Issue #970, opened 2016-08-25
Bad links in training event display


gvwilson:
1. Go to https://amy.software-carpentry.org/workshops/trainings?page=2
2. Click on any of the links that supposedly lead to a training event.
3. 404

I understand why we want to retain this view - it's helpful to look at summary stats - but:

1. The links need to resolve.
2. We need to be able to sort chronologically with most recent first.

chrismedrela:
PR: https://github.com/swcarpentry/amy/pull/974

PR #974, opened 2016-08-27
Fix 970 - bad links and sort order in training event display

chrismedrela:


pbanaszkiewicz:
Ref #970

workshops/templates/workshops/all_trainings.html
```diff
@@ -12,7 +12,7 @@
       </tr>
       {% for training in all_trainings %}
       <tr>
-        <td><a href="{% url 'event_details' training.event %}">{{ training.event__slug }}</a></td>
+        <td><a href="{% url 'event_details' training.event__slug %}">{{ training.event__slug }}</a></td>
```
pbanaszkiewicz: Can we use `.` instead of `__` here?

workshops/views.py
```diff
@@ -1847,7 +1847,7 @@ def all_trainings(request):
         .values('event').annotate(finished=Count('person'))
     finished = {f['event']: f['finished'] for f in finished}
 
-    trainings = Task.objects.filter(role=learner).filter(event__tags=ttt).order_by('event__start') \
+    trainings = Task.objects.filter(role=learner).filter(event__tags=ttt).order_by('-event__start') \
```
pbanaszkiewicz: Not sure if Greg wanted to change the only sorting way or to have an option to choose sorting. @gvwilson ?

workshops/templates/workshops/all_trainings.html
chrismedrela: No, see https://github.com/swcarpentry/amy/blob/develop/workshops/views.py#L1850

workshops/views.py
pbanaszkiewicz: @gvwilson can you please specify what you had in mind [here](https://github.com/swcarpentry/amy/issues/970#issue-173275389)?

gvwilson: I wanted the default display to be most-recent first.

PR #1002, opened 2016-09-14
Add privacy consent field to multiple forms

### LINKED ISSUE #792: Add checkbox and link to privacy policy when people fill out `update_profile/`
We should have a checkbox that says "I have read and agree to the privacy policy" http://software-carpentry.org/privacy/ when people fill out their profile updates.

pbanaszkiewicz:
This fixes #792 by adding a separate field (via a mixin) to multiple
forms that should ask the user for a privacy policy agreement.

This approach doesn't store any additional data in the DB, but forces
user to agree on the privacy policy instead. Such behavior may not be
desired (ie. do we by default assume everyone already in the database
agreed on that policy?).

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -136,6 +136,14 @@ bootstrap_helper_filter = BootstrapHelperFilter()
 bootstrap_helper_inline_formsets = BootstrapHelperFormsetInline()
 
 
+class PrivacyConsentMixin(forms.Form):
+    privacy_consent = forms.BooleanField(
+        label='*I have read and agree to <a href='
+              '"https://software-carpentry.org/privacy/" target="_blank">'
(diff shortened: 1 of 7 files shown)


jduckles:
@pbanaszkiewicz looks good to me. Thanks!

gvwilson:
:+1: from me - once it's merged, please deploy to the dev server so that @jduckles @maneesha and @gvwilson can all test it. Thanks much.

pbanaszkiewicz:
@jduckles @maneesha @gvwilson this PR now lives on http://amy-dev.software-carpentry.org/workshops/

Please test, thank you!

tracykteal:
We don't actually assume that everyone in the database has agreed to the privacy policy, because no one in the database right now has agreed to it. We'll think about what we can do that.

Issue #955, opened 2016-08-12
Deployment on production


chrismedrela:

pbanaszkiewicz:
And describe backup system as well.

pbanaszkiewicz:
Here's what I have so far:

```
# Close milestone
Use GitHub UI: https://github.com/swcarpentry/amy/milestones

-------------------------

# Release AMY v1.7.1 (+ version bump on master)
(on `develop` branch)
git pull upstream develop
git push origin develop

git checkout master

(on `master` branch)
git merge --no-ff develop  # sometimes there are conflicts in files with release version, but most often we need to edit them and commit
vim bower.json
vim workshop/__init__.py
git add workshops/__init__.py bower.json
git commit -m "Bumping version to v1.7.1"

# sometimes there are errors and you'll have hard time removing remote tags etc.
# so just to be safe: TEST
make test

# finally tagging a release
git tag -a v1.7.1 -s  # -s means signed; tag message is usually "AMY release vx.y.z"

# pushing everything everywhere
git push origin master --tags
git push upstream master --tags

--------------------------

# Version bump on develop

git checkout develop
(on `develop` branch)
vim bower.json
vim workshop/__init__.py
git add workshops/__init__.py bower.json
git commit -m "Bumping version to v1.8.0-dev"

git push upstream
git push origin

--------------------------

# Update production

ssh amy@amy.software-carpentry.org

(via SSH)
source ~/amy_site/venv/bin/activate
cd amy_site/amy
git fetch
git checkout tags/v1.7.1  # Python files updated
# worthwhile to create a database backup here!
cp db.sqlite3 db.sqlite3.bak
./manage.py migrate
make serve  # regenerates version in the footer as a side effect
# PUT HERE: additional stuff like one-off commands
make upgrade  # added / updated requirements

# refresh the webserver 
touch ../amy_uwsgirc.ini  # reloads some uWSGI magic
```

chrismedrela:
I can do the next release, so that I'll check if there is anything missing, and then we can put these instructions in i.e. `docs/deploy-procedure.md` file.

pbanaszkiewicz:
@chrismedrela as per email with Chris, the next release (v1.8) will be deployed by me.

chrismedrela:
@pbanaszkiewicz, I'll deploy v1.9 with your help, ok?

pbanaszkiewicz:
Sure!

pbanaszkiewicz:
[...]

PR #1003, opened 2016-09-14
Event import: Prefer GET `url` from POST one

### LINKED ISSUE #999: Update from valid URL causes "Network Connection Error"
(title says it all)

pbanaszkiewicz:
This fixes #999.

The error in #999 was that somehow the `url` parameter sent to
`fetch_event_metadata` function was received from GET, while we used the
POST method.

This fix makes sure we accept any `url` (GET is still preferred).

### DIFF
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -1367,7 +1367,8 @@ def event_import(request):
     This is used to read metadata from workshop website and then fill up fields
     on event_create form."""
 
-    url = request.GET.get('url', '').strip()
+    url = (request.POST.get('url', '').strip() or
+           request.GET.get('url', '').strip())
 
     try:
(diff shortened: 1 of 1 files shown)


gvwilson:
:+1: - please deploy and ask @maneesha and @apawlik to verify.

narayanaditya95:
@pbanaszkiewicz How is that possible?
Both `import_from_url.js` and `update_from_url.js` use the GET method on submission.

pbanaszkiewicz:
@narayanaditya95 was right; the production server was using outdated static files. I updated them during today's software update.

I already tested that the fix is no longer needed, @maneesha and @apawlik can test independently but I'm going to close this PR.

The instructions in #955 were updated to reflect this missing static files collection step.

PR #1013, opened 2016-09-20
Change the 'Reset' button to 'Cancel' button

### LINKED ISSUE #984: Change the way "reset" buttons work
- Rename button from "reset" to "cancel".
- Make the button white instead of red.
- Redirect back to the last page in history.

narayanaditya95:
- Renames red 'Reset' button to white 'Cancel' button
- Redirects to the last page in history

Fixes #984

However, if there is no browser history, the button does nothing.

### DIFF
--- a/pydata/forms.py
+++ b/pydata/forms.py
@@ -21,7 +21,7 @@ class BaseModelAddFormSet(forms.models.BaseModelFormSet):
     validate_min = False
 
     helper = BootstrapHelper(
-        form_tag=False, add_submit_button=False, add_reset_button=False)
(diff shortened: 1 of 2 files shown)


narayanaditya95:
On the `event_edit` and `person_edit` views, switching between tabs cause history to be appended. The cancel button would then simply switch between tabs.

I do not think this is a good enough implementation of a cancel button.

chrismedrela:
> On the event_edit and person_edit views, switching between tabs cause history to be appended. The cancel button would then simply switch between tabs.

Yes, I agree it's a bit buggy in person edit view, but I think this PR is good enough.

PR #1014, opened 2016-09-22
Alter choices in partner form.

### LINKED ISSUE #1011: Remove $1250 from partner form item
Your call @maneesha, but shouldn't we remove the "$1250" from this line as our partnerships include coordinated workshops now?

https://github.com/swcarpentry/amy/blob/da20218f2497792962c96b42ecc21759cf928687/workshops/models.py#L1218

narayanaditya95:
Remove $1250 choice.
Fixes #1011


workshops/models.py
```diff
@@ -1215,7 +1215,6 @@ class EventRequest(AssignmentMixin, ActiveMixin, CreatedUpdatedMixin,
 
     ADMIN_FEE_PAYMENT_CHOICES = (
         ('NP1', 'Non-profit / non-partner: US$2500'),
-        ('partner', 'Partner: US$1250'),
```
pbanaszkiewicz: What happens when we have a historic data with "partner" option in this table? Shouldn't we "hide" it from the form only?

narayanaditya95: They remain unchanged, although we lose out on the verbose representation.

```py
In [5]: er.admin_fee_payment
Out[5]: 'partner'

In [6]: er.get_admin_fee_payment_display()
Out[6]: 'partner'
```

pbanaszkiewicz: In that case :+1:

PR #1018, opened 2016-09-24
Fix #981 -- add more ways to issue a badge

### LINKED ISSUE #981: More ways to get to "issue badge"
1. When I click on `[+]` in the "Progress" column and am taken to the "New training progress" page, one of the options there should be "issue badge".  I realize this isn't training progress, but it's a natural time and place to issue the badge.

2. Similarly, when I'm on the "Trainees" page, if I see someone is eligible but doesn't yet have a badge (i.e., they have a green box in the "Eligible?" column), I should be able to click on that and be taken to the "issue a badge" page.

chrismedrela:
"SWC" and "DC" labels in "eligible" column in trainees list view are
now clickable when a person is eligible to become SWC/DC instructor, but
doesn't have a swc/dc-instructor badge yet. The user is moved to person
edit view where the badge can be awarded.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -464,7 +464,7 @@ class PersonForm(forms.ModelForm):
         widget=selectable.AutoComboboxSelectMultipleWidget,
     )
 
-    helper = bootstrap_helper
+    helper = BootstrapHelper(form_id='person-edit-form')
 
     class Meta:
         model = Person
@@ -647,7 +647,7 @@ class PersonAwardForm(forms.ModelForm):
         widget=selectable.AutoComboboxSelectWidget,
     )
 
-    helper = BootstrapHelper(submit_label='Add')
+    helper = BootstrapHelper(submit_label='Add', form_id='person-awards-form')
 
     class Meta:
         model = Award
@@ -664,7 +664,7 @@ class PersonTaskForm(forms.ModelForm):
         widget=selectable.AutoComboboxSelectWidget,
     )
 
-    helper = BootstrapHelper(submit_label='Add')
+    helper = BootstrapHelper(submit_label='Add', form_id='person-tasks-form')
 
     class Meta:
         model = Task
@@ -1338,9 +1338,6 @@ class BulkDiscardProgressesForm(forms.Form):
         HTML('&nbsp;<a bulk-email-on-click class="btn btn-primary">'
              'Mail selected trainees</a>'),
     )
-
-    class Meta:
-        model = TrainingProgress
 
 
 class BulkChangeTrainingRequestForm(forms.Form):
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -44,7 +44,8 @@ class BootstrapHelper(FormHelper):
                  additional_form_class='',
                  form_tag=True,
                  display_labels=True,
-                 form_action=None):
+                 form_action=None,
+                 form_id=None):
         """
         `duplicate_buttons_on_top` -- Whether submit buttons should be
         displayed on both top and bottom of the form.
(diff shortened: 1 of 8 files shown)


workshops/templates/workshops/all_trainees.html
```diff
@@ -107,8 +107,9 @@
                       </ul>
                   ">SWC</span>
           {% else %}
-            <span class="label label-success" data-toggle="popover"
-                  data-content="Eligible to be certified as Software Carpentry Instructor.">SWC</span>
+            <a class="label label-success" data-toggle="popover"
+               href="{% url 'person_edit' trainee.pk %}?badge=swc-instructor&next={{ request.get_full_path|urlencode }}#awards"
```
pbanaszkiewicz: GH suggests `&amp;` instead of `&` here.

workshops/templates/workshops/all_trainees.html
```diff
@@ -124,8 +125,9 @@
                     </ul>
                   ">DC</span>
           {% else %}
-            <span class="label label-success" data-toggle="popover"
-                  data-content="Eligible to be certified as Data Carpentry Instructor.">DC</span>
+            <a class="label label-success" data-toggle="popover"
+               href="{% url 'person_edit' trainee.pk %}?badge=dc-instructor#awards"
```
pbanaszkiewicz: Why the links are different in case of SWC and DC?

workshops/views.py
```diff
@@ -862,103 +864,113 @@ def form_valid(self, form):
...
+    # Determine initial badge in PersonAwardForm
+    try:
+        initial_badge_name = request.GET['badge']
+    except KeyError:
+        initial_badge = None
+    else:
+        try:
+            initial_badge = Badge.objects.get(name=initial_badge_name)
+        except Badge.DoesNotExist:
+            initial_badge = None
```
pbanaszkiewicz: You can definitely compress these nested try-excepts into one.

workshops/util.py
```diff
@@ -946,3 +947,20 @@ def homework2state(homework):
...
+
+    next_url = request.GET.get('next', None)
+    if next_url is not None and is_safe_url(next_url):
+        return redirect(next_url)
+    else:
+        return redirect(*args, **kwargs)
+
+
+def dict_without_Nones(**keys):
+    return {k: v for k, v in keys.items() if v is not None}
```
pbanaszkiewicz: Please check if we don't have something similar used somewhere, because I do seem to recall.

workshops/forms.py
```diff
@@ -1333,9 +1333,6 @@ class BulkDiscardProgressesForm(forms.Form):
              'Mail selected trainees</a>'),
     )
 
-    class Meta:
-        model = TrainingProgress
-
```
pbanaszkiewicz: Why remove this?

pbanaszkiewicz:
Please also make sure to add tests for the person-edit page because you altered its behavior a little bit.

workshops/util.py
chrismedrela: @pbanaszkiewicz, I cannot find anything like that. Any idea where such utility could live?

workshops/forms.py
chrismedrela: Because it's a `Form`, not `ModelForm`.

chrismedrela:
@pbanaszkiewicz, ping. One thing that I really don't like in this PR is `Person.get_training_slug` (it's really inefficient). Any ideas?

chrismedrela:
@pbanaszkiewicz, I've updated the PR. There is no longer any issue with `Person.get_training_slug`. Now this PR is ready for review.

workshops/util.py
pbanaszkiewicz: If you didn't find it, I wouldn't worry about reusing this particular oneliner.

PR #1023, opened 2016-09-27
Add instructions in bulk add person.

### LINKED ISSUE #1015: In bulk add person, interpret "role" column and use <select>.
In bulk add persons (https://amy.software-carpentry.org/workshops/persons/bulkadd/) we are given a csv template which is useful.

However, the names of roles as they display to the end user are not the same as the names of roles as stored in the database.  Can we either have instructions on what the roles need to be or have the csv importer translate this?

chrismedrela:
Partially fixes #1015.

### DIFF
--- a/workshops/templates/workshops/person_bulk_add_form.html
+++ b/workshops/templates/workshops/person_bulk_add_form.html
@@ -3,6 +3,22 @@
 {% block content %}
 {% load staticfiles %}
 <p><a href="{% url 'person_bulk_add_template' %}">BulkPersonAddTemplate.csv</a> is a blank template of a well-formed .CSV file.</p>
+
+<p>Hint: In the column <code>role</code>, use the names from the left column of the following table:</p>
+
+<table class="table table-striped table-bordered table-condensed">
+  <tr>
+    <th>name to be used in .csv file</th>
+    <th>verbose name</th>
+  </tr>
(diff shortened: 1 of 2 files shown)


chrismedrela:
Here is how the bulk add person page looks like:
image

workshops/templates/workshops/person_bulk_add_form.html
```diff
@@ -3,6 +3,22 @@
...
+    <th>name to be used in .csv file</th>
+    <th>verbose name</th>
+  </tr>
+  <tr><th class="success">contributor</th><td class="danger">Contributed to lesson materials</td></tr>
+  <tr><th class="success">organizer</th><td class="danger">Workshop organizer</td></tr>
+  <tr><th class="success">learner</th><td class="danger">Learner</td></tr>
+  <tr><th class="success">host</th><td class="danger">Workshop host</td></tr>
+  <tr><th class="success">instructor</th><td class="danger">Instructor</td></tr>
+  <tr><th class="success">helper</th><td class="danger">Helper</td></tr>
+</table>
```
gvwilson: Good catch - @chrismedrela ?

chrismedrela: Yes, that's right, replaced with looping over roles pulled from db.

PR #1025, opened 2016-09-28
Fix #1016 -- display SWC/DC tags in "current" events in admin dashboard

### LINKED ISSUE #1016: Display SWC or DC tag next to workshop slug in "Current" list
On the main page  -  it would be useful to see at a glance whether the workshops listed as "Current" were SWC or DC. 

If you need room for this, the "Uninvoiced" column can be narrower.  The "ID" column in the "Unpublished" column can also be removed - this is unnecessary.

chrismedrela:


workshops/templates/workshops/admin_dashboard.html
```diff
@@ -39,6 +40,8 @@
     <table class="table table-striped">
     {% for event in current_events %}
     <tr>
+      <td class="text-center" width="25px">{% if event.is_swc %}{% bootstrap_tag "SWC" %}{% endif %}</td>
+      <td class="text-center" width="25px">{% if event.is_dc %}{% bootstrap_tag "DC" %}{% endif %}</td>
```
pbanaszkiewicz: 1. You forgot about "LC".
2. That's a lot of space.

And you had to add an annotation. Maybe we can fix this by extending `Tag` model? E.g. add `Tag.objects.carpentries()` that's equivalent to `Tag.objects.filter(name__in=['SWC', 'DC', 'LC'])`, and later we could use `{% for tag in event.tags.carpentries %}{% bootstrap_tag tag %}{% endfor %}`, although this may generate quite a few queries. Anyway, my point is that there might be a smarter way.

gvwilson: @chrismedrela is there a quick fix using a loop over available types of carpentry? (I have no comment on the amount of space required.)

chrismedrela: @pbanaszkiewicz , @maneesha wants to have the labels aligned.

I replaced hard-coding with looping. Any opinions?

pbanaszkiewicz: @chrismedrela if they need to be aligned, then please make more space for them on the dashboard (as @maneesha suggested in the original post).

As for the looping, it looks like it will yield a lot of queries, but let me check first. Also, currently you add code that's not being used in this PR.

workshops/templates/workshops/admin_dashboard.html
```diff
@@ -39,6 +40,13 @@
     <table class="table table-striped">
     {% for event in current_events %}
     <tr>
+      {% for tag in carpentries %}
+        <td class="text-center" width="25px">
+          {% if tag in event.tags.all %}
+            {% bootstrap_tag tag.name %}
+          {% endif %}
+        </td>
+      {% endfor %}
```
pbanaszkiewicz: This approach yields a lot of queries.

The best would be to drop "must-be-aligned" constraint and prefetch all related tags and then loop over them.

workshops/views.py
```diff
@@ -344,7 +344,14 @@ def admin_dashboard(request):
...
         Event.objects.upcoming_events() | Event.objects.ongoing_events()
-    ).active()
+    ).active().annotate(
+        is_swc=Sum(Case(When(tags__name='SWC', then=1),
+                        default=0,
+                        output_field=IntegerField())),
+        is_dc=Sum(Case(When(tags__name='DC', then=1),
+                       default=0,
+                       output_field=IntegerField())),
+    )
```
pbanaszkiewicz: UPDATE: actually replacing this annotation with `prefetch_related('tags')` helps with the number of queries! I'd go this way and then merge.

workshops/templates/workshops/admin_dashboard.html
pbanaszkiewicz: > Which code exactly?

You removed it already (the `is_dc`/`is_swc` annotations).

workshops/templates/workshops/admin_dashboard.html
pbanaszkiewicz: Please make sure that this doesn't yield N queries here. I know I got it down to `const` queries per this "current events" column, but I don't remember with which code.

chrismedrela: It makes exactly 3 queries, no matter how many current events are there.

pbanaszkiewicz: Nice!

gvwilson:
I think this is ready to merge - @chrismedrela @pbanaszkiewicz please push the big green button if you agree.

pbanaszkiewicz:
Thanks @chrismedrela !

PR #1026, opened 2016-09-28
Fix #980 -- filter trainees by "ready to be badged" status

### LINKED ISSUE #980: Filter trainees by "ready to be badged" status
On the "Trainees" page, we'd like "Is SWC Instructor?" and "Is DC Instructor?" to have one more option each: "Eligible".  This would filter by people who have fulfilled requirements, but do not yet have the badge.  I know we can calculate this (it's what turns their marker green), and it would be very helpful when one person is marking completion of work and another is issuing badges.

chrismedrela:
~~Don't merge yet.~~

### DIFF
--- a/workshops/filters.py
+++ b/workshops/filters.py
@@ -262,6 +262,25 @@ def filter_trainees_by_training_request_presence(queryset, flag):
         return queryset.filter(trainingrequest__isnull=True)
 
 
+def filter_trainees_by_instructor_status(queryset, choice):
+    if choice == '':
+        return queryset
+    elif choice == 'swc-and-dc':
+        return queryset.filter(is_swc_instructor=True, is_dc_instructor=True)
+    elif choice == 'swc-or-dc':
+        return queryset.filter(Q(is_swc_instructor=True) |
+                               Q(is_dc_instructor=True))
+    elif choice == 'swc':
+        return queryset.filter(is_swc_instructor=True)
+    elif choice == 'dc':
+        return queryset.filter(is_dc_instructor=True)
+    elif choice == 'eligible':
+        return queryset.filter(Q(swc_eligible=True, is_swc_instructor=False) |
+                               Q(dc_eligible=True, is_dc_instructor=False))
+    else:  # choice == 'no'
+        return queryset.filter(is_swc_instructor=False, is_dc_instructor=False)
+
+
 def filter_trainees_by_training(queryset, training):
     if training is None:
         return queryset
@@ -291,14 +310,18 @@ class TraineeFilter(AMYFilterSet):
         action=filter_trainees_by_training_request_presence,
     )
 
-    is_swc_instructor = django_filters.BooleanFilter(
-        label='Is SWC Instructor?',
-        name='is_swc_instructor',
-    )
-
-    is_dc_instructor = django_filters.BooleanFilter(
-        label='Is DC Instructor?',
-        name='is_dc_instructor',
+    is_instructor = django_filters.ChoiceFilter(
+        label='Is SWC/DC instructor?',
+        action=filter_trainees_by_instructor_status,
+        choices=[
+            ('', 'Unknown'),
+            ('swc-and-dc', 'Both SWC and DC'),
+            ('swc-or-dc', 'SWC or DC '),
+            ('swc', 'SWC instructor'),
+            ('dc', 'DC instructor'),
+            ('eligible', 'No, but eligible to be certified'),
+            ('no', 'No'),
+        ]
     )
 
     training = django_filters.ModelChoiceFilter(
@@ -312,8 +335,7 @@ class TraineeFilter(AMYFilterSet):
             'search',
             'all_persons',
             'homework',
-            'is_swc_instructor',
-            'is_dc_instructor',
+            'is_instructor',
             'training',
         ]
         order_by = ["-last_login", "lastname", "-lastname", "firstname", "-firstname",
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -9,7 +9,7 @@ from django.core.exceptions import ValidationError
 from django.core.urlresolvers import reverse
 from django.core.validators import MinValueValidator, RegexValidator
 from django.db import models
-from django.db.models import Q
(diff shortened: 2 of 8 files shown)


chrismedrela:
Ready for review.

This PR replaces the two "is SWC instructor" and "is DC instructor" filters with a new one:
image

gvwilson:
Travis-CI reports failure.

chrismedrela:
@gvwilson, yes, I forget to update some tests. Now it's ready for review.

workshops/filters.py
```diff
@@ -261,6 +261,25 @@ def filter_trainees_by_training_request_presence(queryset, flag):
...
+    elif choice == 'swc-or-dc':
+        return queryset.filter(Q(is_swc_instructor=True) |
+                               Q(is_dc_instructor=True))
+    elif choice == 'swc':
+        return queryset.filter(is_swc_instructor=True)
+    elif choice == 'dc':
+        return queryset.filter(is_dc_instructor=True)
+    elif choice == 'eligible':
+        return queryset.filter(Q(swc_eligible=True) | Q(dc_eligible=True),
+                               is_swc_instructor=False, is_dc_instructor=False)
```
narayanaditya95: This would end up filtering out people with an SWC badge but still eligible for obtaining a DC badge.

Is that the intended functionality?

workshops/models.py
```diff
@@ -263,6 +263,33 @@ def get_by_natural_key(self, username):
...
+            passed_dc_demo=passed('DC Demo'),
+        ).annotate(
+            swc_eligible=(F('passed_training') *
+                          F('passed_swc_homework') *
+                          F('passed_discussion') *
+                          F('passed_swc_demo')),
+            dc_eligible=(F('passed_training') *
+                         F('passed_dc_homework') *
+                         F('passed_discussion') *
+                         F('passed_dc_demo')),
```
pbanaszkiewicz: Isn't this some sort of hard-coding?

pbanaszkiewicz:
There's [confusion](https://github.com/swcarpentry/amy/pull/1026#discussion_r81327735) to be cleared out and GitHub reports conflicts.

workshops/filters.py
gvwilson: @chrismedrela can you please check?

workshops/models.py
gvwilson: Agreed - we should loop over something here, no?

workshops/filters.py
chrismedrela: @narayanaditya95 , you're right, good catch!

workshops/models.py
chrismedrela: Yes, but I don't know how we can improve it (without using `reduce`, which makes code much harder to understand).

pbanaszkiewicz: Are there any other places that hardcode what requirements are needed for someone to pass the training?

chrismedrela: There:
- [`Person.get_missing_swc/dc_instructor_requirements`](https://github.com/swcarpentry/amy/blob/develop/workshops/models.py#L436)
- the trainee dashboard ([template](https://github.com/swcarpentry/amy/blob/develop/workshops/templates/workshops/trainee_dashboard.html), [view](https://github.com/swcarpentry/amy/blob/develop/workshops/views.py#L3415))

There may be other places, but I cannot recall any other one now.

gvwilson:
Tests are passing, but this still feels pretty complicated - if you both think there's a way to simplify it, we can push this one to V1.10 rather than merging this and fixing later. Please let me know.

chrismedrela:
I've refactored trainee dashboard a bit, so that it reuses logic from `annotate_with_instructor_eligibility`. I don't see any easy way to simplify it -- we just need to ensure that the number of places where we hard-code requirements set is small and well known.

gvwilson:
OK - please go ahead and merge.

chrismedrela:
@pbanaszkiewicz , did you completed your code review and can we merge?

pbanaszkiewicz:
@chrismedrela I'll take a look at the hardcoded requirements and get back to you today.

pbanaszkiewicz:
I don't think we should change the hardcoded requirements now, but eventually we'll have to.

When that day comes I think we should go this way:

* link requirements with training instance (eg. *this particular SWC instructor training will require trainees to pass: homework & discussion session & teaching live*),
* once a trainee is added to the training, they should be created 1 requirement instance (not-passed by default) per each requirement (eg. if Jan Kowalski [==*John Smith*] attends this particular SWC instructor training, when he enrols and gets accepted, he should be created empty/non-passed `SWC homework`, `Discussion session` and `Teaching live` requirement instances). This approach saves us from some terrible DB queries, like finding trainees eligible for instructor badges,
* we should disallow changing requirements for ongoing trainings, or for any trainings with accepted people - alternatively we should have procedure for checking integrity between training requirements and requirements instances assigned to each trainee.

Tl;dr; I think @chrismedrela can merge this.

chrismedrela:
> once a trainee is added to the training, they should be created 1 requirement instance (not-passed by default) per each requirement (eg. if Jan Kowalski [==John Smith] attends this particular SWC instructor training, when he enrols and gets accepted, he should be created empty/non-passed SWC homework, Discussion session and Teaching live requirement instances). This approach saves us from some terrible DB queries, like finding trainees eligible for instructor badges,

@pbanaszkiewicz, I'm :-1: for that. I agree that some queries are terrible, but your approach will result in invalid query results when database become inconsistent - I mean when there is no empty/non-passed records when there should be. And there are a lot of scenarios how this could happen.

@gvwilson @pbanaszkiewicz We can keep the implementation simple if we assume that there is always only one set of binding requirements (for each carpentry). So, when we want to change requirements (let say, add second SWC homework), we say "starting from November 2016, checkout procedure is changed, so if you want to become SWC instructor, you need to pass training, first homework, second homework, discussion and demo; it doesn't matter which TTT training you attended". I think it's ok as long as 1) we warn trainees at TTT workshops that checkout procedure may change and new rules will apply to them and 2) we tell them about changes at least N days in advance. It may even help trainees who procrastinate on going through checkout procedure.

PR #1028, opened 2016-09-28
Fix #976 -- let users edit training progress in detailed person view

### LINKED ISSUE #976: Add training status items to person display
The page that displays information about a Person should also display their training status items, and allow them to be edited, just like the 'Training' display page does, so that if I search for a person by name and bring up that page, I can edit their information.

chrismedrela:


workshops/templates/workshops/training_progresses_inline.html
```diff
@@ -0,0 +1,17 @@
+{% load training_progress %}
```
narayanaditya95: I believe that the `training_progress` template tag can be removed from `all_trainees.html`

workshops/templates/workshops/training_progresses_inline.html
```diff
@@ -0,0 +1,17 @@
...
+     data-html="true">
+    {% if progress.discarded %}<strike>{% endif %}
+      {{ progress.requirement }}
+    {% if progress.discarded %}</strike>{% endif %}
+  </a>
+  &nbsp;
+{% endfor %}
+<a href="{% url 'trainingprogress_add' person.id %}?next={{ request.get_full_path|urlencode }}"
+   class="label label-default"
+   data-toggle="popover" data-content="Click to add new progress">+</a>
```
narayanaditya95: Newline, maybe?

gvwilson:
Looks good to me once @narayanaditya95's small comments are addressed - thanks.

chrismedrela:
@narayanaditya95 , thank you for the review!

> Should we consider moving all templates that are explicitly intended to be included within another template under an includes directory?

Yes, although let's do this in a separate issue.

PR #1029, opened 2016-09-28
Install django debug toolbar

chrismedrela:
The toolbar is enabled only when `DEBUG == True`. By default, almost all its functionality is disabled in order not to slow down page reloads. They can be turned on using checkboxes in the toolbox. The only one functionality that is turned on is "timing panel" that shows how long it took to process the last request.

The most useful feature of the toolbar is SQL panel which shows how many queries have been executed and how long did they take. It's important to track how long they execute to ensure that AMY scales up with growing database. In addition, it's much easier to make performance experiments now.

### DIFF
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -141,11 +141,13 @@ INSTALLED_APPS = (
     'captcha',
     'compressor',
     'social.apps.django_app.default',
+    'debug_toolbar',
 )
 
 CRISPY_TEMPLATE_PACK = 'bootstrap3'
 
 MIDDLEWARE_CLASSES = (
+    'debug_toolbar.middleware.DebugToolbarMiddleware',
     'reversion.middleware.RevisionMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.common.CommonMiddleware',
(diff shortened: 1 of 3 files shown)


pbanaszkiewicz:
I tried including DDT in AMY but @gvwilson was so against it :-)

@chrismedrela if you convince @gvwilson, that would be awesome.

amy/settings.py
```diff
@@ -340,3 +342,25 @@
...
+        'debug_toolbar.panels.request.RequestPanel',
+        'debug_toolbar.panels.sql.SQLPanel',
+        'debug_toolbar.panels.staticfiles.StaticFilesPanel',
+        'debug_toolbar.panels.templates.TemplatesPanel',
+        'debug_toolbar.panels.cache.CachePanel',
+        'debug_toolbar.panels.signals.SignalsPanel',
+        'debug_toolbar.panels.logging.LoggingPanel',
+        'debug_toolbar.panels.redirects.RedirectsPanel',
+    ],
+}
```
pbanaszkiewicz: I'd actually leave the original config (since I find it useful).

There's a newline missing at the end of this file.

gvwilson:
If you two both want it, you can outvote me.

pbanaszkiewicz:
@gvwilson your opinion matters too, so are you against incorporating DDT into AMY?

gvwilson:
I don't do dev these days, so if you two find it useful, add it.

gvwilson:
@chrismedrela please merge once you've addressed @pbanaszkiewicz's comment.

chrismedrela:
@pbanaszkiewicz , I'll merge it. If you really want to turn on SQL panel by default, please create an issue so we can discuss it.

amy/settings.py
pbanaszkiewicz: This means that trainees list is a very unoptimised page. I checked and yes indeed, 28 queries when the page loads only 3 trainees. That's way too much.

chrismedrela: That's true, it makes ~200 queries in 600 ms on production data.

chrismedrela: I've created an issue for that https://github.com/swcarpentry/amy/issues/1047

Issue #1032, opened 2016-09-29
Enable filtering by multiple partnership levels on organizations.


gvwilson:
We should be able to filter organizations according to their partnership level.  User should be able to select multiple partnership levels, which should be combined with 'or' (i.e., show me silver or bronze).  For now, don't worry about temporal nature of partnerships: show all organizations that have, or have ever had, a particular partnership level.

chrismedrela:
@gvwilson, when you say partnership, do you mean `Membership`? 

Also, I don't think we track partnership levels (silver, bronze etc.) -- we only track variant (partner/affiliate/sponsor) and contribution type (financial/person-days/other).

chrismedrela:
Ahh, ok, I've just had a look at #1034 and I see there are new variants (bronze, silver etc.).

gvwilson:
Yes, sorry, Membership, and I've added the new levels.

PR #1034, opened 2016-09-29
Displaying memberships as first-class objects

gvwilson:
This is related to #1032, which asks for filtering of organizations by partnership levels.  It adds pages to display all partnerships and to view the details of a single partnership.

**NOTE:** `workshops/urls.py` now has `organization/membership/add` and `memberships/add` - should the former be removed?


workshops/migrations/0119_auto_20160929_1131.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('workshops', '0118_auto_20160922_0009'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='membership',
+            name='variant',
+            field=models.CharField(choices=[('partner', 'Partner'), ('affiliate', 'Affiliate'), ('sponsor', 'Sponsor'), ('bronze', 'Bronze'), ('silver', 'Silver'), ('gold', 'Gold'), ('platinum', 'Platinum')], max_length=40),
```
pbanaszkiewicz: This doesn't mess up with @narayanaditya95's Sponsorships, does it?

workshops/views.py
```diff
@@ -496,6 +509,13 @@ def membership_create(request, org_domain):
     return render(request, 'workshops/generic_form.html', context)
 
 
+class MembershipCreate(OnlyForAdminsMixin, PermissionRequiredMixin,
+                 CreateViewContext):
+    permission_required = 'workshops.add_membership'
+    model = Membership
+    form_class = MembershipForm
```
pbanaszkiewicz: Perhaps it'd useful to have an autocompletion field `organization` in the `MembershipForm`.

gvwilson:
I don't know - @narayanaditya95 can you please review as well?
<https://github.com/narayanaditya95>

gvwilson:
:+1: but I don't know how to implement that.

workshops/migrations/0119_auto_20160929_1131.py
narayanaditya95: This is independent of the `Sponsorship` model, isn't it?

Memberships represent contracts between an org and SWC/LC.
Sponsorships represent contracts between an org and a particular event.

workshops/templates/workshops/all_memberships.html
```diff
@@ -0,0 +1,39 @@
...
+    {% for membership in all_memberships %}
+        <tr>
+            <td><a href="{% url 'organization_details' membership.organization.domain %}">{{ membership.organization.fullname }}</a></td>
+	    <td>{{ membership.variant }}</td>
+	    <td>{{ membership.agreement_start }}</td>
+	    <td>{{ membership.agreement_end }}</td>
+	    <td>{{ membership.contribution_type }}</td>
+        </tr>
+    {% endfor %}
+    </table>
```
narayanaditya95: Please also add actions buttons to allow access to `membership_detail` and `membership_edit` views.

workshops/views.py
narayanaditya95: @pbanaszkiewicz We already have a view called `membership_create` [here](https://github.com/swcarpentry/amy/blob/develop/workshops/views.py#L464).

We should extend it to allow populating the `organization` field based on a URL or GET parameter.

workshops/views.py
```diff
@@ -456,6 +457,32 @@ class OrganizationDelete(OnlyForAdminsMixin, PermissionRequiredMixin,
...
+    return render(request, 'workshops/all_memberships.html', context)
+
+
+@admin_required
+def membership_details(request, membership_id):
+    '''List details of a particular membership.'''
+    membership = get_object_or_404(Membership, id=membership_id)
+    context = {'title' : '{0}'.format(membership),
+               'membership' : membership}
+    return render(request, 'workshops/membership.html', context)
```
narayanaditya95: Should we add a link to `membership_details` view from the `organization_details` view?
Although we already have a table there, a link wouldn't hurt.

workshops/urls.py
```diff
@@ -20,7 +20,12 @@
         url(r'^membership/add/$', views.membership_create, name='membership_add'),
     ])),
 
+    url(r'^memberships/', include([
+        url(r'^$', views.all_memberships, name='all_memberships'),
+        url(r'^add/$', views.MembershipCreate.as_view(), name='membership_add'),
+    ])),
```
narayanaditya95: We can do away with the other `membership_add` urlpattern by extending the existing `membership_create` view.

workshops/views.py
pbanaszkiewicz: @narayanaditya95 good catch & :+1: for single view

workshops/urls.py
pbanaszkiewicz: I'd try to use the CBV `MembershipCreate` and change the earlier url `^organization/(?P<org_domain>[\w\.-]+)/membership/add/$` to `^memberships/add/(?P<org_domain>[\w\.-]+)/$`.

gvwilson:
@pbanaszkiewicz do you have time to finish this off for V1.9 ?

pbanaszkiewicz:
@gvwilson I can work on this and #1040 on Saturday.

pbanaszkiewicz:
@gvwilson here you go https://github.com/gvwilson/amy/pull/8 please merge -- this PR should be automatically updated.

gvwilson:
@pbanaszkiewicz thanks - please give it a final look in case I broke anything in the merge?

workshops/templates/workshops/all_memberships.html
```diff
@@ -0,0 +1,39 @@
...
+        <tr>
+            <th>organization</th>
+            <th>kind</th>
+            <th>start date</th>
+            <th>end date</th>
+            <th>contribution</th>
+        </tr>
+    {% for membership in all_memberships %}
+        <tr>
+            <td><a href="{{ membership.get_absolute_url }}">{{ membership.organization.fullname }}</a></td>
```
narayanaditya95: Isn't this confusing to add a hyperlink to the `membership_details` view here? I would suggest adding view and edit glyphicons like the ones found in `all_tasks` view.

workshops/templates/workshops/organization.html
```diff
@@ -72,27 +72,26 @@
...
   <tr>
-    <th></th><th></th><th></th><th></th>
     <th><small>Imposed</small></th><th><small>Completed</small></th><th><small>Remaining</small></th>
     <th><small>Imposed</small></th><th><small>Completed</small></th><th><small>Remaining</small></th>
     <th></th>
   </tr>
   {% for membership in organization.membership_set.all %}
   <tr>
-    <td>{{ membership.get_variant_display }}</td>
+    <td><a href="{{ membership.get_absolute_url }}">{{ membership.get_variant_display }}</a></td>
```
narayanaditya95: Why not add a glyphicon to the right of the row to link to `membership_details` view?

workshops/views.py
```diff
@@ -486,6 +513,32 @@ def membership_create(request, org_domain):
...
+
+    def get_initial(self):
+        initials = super().get_initial()
+
+        if self.org_domain:
+            organization = get_object_or_404(Organization,
+                                             domain=self.org_domain)
+            initials.update({'organization': organization})
+
+        return initials
```
narayanaditya95: We can get rid of the `get` and `post` methods:

```py
    def get_initial(self):
        initials = super().get_initial()
        org_domain = self.kwargs.get('org_domain', None)
        if org_domain:
            organization = get_object_or_404(Organization,
                                             domain=org_domain)
            initials.update({'organization': organization})

        return initials
```

workshops/views.py
narayanaditya95: @pbanaszkiewicz You forgot to add an autocomplete `organization` field.

narayanaditya95: Also trim out the existing `membership_create` function.

workshops/views.py
```diff
@@ -486,6 +513,32 @@ def membership_create(request, org_domain):
     return render(request, 'workshops/generic_form.html', context)
 
 
+class MembershipCreate(OnlyForAdminsMixin, PermissionRequiredMixin,
+                 CreateViewContext):
+    permission_required = 'workshops.add_membership'
```
narayanaditya95: The existing `membership_create` view has an extra permission required. Is that of any concern?

narayanaditya95:
@gvwilson Please merge [amy#9](https://github.com/gvwilson/amy/pull/9).
@chrismedrela @pbanaszkiewicz Please review the changes I proposed.

gvwilson:
One question, then it's good to merge as far as I can tell.

workshops/templates/workshops/all_memberships.html
chrismedrela: It'd be better to use `url` tag:
```
<a href="{% url 'membership_details' membership.id %}"></a>
```
or something similar.

workshops/templates/workshops/organization.html
chrismedrela: Yes, to keep list views consistent, it'd be better to use glyphicons.

chrismedrela:
> @chrismedrela @pbanaszkiewicz Please review the changes I proposed.

Do you mean reviewing your review? :-)

chrismedrela:
@gvwilson or @pbanaszkiewicz, could you also change `workshops/templates/workshops/all_organizations.html:26` so that it links to `membership_details` instead of `membership_edit`?

workshops/views.py
gvwilson: I don't know - @chrismedrela ?

gvwilson:
I don't understand the request to link to `membership_details` - do you mean `organization_details` (since this is the organization page, and there isn't any reference to membership in it)?

chrismedrela:
> I don't understand the request to link to membership_details - do you mean organization_details (since this is the organization page, and there isn't any reference to membership in it)?

[Here](https://github.com/swcarpentry/amy/blob/develop/workshops/templates/workshops/all_organizations.html#L26) we need to replace `membership_edit` with `membership_details`.

workshops/views.py
chrismedrela: Yes, I think that we should require `change_organization` permission:
```python
    permission_required = [
        'workshops.add_membership',
        'workshops.change_organization'
    ]
```

gvwilson:
Thanks - I was looking in the wrong file *blush*

workshops/templates/workshops/organization.html
chrismedrela: In the first `<tr>`:
```html
<th class="additional-links"></th>
```
For each row:
```html
<td>
  <a href="{% url 'membership_details' membership.pk %}" title="View membership"><span class="glyphicon glyphicon-info-sign"></span></a>
  &nbsp;
  {% if perms.workshops.change_membership %}
    <a href="{% url 'membership_edit' membership.pk %}" title="Edit membership"><span class="glyphicon glyphicon-pencil"></span></a>
  {% endif %}
</td>
```

workshops/views.py
chrismedrela: After looking at other views, I've changed my mind. I don't think we need `change_organization` permission. @pbanaszkiewicz could you confirm that?

workshops/templates/workshops/all_memberships.html
```diff
@@ -19,7 +19,7 @@
         </tr>
     {% for membership in all_memberships %}
         <tr>
-            <td><a href="{{ membership.get_absolute_url }}">{{ membership.organization.fullname }}</a></td>
+            <td><a href="{% url 'membership_details' membership.id %}"></a>{{ membership.organization.fullname }}</a></td>
```
narayanaditya95: `membership.get_absolute_url` does resolve to the view `membership_details` right?

chrismedrela:
@narayanaditya95, could you do the final review?

workshops/models.py
```diff
@@ -119,7 +119,9 @@ class Membership(models.Model):
                              on_delete=models.PROTECT)
 
     def __str__(self):
-        return "{} Membership of <{}>".format(self.variant, str(self.organization))
+        return "{} membership: {} <{}>".format(self.variant.title(),
+                                               self.organization.fullname,
+                                               self.organization.domain)
```
pbanaszkiewicz: You could have left `str(self.organization)` here because you changed `Organization.__str__`.

pbanaszkiewicz:
@chrismedrela in general, changes look good. I have some small comments, but they shouldn't stop you from merging. 

Notes:
* you should not make one big commit instead of many smaller commits. 3e32876901682577503330c21febfb6173e7c47b is a good example of this case; you even described small changes in the commit message – they should become individual commits
* usually you should refrain from making indentation changes to existing code.

chrismedrela:
@pbanaszkiewicz, ok, in the future, I'll create smaller commits and stop correct indentation in templates.

PR #1037, opened 2016-10-03
Search in training applications

### LINKED ISSUE #968: Search must look in applications for training
Search doesn't currently pull up matches in training applications - it should.

pbanaszkiewicz:
This fixes #968 by extending current search capabilities to training applications.

### DIFF
--- a/workshops/models.py
+++ b/workshops/models.py
@@ -2052,7 +2052,16 @@ class TrainingRequest(ActiveMixin, CreatedUpdatedMixin, models.Model):
         return reverse('trainingrequest_details', args=[self.pk])
 
     def __str__(self):
-        return "Training request #{}".format(self.pk)
+        return (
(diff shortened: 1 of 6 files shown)


workshops/models.py
```diff
@@ -2054,6 +2054,18 @@ def clean(self):
...
+            '{state} training request by {personal} {family} <{email}>'
+            .format(
+                state=self.get_state_display(),
+                personal=self.personal,
+                family=self.family,
+                email=self.email,
+                timestamp=self.created_at,
+            )
+        )
+
```
chrismedrela: In search results, it would be better to see something like: 
```
Bob Smith <bob.smith@example.com>
```
or
```
Bob Smith <bob.smith@example.com> - pending
```
so that if you have i.e. 20 matches, it's easy to spot on the right one.

pbanaszkiewicz:
@chrismedrela fixed

PR #1048, opened 2016-10-11
Make database filename customizable via environment variables

chrismedrela:
I like to test AMY on local machine on production data. That means, that I need to change `DATABASES['default']['NAME']` setting very often (i.e. every time I switch branches).

This PR let you overwrite the default (`db.sqlite3`) with whatever you put in `DB_NAME` env var. But only if `DEBUG is True` to prevent surprises on production.

### DIFF
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -13,6 +13,7 @@ import json
 import os
 import sys
 
+from django.core.exceptions import ImproperlyConfigured
 from django.utils.translation import ugettext_lazy as _
 
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
@@ -190,10 +191,20 @@ MESSAGE_TAGS = {
 # Database
 # https://docs.djangoproject.com/en/1.7/ref/settings/#databases
 
+if DEBUG:
+    DB_FILENAME = os.environ.get('AMY_DB_FILENAME', 'db.sqlite3')
+else:
+    try:
+        DB_FILENAME = os.environ['AMY_DB_FILENAME']
+    except KeyError as ex:
+        raise ImproperlyConfigured(
+            'You must specify AMY_DB_FILENAME environment variable '
+            'when DEBUG is False.') from ex
+
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
+        'NAME': os.path.join(BASE_DIR, DB_FILENAME),
         'TEST': {},
     }
 }


gvwilson:
Let's defer this to v1.10 so that we can test more - I realize it's a small thing, but it directly affects production.

chrismedrela:
@gvwilson, how does it affects production? The behavior is changed only when `DEBUG is True`.

gvwilson:
Yeah - so if somebody sets DEBUG to True accidentally on the production 
server, we could be in trouble.  No reason to be skeptical about this in 
particular - I've just had bad experiences in general.
Cheers,
Greg

amy/settings.py
```diff
@@ -175,10 +175,15 @@
 # Database
 # https://docs.djangoproject.com/en/1.7/ref/settings/#databases
 
+if DEBUG:
+    database_name = os.environ.get('DB_NAME', 'db.sqlite3')
+else:
+    database_name = 'db.sqlite3'
```
pbanaszkiewicz: I think it should be:
```python
database_name = 'db.sqlite3'
if DEBUG:
    database_name = os.environ.get('DB_NAME', database_name)
```

narayanaditya95:
Proposal: Extend the PR to use 12-factor inspired environment variables for most settings. See [django-environ](https://github.com/joke2k/django-environ)

chrismedrela:
@narayanaditya95 , what's the advantage of using `django-environ` over simple `env.get('WHATEVER_SETTING', default_value)`?

narayanaditya95:
Except for being less verbose, it can parse other parameters from the env variables as well.
Eg `DATABASE_URL=psql://urser:un-githubbedpassword@127.0.0.1:8458/database` parses the host, port, username and password.

chrismedrela:
@narayanaditya95 , IMO we can switch to `django-environ` when we need its unique features. Right now, doing it means that we add one layer of abstraction more without any advantages.

amy/settings.py
pbanaszkiewicz: We should keep the UPPERCASE style for that var name.

chrismedrela:
@pbanaszkiewicz what about changing the current approach in this PR and:

- make `DB_NAME` envvar required
- use it for both deploy (`DEBUG == True`) and production (`DEBUG == False`).

So that it's closer to [12-factor app](https://12factor.net/config). The main drawback is that we need to change Travis configuration and deployment procedure, but once we figure it out, we can seamlessly add more envvars (like `ENABLE_PYDATA` and others), which is inevitable if we want to have only one `settings.py` file and one branch (that is, no `pydata-deploy` branch).

pbanaszkiewicz:
@chrismedrela:
> @pbanaszkiewicz what about changing the current approach in this PR and:
> 
> - make `DB_NAME` envvar required
> - use it for both deploy (`DEBUG == True`) and production (`DEBUG == False`).
> 
> So that it's closer to [12-factor app](https://12factor.net/config). The main drawback is that we need to > change Travis configuration and deployment procedure, but once we figure it out, we can seamlessly > add more envvars (like `ENABLE_PYDATA` and others), which is inevitable if we want to have only one > `settings.py` file and one branch (that is, no `pydata-deploy` branch).

:+1: for `AMY_DB_NAME` envvar required for deployment, :+1: for `AMY_DB_NAME` **optional** for development.

Making it required for development is just pain in the butt, and doesn't have any real advantages.

Also please note that:

1. we are using other envvars already
2. they have `AMY_` prefix

chrismedrela:
ping @pbanaszkiewicz

amy/settings.py
```diff
@@ -190,10 +191,20 @@
...
 
+if DEBUG:
+    DB_FILENAME = os.environ.get('AMY_DB_FILENAME', 'db.sqlite3')
+else:
+    try:
+        DB_FILENAME = os.environ['AMY_DB_FILENAME']
+    except KeyError as ex:
+        raise ImproperlyConfigured(
+            'You must specify AMY_DB_FILENAME environment variable '
+            'when DEBUG is False.') from ex
```
pbanaszkiewicz: Sorry, I rarely see "raise ex1 from ex2", can you please explain to me in this thread what it does?

chrismedrela: It's an explicit way to chain exceptions, so that traceback are printed in this way:
```
>>> try:
...     print(1 / 0)
... except Exception as exc:
...     raise RuntimeError("Something bad happened") from exc
...
  File "<stdin>", line 2, in <module>
ZeroDivisionError: int division or modulo by zero

The above exception was the direct cause of the following exception:

  File "<stdin>", line 4, in <module>
RuntimeError: Something bad happened
```

However, it doesn't make much difference. If you omit `from ex`, both tracebacks are still printed, except that the message between them is a bit different:

```
>>> try:
...     print(1 / 0)
... except:
...     raise RuntimeError("Something bad happened")
...
  File "<stdin>", line 2, in <module>
ZeroDivisionError: int division or modulo by zero

During handling of the above exception, another exception occurred:

  File "<stdin>", line 4, in <module>
RuntimeError: Something bad happened
```

PR #1060, opened 2016-10-23
Fix #782 -- use hyperlinks in report on all activity

### LINKED ISSUE #782: Report on all activity should use hyperlinks for workshops
On https://amy.software-carpentry.org/workshops/reports/all_activity_over_time, the references to workshops should be hyperlinks, not just plain-text slugs.

chrismedrela:
Fixes #782.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -386,6 +386,14 @@ class ReportsViewSet(ViewSet):
             start=request.query_params.get('start', None),
             end=request.query_params.get('end', None))
 
+        data = self.get_all_activity_over_time(start, end)
+        data['missing']['attendance'] = self.listify(
+            data['missing']['attendance'], request, format=format)
+        data['missing']['instructors'] = self.listify(
+            data['missing']['instructors'], request, format=format)
+        return Response(data)
+
+    def get_all_activity_over_time(self, start, end):
         events_qs = Event.objects.filter(start__gte=start, start__lte=end)
         swc_tag = Tag.objects.get(name='SWC')
         dc_tag = Tag.objects.get(name='DC')
@@ -422,10 +430,6 @@ class ReportsViewSet(ViewSet):
         dc_total_learners = dc_total_learners['count']
 
         # Workshops missing any of this data.
-        # There's no point in using hyperlinks here, because it would:
-        # a) require using reverse() unless we somehow managed to switch to
-        #    serializer for this view
-        # b) make JS part even harder (what is available right now just works)
         missing_attendance = events_qs.filter(attendance=None) \
                                       .values_list('slug', flat=True)
         missing_instructors = events_qs.annotate(
@@ -438,16 +442,26 @@ class ReportsViewSet(ViewSet):
             )
         ).filter(instructors=0).values_list('slug', flat=True)
 
-        return Response({
+        return {
             'start': start,
             'end': end,
             'workshops': {
                 'SWC': swc_workshops.count(),
                 'DC': dc_workshops.count(),
+                # This dictionary is traversed in a template where we cannot
+                # write "{{ data.workshops.SWC,DC }}", because commas are
+                # disallowed in templates. Therefore, we include
+                # swc_dc_workshops twice, under two different keys:
+                # - 'SWC,DC' - for backward compatibility,
+                # - 'SWC_or_DC' - so that you can access it in a template.
                 'SWC,DC': swc_dc_workshops,
+                'SWC_or_DC': swc_dc_workshops,
                 'WiSE': wise_workshops,
                 'TTT': ttt_workshops,
+                # We include self_organized_workshops twice, under two
+                # different keys, for the same reason as swc_dc_workshops.
                 'self-organized': self_organized_workshops,
+                'self_organized': self_organized_workshops,
             },
             'instructors': {
                 'SWC': {
@@ -464,14 +478,10 @@ class ReportsViewSet(ViewSet):
                 'DC': dc_total_learners,
             },
             'missing': {
-                # qs.values_list returns an iterator, so we need to listify it
-                # for YAML
-                'attendance': self.listify(missing_attendance, request,
-                                           format),
-                'instructors': self.listify(missing_instructors, request,
-                                            format),
+                'attendance': missing_attendance,
+                'instructors': missing_instructors,
             }
-        })
+        }
 
     def instructors_by_time_queryset(self, start, end):
         """Just a queryset to be reused in other view."""
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -1500,3 +1500,16 @@ class SendHomeworkForm(forms.ModelForm):
         fields = [
             'url',
         ]
+
+
+class AllActivityOverTimeForm(forms.Form):
+    start = forms.DateField(
+        label='Begin date as YYYY-MM-DD',
+        input_formats=['%Y-%m-%d', ],
+    )
+    end = forms.DateField(
+        label='End date as YYYY-MD-DD',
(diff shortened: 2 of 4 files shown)


api/views.py
```diff
@@ -266,16 +267,17 @@ def _only_latest_date(self, iterable):
...
-        format_ = (request.query_params.get('format') or format or '').lower()
+        if request is not None:
+            format_ = (request.query_params.get('format') or format or '').lower()
 
-        if format_ in self.formats_requiring_lists:
-            # list-ify the generator for renderers requiring lists
-            return list(iterable)
+            if format_ in self.formats_requiring_lists:
+                # list-ify the generator for renderers requiring lists
+                return list(iterable)
```
chrismedrela: @pbanaszkiewicz , could we always listify, that is, always return `list(iterable)`?

api/views.py
```diff
@@ -430,16 +439,16 @@ def all_activity_over_time(self, request, format=None):
...
 
-        return Response({
+        return {
             'start': start,
             'end': end,
             'workshops': {
                 'SWC': swc_workshops.count(),
                 'DC': dc_workshops.count(),
-                'SWC,DC': swc_dc_workshops,
+                'SWC_or_DC': swc_dc_workshops,
```
chrismedrela: @pbanaszkiewicz, We change the API here. Is it use anywhere?

api/views.py
pbanaszkiewicz: No, this is not a good solution performance-wise.

api/views.py
```diff
@@ -268,16 +269,17 @@ def _only_latest_date(self, iterable):
...
             yield next_
 
-    def listify(self, iterable, request, format=None):
+    def listify(self, iterable, request=None, format=None):
         """Some renderers require lists instead of any iterables for rendering.
         This function conditionally turns iterables into lists based on the
         format requested by browser."""
         # choose either '?format=...' or '/url.format' or None
-        format_ = (request.query_params.get('format') or format or '').lower()
+        if request is not None:
```
pbanaszkiewicz: Why is this here?

api/views.py
pbanaszkiewicz: It was used here: https://github.com/swcarpentry/amy/pull/1060/commits/6239fa34d6fb0c4798a5ac2831a5040623493ffe#diff-345991cbd83965baed2fa14d0e2f37ceL106

pbanaszkiewicz:
:-1: because:
* missing explanation of changes
* actual changes make code harder to follow
* changes in data structure

api/views.py
chrismedrela: You're right, this change is unnecessary.

pbanaszkiewicz: Sorry, I probably sounded rude, but I actually meant to ask what was this code doing and why, because you hadn't included any explanation.

chrismedrela:
> changes in data structure

Do you mean changes in the schema of returned JSON?

> actual changes make code harder to follow

I completely agree, however I don't know how to make the code easier to follow. Any ideas?

pbanaszkiewicz:
> Do you mean changes in the schema of returned JSON?

Yes.

> I completely agree, however I don't know how to make the code easier to follow. Any ideas?

I don't know what you're achieving with this code (== "I don't understand it"), so I don't know if we can do better.

api/views.py
chrismedrela: But this change *really* was unnecessary, so I've reverted it. :-)

chrismedrela:
> I don't know what you're achieving with this code (== "I don't understand it"), so I don't know if we can do better.

So far, what we did is we returned HTML page without data with a bit of JS that asks API for the data and, then, renders the data.

Now, everything is rendered on server. When triggering `all_activity_over_time` view, server asks API for the data. However, we don't make a request, we just call `ReportsViewSet.get_all_activity_over_time` to get the data as Python dictionary without network overhead. Then, we pass the dictionary to `all_activity_over_time.html` template where it's traversed. We had to change data structure, because django templates don't allow special characters when traversing the structure (i.e., you cannot write `{{ data.workshops['SWC,DC'] }}`).

Yes, I should have included a description like the above one in the commit message. Sometimes I forget that other people don't have access to my mind. :-)

Is there anything else you want me to clarify?

pbanaszkiewicz:
> Is there anything else you want me to clarify?

Now it's clear to me, thank you. I'll get back to you about the solution soon.

pbanaszkiewicz:
Okay, so here's I think the sanest solution I came up with so far.

1) Drop `listify` from `get_all_activity_over_time` and make it only return "pure" dict:

```python
    def get_all_activity_over_time(self, start, end):
        events_qs = Event.objects.filter(start__gte=start, start__lte=end)
        …

        # Workshops missing any of this data.
        # --- REMOVING COMMENT ABOUT JS FROM HERE ---
        missing_attendance = …

        return {
            'start': start,
            'end': end,
            'workshops': {
                'SWC': swc_workshops.count(),
                'DC': dc_workshops.count(),
                'SWC_or_DC': swc_dc_workshops,
                'WiSE': wise_workshops,
                'TTT': ttt_workshops,
                'self_organized': self_organized_workshops,
            },
            'instructors': {
                'SWC': {
                    'total': swc_total_instr,
                    'unique': swc_unique_instr,
                },
                'DC': {
                    'total': dc_total_instr,
                    'unique': dc_unique_instr,
                },
            },
            'learners': {
                'SWC': swc_total_learners,
                'DC': dc_total_learners,
            },
            'missing': {
                'attendance': missing_attendance,
                'instructors': missing_instructors,
            }
        }
```

2) use `listify` in `all_activity_over_time` view on `get_all_activity_over_time` output:

```python
    @list_route(methods=['GET'])
    def all_activity_over_time(self, request, format=None):
        """Workshops, instructors, and missing data in specific periods."""
        start, end = self._default_start_end_dates(
            start=request.query_params.get('start', None),
            end=request.query_params.get('end', None))

        data = self.get_all_activity_over_time(start, end)
        data['missing']['attendance'] = self.listify(…)
        data['missing']['instructors'] = self.listify(…)
        return Response(data, …?)
```

3) Instead of changing the original structure, extend it: (in `get_all_activity_over_time`)

```python
            'workshops': {
                'SWC': swc_workshops.count(),
                'DC': dc_workshops.count(),
                'SWC_or_DC': swc_dc_workshops,
                'SWC,DC': swc_dc_workshops,  # for backwards-compatibility
                'WiSE': wise_workshops,
                'TTT': ttt_workshops,
                'self_organized': self_organized_workshops,
                'self-organized': self_organized_workshops,  # for backwards-compatibility
            },
```

Now we should not have any problems with API structure, and we don't need any partials, and we keep your design (ie. server-side rendering).

pbanaszkiewicz:
@chrismedrela is this ready?

chrismedrela:
@pbanaszkiewicz yes.

api/views.py
```diff
@@ -438,16 +442,18 @@ def all_activity_over_time(self, request, format=None):
...
             'end': end,
             'workshops': {
                 'SWC': swc_workshops.count(),
                 'DC': dc_workshops.count(),
                 'SWC,DC': swc_dc_workshops,
+                'SWC_or_DC': swc_dc_workshops,
                 'WiSE': wise_workshops,
                 'TTT': ttt_workshops,
                 'self-organized': self_organized_workshops,
+                'self_organized': self_organized_workshops,
```
pbanaszkiewicz: One small final bit: can you please add a comment here explaining why we need `SWC,DC` and `SWC_or_DC` (same for self organized)?

chrismedrela: @pbanaszkiewicz ping

PR #1065, opened 2016-10-27
Rewrite the query to fetch an organization.

### LINKED ISSUE #964: Bulk import workflow encounters IntegrityError when saving an organization
Currently, we allow organizations with the domain that contains the `www` subdomain. For eg: Google can exist as `www.google.com` as well as `google.com`, leading to `IntegrityError` while saving the first while the second exists.

Shouldn't we enforce one URL pattern and trim/add `www` to the `domain` field when saving an organization?

Testcase:

``` py
In [5]: Organization.objects.create(fullname='Google', domain='google.com')
Out[5]: <Organization: google.com>

In [6]: Organization.objects.create(fullname='Google', domain='www.google.com')
---------------------------------------------------------------------------
IntegrityError                            Traceback (most recent call last)
```

narayanaditya95:
Prevents IntegrityError while trying to save an organization
with conflicting fullname but different domains.
Fixes #964

### DIFF
--- a/pydata/api.py
+++ b/pydata/api.py
@@ -4,6 +4,7 @@ from urllib.parse import urljoin, urlparse
 
 import requests
 from django.conf import settings
+from django.db.models import Q
 
 from workshops.models import (
     Person,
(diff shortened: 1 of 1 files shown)


pydata/api.py
```diff
@@ -118,14 +119,19 @@ class SponsorshipAPIClient(BaseAPIClient):
     model = Sponsorship
 
     def parse(self, sponsor):
+        domain = urlparse(sponsor['external_url']).netloc
+        try:
+            organization = Organization.objects.filter(
+                Q(fullname=sponsor['name']) | Q(domain=domain)
+            ).first()
```
chrismedrela: `.first()` returns the first object matched by the queryset, **or None if there is no matching object**.

chrismedrela:
@narayanaditya95 , does this PR really fixes #964, that is, can we close #964 when we merge this PR?

narayanaditya95:
@chrismedrela Yes, this PR fixes #964.

I will be creating a new PR for tests as it involves a change to `travis.yml`.

chrismedrela:
> I will be creating a new PR for tests as it involves a change to travis.yml.

Why is a change to `travis.yml` necessary?

narayanaditya95:
@chrismedrela 
The test suite tests the project without adding `pydata` to the list of installed apps. I will create a new job that makes changes to the codebase necessary for including the `pydata` app within the project and then run the test suite. Also, it would involve marking certain tests as expected failures when run with `pydata` installed.

So, the current plan of action is to create three jobs on travis:
1. Runs the test suite without making any changes.
2. Merges the branch `pydata-deploy` and then runs the test suite.
3. Runs the check for missing migrations.

PR #1067, opened 2016-10-31
First class membership objects extended

narayanaditya95:
This PR builds upon @gvwilson's PR.


chrismedrela:
@narayanaditya95 , I'm sorry, I've started working at #1034 and forgot about this PR. I think that I've included most of your changes there, but could you have a look at #1034 and do a review? In the meantime, I'm closing this one, so that we all work on #1034.

PR #1071, opened 2016-11-02
Querying historical data with parameters

### LINKED ISSUE #979: Querying historical data with parameters
Looking in `workshops/views.py` at `instructors_over_time`, it looks like there's no way to pass a particular badge or set of badges to the view.  I need to be able to query with something like `...url?badge=swc` to get instructors of a particular kind, both in the JSON/CSV/whatever data dump and in the graph. I will also be wanting to query workshops over time and learners over time with specific tags so that I can view SWC or DC or TTT workshops over time, participants in instructor training over time, etc.

chrismedrela:
Fixes #979.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -58,6 +58,7 @@ from .filters import (
     TaskFilter,
     PersonFilter,
     InstructorsOverTimeFilter,
+    WorkshopsOverTimeFilter,
 )
 
 
@@ -290,7 +291,10 @@ class ReportsViewSet(ViewSet):
     def workshops_over_time(self, request, format=None):
         """Cumulative number of workshops run by Software Carpentry over
         time."""
-        qs = self.queryset1.annotate(count=Count('id'))
+        qs = self.queryset1
+        qs = WorkshopsOverTimeFilter(request.GET, queryset=qs).qs
+        qs = qs.annotate(count=Count('id'))
+
         serializer = WorkshopsOverTimeSerializer(qs, many=True)
 
         # run a cumulative generator over the data
--- a/api/views.py
+++ b/api/views.py
@@ -53,7 +53,12 @@ from .serializers import (
     PersonSerializer,
 )
 
-from .filters import EventFilter, TaskFilter, PersonFilter
+from .filters import (
+    EventFilter,
+    TaskFilter,
+    PersonFilter,
+    InstructorsOverTimeFilter,
+)
 
 
 class IsAdmin(BasePermission):
@@ -317,11 +322,12 @@ class ReportsViewSet(ViewSet):
 
         badges = Badge.objects.instructor_badges()
 
-        qs = Person.objects.filter(badges__in=badges)  \
-                           .annotate(date=Min('award__awarded'),
-                                     count=Value(1,
-                                                 output_field=IntegerField())) \
-                           .order_by('date')
+        qs = Person.objects.filter(badges__in=badges)
+        filter = InstructorsOverTimeFilter(request.GET, queryset=qs)
+        qs = filter.qs.annotate(date=Min('award__awarded'),
+                                count=Value(1,
+                                            output_field=IntegerField())) \
+                      .order_by('date')
 
         serializer = InstructorsOverTimeSerializer(qs, many=True)
 
--- a/api/views.py
+++ b/api/views.py
@@ -59,6 +59,7 @@ from .filters import (
     PersonFilter,
     InstructorsOverTimeFilter,
     WorkshopsOverTimeFilter,
+    LearnersOverTimeFilter,
 )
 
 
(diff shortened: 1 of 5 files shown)


api/views.py
```diff
@@ -285,7 +292,10 @@ def listify(self, iterable, request, format=None):
     def workshops_over_time(self, request, format=None):
         """Cumulative number of workshops run by Software Carpentry over
```
gvwilson: Or Data Carpentry? What if I want those?

api/views.py
```diff
@@ -285,7 +292,10 @@ def listify(self, iterable, request, format=None):
...
         time."""
-        qs = self.queryset1.annotate(count=Count('id'))
+        qs = self.queryset1
```
gvwilson: `queryset1` is a pretty meaningless variable name...

api/views.py
```diff
@@ -299,7 +309,10 @@ def workshops_over_time(self, request, format=None):
     def learners_over_time(self, request, format=None):
         """Cumulative number of learners attending Software-Carpentry workshops
```
gvwilson: 1. "Software Carpentry" isn't hyphenated.
2. What about Data Carpentry? Library Carpentry? etc.

workshops/templates/workshops/time_series.html
```diff
@@ -17,7 +17,7 @@
     <script src="{% static 'metrics-graphics/dist/metricsgraphics.js' %}"></script>
 
     <script type="text/javascript">
-        d3.json('{{ api_endpoint }}?format=json', function(data) {
+        d3.json('{{ api_endpoint }}&format=json', function(data) {
```
gvwilson: Does this imply that the `api_endpoint` variable now includes the start of the query parameter string?

workshops/views.py
```diff
@@ -2185,8 +2198,11 @@ def export_members(request):
 @admin_required
 def workshops_over_time(request):
     '''Export JSON of count of workshops vs. time.'''
+    endpoint = '{}?{}'.format(reverse('api:reports-workshops-over-time'),
```
gvwilson: OK, this answers the question above. But isn't there something in the library to construct the URL with query parameters, rather than us using string formatting?

workshops/views.py
```diff
@@ -2195,8 +2211,11 @@ def workshops_over_time(request):
 @admin_required
 def learners_over_time(request):
     '''Export JSON of count of learners vs. time.'''
+    endpoint = '{}?{}'.format(reverse('api:reports-learners-over-time'),
```
gvwilson: As above.

workshops/views.py
```diff
@@ -2205,8 +2224,11 @@ def learners_over_time(request):
 @admin_required
 def instructors_over_time(request):
     '''Export JSON of count of instructors vs. time.'''
+    endpoint = '{}?{}'.format(reverse('api:reports-instructors-over-time'),
```
gvwilson: As above.

api/views.py
chrismedrela: This is an outdated docstring. I'll update it.

workshops/templates/workshops/time_series.html
chrismedrela: Yes.

workshops/views.py
chrismedrela: I can't find anything like that. Stack Overflow suggests string formatting.

chrismedrela:
@gvwilson, updated.

@pbanaszkiewicz please merge if you like the design. The design is that we display filter forms (`AMYFilterSet` children) in the sidebar. When the form is submitted, we pass GET parameters to API where the same `AMYFilterSet`s are reused. The issues with this PR are:
- Where should `InstructorOverTimeFilter` (and other filters) live? It's used in both `workshops` and `api` apps.
- If you visit [this API](https://amy.software-carpentry.org/api/v1/events/), you can filter events in your browser using the modal form (just click "filter" button). However, we don't display similar form for reports API. Any ideas how we could do this?

api/views.py
```diff
@@ -317,11 +322,12 @@ def instructors_over_time(self, request, format=None):
...
-                           .annotate(date=Min('award__awarded'),
-                                     count=Value(1,
-                                                 output_field=IntegerField())) \
-                           .order_by('date')
+        qs = Person.objects.filter(badges__in=badges)
+        filter = InstructorsOverTimeFilter(request.GET, queryset=qs)
+        qs = filter.qs.annotate(date=Min('award__awarded'),
+                                count=Value(1,
+                                            output_field=IntegerField())) \
+                      .order_by('date')
```
pbanaszkiewicz: Minor note: this looks strange. I think you can write the same code in a little bit more appealing way:

```python
    filter = InstructorsOverTimeFilter(
        request.GET,
        queryset=Person.objects.filter(badges__in=badges),
    )
    qs = filter.qs.annotate(
        date=Min('award__awarded'),
        count=Value(1, output_field=IntegerField()),
    ).order_by('date')
```

pbanaszkiewicz:
> If you visit this API, you can filter events in your browser using the modal form (just click "filter" button). However, we don't display similar form for reports API. Any ideas how we could do this?

@chrismedrela there doesn't seem to be any way to do this with the `ReportsViewSet` view set we have.

PR #1072, opened 2016-11-02
Views refactoring + one bug fix

chrismedrela:
- Bug fix in `all_trainings`: The number of trainees who became SWC or DC instructor was overestimated due to a bug in a query.
- Switching to class-based views in case of simple `ListView`s and some simple `DetailView`s.
- Adopting new name convention (`AMYWhateverView`) for our generic views.
- Moving generic views from `views.py` to `base_views.py`.

@pbanaszkiewicz what do you think about these changes?

### DIFF
--- a/workshops/base_views.py
+++ b/workshops/base_views.py
@@ -123,7 +123,7 @@ class AMYDeleteView(DeleteView):
         return self.http_method_not_allowed(request, *args, **kwargs)
 
 
-class FormViewContext(FormView):
+class AMYFormView(FormView):
     """
     Class-based view to allow displaying of forms with bootstrap form helper.
     """
--- a/workshops/base_views.py
+++ b/workshops/base_views.py
@@ -128,7 +128,7 @@ class FormViewContext(FormView):
         return context
 
 
-class FilteredListView(ListView):
+class AMYListView(ListView):
     paginator_class = Paginator
     filter_class = None
     queryset = None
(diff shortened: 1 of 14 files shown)


workshops/base_views.py
```diff
@@ -0,0 +1,208 @@
...
+            return HttpResponseRedirect(success_url)
+        except ProtectedError as e:
+            return failed_to_delete(self.request, self.object,
+                                    e.protected_objects)
+
+    def get(self, request, *args, **kwargs):
+        return self.http_method_not_allowed(request, *args, **kwargs)
+
+
+class FormViewContext(FormView):
```
narayanaditya95: `FormViewContext` -> `AMYFormView`?

narayanaditya95:
@chrismedrela Can you make the corresponding changes in `pydata/views.py` in this PR?

chrismedrela:
@narayanaditya95 what changes do you mean?

pbanaszkiewicz:
@chrismedrela:
> @narayanaditya95 what changes do you mean?

Aditya suggested renaming `FormViewContext` to `AMYFormView` (seems reasonable). He also asked if we should make `views` a package, but I think it's a little bit too much work - we could do this for v1.10, though.

chrismedrela:
> Aditya suggested renaming FormViewContext to AMYFormView (seems reasonable).
OK

So now it's ready. I've added a few more commits, mainly converting some detailed view to class-based views.

pbanaszkiewicz:
Thanks @chrismedrela

Issue #1066, opened 2016-10-29
Configure Travis so that it tests merged changes


chrismedrela:
Right now, if somebody proposes merging `my-patch` branch, Travis runs tests on this branch. Instead, we should first merge the branch with `develop` and then run tests. If both branches have new migrations, running tests without merging doesn't detect conflicting migrations.

narayanaditya95:
Suggested a fix at #1073

narayanaditya95:
As @pbaneswicz pointed out at #1073, Travis does merge the develop branch into the branch contained in the PR.
However, our problem is somewhat different. Although a PR can be marked safe by Travis today (after merging from develop), a change to develop tomorrow that does not cause a merge conflict with the PR may lead to a migration clash or other test failures. I think the solution to this issue is to somehow rerun the Travis test suite just before merge. Does the web interface allow us to do that?

chrismedrela:
> Yes, it does. So, we must manually restart the test suite for a PR after every change to the develop branch to ensure that the tests always pass.

How can we do that?

narayanaditya95:
At the PR page, use the following workflow:

1. Click of "Show all checks"
2. Click on "Details" to the right of the Travis test list option.
3. At the build page of Travis, click on "Restart build".

chrismedrela:
@narayanaditya95 thank you very much for instructions. I'm closing it, because we found the solution: we need to manually restart build before merging.

Issue #938, opened 2016-08-05
Splitting amy.settings


narayanaditya95:
Currently, we use a single settings file.
How about we divide `amy.settings.py` into a hierarchical structure to allow us to easily override settings according to different environments.

```
settings/
    base.py
    local.py
    prod.py
    pydata.py
```

pbanaszkiewicz:
-1 (unless you have other reasons) because the preferred way is to vary params via environment variables.

narayanaditya95:
There exist a lot of conditionals in `settings.py` of the form:

```py
if DEBUG:
     # override something
```

We can move these overrides to `local.py` as:

`local.py`
```py
from .base import *

# override something
```

chrismedrela:
I'm :+1: for having only one `settings.py` file and overriding params via environment variables. This is also recommended by [The Twelve-Factor App](https://12factor.net/config).

PR #1073, opened 2016-11-04
Test pydata application

narayanaditya95:
PR created to test whether Travis creates multiple jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,33 @@
 sudo: false
+
 language: python
+
 cache: pip 
+
 python:
   - 3.5
+
+env:
+  - AMY_ENABLE_PYDATA=true
+    AMY_PYDATA_USERNAME=username
+    AMY_PYDATA_PASSWORD=password
+  - CHECK_MIGRATION=true
+
 install:
   - pip install -r requirements.txt
   - pip install coveralls
+
+before_script:
+  - if [[ $CHECK_MIGRATION == true ]]; then
+      python manage.py makemigrations --dry-run -e;
+      export STATUS_CODE=$?;
+      if [[ "$STATUS_CODE" == "0" ]]; then
+        exit 1;
+      fi;
+    fi;
+
 script:
-  - python manage.py makemigrations --dry-run -e; export STATUS_CODE=$?; if [[ "$STATUS_CODE" == "0" ]]; then exit 1; fi
   - coverage run --source=amy,api,workshops manage.py test
+
 after_success:
   coveralls
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -33,6 +33,13 @@ else:
     SECRET_KEY = None
 SECRET_KEY = os.environ.get('AMY_SECRET_KEY', SECRET_KEY)
 
+# settings for PyData application
+ENABLE_PYDATA = json.loads(os.environ.get('AMY_ENABLE_PYDATA', 'false'))
+
+if ENABLE_PYDATA:
+    PYDATA_USERNAME_SECRET = os.environ.get('AMY_PYDATA_USERNAME')
+    PYDATA_PASSWORD_SECRET = os.environ.get('AMY_PYDATA_PASSWORD')
+
 # be sure to put these values in your envvars, even for development
 RECAPTCHA_PUBLIC_KEY = os.environ.get('AMY_RECAPTCHA_PUBLIC_KEY', None)
 RECAPTCHA_PRIVATE_KEY = os.environ.get('AMY_RECAPTCHA_PRIVATE_KEY', None)
@@ -122,12 +129,18 @@ if DEBUG:
 
 # Application definition
 
-INSTALLED_APPS = (
+INSTALLED_APPS = [
     'django.contrib.auth',
     'django.contrib.contenttypes',
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+]
+if ENABLE_PYDATA:
+    INSTALLED_APPS += [
+        'pydata',
+    ]
+INSTALLED_APPS += [
     'workshops',
     # this should be after 'workshops' because templates in
     # 'templates/registration/' clash
@@ -144,7 +157,7 @@ INSTALLED_APPS = (
     'compressor',
     'social.apps.django_app.default',
     'debug_toolbar',
-)
+]
 
 CRISPY_TEMPLATE_PACK = 'bootstrap3'
 
--- a/amy/urls.py
+++ b/amy/urls.py
@@ -5,6 +5,12 @@ from django.contrib.auth import views as auth_views
 
 urlpatterns = [
     url(r'^workshops/admin/', include(admin.site.urls)),
+]
+if settings.ENABLE_PYDATA:
+    urlpatterns += [
+        url(r'^workshops/', include('pydata.urls')),
+    ]
+urlpatterns += [
     url(r'^workshops/', include('workshops.urls')),
     # url(r'^account/', include('django.contrib.auth.urls')),
 
--- a/workshops/test/test_sponsorship.py
+++ b/workshops/test/test_sponsorship.py
@@ -72,7 +72,7 @@ class TestSponsorshipViews(TestBase):
         self.sponsorship = Sponsorship.objects.create(
             organization=self.org_alpha,
             event=self.event,
-            amount=100.00,
+            amount=1500,
         )
         self._setUpUsersAndLogin()
 
@@ -116,7 +116,7 @@ class TestSponsorshipViews(TestBase):
         payload = {
             'sponsor-organization': self.org_beta.pk,
             'sponsor-event': self.event.pk,
-            'sponsor-amount': 100.00,
+            'sponsor-amount': 1500,
         }
         response = self.client.post(
             reverse('event_edit', kwargs={'slug': self.event.slug}),
@@ -132,7 +132,7 @@ class TestSponsorshipViews(TestBase):
         self.assertTrue(response.context['object'].sponsorship_set.all())
         self.assertEqual(response.context['object'].sponsors.count(), 2)
         self.assertIn(
-            100.00,
+            1500,
             response.context['object'].sponsorship_set.values_list('amount', flat=True)
         )
 
@@ -142,6 +142,7 @@ class TestSponsorshipViews(TestBase):
             'sponsor-organization': self.org_beta.pk,
             'sponsor-event': self.event.pk,
             'sponsor-contact': self.harry.pk,
+            'sponsor-amount': 1500,
         }
         response = self.client.post(
             reverse('event_edit', kwargs={'slug': self.event.slug}),
--- a/docs/PyData_deploy.md
+++ b/docs/PyData_deploy.md
@@ -1,40 +1,17 @@
 # Deploy AMY for PyData
 
- -  Add `pydata` to `INSTALLED_APPS`. Ensure that the `pydata` app is listed 
-    before `workshops` app.
+ -  Export `AMY_ENABLE_PYDATA` environment variable.
 
-    ```py
-    INSTALLED_APPS = (
-        ...
-        'pydata',
-        'workshops',
-        ...
-    )
+    ```sh
+    export AMY_ENABLE_PYDATA=true
     ```
 
- -  Include `pydata.urls` in the `urlpatterns` of `amy.urls` before `workshops.url`.
+ -  Export the username and password of a superuser of the conference site 
+    as environment variables.
 
-    ```py
-    urlpatterns = [
-        ...
-        url(r'^workshops/', include('pydata.urls')),
-        url(r'^workshops/', include('workshops.urls')),
-        ...
-    ]
-    ```
-
- -  Add the username and password of a superuser of the conference site to `amy/settings.py`.
-
-    ```py
-    PYDATA_USERNAME_SECRET = 'username'
-    PYDATA_PASSWORD_SECRET = 'password'
-    ```
-
-    You can also fetch them from the environment variables.
-
-    ```py
-    PYDATA_USERNAME_SECRET = os.environ.get('PYDATA_USERNAME_SECRET')
-    PYDATA_PASSWORD_SECRET = os.environ.get('PYDATA_PASSWORD_SECRET')
+    ```sh
+    export AMY_PYDATA_USERNAME = "username"
+    export AMY_PYDATA_PASSWORD = "password"
     ```
 
  -  Install fixtures from the `pydata/fixtures/` directory.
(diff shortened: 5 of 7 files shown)


.travis.yml
```diff
@@ -1,13 +1,38 @@
...
+
 language: python
+
 cache: pip 
+
 python:
   - 3.5
+
+env:
+  - PYDATA=true
```
narayanaditya95: PyData app introduces a migration since it adds a `choice` argument to the sponsorship amount field. So, I do not check for missing migrations here.

narayanaditya95:
@pbanaszkiewicz @chrismedrela 
I have added another job on Travis that runs the test suite while the `pydata` app is deployed.
Please review.

.travis.yml
```diff
@@ -1,13 +1,41 @@
...
+  - PYDATA=false CHECK_MIGRATION=true
+
+before_install:
+  - git config --global user.email "you@example.com"
+  - git config --global user.name "Your Name"
+  - if [[ $TRAVIS_PULL_REQUEST != false ]]; then
+      git pull -q --no-edit origin develop;
+    fi;
+  - if [[ $PYDATA == true ]]; then
+      git pull -q --no-edit origin deploy-pydata;
```
pbanaszkiewicz: What exactly is in `deploy-pydata`?

Are you sure `git pull` is the right way to go?

Can we use `git config --local` instead of `git config --global`?

narayanaditya95: >What exactly is in deploy-pydata?

Adds `pydata` to INSTALLED_APPS and urlconfig. See the commit [cf8603aa](https://github.com/swcarpentry/amy/commit/cf8603aa5f6cc8fc33a829735e418ddb0e6b42dc)

>Are you sure git pull is the right way to go?

Isn't `pull` an aggregate command that executes `fetch` and `merge`. Shouldn't be a problem.

>Can we use git config --local instead of git config --global?

The instances destruct after every test. It does not make a difference.

pbanaszkiewicz: > Adds pydata to INSTALLED_APPS and urlconfig. See the commit cf8603aa

Why not conditionally include it in the config? It could be controlled via envvar.

> Isn't pull an aggregate command that executes fetch and merge. Shouldn't be a problem.

Hmm well I can see some issues and questions here.

You're assuming that someone wants to merge into develop, which is not always true.

You're merging develop into PR branch (this may be the same as merging PR branch into develop - I'm not sure).

What happens when you get merge conflict. The test suite doesn't run?

And finally, according to the travis-ci [docs](https://docs.travis-ci.com/user/pull-requests):

> Rather than test the commits that have been pushed to the branch the pull request is from, we test the merge between the origin and the upstream branch. To only build on push events, you can disable Build on Pull Requests from your repository settings.

So… is this what @chrismedrela intended in #1066?

narayanaditya95: >Why not conditionally include it in the config? It could be controlled via envvar.

Nice suggestion. I am starting to reconsider #938. By overriding the envvar `DJANGO_SETTINGS_MODULE`, we can switch between a normal deployment and a PyData deployment.

>So… is this what @chrismedrela intended in #1066?

Please refer to the discussion at #1066. As you pointed out, Travis does merge the develop branch before testing.

narayanaditya95: I have removed the fix for #1066. Also, I am moving the discussion out of this comment thread.

narayanaditya95:
@pbanaszkiewicz 

>Why not conditionally include it in the config? It could be controlled via envvar.

Nice suggestion. I am starting to reconsider #938. By overriding the envvar DJANGO_SETTINGS_MODULE, we can switch between a normal deployment and a PyData deployment.

.travis.yml
chrismedrela: Where is that migration? I can't find it.

chrismedrela:
My general question is: why can't we have just one job on Travis with PYDATA=true? Do I understand correctly that:
- This job runs tests of *all* apps.
- If AMY works with pydata enabled, it also works with pydata disabled.

Having one job makes testing and deployment much easier - we don't need `deploy-pydata` branch, separate Travis jobs etc.

chrismedrela:
OK, I understand why we need two jobs now. However, I think we need to address a much more fundamental issue - the way we monkey patch everything in `pydata/apps.py` scares me, especially that we monkey patch models. It's not going to scale at all. 

I think what we need to do is to reuse the same `Sponsorship` model for both standard and pydata deployment. This can be done by converting it from an enum to a ForeignKey to a new model `SponsorshipOption`, so that the choices can be dynamically changed by altering the records of `SponsorshipOption` table.

If for some reason we need different models for standard and pydata deployment, a model in `pydata` app would inherit from a model in `workshops` app.

Any ideas how we could get rid of monkey patching templates, forms and views? This could be probably done be some kind of inheritance again.

Can we postpone it to v1.10?

pbanaszkiewicz:
@chrismedrela:
> OK, I understand why we need two jobs now. However, I think we need to address a much more fundamental issue - the way we monkey patch everything in `pydata/apps.py` scares me, especially that we monkey patch models. It's not going to scale at all.

That's true, but it was a deliberate choice. Maybe it's time to revisit it, maybe it's too soon.

> Can we postpone it to v1.10?

Yes, this is definitely too much to merge this late in v1.9 development cycle.

> Any ideas how we could get rid of monkey patching templates, forms and views? This could be probably done be some kind of inheritance again.

Perhaps we should have a discussion somewhere else, and only focus on [that solution](https://github.com/swcarpentry/amy/pull/1073#issuecomment-258798783) in this thread. @narayanaditya95 can you please make a separate issue if you feel we should discuss PyData application and how it integrates with other parts of AMY?

chrismedrela:
I'm :+1: for merging when these three issues above are addressed. I've created a new issue #1075 to address monkey patching issue.

workshops/test/test_sponsorship.py
```diff
@@ -72,7 +72,7 @@ def setUp(self):
         self.sponsorship = Sponsorship.objects.create(
             organization=self.org_alpha,
             event=self.event,
-            amount=100.00,
```
narayanaditya95: `100.00` is not a valid choice variable to the `amount` field when we run the tests with PyData deployed.

.travis.yml
```diff
@@ -1,13 +1,38 @@
...
 python:
   - 3.5
+
+env:
+  - PYDATA=true
+  - PYDATA=false CHECK_MIGRATION=true
+
+before_install:
+  - git config --global user.email "you@example.com"
+  - git config --global user.name "Your Name"
```
narayanaditya95: git refuses to pull a branch without these configs. It is probably because git needs to add a merge commit and it requires credentials.

.travis.yml
```diff
@@ -1,13 +1,38 @@
...
+
+before_script:
+  - if [[ $CHECK_MIGRATION == true ]]; then
+      python manage.py makemigrations --dry-run -e;
+      export STATUS_CODE=$?;
+      if [[ "$STATUS_CODE" == "0" ]]; then
+        exit 1;
+      fi;
+    fi;
+
```
narayanaditya95: Okay

narayanaditya95:
@chrismedrela 
Are you fine with the current implementation of merging a branch, or would you rather consider a different approach that I suggested to @pbanaszkiewicz 

>Nice suggestion. I am starting to reconsider #938. By overriding the envvar DJANGO_SETTINGS_MODULE, we can switch between a normal deployment and a PyData deployment.

.travis.yml
narayanaditya95: See #1078

chrismedrela:
> Nice suggestion. I am starting to reconsider #938. By overriding the envvar DJANGO_SETTINGS_MODULE, we can switch between a normal deployment and a PyData deployment.

@narayanaditya95 I'd like to have everything on one branch (`develop`). Maintaining second branch (`deploy-pydata`) would be a lot of work, because of backporting everything from `develop` to `deploy-pydata`. So, yes, I'm :+1: for varying between deployments with envvars, with one important modification: I'd like to have one `settings.py` file that reads from an `ENABLE_PYDATA` envvar and contains conditionals like `if ENABLE_PYDATA:`. IMO that is simpler than having two separate `settings.py` files and keeping them in sync.

chrismedrela:
BTW this is exactly what @pbanaszkiewicz has proposed:

>>> What exactly is in deploy-pydata?

>> Adds pydata to INSTALLED_APPS and urlconfig. See the commit cf8603a

> Why not conditionally include it in the config? It could be controlled via envvar.

narayanaditya95:
@chrismedrela Please review.

amy/settings.py
```diff
@@ -338,6 +338,17 @@
...
 }
 
+# settings for PyData application
+ENABLE_PYDATA = os.environ.get('AMY_ENABLE_PYDATA', False)
+
+if ENABLE_PYDATA:
+    INSTALLED_APPS.insert(
+        INSTALLED_APPS.index('workshops'),
+        'pydata'
+    )
```
chrismedrela: Could you replace it with:

```python
INSTALLED_APPS = [
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.messages',
    'django.contrib.staticfiles',
    'workshops',
]
if ENABLE_PYDATA:
    INSTALLED_APPS += [
        'pydata',
    ]
INSTALLED_APPS += [
    # this should be after 'workshops' because templates in
    # 'templates/registration/' clash
    'django.contrib.admin',
    'crispy_forms',
    'selectable',
    'django_countries',
    'django_filters',
    'reversion',
    'rest_framework',
    'api',
    'extforms',
    'captcha',
    'compressor',
    'social.apps.django_app.default',
    'debug_toolbar',
]
```

So that `INSTALLED_APPS` is defined in one place.

amy/settings.py
```diff
@@ -338,6 +338,17 @@
     },
 }
 
+# settings for PyData application
+ENABLE_PYDATA = os.environ.get('AMY_ENABLE_PYDATA', False)
```
chrismedrela: What are valid values for `AMY_ENABLE_PYDATA` envvar? Which ones are truthy and which ones are falsy? What if somebody sets `AMY_ENABLE_PYDATA=false`?

.travis.yml
```diff
@@ -1,13 +1,33 @@
...
 script:
-  - python manage.py makemigrations --dry-run -e; export STATUS_CODE=$?; if [[ "$STATUS_CODE" == "0" ]]; then exit 1; fi
   - coverage run --source=amy,api,workshops manage.py test
```
chrismedrela: @pbanaszkiewicz, BTW where does it sends the coverage data?

@narayanaditya95 should we include `pydata` in `--source`?

amy/urls.py
```diff
@@ -55,3 +55,7 @@
     urlpatterns += [
         url(r'^__debug__/', include(debug_toolbar.urls)),
     ]
+
+
+if settings.ENABLE_PYDATA:
+    urlpatterns.insert(1, url(r'^workshops/', include('pydata.urls')))
```
chrismedrela: Could you change it in a similar way to `INSTALLED_APPS`?

amy/settings.py
```diff
@@ -338,6 +338,17 @@
...
+
+if ENABLE_PYDATA:
+    INSTALLED_APPS.insert(
+        INSTALLED_APPS.index('workshops'),
+        'pydata'
+    )
+    PYDATA_USERNAME_SECRET = os.environ.get('AMY_PYDATA_USERNAME')
+    PYDATA_PASSWORD_SECRET = os.environ.get('AMY_PYDATA_PASSWORD')
```
chrismedrela: Are these two settings optional? That is, is `None` valid value for both username and password?

.travis.yml
```diff
@@ -1,13 +1,33 @@
...
 cache: pip 
+
 python:
   - 3.5
+
+env:
+  - AMY_ENABLE_PYDATA=true
+    AMY_PYDATA_USERNAME=username
+    AMY_PYDATA_PASSWORD=password
+  - CHECK_MIGRATION=true
```
chrismedrela: 1) Is this necessary? We don't have two Travis jobs anymore.
2) Previously, you couldn't enable pydata and, at the same time, enable migrations checks. What did you changed, that it works now?

amy/settings.py
narayanaditya95: No. But we have checks in `pydata/checks.py` to ensure these variables are not None.

.travis.yml
narayanaditya95: >@pbanaszkiewicz, BTW where does it sends the coverage data?

The `coverage` command generates a report. The `coveralls` command sends it to [Coveralls](https://coveralls.io/).

>@narayanaditya95 should we include pydata in --source?

I am not sure what will happen if two Travis jobs send coverage data to Coveralls for a single commit. Will it break?
@pbanaszkiewicz

amy/settings.py
chrismedrela: `check_pydata_installed_before_workshops` checks only if these variables are set, but doesn't check if they're not `None`.

.travis.yml
chrismedrela: > The coverage command generates a report. The coveralls command sends it to Coveralls.

Yes, and where is the report available? https://coveralls.io/github/swcarpentry/amy results in 404.

chrismedrela:
@pbanaszkiewicz could you review the design?

amy/settings.py
narayanaditya95: Resolved

.travis.yml
narayanaditya95: Only @pbanaszkiewicz has the permission to access coveralls.io

amy/urls.py
narayanaditya95: Resolved

narayanaditya95:
@chrismedrela 
Addressed all concerns.

chrismedrela:
@narayanaditya95 thank you for your work on this PR! @pbanaszkiewicz please merge if you like it.

amy/settings.py
pbanaszkiewicz: We should probably parse `AMY_ENABLE_PYDATA` the same way we parse `AMY_DEBUG`:
```
json.loads(os.environ.get('AMY_DEBUG', 'true'))
```

.travis.yml
pbanaszkiewicz: I don't think I ever set up coveralls, but the codebase was ready for it.

amy/settings.py
```diff
@@ -33,6 +33,13 @@
     SECRET_KEY = None
 SECRET_KEY = os.environ.get('AMY_SECRET_KEY', SECRET_KEY)
 
+# settings for PyData application
+ENABLE_PYDATA = json.loads(os.environ.get('AMY_ENABLE_PYDATA', 'false'))
```
pbanaszkiewicz: Maybe just add a comment? :-)

PR #1077, opened 2016-11-08
Fix #955 -- add description of deployment procedure

### LINKED ISSUE #955: Deployment on production

chrismedrela:
Fixes #955.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -1,0 +1,163 @@
+# On the End of Each Milestone
+
+1.  Close milestone using [GitHub UI](https://github.com/swcarpentry/amy/milestones).
+
+2.  Write down release notes and add it to the repository in `docs/releases` directory.
+    Release notes should have two sections: new features and bugfixes, each one enumerating changes and mentioning their authors.
+    Base your work on the list of all closed pull requests for given milestone, available on GitHub.
+
+3.  Follow Release Procedure (see below).
+
+4.  Follow Deployment Procedure (see below).
+
+# Release Procedure
+
+We assume that you want to release AMY v1.X.Y.
+
+Execute the following commands on your local machine, not production.
+
+1.  Move to AMY root directory (the one with `manage.py` file) using `cd` command.
+
+2.  Make sure you have configured repositories:
+
+    - `origin` for your repository on GitHub
+    - `upstream` for [`swcarpentry/amy` repo on GitHub](https://github.com/swcarpentry/amy)
+
+    For example, this is the correct configuration for `chrismedrela`:
+
+        $ git remote -v
+        origin	git@github.com:chrismedrela/amy.git (fetch)
+        origin	git@github.com:chrismedrela/amy.git (push)
+        upstream	git@github.com:swcarpentry/amy.git (fetch)
+        upstream	git@github.com:swcarpentry/amy.git (push)
+
+3.  Make sure your local `develop` and `master` branches are up to date:
+
+        $ git checkout develop
+        $ git pull upstream develop
+        $ git push origin develop
+
+        $ git checkout master
+        $ git pull upstream master
+        $ git push origin master
+
+    Pushes to your `origin` repository are optional.
+
+4.  Merge `develop` into `master` branch:
+
+        $ git checkout master
+        $ git merge --no-ff develop  # sometimes there are conflicts in files with release version, but most often we need to edit them and commit
+
+5.  Bump version on `master`:
+
+        $ vim bower.json  # change 'version' to 1.X.Y
+        $ vim workshop/__init__.py  # change version to 1.X.Y
+        $ git add workshops/__init__.py bower.json
+        $ git commit -m "Bumping version to v1.X.Y"
+
+6.  Just to be safe, run tests:
+
+        $ make test
+
+7.  Tag a release. Tag message should be `AMY release v1.X.Y`.
+
+        $ git tag -a v1.X.Y -s
+
+    Omit `-s` flag if you cannot create signed tags.
+    See [Git documentation](https://git-scm.com/book/tr/v2/Git-Tools-Signing-Your-Work) for more info about signed tags.
+
+8.  Push `master` and the new tag everywhere:
+
+        $ git push origin master --tags
+        $ git push upstream master --tags
+
+9.  Bump version on `develop`:
+
+        $ git checkout develop
+        $ vim bower.json  # change version to the next one after 1.X.Y (i.e. 1.X+1.0-dev)
+        $ vim workshop/__init__.py  # change version to the next one
+        $ git add workshops/__init__.py bower.json
+        $ git commit -m "Bumping version to v1.X+1.0-dev"
+
+10. And push it everywhere:
+
+        $ git push upstream develop
+        $ git push origin develop
+
+# Deployment Procedure
+
+1.  Log into production:
+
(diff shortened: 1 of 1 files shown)


chrismedrela:
@pbanaszkiewicz please review.

docs/deploy.md
```diff
@@ -0,0 +1,131 @@
+# Procedure for releasing new AMY version
+
+1. Close milestone using [GitHub UI](https://github.com/swcarpentry/amy/milestones).
+
+1. Follow procedure for deployment on production.
+
+1. Write down release notes and publish it on [Piotr's blog](http://piotr.banaszkiewicz.org/). Release notes should have two sections: new features and bugfixes, each one enumerating changes and mentioning their authors. Base your work on the list of all closed pull request for given milestone, available on GitHub.
```
pbanaszkiewicz: 1. "all closed pull request**s**"
2. Can you split long lines? It will be easier to comment on single lines.

docs/deploy.md
```diff
@@ -0,0 +1,131 @@
...
+            $ git pull upstream develop
+            $ git push origin develop
+            
+            $ git checkout master
+            $ git pull upstream master
+            $ git push origin master
+        
+        Pushes to your `origin` repository are optional.
+        
+    1.  Update `master` branch:
```
pbanaszkiewicz: Merge `develop` into `master` branch.

docs/deploy.md
```diff
@@ -0,0 +1,131 @@
...
+    
+    1.  Just to be safe, run tests.
+    
+            $ make test
+    
+    1.  Tag a release. Tag message should be `AMY release v1.X.Y`. 
+        
+            $ git tag -a v1.7.1 -s
+            
+        Omit `-s` flag if you cannot create signed tags.
```
pbanaszkiewicz: Can you please add a link to some help regarding signed tags here? thanks

docs/deploy.md
```diff
@@ -0,0 +1,131 @@
...
+            $ git add workshops/__init__.py bower.json
+            $ git commit -m "Bumping version to v1.X.Y"
+    
+    1.  Just to be safe, run tests.
+    
+            $ make test
+    
+    1.  Tag a release. Tag message should be `AMY release v1.X.Y`. 
+        
+            $ git tag -a v1.7.1 -s
```
pbanaszkiewicz: In [line 52](https://github.com/swcarpentry/amy/pull/1077/files#diff-ef860daacd753531645fa756fbe5b836R52) you have `v1.X.Y` but here you use explicit values for X and Y: `v1.7.1`. Maybe it's a good idea to keep X and Y everywhere or the same example `v1.7.1` everywhere?

docs/deploy.md
```diff
@@ -0,0 +1,131 @@
...
+    
+    1.  Push `master` and the new tag everywhere:
+    
+            $ git push origin master --tags
+            $ git push upstream master --tags
+    
+1.  Bump version on `develop`:
+
+        $ git checkout develop
+        $ vim bower.json  # change version to the next one after 1.X.Y (i.e. 1.X+1.Y)
```
pbanaszkiewicz: That would be `1.X+1.0`.

Additional note: we won't have major version ==1 forever.

docs/deploy.md
```diff
@@ -0,0 +1,131 @@
...
+            ssh amy@amy.software-carpentry.org
+
+        All the following commands will be executed on production, not your local machine.
+        
+    1.  Execute:
+        
+            source ~/amy_site/venv/bin/activate
+            cd amy_site/amy
+            git fetch
+            git checkout tags/v1.7.1  # Python files updated
```
pbanaszkiewicz: "Python files updated, but not the served files - server keeps them in memory" or something similar

docs/deploy.md
```diff
@@ -0,0 +1,131 @@
...
+            
+            $ cp db.sqlite3 db.sqlite3.bak
+            
+    1.  Update dependencies:
+            
+            $ make upgrade
+            
+    1.  Migrate database online:
+    
+            $ ./manage.py migrate
```
pbanaszkiewicz: What about `AMY_DB_NAME`?

docs/deploy.md
```diff
@@ -0,0 +1,131 @@
...
+            
+    1.  Regenerates version in the footer:
```
pbanaszkiewicz: "version number".

This is easier than writing `make workshops/git_version.py` plus it runs checks.

chrismedrela:
@pbanaszkiewicz I've made changes you requested. I don't know if the way I added `AMY_DB_FILENAME` is correct, but I cannot verify it and I don't want to experiment on production now.

> I think it's worth to have an additional step that tries to apply migrations on a production database copy.

I don't think so - I think we should temporarily stop server for upgrade. When we do that, it doesn't matter if you apply migrations to `db.sqlite3` or its copy, because the server is off and none of these dbs can change during migration.

pbanaszkiewicz:
@chrismedrela:

> I don't know if the way I added `AMY_DB_FILENAME` is correct, but I cannot verify it and I don't want to experiment on production now.

It's not correct, because you're setting it in your own environment, not server's (or, not uWSGI's, to be precise). `AMY_DB_FILENAME`, along with other envvars, is set up in `amy_uwsgi.ini` config file.

>> I think it's worth to have an additional step that tries to apply migrations on a production database copy.

> I don't think so - I think we should temporarily stop server for upgrade. When we do that, it doesn't matter if you apply migrations to `db.sqlite3` or its copy, because the server is off and none of these dbs can change during migration.

I changed a few things on the production. First of all, I added a `/etc/init` (or Upstart) script for running uWSGI in Emperor mode. This way admin can start/stop uWSGI easily (`start uwsgi-emperor` and `stop uwsgi-emperor`). I also altered `nginx` configuration so that it will show maintenance page `/home/amy/amy_site/static_html/maintenance.html` when `uwsgi` is not responding.

There is one caveat, though. "Normal" user cannot issue `start/stop uwsgi-emperor` commands, only admin can. If anyone has a solution to this issue, I'll be grateful, because for now we cannot stop the server for upgrade.

While I agree with you that being able to stop uWSGI for AMY upgrade is nice, I didn't intend to use `AMY_DB_FILENAME` because the DB would change during migration; instead I was thinking about migrations failing. It happened a few times, mostly due to data migrations and not all cases covered. I don't think this would be the case if we could stop uWSGI for the upgrade.

Finally I noticed that we should be running `./manage.py` with the same environmental variables as we run the production; this is something we should think about…

Anyway, to sum up:

1. changes on the production to: start/stop uWSGI easily, and have a maintenance page show up when uWSGI is off
2. need to be able to start/stop uWSGI by normal user (`amy`), not only by root
3. need to read and set up envvars from `amy_uwsgi.ini` either when sourcing the virtualenv or when issuing any `./manage.py` command
4. also, the part about `AMY_DB_FILENAME` will have to change.

chrismedrela:
> 3. need to read and set up envvars from amy_uwsgi.ini either when sourcing the virtualenv or when issuing any ./manage.py command

According to [this](https://github.com/unbit/uwsgi/issues/629#issuecomment-43303330), you can factor out envvars to `.env` file, and modify `amy_uwsgirc.ini`:
```
[uwsgi]
for-readline = ../.env
  env = %(_)
endfor = 
```

chrismedrela:
@pbanaszkiewicz please review again.

Issue #170, opened 2015-01-27
Add "Add new" link also at the top of the page


r-gaia-cs:
I know that user interface isn't a priority right know but what about add "Add new" link also at the top of the page, e.g.

![s](https://cloud.githubusercontent.com/assets/1506457/5910992/b743f306-a5a7-11e4-8c55-23a7f70657a6.jpg)

Do that should be easy. If all agree to it I can open a pull request to it tomorrow.

gvwilson:
Post-v1.0 please.

wking:
On Mon, Jan 26, 2015 at 04:35:29PM -0800, Greg Wilson wrote:
> Post-v1.0 please.

We're not at v1.0 yet [1] and this is tagged v1.1, but it's also
tagged essential?  Shouldn't all of the v1.1 issues have the same
priority?  Should we be dealing with them now, or waiting for v1.0?

pbanaszkiewicz:
Added in #222

PR #1079, opened 2016-11-08
Refactor views to allow better form prepopulation.

narayanaditya95:
Changes:
1. Adds a `PrepopulationSupportMixin` view.
   The mixin prepopulates and disables fields based on GET parameters.
2. Refactor `MembershipCreate` view to use `PrepopulationSupportMixin`

@pbanaszkiewicz @chrismedrela please review.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ PyYAML
 requests>=2.0
 requests-mock>=0.7,<0.8
 django-crispy-forms>=1.6,<1.7
-django-selectable>=0.9.0,<1.0
+git+https://github.com/mlavin/django-selectable.git@45b0dc8c4e6351afc9986b7edf2e45c3416cd66e#egg=django-selectable
 django-countries>=3.4,<3.5
 django-filter>=0.12,<0.13
 django-reversion>=1.10,<1.11
--- a/workshops/base_views.py
+++ b/workshops/base_views.py
@@ -206,3 +206,19 @@ class RedirectSupportMixin:
             return next_url
         else:
             return default_url
+
+
+class PrepopulationSupportMixin:
+
+    def get_initial(self):
+        return {
+            field: self.request.GET.get(field) for field in self.populate_fields
+        }
+
+    def get_form(self, *args, **kwargs):
+        """Disable fields that are pre-populated."""
+        form = super().get_form(*args, **kwargs)
+        for field in self.populate_fields:
+            if field in self.request.GET:
+                form.fields[field].disabled = True
+        return form
--- a/workshops/templates/workshops/training_progresses_inline.html
+++ b/workshops/templates/workshops/training_progresses_inline.html
@@ -12,6 +12,6 @@
   </a>
   &nbsp;
 {% endfor %}
-<a href="{% url 'trainingprogress_add' person.id %}?next={{ request.get_full_path|urlencode }}"
+<a href="{% url 'trainingprogress_add' %}?trainee={{ person.id|urlencode }}&amp;next={{ request.get_full_path|urlencode }}"
    class="label label-default"
    data-toggle="popover" data-content="Click to add new progress">+</a>
--- a/workshops/templates/workshops/organization.html
+++ b/workshops/templates/workshops/organization.html
@@ -118,7 +118,7 @@
 {% endif %}
 
 {% if perms.workshops.add_membership and perms.workshops.change_organization %}
-  <a href="{% url 'membership_add' organization.domain %}" class="btn btn-success">Add a membership</a>
+  <a href="{% url 'membership_add' %}?organization={{ organization.pk }}" class="btn btn-success">Add a membership</a>
 {% else %}
   <a href="#" class="btn btn-success disabled">Add a membership</a>
 {% endif %}
--- a/workshops/test/test_training_progress.py
+++ b/workshops/test/test_training_progress.py
@@ -219,11 +219,12 @@ class TestCRUDViews(TestBase):
         self.assertEqual(rv.context['form'].initial['evaluated_by'], self.admin)
 
     def test_create_view_works_with_initial_trainee(self):
-        rv = self.client.get(reverse('trainingprogress_add',
-                                     args=[self.ironman.pk]))
+        rv = self.client.get(reverse('trainingprogress_add'), {
+            'trainee': self.ironman.pk
+        })
         self.assertEqual(rv.status_code, 200)
         self.assertEqual(rv.context['form'].initial['evaluated_by'], self.admin)
-        self.assertEqual(rv.context['form'].initial['trainee'], self.ironman)
+        self.assertEqual(int(rv.context['form'].initial['trainee']), self.ironman.pk)
 
     def test_create_view_works(self):
         data = {
--- a/workshops/urls.py
+++ b/workshops/urls.py
@@ -23,7 +23,6 @@ urlpatterns = [
     url(r'^memberships/', include([
         url(r'^$', views.AllMemberships.as_view(), name='all_memberships'),
         url(r'^add/$', views.MembershipCreate.as_view(), name='membership_add'),
-        url(r'^add/(?P<org_domain>[\w\.-]+)/$', views.MembershipCreate.as_view(), name='membership_add'),
     ])),
     url(r'^membership/(?P<membership_id>\d+)/', include([
         url(r'^$', views.MembershipDetails.as_view(), name='membership_details'),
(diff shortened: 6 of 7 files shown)


workshops/base_views.py
```diff
@@ -206,3 +206,19 @@ def get_success_url(self):
...
+
+    def get_initial(self):
+        return self.request.GET.dict()
+
+    def get_form(self, *args, **kwargs):
+        """Disable fields that are pre-populated."""
+        form = super().get_form(*args, **kwargs)
+        for field in self.get_initial():
+            try:
+                form.fields[field].disabled = True
```
chrismedrela: This doesn't make autocomplete fields disabled. What about leaving pre-populated fields enabled?

workshops/base_views.py
```diff
@@ -206,3 +206,19 @@ def get_success_url(self):
             return next_url
         else:
             return default_url
+
+
+class PrepopulationSupportMixin:
+
+    def get_initial(self):
+        return self.request.GET.dict()
```
chrismedrela: IMO it'd be better to explicitly specify which fields can be pre-populated, i.e. introduce `ALLOW_PREPOPULATE` class attribute that must be set in children classes.

workshops/base_views.py
narayanaditya95: >This doesn't make autocomplete fields disabled.

It works for me. Can you point out the form where it fails to disable a field?

 >What about leaving pre-populated fields enabled?

I like having them disabled so that I do not accidently change them.

workshops/base_views.py
narayanaditya95: Can you show an use case where we would not like to have a field prepopulated even though it exists as a GET parameter?
I think we can do that when we arrive at such a use case.

workshops/base_views.py
chrismedrela: >> This doesn't make autocomplete fields disabled.

> It works for me. Can you point out the form where it fails to disable a field?

When I go to `http://127.0.0.1:8000/workshops/memberships/add/?organization=435`, the field is disabled, but the arrow at the right side is enabled. I can click it and a list of organizations is displayed. Then, I can select one of them. Does it work in the same way on your machine?

>> What about leaving pre-populated fields enabled?

> I like having them disabled so that I do not accidently change them.

I'm :+1: for disabling them if only they're *really* disabled. :-)

workshops/base_views.py
chrismedrela: > Can you show an use case where we would not like to have a field prepopulated even though it exists as a GET parameter?

I cannot right now, but I'm worrying about security issues - like if it's possible to override value for `id` field (seems not possible). We may have name collisions, i.e. when we define `next` field on a model and, at the same time, we support redirection (`RedirectSupportMixin`). Right now I cannot imagine any scenario where this behavior could be dangerous, so let's stick with your way. We can always add `ALLOW_PREPOPULATE` later.

workshops/base_views.py
narayanaditya95: The issue was fixed at [django-selectable #170](https://github.com/mlavin/django-selectable/pull/170)

chrismedrela: I agree this is an issue with `django-selectable`, but it'd be very confusing to have an enabled field that is ignored finally. IMO fields should be left enabled or made 100% disabled.

> The issue was fixed at django-selectable #170

It seems that this fix wasn't released yet (django-selectable 0.9 is from 2014), so we can't fix it by upgrading `django-selectable`. So, if we want to make them disabled, we need a workaround like [this one](https://github.com/mlavin/django-selectable/issues/163). Alternatively, we can leave prepopulated fields enabled and disable them when the fix is released.

narayanaditya95: >It seems that this fix wasn't released yet (django-selectable 0.9 is from 2014), so we can't fix it by upgrading django-selectable.

We can instruct pip to fetch the `develop` branch of the repository so that we have the fix. Is that fine?

chrismedrela: > We can instruct pip to fetch the develop branch of the repository so that we have the fix. Is that fine?
Is it possible to specify exact commit?

narayanaditya95: Yes. See [SO thread](http://stackoverflow.com/questions/13685920/install-specific-git-commit-with-pip)

chrismedrela: OK, so I'm :+1: for fetching today's latest commit from `develop` branch.

narayanaditya95:
@chrismedrela 
I made a slight change. Now, only fields that are specified in the GET parameters are disabled. Fields that are prepopulated by overriding the `get_initial` method is not disabled.
This was done so that the `evaluated_by` field on the TrainingProgressCreate view was not disabled and could be changed by the user.

narayanaditya95:
I seem to have run into the dreaded XML parse failures.
@chrismedrela any idea what caused the tests on the PersonDetailView to fail?

chrismedrela:
It seems that this part is invalid:
```html
<a href="/workshops/training_progresses/add/?trainee=5&next=/workshops/person/5/"
   class="label label-default"
   data-toggle="popover" data-content="Click to add new progress">+</a>
```
Probably you cannot have new line characters inside a tag. This summer I hit similar issue with invalid HTML. What I'd do is to rewrite broken tests using `django-webtest`. See for example `TestPerson.test_awarding_instructor_badge_workflow`.

workshops/base_views.py
pbanaszkiewicz: I'm :+1: for better control by the programmer. We can have `prepopulated_fields` (or similar) class variable that lists all fields we allow for prepopulation.

workshops/templates/workshops/training_progresses_inline.html
```diff
@@ -12,6 +12,6 @@
   </a>
   &nbsp;
 {% endfor %}
-<a href="{% url 'trainingprogress_add' person.id %}?next={{ request.get_full_path|urlencode }}"
+<a href="{% url 'trainingprogress_add' %}?trainee={{ person.id }}&next={{ request.get_full_path|urlencode }}"
```
pbanaszkiewicz: Maybe we could have a new `url` template tag that generates query string as well?

requirements.txt
```diff
@@ -3,7 +3,7 @@ PyYAML
 requests>=2.0
 requests-mock>=0.7,<0.8
 django-crispy-forms>=1.6,<1.7
-django-selectable>=0.9.0,<1.0
+git+https://github.com/mlavin/django-selectable.git@master#egg=django-selectable
```
pbanaszkiewicz: Okay, so this one is very tempting, but it's a serious security threat because we're not pinning the version here.

How much work could it take to keep the current version and add a workaround for this `django-selectable` issue?

narayanaditya95: I will proceed to pin the version correct to the latest commit.
```
git+https://github.com/mlavin/django-selectable.git@<commit_id>#egg=django-selectable
```

Does that solve the security issue?

workshops/templates/workshops/training_progresses_inline.html
narayanaditya95: I discovered [django-spurl](https://github.com/j4mie/django-spurl). It allows manipulating urls in templates.

```
{% url 'trainingprogress_add' as trainingprogress_add_url %}
<a href="{% spur base=trainingprogress_add_url add_query="trainee={{ person.id }}&next={{ request.get_full_path|urlencode }}" }">
```

I don't think it makes it any easier. We should probably stick to doing it the pure Django way.

narayanaditya95:
>It seems that this part is invalid:

><a href="/workshops/training_progresses/add/?trainee=5&next=/workshops/person/5/"
   class="label label-default"
   data-toggle="popover" data-content="Click to add new progress">+</a>
Probably you cannot have new line characters inside a tag. This summer I hit similar issue with invalid HTML. What I'd do is to rewrite broken tests using django-webtest. See for example TestPerson.test_awarding_instructor_badge_workflow.

@chrismedrela 
The `&` in the `href` attribute must be escaped using `&amp;`. That fixed the problem. :)

narayanaditya95:
@pbanaszkiewicz 

>Two issues that I have with this PR:

> 1. I feel like an explicit list of fields per class is a way to go here (we'll avoid some "magic" or "wtf"s later)

I have implemented it. The class now requires a list of fields that are allowed to be prepopulated.

> 2. django-selectable from git is a big warning

I have pinned `django-selectable` to the current commit.

> One feature that I'd love to have: url template tag that generates query params. (This may be a lot of work so don't worry too much about it).

I am not sure whether adding a layer of abstraction will make it easier to read. :(

workshops/templates/workshops/training_progresses_inline.html
chrismedrela: I'm :+1: for doing it the pure Django way.

requirements.txt
```diff
@@ -3,7 +3,7 @@ PyYAML
 requests>=2.0
 requests-mock>=0.7,<0.8
 django-crispy-forms>=1.6,<1.7
-django-selectable>=0.9.0,<1.0
+git+https://github.com/mlavin/django-selectable.git@45b0dc8#egg=django-selectable
```
chrismedrela: Please use full commit hash. I don't know what would happen if there were multiple commits starting with `45b0dc8` in the future.

narayanaditya95: Added full commit hash.

pbanaszkiewicz:
I don't feel safe including a development version of a library. But then, why do I feel safe including a specific version of that library…? I guess the worst that could happen is that we revert this PR.

PR #1080, opened 2016-11-08
Allow requests from localhost while in DEBUG mode.

narayanaditya95:
From Django official docs:
> Changed in Django 1.10.3:
> In older versions, ALLOWED_HOSTS wasn’t checked if DEBUG=True.
> This was also changed in Django 1.9.11 and 1.8.16 to prevent a DNS rebinding attack.

### DIFF
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -117,7 +117,8 @@ TEMPLATES = [
 ALLOWED_HOSTS = [
     'amy.software-carpentry.org',
 ]
-
+if DEBUG:
+    ALLOWED_HOSTS.append('127.0.0.1')
 
 # Application definition
 


pbanaszkiewicz:
Gentlemen, first of all we're not on the v1.10 yet, and second of all: please don't merge PRs that aren't triaged (like this one).

narayanaditya95:
@pbanaszkiewicz The change was also down ported to Django 1.9.11.

chrismedrela:
> please don't merge PRs that aren't triaged (like this one).

If you mean attaching milestone, then sorry about that.

> The code here should have an explanatory comment.

How exactly would you like to comment it?

PR #1090, opened 2016-11-19
Backporting release notes

### LINKED ISSUE #1082: Import old release notes
Add them to `docs/releases` dictionary. Get old release notes from [Piotr's blog](http://piotr.banaszkiewicz.org/).

It'd be better to keep release notes along with source code, so that they won't get lost, and to have everything in one place. @pbanaszkiewicz what's your opinion?

pbanaszkiewicz:
This fixes #1082 by backporting release notes from my blog.

I decided to switch from "I" to "[Piotr](https://github.com/pbanaszkiewicz)", and to remove most of the personal stuff. I also tried to arrange all release notes in specific template (which was not always possible).

We're losing timestamps with this approach. I'm not sure if we need them. Maybe you could vote for having them? And point for a way to add them to the template?

### DIFF
--- a/docs/releases/README.md
+++ b/docs/releases/README.md
@@ -1,0 +1,6 @@
+# Release notes
+
+Release notes since v0.9.1 to v1.8.0 were taken from
+[Piotr Banaszkiewicz](https://github.com/pbanaszkiewicz)'s
+[blog](http://piotr.banaszkiewicz.org/) and adapted to a general template
+used here.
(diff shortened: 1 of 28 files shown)


chrismedrela:
> We're losing timestamps with this approach. I'm not sure if we need them. Maybe you could vote for having them? And point for a way to add them to the template?

:+1: for adding them to a template.

PR #1091, opened 2016-11-20
Export badges: group by person

### LINKED ISSUE #961: Export badge: group by person
This link:

https://amy.software-carpentry.org/workshops/export/badges

returns a JSON or YAML file that does not separate DC or SWC badges.

pbanaszkiewicz:
This fixes #961 by adding a new API endpoint for exporting badges grouped by person. So the new structure is: person→badges; the old was badge→persons.


workshops/templates/workshops/export_badges.html
```diff
@@ -0,0 +1,25 @@
+{% extends "base_nav_fixed.html" %}
+
+{% block content %}
+<h3>Group by badge name</h3>
+<a href="{{ badges_api_link }}">API link.</a>
+<p>
+  {% if badges_json_link %}
```
chrismedrela: Isn't it always true? When I look into the view:
```python
badges_json_link = reverse('api:export-badges', kwargs={'format': 'json'})
```
When `reverse` can return something falsy?

pbanaszkiewicz:
@chrismedrela:
> 1. Export badges by person looks very similar to `api/v1/persons/`. The only difference is that the `badges` field is a list of badges' names instead of list of dictionaries describing badges in more details. Could we reuse this API endpoint? And you can filter out people without any badge by selecting all badges in `filter` modal window. 

The other API endpoint that you mentioned is intended for future use in case we add AngularJS or some other front framework. I'd leave that part of API be for now.

> 2. This PR results in `django.db.utils.OperationalError: too many SQL variables` on production data when you open `api/v1/export/badges_by_person/`.

This is interesting, I'll check and get back to you.

> Isn't it always true? When I look into the view:
> ```python
> badges_json_link = reverse('api:export-badges', kwargs={'format': 'json'})
> ```
> When `reverse` can return something falsy?

This part is a heritage from `export.html` template, I'll update it.

pbanaszkiewicz:
@chrismedrela updated - please take a look at the commits.

pbanaszkiewicz:
I'm going to merge this, and if Chris has any objections we'll fix them later on.

PR #1092, opened 2016-11-20
Bulk upload: fix #1000 ("use that person") and #1015 (list of roles)

### LINKED ISSUE #1000: Provide "use that person" during bulk upload.
1. Create CSV file.
2. Bulk upload.
3. Get told "There's a matching person in the database: Alan Turing. Use their email to merge."
4. But their email isn't displayed.
5. And it would be really helpful if there a button there that would switch the info to _that person_ in place.

pbanaszkiewicz:
This PR fixes #1000 and #1015 (in one of the commits).

88c3e1e, c0ecdfa, e28c85e and fa96e0e are cleaning up and / or fixing small bugs. Other commits change the workflow for bulk-upload (slightly). Some parts, like magic-matching or showing if `personal` or `family` is different in DB/uploaded file, were removed; it looks like we ended up with slightly more maintainable code.

### DIFF
--- a/workshops/templates/workshops/person_bulk_add_results.html
+++ b/workshops/templates/workshops/person_bulk_add_results.html
@@ -43,7 +43,18 @@
             </td>
             <td class="editable">
                 <span>{{ entry.role|default:"—" }}</span>
-                <input type="text" name="role" value="{{ entry.role|default:"" }}" class="hidden">
+                <select name="role" class="hidden">
+                {% if entry.role in possible_roles %}
+                    {% for role in possible_roles %}
+                    <option value="{{ role }}" {% if role == entry.role %}selected="selected"{% endif %}>{{ role }}</option>
+                    {% endfor %}
+                {% else %}
+                    <option selected="selected" value="{{ entry.role }}">{{ entry.role }}</option>
+                    {% for role in possible_roles %}
+                    <option value="{{ role }}">{{ role }}</option>
+                    {% endfor %}
+                {% endif %}
+                </select>
                 </td>
             <td class="text-danger">
             {% if entry.errors %}
@@ -100,8 +111,10 @@
             var cell = $(e.currentTarget).parent();
             var span = $(cell).find("span");
             var input = $(cell).find("input");
+            var select = $(cell).find("select");
             $(span).addClass("hidden");
             $(input).removeClass("hidden");
+            $(select).removeClass("hidden");
         }
     });
 
--- a/workshops/templates/workshops/person_bulk_add_results.html
+++ b/workshops/templates/workshops/person_bulk_add_results.html
@@ -55,7 +55,7 @@
                     {% endfor %}
                 {% endif %}
                 </select>
-                </td>
+            </td>
             <td class="text-danger">
             {% if entry.errors %}
                 <ul>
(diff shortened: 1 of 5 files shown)


workshops/util.py
```diff
@@ -137,47 +137,29 @@ def verify_upload_person_task(data):
...
-                        errors.append('{0} mismatch: database "{1}" '
-                                      'vs uploaded "{2}".'
-                                      .format(which, actual, uploaded))
-
-            except Person.DoesNotExist:
-                # in this case we need to add a new person
-                pass
-
+                person = Person.objects.get(id=int(person_id))
+            except (ValueError, TypeError, Person.DoesNotExist):
```
chrismedrela: You validate `existing_person_id` inside `person_bulk_add_match_person` to make sure it's an `int`. So you can use `id=person_id` and maybe add an `assert isinstance(person_id, int)`.

workshops/util.py
```diff
@@ -137,47 +137,29 @@ def verify_upload_person_task(data):
         email = item.get('email', None)
         personal = item.get('personal', None)
         family = item.get('family', None)
+        person_id = item.get('existing_person_id', None)
         person = None
```
chrismedrela: I would delete this line (`person = None`), it's confusing.

workshops/util.py
```diff
@@ -137,47 +137,29 @@ def verify_upload_person_task(data):
...
-                    # let's check if the task exists
-                    try:
-                        Task.objects.get(event=existing_event, person=person,
-                                         role=existing_role)
-                    except Task.DoesNotExist:
-                        info.append('Task will be created.')
-                    else:
-                        info.append('Task already exists.')
-        else:
+                info.append('Existing record for person will be used.')
```
chrismedrela: It'd be actually nice to display which Person record will be used, especially when there is more than one matching person.

workshops/util.py
```diff
@@ -137,47 +137,29 @@ def verify_upload_person_task(data):
...
+
+        if not email and not person:
             info.append('It\'s highly recommended to add an email address.')
 
         if person:
-            # force username from existing record
+            # force details from existing record
+            item['personal'] = person.personal
+            item['family'] = person.family
+            item['email'] = person.email
```
chrismedrela: In lines 192-194 you make validation using `personal`, but here you use `person.personal`. So when I click "use that person", and then change personal name, and then click "verify", then: 1) my changes are ignored 2) the changed value is validated instead of the one that is actually used (that is, the one from db). The same applies to family and email.

workshops/templates/workshops/person_bulk_add_results.html
```diff
@@ -43,8 +43,19 @@
...
                 <span>{{ entry.role|default:"—" }}</span>
-                <input type="text" name="role" value="{{ entry.role|default:"" }}" class="hidden">
-                </td>
+                <select name="role" class="hidden">
+                {% if entry.role in possible_roles %}
+                    {% for role in possible_roles %}
+                    <option value="{{ role }}" {% if role == entry.role %}selected="selected"{% endif %}>{{ role }}</option>
+                    {% endfor %}
+                {% else %}
+                    <option selected="selected" value="{{ entry.role }}">{{ entry.role }}</option>
```
pbanaszkiewicz: Here's the logic: if user submits a valid option, I choose it. If they submit an invalid option, I show it them and tell them it's a wrong option.
zrzut ekranu z 2016-11-28 18-47-27

pbanaszkiewicz:
@chrismedrela updated - please take a look.

pbanaszkiewicz:
I'm going to merge this; if there are any issues, they will be solved separately.

Issue #1098, opened 2016-11-25
Add a template filter to allow sorting of FKs within templates


narayanaditya95:
@pbanaszkiewicz @chrismedrela 
What do you think about incorporating https://www.djangosnippets.org/snippets/741/ into AMY?

chrismedrela:
@narayanaditya95 where could we use it?

narayanaditya95:
For eg: in the view `person_edit` we pass the context variables `awards` and `tasks` from the view to the template. We can move the entire logic to the templates and use `object.task_set|order_by:"..."` in the templates.

I find it to be much cleaner.

pbanaszkiewicz:
@narayanaditya95 can you provide a proof-of-concept? Thanks!

narayanaditya95:
@pbanaszkiewicz Okay, it seems like DTL has support for sorting relations. See [dictsort](https://docs.djangoproject.com/es/1.9/ref/templates/builtins/#dictsort).

So, we do not need to add a new template filter. I have updated the proof of concept.

PR #1106, opened 2016-11-29
Template filter dictsort

narayanaditya95:
Proof of concept for #1098

Should I continue working on this?

### DIFF
--- a/workshops/templates/workshops/person_edit_form.html
+++ b/workshops/templates/workshops/person_edit_form.html
@@ -2,6 +2,7 @@
 
 {% load crispy_forms_tags %}
 {% load selectable_tags %}
+{% load order_by %}
 
 {% block content %}
 <div>
@@ -18,7 +19,7 @@
   </div>
   <div role="tabpanel" class="tab-pane" id="awards">
     {% crispy award_form %}
-    {% if awards %}
+    {% if object.award_set %}
     <table class="table table-striped">
       <thead>
         <tr>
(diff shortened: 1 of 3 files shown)


pbanaszkiewicz:
> Should I continue working on this?

Please don't. There's a few issues I can see here:

* We're shifting sorting from DB to Python; I would expect the latter to be slower at this.
* By removing `awards` and `tasks` from context variables, you're not able to optimize that page; I trimmed the number of queries from 16 to 13 just by adding `select_related` on these variables.

If you want to remove sorting from our views, maybe a default sorting order for model should be considered?

pbanaszkiewicz:
Ping @narayanaditya95

narayanaditya95:
@pbanaszkiewicz 

I have reverted back from using DTL's offering to a custom template tag that modifies the queryset before evaluation. That should ensure that the sorting takes place within the DB.
However, I am inclined towards declaring a default ordering for respective models.

The main objective of this proposal is to ensure that most templating logic (like the order of events for eg.) stays within the templates.
What is your opinion?

pbanaszkiewicz:
> The main objective of this proposal is to ensure that most templating logic (like the order of events for eg.) stays within the templates.
>
> What is your opinion?

I don't really weigh on either option. Do you specifically want this in ListViews or in any view? Because if you think about ListViews then I'd much rather like another take on DataTables (or similar solution).

narayanaditya95:
Closing the issue. `order_by` is usually called on a queryset along with `select_related` and other queryset modifiers. Moving just one such modifier to the template will offer no benefit.

PR #1111, opened 2016-12-11
Search: switch to GET, disable CSRF protection

### LINKED ISSUE #1069: Search form should use GET method
Benefits:
- the browser won't ask you to confirm form resubmission when refreshing the search result page
- search result page is stateless -- the URL includes the term you search for, so that you can send the link to other people

pbanaszkiewicz:
Since this is a purely GET (no side effects) page, we don't need
the CSRF protection, and it will look better in the address bar.

This commit fixes #1069, and also provides working tests against
that search page.

### DIFF
--- a/workshops/templates/navigation_fixed.html
+++ b/workshops/templates/navigation_fixed.html
@@ -71,8 +71,7 @@
               </ul>
             </li>
           </ul>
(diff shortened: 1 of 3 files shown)


pbanaszkiewicz:
Thanks @narayanaditya95 !

PR #1113, opened 2016-12-12
remove word "imposed"

jduckles:
I probably shouldn't be hacking at the migrations for this, and I think it may need another migration with the way y'all do things. However, this represents a change in the template language around organizational memberships I'd like to see.


pbanaszkiewicz:
@jduckles can you please ensure that you have "Allow edits from maintainers" ticked off? This will allow me to send commits to your branch (and therefore update this PR).

jduckles:
@pbanaszkiewicz Allow edits is ticked.

PR #1115, opened 2016-12-17
Generating list of new certificates with awarders instead of event names

gvwilson:


chrismedrela:
@maneesha how is the CSV generated by `check_certificates` command used?

PR #1120, opened 2016-12-20
Fix the delete button on the TrainingProgressUpdate view.

### LINKED ISSUE #1085: Delete button doesn't work in edit views
This is because delete views redirects to URL with appended `/delete`, i.e. `workshops/training_progress/1172/edit/delete` instead of `https://amy.software-carpentry.org/workshops/training_progress/1172/delete`.

A workaround is to delete objects from detailed view. This workaround doesn't work for TrainingProgresses, because they don't have detailed view (see #1084).

narayanaditya95:
`delete-form` now submits to the delete view of the object.

Fixes #1085

### DIFF
--- a/workshops/templates/workshops/trainingprogress_form.html
+++ b/workshops/templates/workshops/trainingprogress_form.html
@@ -4,7 +4,7 @@
 {% load selectable_tags %}
 
 {% block content %}
-<form action="delete?next={{ request.GET.next|urlencode }}" method="POST" id="delete-form">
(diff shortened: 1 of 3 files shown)


pbanaszkiewicz:
@chrismedrela can you please add an issue for `TrainingProgress` detail view if there isn't one?

apawlik:
Hi, I am trying to delete one a training progress element which I added by mistake and still get "Not found".  I browsed through the related issues in the repo and I'm still confused if it got fixed so just in case - I am still able to generate the error :-| Fiddling with the URL kinda works, though.

pbanaszkiewicz:
@apawlik This bug has been fixed, however AMY version containing that fix was not released / deployed.

PR #1123, opened 2016-12-26
WIP: Allow AMY to create PDF certificates for awards.

### LINKED ISSUE #656: Generate certificates from AMY

aditnryn:
Changes:
- Pull certificates from swcarpentry/certification repo.
- Add cairosvg to requirements.txt
- Add a detail view to exporting PDFs.

Fixes #656 

@gvwilson I fetched certificates from swcarpentry/certification repo. What is the mapping between the badge that a person can be awarded with and the certificate SVG that is printed? Do all awards have a corresponding SVG file?
For eg: there isn't any `creator.svg` or `member.svg`, while SVGs like core.svg and attendance.svg do not have their corresponding badge.


workshops/views.py
```diff
@@ -1514,6 +1516,22 @@ class AwardDelete(RedirectSupportMixin, MockAwardDelete):
...
 
+class AwardCertification(OnlyForAdminsMixin, AMYDetailView):
+    queryset = Award.objects.all()
+
+    def get_template_names(self):
+        return 'workshops/certificates/{}.svg'.format(self.object.badge.name)
+
+    def render_to_response(self, context, **response_kwargs):
+        return HttpResponse(
+            cairosvg.svg2pdf(
```
gvwilson: Hm - are we sure PDF generation will be fast enough that the HTTP response doesn't time out?

workshops/views.py
```diff
@@ -1514,6 +1516,22 @@ class AwardDelete(RedirectSupportMixin, MockAwardDelete):
     pass
 
 
+class AwardCertification(OnlyForAdminsMixin, AMYDetailView):
+    queryset = Award.objects.all()
+
+    def get_template_names(self):
+        return 'workshops/certificates/{}.svg'.format(self.object.badge.name)
```
gvwilson: And if the requested certificate doesn't exist?  (People can hack URLs...)

narayanaditya95:
@gvwilson Can you address this query.

>What is the mapping between the badge that a person can be awarded with and the certificate SVG that is printed? Do all awards have a corresponding SVG file?
>For eg: there isn't any creator.svg or member.svg, while SVGs like core.svg and attendance.svg are never used.

workshops/views.py
narayanaditya95: Yes. After profiling:
`1 loop, best of 3: 222 ms per loop`

gvwilson:
We don't award certificates for being a creator or member, though we 
could if someone wanted to bother to.  The core and attendance SVG files 
were created by people who needed awards so that workshop participants 
could count the class toward university requirements; they're not 
"official".  Simplest solution would be to add a flag to the Badge 
object to say "does this get a certificate?"

narayanaditya95:
> The core and attendance SVG files
were created by people who needed awards so that workshop participants
could count the class toward university requirements; they're not
"official".

So we are fine if AMY does not have the ability to generate such certificates?

> Simplest solution would be to add a flag to the Badge
object to say "does this get a certificate?"

That certainly is the easier way out.

gvwilson:
On 2017-01-01 3:40 AM, Aditya Narayan wrote:

>     The core and attendance SVG files
>     were created by people who needed awards so that workshop participants
>     could count the class toward university requirements; they're not
>     "official".
>
> So we are fine if AMY does not have the ability to generate such 
> certificates?
>
Yes.
>
>     Simplest solution would be to add a flag to the Badge
>     object to say "does this get a certificate?"
>
> That certainly is the easier way out.
>
Agreed - please implement.  We don't have a way to edit badge specs 
right now (at least I don't think we do), so we'll need a migration to 
set this field for all existing badges.

narayanaditya95:
@gvwilson 

> We don't have a way to edit badge specs
right now (at least I don't think we do), so we'll need a migration to
set this field for all existing badges.

I added a Boolean property to the Badge object, instead of a BooleanField. No migration :)

There is something wrong about the `organizer.svg` file. It cannot be rendered with `cairosvg` (even outside of the Django request-response cycle). 

@pbanaszkiewicz How should I go about with testing PDF outputs? Checking the checksum hash of the output against a fixed output?

gvwilson:
But how do we set the value for each existing Badge without a migration?

narayanaditya95:
If the SVG file of the same name exists, the property `has_certificate` returns `True`.

gvwilson:
Urp - I would rather have something explicit in the database, and then 
raise an error if we can't find a matching certificate, since the 
mechanism for awarding certs may change over time.  It also means that 
the database is self-contained - storing key information externally is 
problematic.  @pbanaszkiewicz, your call.

pbanaszkiewicz:
I'd suggest 1-to-1 relation between `Award` and `Certificate` (new model), with hard-coded list of accepted `Badge` objects.

There's one serious advantage with this approach: we can easily check thousands of certificates (if they exist) just by asking the database; can't expect to do that with disk IO (currect implementation).

@narayanaditya95 
> @pbanaszkiewicz How should I go about with testing PDF outputs? Checking the checksum hash of the output against a fixed output?

We can try that; we should cover some corner cases, like missing names or long names. We could generate PDFs locally, confirm that output looks okay and then capture the checksum and compare it in the tests.

narayanaditya95:
@pbanaszkiewicz 

>There's one serious advantage with this approach: we can easily check thousands of certificates (if they exist) just by asking the database; can't expect to do that with disk IO (currect implementation).

The current implementation does not save PDFs to disk, but rather returns them within the request-response cycle.

pbanaszkiewicz:
@narayanaditya95 definitely save then.

narayanaditya95:
@pbanaszkiewicz Any specific reason to save them on disk on AMY server? The time taken to generate PDFs is `1 loop, best of 3: 222 ms per loop`. I think saving them to disk is an unnecessary overhead.

pbanaszkiewicz:
@narayanaditya95 you can generate once for all people, save on disk and be only IO-dependant.

Generation upon request simply doesn't scale well.

narayanaditya95:
>@narayanaditya95 you can generate once for all people, save on disk and be only IO-dependant.

Then we would have to set up a cron job as well. On requesting a certificate, the admin would then have to wait till the cron job is executed, which is anyway worse than having to wait over few seconds for AMY to generate the certificate at the same moment.

>Generation upon request simply doesn't scale well.
 
Maybe future @swcarpentry/amy-maintainers can ponder about it. ;-)

chrismedrela:
@narayanaditya95 are you working on this PR?

>> @narayanaditya95 you can generate once for all people, save on disk and be only IO-dependant.
> 
> We would have to set up a cron job as well. On requesting a certificate, the admin would then have to wait till the cron job is executed, which is anyway worse than having to wait over few seconds (worst case) for AMY to generate the certificate at the same moment.

What about lazy generation of certificates, so that we treat filesystem as a cache rather than storage? When somebody wants to download the certificate, AMY checks if there is already PDF in filesystem and returns it, otherwise it generates PDF, saves to disk and returns the PDF. We may want to be eager when a badge is awarded and force certificate generation. And generate existing certificates.

> I'd suggest 1-to-1 relation between Award and Certificate (new model), with hard-coded list of accepted Badge objects.
> 
> There's one serious advantage with this approach: we can easily check thousands of certificates (if they exist) just by asking the database; can't expect to do that with disk IO (currect implementation).

Why would we ever want to check if certificates exists? We can just check if corresponding Award records exists. I'd rather go with the simpler approach (adding BooleanField to Badge).

workshops/views.py
```diff
@@ -1514,6 +1516,22 @@ class AwardDelete(RedirectSupportMixin, MockAwardDelete):
...
+        return HttpResponse(
+            cairosvg.svg2pdf(
+                render_to_string(
+                    self.get_object().badge.certificate,
+                    self.get_context_data()
+                ).encode('utf-8'),
+                dpi=90,
+            ),
+            content_type='application/pdf',
+        )
```
chrismedrela: Please add a meaningful filename rather than `download.pdf`.

narayanaditya95:
@chrismedrela 

> are you working on this PR?

Yes (was busy for the first week of the semester). I would be happy if we could arrive at the best solution.
 
>What about lazy generation of certificates, so that we treat filesystem as a cache rather than storage? When somebody wants to download the certificate, AMY checks if there is already PDF in filesystem and returns it, otherwise it generates PDF, saves to disk and returns the PDF.

How would we handle cache invalidation? The PDF generated by CairoSVG cannot be parsed as text. The only way to handle cache invalidation would be to add a `modified_at` field to `Award` and check if the PDF was generated before the last time the Award object was modified OR we can provide a manual `Regenerate` button for admins. This still seems like an overkill considering PDF generation take <250ms.
Opinion?

workshops/views.py
narayanaditya95: Done

narayanaditya95:
Funny thing.
Test pass on my Mac OSX 10.12.

```
python manage.py test --keepdb workshops.test.test_badge                                              certificate ✭
/Users/aditya/.virtualenvs/amy/lib/python3.6/site-packages/six.py:808: RemovedInDjango110Warning: SubfieldBase has been deprecated. Use Field.from_db_value instead.
  return meta(name, bases, d)

Using existing test database for alias 'default'...
......x.
Ran 8 tests in 1.787s

OK (expected failures=1)
Preserving test database for alias 'default'...
```

But they fail on Travis running Ubuntu. :(

chrismedrela:
@narayanaditya95 you're right, I forgot about cache invalidation which makes it more complicated.

@narayanaditya95 could you add tests:
- when a person doesn't have a badge, he cannot download related certificate
- when a badge is awarded, and then cancelled, it's not possible to download the no-longer-valid certificate.
- a certificate cannot be downloaded by a trainee (only by admins), even if it's his/her certificate.

@pbanaszkiewicz what about sticking with the current approach (generating certificates on each request) and worry about storing them in files when this approach stops scale? I don't think we'll hit this point quickly. Any other approach would be much more complicated.

> If the SVG file of the same name exists, the property has_certificate on every Badge object returns True.
Definitively add BooleanField to Badge for the reasons gvwilson raised.

>  If you use Ubuntu, can you run the tests on your machine?
Yes, they failed, and I got different md5s than Travis CI. Maybe let's test if SVG is rendered correctly and ignore PDFs? I don't see any other approach than to test PDFs manually with your own eye.

And, a minor issue, what about `DownloadCertificate` instead of `AwardCertification`?

narayanaditya95:
>@narayanaditya95 could you add tests:
> - when a person doesn't have a badge, he cannot download related certificate
> - when a badge is awarded, and then cancelled, it's not possible to download the no-longer-valid certificate.
> - a certificate cannot be downloaded by a trainee (only by admins), even if it's his/her certificate.

- There wouldn't exist any URL that could return a certificate for the user. -1.
- Awards can only be deleted. The same download link would return a 404 then. -1.
- +1. On it.

>Definitively add BooleanField to Badge for the reasons gvwilson raised.

You mean replacing the `has_certificate` property by a ` has_certificate = models.BooleanField`, right?

> Yes, they failed, and I got different md5s than Travis CI. Maybe let's test if SVG is rendered correctly and ignore PDFs? I don't see any other approach than to test PDFs manually with your own eye.

Maybe, I will write simple test for HTTP 200 for the views. :/

>And, a minor issue, what about DownloadCertificate instead of AwardCertification?

+1

chrismedrela:
> - There wouldn't exist any URL that could return a certificate for the user. -1.
> - Awards can only be deleted. The same download link would return a 404 then. -1.

You're right.

>> Definitively add BooleanField to Badge for the reasons gvwilson raised.
> 
> You mean replacing the has_certificate property by a has_certificate = models.BooleanField, right?

Yes

>> Yes, they failed, and I got different md5s than Travis CI. Maybe let's test if SVG is rendered correctly and ignore PDFs? I don't see any other approach than to test PDFs manually with your own eye.
>
> Maybe, I will write simple test for HTTP 200 for the views. :/

Yes, that's a good idea, and you can check if a file is returned.

pbanaszkiewicz:
@narayanaditya95:
> How would we handle cache invalidation? The PDF generated by CairoSVG cannot be parsed as text. The only way to handle cache invalidation would be to add a modified_at field to Award and check if the PDF was generated before the last time the Award object was modified OR we can provide a manual Regenerate button for admins. This still seems like an overkill considering PDF generation take <250ms.
Opinion?

@chrismedrela:
> @pbanaszkiewicz what about sticking with the current approach (generating certificates on each request) and worry about storing them in files when this approach stops scale? I don't think we'll hit this point quickly. Any other approach would be much more complicated.

Hi both, if you don't want to change the behavior too much then there's this simple-ish solution:

* make `Award` (or `Certificate` if we have one) a hashable object
* upon download request, check for `{hash}.pdf` on the disk
* if there's one, read it to the browser
* if there isn't one, generate it, save on the disk and read it to the browser

What do you think? Assuming we don't hit some terrible collisions and that we have infinite disk space, this could be solid.

@narayanaditya95:
> * There wouldn't exist any URL that could return a certificate for the user. -1.
> * Awards can only be deleted. The same download link would return a 404 then. -1.

Please make tests for the cases Chris mentioned. We want to make sure that we don't accidentally "enable" these "features" in the future. So in case of both, simple tests ensuring 404 would suffice.

@chrismedrela, @narayanaditya95:
> Yes, they failed, and I got different md5s than Travis CI. Maybe let's test if SVG is rendered correctly and ignore PDFs? I don't see any other approach than to test PDFs manually with your own eye.

Can we test SVGs only, not the resulting PDFs?

@chrismedrela:
> Yes, and make Badge editable in Django Admin.

Badge is already editable in Django admin.

chrismedrela:
> Hi both, if you don't want to change the behavior too much then there's this simple-ish solution:
> 
> * make `Award` (or `Certificate` if we have one) a hashable object
> * upon download request, check for `{hash}.pdf` on the disk
> * if there's one, read it to the browser
> * if there isn't one, generate it, save on the disk and read it to the browser
> 
> What do you think? Assuming we don't hit some terrible collisions and that we have infinite disk space, this could be solid.

:+1: just make sure that hash changes if award.awarded_by.get_full_name() changes

narayanaditya95:
@pbanaszkiewicz I realized that AMY does not use any sort of caching. Is it because you never felt the need for it?

chrismedrela:
@narayanaditya95 I cannot speak for @pbanaszkiewicz, but I've never needed caching in AMY.

narayanaditya95:
@chrismedrela The problem here is django does not allow me to define a FileSystem-based cache named `certificate`, without defining a `default` cache. So, I ended up adding a `default` memcache. 
Do you any other way out?

pbanaszkiewicz:
We never needed any cache. @narayanaditya95 did you try [this](https://github.com/swcarpentry/amy/pull/1123#issuecomment-274316796) simple cache?

pbanaszkiewicz:
A corresponding issue #656 was closed.

PR #1125, opened 2017-01-07
Event form tags: bigger list; resizable <select>s

### LINKED ISSUE #898: Better format to select tags
Can we have a better format to select tags?  The current scrolling list makes it difficult to see which tags are selected when in 'edit' mode.

pbanaszkiewicz:
This fixes #898 by:

* changing default size of "Tags" field on Event add/edit forms to 10
  (the current number of tags in the database),
* adding "resize: vertical" option to CSS3 for <select> tags; therefore
  making any <select> resizable in vertical if user wants to just see
  more of the options.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -10,6 +10,7 @@ from django.core.exceptions import ValidationError
 from django.core.validators import RegexValidator
 from django.forms import (
     HiddenInput,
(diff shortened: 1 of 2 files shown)


chrismedrela:
Looks good.

PR #1126, opened 2017-01-07
Contact selected staff

### LINKED ISSUE #934: Select people to email from "Find staff" page
When I run a search on this page I'd like to be able to select email addresses from the search results.  
Can we have a check box next to each name with a button "Email selected people" or "Select all" that will return a list of all their email addresses?

pbanaszkiewicz:
This fixes #934.

This is a pure JavaScript solution, and it only works on a single results page (ie. checkboxes aren't saved across different pages).


pbanaszkiewicz:
@chrismedrela submitted for a re-review.

pbanaszkiewicz:
>  Even I wasn't aware of many things under amy_utils.js.
> Good work @chrismedrela

Yes indeed, @chrismedrela the plugins in `amy_utils.js` look 10× better than my previous JS code.

Issue #1110, opened 2016-12-07
cancelled workshops should not show up as not invoiced in dashboard


maneesha:
Title says it all.
This probably relates to #1064

Issue #1064, opened 2016-10-25
"Cancelled" workshops should not show up in "Unpublished" column


maneesha:
Title says it all.

PR #1127, opened 2017-01-07
Speed up person details & update, and admin dashboard pages

pbanaszkiewicz:
On a test person (@gvwilson) profile pages get to ~20 queries from >220.

Similarly dashboard is now O(const) instead of O(n) due to annotation of `num_instructors`.

### DIFF
--- a/workshops/views.py
+++ b/workshops/views.py
@@ -443,7 +443,10 @@ class PersonDetails(OnlyForAdminsMixin, AMYDetailView):
                 output_field=IntegerField()
             )
         )
-    )
+    ).prefetch_related(
+        'award_set__badge', 'award_set__awarded_by', 'award_set__event',
+        'task_set__role', 'task_set__event',
+    ).select_related('airport')
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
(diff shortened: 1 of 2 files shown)


workshops/views.py
```diff
@@ -183,8 +183,15 @@ def admin_dashboard(request):
...
-                                      .select_related('host')
+    unpublished_events = Event.objects \
+        .active().unpublished_events().select_related('host').annotate(
+            num_instructors=Count(
+                Case(
+                    When(task__role__name='instructor', then=Value(1)),
+                    output_field=IntegerField()
+                )
+            ),
+        )
```
pbanaszkiewicz: This actually produces wrong number of instructors for events that have "cancelled" tag. This is due to `unpublished_events` having `Q(tags__name='cancelled')`.

chrismedrela: This is not a problem since we'll filter out cancelled events (#1064). Just add a comment explaining that this annotation produces wrong number for cancelled events and that it's ok, because `unpublished_events()` filters out these events.

pbanaszkiewicz:
@chrismedrela not sure if you have the timeline here right; this (#1127) was submitted first, then I commented about the bug and then I fixed it (#1130) since apparently it was reported often (#1064, #1110, #1128).

I decided to add a comment anyway.

PR #1131, opened 2017-01-12
Make select widget filtering events by tags bigger

chrismedrela:
Extension to #1125.

### DIFF
--- a/workshops/forms.py
+++ b/workshops/forms.py
@@ -40,6 +40,7 @@ from workshops.models import (
     TrainingRequest,
     DCSelfOrganizedEventRequest,
     TrainingProgress,
+    Tag,
 )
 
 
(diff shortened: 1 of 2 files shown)


workshops/filters.py
```diff
@@ -119,6 +119,13 @@ class EventFilter(AMYFilterSet):
     ]
     status = EventStateFilter(choices=STATUS_CHOICES)
 
+    tags = django_filters.ModelMultipleChoiceFilter(
```
pbanaszkiewicz: This is disputable, but I left this out by purpose, so that the sidebar doesn't take one million pixels of vertical space.

chrismedrela: ok, we can left it out since select widgets are resizable.

chrismedrela:
> Have you tried taking this further? For example, creating our own `SelectWidget` that reads `ITEMS_VISIBLE_IN_SELECT_WIDGET` from associated object?

We can refactor it in this way later, when we use it in more than one model. Now, we use it in only one model, so I don't think we need additional layers of indirection.

> Maybe also switch from Model.CONSTANT to Model.Meta.variable?

You cannot add your own variables to `Model.Meta` (although [it's a requested feature](https://code.djangoproject.com/ticket/5793)).

PR #1133, opened 2017-01-17
Install django_extensions

chrismedrela:
`django_extensions` provide `shell_plus --notebook` command that open jupyter notebook in a browser, which is IMO much more convenient than command line shell.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,4 +22,3 @@ django-app-namespace-template-loader>=0.4
 django-webtest==1.8.0
 django-debug-toolbar==1.5
 django-extensions>=1.7,<1.8
-jupyter>=1.0,<1.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,3 +21,5 @@ Requests-OAuthlib>=0.6.1,<0.7
 django-app-namespace-template-loader>=0.4
 django-webtest==1.8.0
 django-debug-toolbar==1.5
+django-extensions>=1.7,<1.8
+jupyter>=1.0,<1.1
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -158,6 +158,7 @@ INSTALLED_APPS += [
     'compressor',
     'social.apps.django_app.default',
     'debug_toolbar',
+    'django_extensions',
 ]
 
 CRISPY_TEMPLATE_PACK = 'bootstrap3'


requirements.txt
```diff
@@ -21,3 +21,5 @@ Requests-OAuthlib>=0.6.1,<0.7
 django-app-namespace-template-loader>=0.4
 django-webtest==1.8.0
 django-debug-toolbar==1.5
+django-extensions>=1.7,<1.8
+ipython[notebook]>=5.1,<5.2
```
pbanaszkiewicz: Isn't this `jupyter[notebook]` nowadays?

narayanaditya95:
@chrismedrela I usually use the IPython shell. Is there a specific reason to switch from the terminal to the brower?

pbanaszkiewicz:
@narayanaditya95 for me the reason is mostly in `shell_plus` command itself, but `django_extensions` may contain some other nice things.

chrismedrela:
> @chrismedrela I usually use the IPython shell. Is there a specific reason to switch from the terminal to the brower?

- Editing in browser is much more convenient than in console (especially if you have multiline snippet of code).
- Your session is saved in a notebook. A few days ago I created a notebook named issue-859 where I put experimental code needed to fix that issue. I love that today (when I came back to this issue) I could open this notebook and execute this code again.
- I can delete all unnecessary snippets of code. Or edit them in-place.
- If you want to restart shell, you just restart kernel and all your snippets of code are preserved.

chrismedrela:
Additionally, django-extensions introduce `runserver_plus` that give you access to REPL in the case of an exception, so that you can debug in browser.

pbanaszkiewicz:
My only concern is that we're forcing everyone to install jupyter even though it's strictly for Chris's development ease. We could add `dev_requirements.txt`.

narayanaditya95:
>My only concern is that we're forcing everyone to install jupyter even though it's strictly for Chris's development ease. We could add dev_requirements.txt.

But then that would require manually adding `django_extensions` to the list of `INSTALLED_APPS`.

chrismedrela:
> My only concern is that we're forcing everyone to install jupyter even though it's strictly for Chris's development ease. 

First of all, you can use these tools too. :-) But even if it's only for me, why is it a problem that we force everyone to install jupyter? It doesn't take a lot of time to install it.

narayanaditya95:
@chrismedrela I think you can remove `jupyter` from `requirements.txt`. If someone is interested in using jupyter notebook, they can install it over the existing virtualenv.

PR #1138, opened 2017-01-18
Fix #1107 -- validate GitHub usernames

### LINKED ISSUE #1107: Validating GitHub user names
If AMY is going to allow login with GitHub usernames then it needs to validate for this.  I believe that only alphanumeric and underscores are allowed but  I can't find documentation on this.

chrismedrela:
Fixes #1107.


pbanaszkiewicz:
Please hold on with merging until we figure out in issue #1107 what to do about invalid GitHub usernames.

narayanaditya95:
Suggestion. You may want to use `PyGitHub`'s [get_user](http://pygithub.readthedocs.io/en/latest/github.html#github.MainClass.Github.get_user) method to verify is the user exists.

chrismedrela:
> Suggestion. You may want to use PyGitHub's get_user method to verify is the user exists.

Let's do this in another issue: #1139

> Any particular reason to define a field in itself, rather than adding a RegexValidator to the existing fields?

No particular reason except that in the future it'll be easier to add a new behavior (i.e. adding new validation rule, like verifying if the user exists) to all GitHub fields.

pbanaszkiewicz:
So what is actually blocking us here?

chrismedrela:
It's blocked by corresponding issue, that is #1107. This issue needs feedback from @maneesha regarding what to do with existing records with invalid GitHub usernames.

pbanaszkiewicz:
Superseded by #1281.

PR #1143, opened 2017-01-21
Add a lookup on the `github` field.

narayanaditya95:
GitHub username is saved only after it is selected from the lookup.


narayanaditya95:
@chrismedrela This build upon your PR #1138. I hope you don't mind.

workshops/lookups.py
```diff
@@ -5,13 +5,22 @@
...
 
 from workshops import models
 from workshops.util import lookup_only_for_admins
 
 
+class GitHubUserLookup(LookupBase):
+    gh = Github()
+
+    def get_query(self, request, username):
+        users = self.gh.search_users(username)[:10]
```
chrismedrela: Please refactor this magic number into a constant.

workshops/forms.py
```diff
@@ -490,6 +477,17 @@ class Meta:
...
                   'twitter', 'url', 'occupation', 'orcid', 'notes', 'lessons',
                   'domains', 'languages']
+        widgets = {
+            'airport': selectable.AutoComboboxSelectWidget(
+                lookup_class=lookups.AirportLookup,
+            ),
+            'languages': selectable.AutoComboboxSelectMultipleWidget(
+                lookup_class=lookups.LanguageLookup,
+            ),
+            'github': selectable.AutoCompleteSelectWidget(
```
chrismedrela: Is there any reason for using `AutoCompleteSelectWidget` instead of `AutoComboboxSelectWidget`?

narayanaditya95: We don't need a button to list GH usernames without any input. So I opted for `AutoCompleteSelectWidget` over `AutoComboboxSelectWidget`.

narayanaditya95:
@chrismedrela 

>Here are my concerns:
> - If you enter username like chrismedrelablablablabla, it throws exceptions.

Fixed.

> - When you enter a not existing username into the autocomplete field, so that it's red, and submit the form, invalid value is ignored and NULL is saved, but no warning is displayed. In such case, as an user, I'd think the invalid username was saved. I think we should check if the value is correct (on frontend) and prevent sending the form if it's not. This applies to all autocomplete fields. This should probably land in a new issue.

Yes, that is an issue. When the autocomplete fields are marked as `required=True`, the frontend ensures that a valid value exists. Can we consider making the `github` field mandatory?

> - We'll make a lot of requests to GitHub API. Are you sure we won't exceed the limit?

Yes, that is a problem (a dealbreaker maybe). GitHub docs state that _The Search API has a custom rate limit. For requests using Basic Authentication, OAuth, or client ID and secret, you can make up to 30 requests per minute. For unauthenticated requests, the rate limit allows you to make up to 10 requests per minute._ 

I added a `minLength` attribute and increased the `delay` (default is 300ms) to 500ms to ensure that we do not generate a lot of requests and switched to authenticated requests.

chrismedrela:
>> When you enter a not existing username into the autocomplete field, so that it's red, and submit the form, invalid value is ignored and NULL is saved, but no warning is displayed. In such case, as an user, I'd think the invalid username was saved. I think we should check if the value is correct (on frontend) and prevent sending the form if it's not. This applies to all autocomplete fields. This should probably land in a new issue.
> 
> Yes, that is an issue. When the autocomplete fields are marked as required=True, the frontend ensures that a valid value exists. Can we consider making the github field mandatory?

Probably not -- if we make this field mandatory, you cannot edit Person records that don't have GitHub username (that is most of them). I've created a new issue to handle it: #1154.

When rate limits are exceeded, `Github.search_users` raises `GithubException` which results in 500 Internal Server Error (just open in your browser http://127.0.0.1:8000/selectable/workshops-githubuserlookup/?term=gggg&timestamp=1489578524852 multiple times). Please fix it (handle `GithubException` and add logging, so that we can later check if we had ever exceeded limits) and add regression test. @pbanaszkiewicz do we have any logs configured?

chrismedrela:
Hint: In the regression test, you need to mock `Github.search_users`. You can do this in the following way (I didn't test the following code):
```python
from unittest.mock import patch
class GitHubUserLookupTests(TestBase):
    @patch('workshops.lookups.Github.search_users')
    def test_rate_limits_exceeded(self, search_users_mock):
        search_users_mock.side_effect = GithubException
        # your code that queries /selectable/workshops-githubuserlookup/
```

pbanaszkiewicz:
@chrismedrela 
> @pbanaszkiewicz do we have any logs configured?

We have some, what do you need?

chrismedrela:
@pbanaszkiewicz 
1. how can you log messages - is it the same way as described in [Django doc](https://docs.djangoproject.com/en/1.10/topics/logging/#using-logging)?
2. how can I access logs on production?

chrismedrela:
@pbanaszkiewicz you moved this PR to v1.11. I propose that v1.11 is only about bugfixes (because this is what Maneesha expects 'till the end of April) and move it to backlog (that is, attach to no milestone).

pbanaszkiewicz:
@chrismedrela 
> @pbanaszkiewicz 
>
> 1. how can you log messages - is it the same way as described in [Django doc](https://docs.djangoproject.com/en/1.10/topics/logging/#using-logging)?
> 2. how can I access logs on production?

We're only storing Nginx logs, look for `~/amy_site/amy.log`.

> @pbanaszkiewicz you moved this PR to v1.11. I propose that v1.11 is only about bugfixes (because this is what Maneesha expects 'till the end of April) and move it to backlog (that is, attach to no milestone).

I was going to triage this PR later since there was no milestone v1.12, but yeah I'll remove the milestone.

PR #1144, opened 2017-01-21
Fix #1109 -- make airport code insensitive in profile update requests

### LINKED ISSUE #1109: airport IATA code should be case insensitive
title says it all

chrismedrela:
Fixes #1109.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -688,5 +688,6 @@ class AirportViewSet(viewsets.ReadOnlyModelViewSet):
     permission_classes = (IsAuthenticated, IsAdmin)
     queryset = Airport.objects.all()
     serializer_class = AirportSerializer
-    lookup_field = 'iata'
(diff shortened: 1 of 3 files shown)


api/views.py
```diff
@@ -688,5 +688,6 @@ class AirportViewSet(viewsets.ReadOnlyModelViewSet):
     permission_classes = (IsAuthenticated, IsAdmin)
     queryset = Airport.objects.all()
     serializer_class = AirportSerializer
-    lookup_field = 'iata'
+    lookup_field = 'iata__iexact'
+    lookup_url_kwarg = 'iata'
```
narayanaditya95: Do we need case insensitive URLs?

chrismedrela: No, but that was suggested by @pbanaszkiewicz

PR #1146, opened 2017-01-21
Fix #1047 -- optimize SQL queries on trainees list view

### LINKED ISSUE #1047: Optimize SQL queries on trainees list view
On production data, we make ~200 queries in 600 ms on my local machine, which is way too much.

chrismedrela:
So that it always results in constant number of queries (16) and all queries execute in 500 ms with debug toolbar enabled (which makes everything much, much slower).

@narayanaditya95 or @pbanaszkiewicz please review.

### DIFF
--- a/workshops/templates/workshops/all_trainees.html
+++ b/workshops/templates/workshops/all_trainees.html
@@ -66,7 +66,7 @@
           {% include "workshops/training_progresses_inline.html" with person=trainee %}
         </td>
         <td>
-          {% for task in trainee.get_training_tasks %}
+          {% for task in trainee.training_tasks %}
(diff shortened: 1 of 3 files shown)


workshops/templates/workshops/all_trainees.html
```diff
@@ -66,7 +66,7 @@
           {% include "workshops/training_progresses_inline.html" with person=trainee %}
         </td>
         <td>
-          {% for task in trainee.get_training_tasks %}
+          {% for task in trainee.training_tasks %}
```
pbanaszkiewicz: Can you fix other places that use `get_training_tasks`?

chrismedrela: Is `3919d60` what you ask for?

PR #1147, opened 2017-01-22
Fix #1005 -- allow prepopulation of group name in training request form

### LINKED ISSUE #1005: Allow pre-population of Group in training request form
It would be nice to be able to allow a pass through of the group_name in order to pre-populate the form with the appropriate group name. I have a group that would like to open up applications in their organization and then be able to go through a process in conjunction with us to help select the instructors who will be trained.

This feature would allow us to sent out a URL which is something like:
`https://amy.software-carpentry.org/workshops/request_training/?id_group_name=Thisisatest`

From my read of [this stackoverflow](http://stackoverflow.com/questions/25173991/how-to-pre-populate-django-form-with-part-of-url) topic I think it would be a slight tweak to the View to make it happen.

chrismedrela:
Fixes #1005.

### DIFF
--- a/extforms/test/test_training_request_form.py
+++ b/extforms/test/test_training_request_form.py
@@ -6,7 +6,6 @@ from workshops.test import TestBase
 
 class TestTrainingRequestForm(TestBase):
     def setUp(self):
-        self._setUpUsersAndLogin()
         self._setUpRoles()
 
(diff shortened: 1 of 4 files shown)


pbanaszkiewicz:
Please go ahead and merge.

PR #1156, opened 2017-03-16
Fix #1141 -- 500 Error when GitHub username contains a space

### LINKED ISSUE #1141: GitHub username containing spaces results in Internal Server Error on person detailed view
Check it out https://amy.software-carpentry.org/workshops/person/10483/

We get an exception when we ask GitHub API for user ID and we pass invalid GitHub username. This doesn't result in the exception for other invalid usernames. We ask GitHub API to check if SocialAuthTable is in sync and if the person can login using their GitHub account.

chrismedrela:
Fixes #1141.


pbanaszkiewicz:
@chrismedrela updated with validators applied to GitHub username fields in models. All the rest looks good. Merge when you ready / comment if my commits doesn't look right.

pbanaszkiewicz:
@chrismedrela:
> @pbanaszkiewicz there is one drawback of enforcing correct GitHub usernames - you cannot edit existing records with invalid GitHub nicks. In production, there are two such people, four closed profile update requests and two training requests. We need to fix that records before we merge your changes.

You can edit them, you just need to fix the errors ;) How are you going to fix the records from programming point of view? This needs human assistance to decide what to do about the values.

I'd say we should email affected people and ask for selecting one GH username.

PR #1168, opened 2017-03-28
Add support for PostgreSQL

chrismedrela:
In the long run, we'd like to switch from SQLite3 to PostgreSQL. However, AMY doesn't work on PostgreSQL due to some backend-specific bugs. This PR:

- fixes these bugs,
- let you use any backend, not only SQLite3, by replacing `AMY_DB_FILENAME` with `DATABASE_URL` env var,
- configures Jenkins CI to run tests on both SQLite3 and PosgreSQL, so that we don't introduce new backend-specific bugs.

Additionally, we need to truncate language names to 40 characters on production (see "fix db bug in migration no 0097").


chrismedrela:
I've reverted this PR, because it broke tests.

chrismedrela:
See #1203 for continuation.

PR #1180, opened 2017-04-13
Fix #1176: Working  functionality.

### LINKED ISSUE #1176: update from URL is not working
...fails silently

narayanaditya95:
Fix _id_s of inputs.

### DIFF
--- a/workshops/static/update_from_url.js
+++ b/workshops/static/update_from_url.js
@@ -21,13 +21,13 @@ $('#update_url_form').submit(function(e) {
     switch (action) {
       case 'overwrite':
         console.log('overwrite');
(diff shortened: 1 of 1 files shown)


chrismedrela:
@narayanaditya95 thank you for quick response. The only thing that I'd improve is the commit message, but this is a minor issue so let's merge it.

chrismedrela:
@narayanaditya95 and I think it would be better to stick with "status: needs code review" label instead of "request a review" functionality.

chrismedrela:
Ref #1176.

@narayanaditya95 one more thing, please always duplicate the issue ID in the description, otherwise it's not crosslinked. Sorry for being so nitpicky.

PR #1182, opened 2017-04-16
Fix #1175 -- fix search page

### LINKED ISSUE #1175: search page is not working
Search from nav bar works.
Search from here: https://amy.software-carpentry.org/workshops/search/ does not.

chrismedrela:
The form on the search page didn't work, because it used POST instead of GET.

Fixes #1175.


chrismedrela:
@narayanaditya95 and now?

narayanaditya95:
Thank you @chrismedrela

PR #1183, opened 2017-04-16
Fix #1044 -- replace contains with icontains

### LINKED ISSUE #1044: Replace contains with icontains
We use `contains` instead of `icontains` in many places. On SQLite, [both matches are case-insensitively](https://docs.djangoproject.com/el/1.10/ref/databases/#substring-matching-and-case-sensitivity), but this will result in bugs when we switch to i.e. PostgreSQL.

chrismedrela:
Fixes #1044.

### DIFF
--- a/pydata/views.py
+++ b/pydata/views.py
@@ -66,7 +66,7 @@ class BaseImport(OnlyForAdminsMixin, View):
         try:
             url = request.GET['url']
             conf_url, pk = self.get_pk(url).groups()
-            event = Event.objects.get(url__contains=conf_url)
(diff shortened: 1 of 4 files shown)


chrismedrela:
@narayanaditya95, no all tests pass on PosgreSQL, so I wrote one that failed and now passes on PostgreSQL. If you want to run tests on PostgreSQL, rebase against `ba924d0` (see #1168), configure PostgreSQL and execute:

    DATABASE_URL=postgres://postgres:password@localhost/testdb ./manage.py test

chrismedrela:
@narayanaditya95 please merge if you like it.

PR #1188, opened 2017-04-27
Fix #1076 -- make family name optional

### LINKED ISSUE #1076: Allow null family name
Sometimes people identify themselves by only one name.
It looks like this was historically allowed as we do have people without a family name in the database, but new records require a family name.

chrismedrela:
Fixes #1076.


chrismedrela:
@narayanaditya95 thank you for checking this PR and merging

PR #1194, opened 2017-06-09
Set Airport model to order objects by IATA code

### LINKED ISSUE #1193: Airports should be sorted
I just went to update my user profile, and having moved, needed to update my nearest airport. Unfortunately, the drop-down list of airports seems to be listed in database insertion order (rather than alphabetical), so I had to meticulously read each of the dozens of entries to find my regional airport.

It should be pretty easy to update the Django Airport model [here](https://github.com/swcarpentry/amy/blob/develop/workshops/models.py#L213) to [sort](https://docs.djangoproject.com/en/1.11/ref/models/options/#ordering) by IATA code by default. Something like this:

```python
class Airport(models.Model):
    ...
    class Meta:
        ordering = ['iata']
```

jameshiebert:
Fixes #1193


chrismedrela:
Closed in favour of #1205.

PR #1217, opened 2017-10-01
Documentation - design notes

pbanaszkiewicz:
Hello all,
this PR covers basics of server-side of AMY deployment; documentation on AMY internal design has also been started.

I'd like to hear any feedback on the server-side. Thanks.


pbanaszkiewicz:
I added a note about future fate of these incomplete DB models docs. I didn't want to remove them in case someone decides to use them for the docstrings.

Issue #1233, opened 2018-03-29
Running make dev_database fails with "TypeError: 'gender' is an invalid keyword argument for this function"


askingalot:
Looks like changes to the `TrainingRequest` model were not reflected in the `fake_database.py` file.

Specifically, `gender`, `gender_other` and `additional_skills` were removed from the model and not removed from the fake_database.py file. Also `underresourced`, `underrepresented`,`nonprofit_teaching_experience`, `training_completion_agreement`, `workshop_completion_agreement` and `notes` were not being set in the `fake_databse.py` file

pbanaszkiewicz:
Closing via #1234.

PR #1234, opened 2018-03-29
fix for TrainingRequest in Fake Database

askingalot:
Fix for #1233


pbanaszkiewicz:
Hello @askingalot. I tested it and it worked well. Thank you for your contribution!

Issue #1210, opened 2017-08-03
Responsive CSS issue


jduckles:
At intermediate window sizes it seems the login and + widgets flow to a second row and overlap half the person name. Very minor, but mildly annoying if you're at an intermediate window size.

![image](https://user-images.githubusercontent.com/119403/28947484-3a3b5a88-7904-11e7-81f7-7912e788782d.png)

pbanaszkiewicz:
@jduckles I'm deferring this issue for now, because I want to make a slight redesign of AMY in near future. However, if this issue is very annoying, please let me know and I'll reopen @askingalot's PR #1235 and merge it.

jduckles:
No worries.

PR #1235, opened 2018-03-29
fixing nav covering content issue #1210

askingalot:
Fix for #1210

### DIFF
--- a/workshops/static/css/amy.css
+++ b/workshops/static/css/amy.css
@@ -90,6 +90,11 @@ pre {
 body {
     padding-top: 60px;
 }
+@media screen and (min-width: 768px) and (max-width: 1199px) {
+    body {
+        padding-top: 120px;
+    }
+}
 
 /* when using nav-tab (http://getbootstrap.com/javascript/#tabs) move the
    crispy form down a few notches */


pbanaszkiewicz:
Hello @askingalot. Thank you for your contribution.

Unfortunately there may be a case when we have low resolution, but the top navbar height is correct (60px). With this solution, we'd end up with a blank gap just below the top navbar in such case.

Is it possible to condition `body`'s `padding-top` not only on screen width, but also navbar height, @askingalot?

askingalot:
@pbanaszkiewicz I'll take a look.

askingalot:
@pbanaszkiewicz, I just did a round of testing based on your feedback. I basically mimicked different resolutions with the browser zoom feature, and I used the mobile view in the browser dev tools. I did see a couple of zoom levels that had a blank gap at the top. I think this (at least roughly) reproduces the issue you raised.

I don't know of a way in pure CSS to set the padding of an element based on the height of another. If you know of one, I'd love to hear about it. All I can think to do is implement a solution in javascript. I'll give it a shot.

BTW, I also noticed that the my solution doesn't take the Filter sidebar into consideration, so I can look into pushing it down when the navbar gets bigger too.

pbanaszkiewicz:
Hello @askingalot,

> @pbanaszkiewicz, I just did a round of testing based on your feedback. I basically mimicked different resolutions with the browser zoom feature, and I used the mobile view in the browser dev tools. I did see a couple of zoom levels that had a blank gap at the top. I think this (at least roughly) reproduces the issue you raised.

Exactly.

> I don't know of a way in pure CSS to set the padding of an element based on the height of another. If you know of one, I'd love to hear about it. All I can think to do is implement a solution in javascript. I'll give it a shot.

I don't know anything like that either.

> BTW, I also noticed that the my solution doesn't take the Filter sidebar into consideration, so I can look into pushing it down when the navbar gets bigger too.

I plan to make a general AMY dashboard redesign (something like this: http://coderthemes.com/velonic_3.0/admin_3/), and the biggest changes would be:
1. navigation bar and sidebar switched
2. darker colors for navigation

I hope we'd end up with cleaner look, nicer navigation and easier to maintain templates (e.g. now we have some views fluid and some non-fluid).

For now I'd say we drop this PR if we're unable to fix this issue.

askingalot:
That’s fine. If you’re gonna do a redesign soon, i agree it’s not worth continuing to work on this.

Issue #1229, opened 2018-02-09
Creating new badges for mentors/mentees


jduckles:
I'd like to create new badges to track the mentorship program.

Is there any problem with me making the following insertions into the AMY DB to create new badges? Or is there an interface I might be missing that would be a cleaner way to do this?

```
INSERT INTO  "workshops_badge" ("name", "title", "criteria", "id") VALUES ('mentor', 'Mentor', 'Mentoring others in the mentorship program', '8');
INSERT INTO  "workshops_badge" ("name", "title", "criteria", "id") VALUES ('mentee', 'Mentee', 'Being mentored through the mentorship program', '9');

```

pbanaszkiewicz:
@jduckles it's not a proper way. It would only be visible in the AMY instance where you made that change.
The proper way would be via [data migration](https://docs.djangoproject.com/en/2.0/topics/migrations/#data-migrations).

jduckles:
Closing as #1241 addresses it.

PR #1241, opened 2018-05-06
Adding badges to track mentorship participation

jduckles:
Based on #1229 @pbanaszkiewicz suggested creating a migration for making new mentorship badges. 

Please review closely. First time making a migration, don't want to break anything 😬.


pbanaszkiewicz:
Thanks, @jduckles!

Issue #1244, opened 2018-05-09
Add "Share Profile on Website" checkbox


jduckles:
Very similar to the "May Contact" checkbox, we'd like to have a checkbox in the user's profile to have their name and social links (Twitter, ORCID, GitHub, URL) included/excluded from the public website. We're going to make a push to get people to update their profiles and consent to new Privacy Policy when doing so. Would this addition be possible on a relatively short timeline?

Looking through where `may_contact` is listed, it looks like we'd add this to the model for `Person`:
`publish_profile = models.BooleanField(default=False)`

From my digging around, I think we'll need to add it to the `PersonForm`, `AutoUpdateProfileForm` and `PersonsMergeForm`. Maybe other places too. 

Then we'll need it accessible as a column via the API so we can feed the information to a JSON or YAML file that builds the website.

jduckles:
@pbanaszkiewicz pinging for your input.

pbanaszkiewicz:
@jduckles:
> Then we'll need it accessible as a column via the API so we can feed the information to a JSON or YAML file that builds the website.

Two questions:
1. I just want to make sure: Carpentries website uses `/api/v1/export/instructors/` endpoint, right?
2. we can easily add `publish_profile` column to the JSON/YAML output, but it's open for reading so I think I'll filter out all `publish_profile=False` - but then there's no point in adding the `publish_profile` column.

For now I'm assuming 1. true and 2. no additional column

pbanaszkiewicz:
@jduckles with #1251 we have two fields:
* `may_contact` - user allows to be contacted
* `publish_profile` - user allows for their profile to appear on The Carpentries website.

If I recall correctly, `may_contact` indicates willingness to being contacted in other manners than purely for organization purposes (or am I wrong? Can't find the source for this claim). In Poland I think services allow themselves to contact users via their service policies, but user needs to explicitly accept (a separate checkbox) to be contacted for marketing purposes.

Anyway, I can add both `publish_profile` and `may_contact` fields to the filters for the endpoints you mentioned.

jduckles:
Yes, please add them both as filters. To my mind may contact allows us to reach out to people who's information we have. It is an explicit opt-in to them providing their email address to us with the assumption that we will use it to contact them. We should not ever use an email record in AMY to send an email to someone who has a "may_contact=False".

pbanaszkiewicz:
Fixed in #1251 .

PR #1251, opened 2018-05-21
GDPR: publish profile option for users

pbanaszkiewicz:
This commit addresses #1244.


pbanaszkiewicz:
I think we need to restrict `export-*` views even further, for example only to users who have a special permission. Therefore this PR will need to be improved.

PR #1296, opened 2018-07-15
Eligible trainees not showing up

### LINKED ISSUE #1226: Trainees eligible for certification are not showing up in list
When filtering the Trainees list - in the "Is SWC/DC instructor?" box, you can select "No, but eligible to be certified". This should bring up any trainees who have completed all checkout steps but have not yet been awarded a certificate. However, the matching parameters appear to be set up incorrectly, as not all eligible trainees are shown. 

As an example, just now I applied this filter and showed up no trainees waiting to be certified: 
screen shot 2018-02-07 at 9 48 54 am

However, looking at some of the individual training events, I found at least one who is eligible but has not been certified. 
screen shot 2018-02-07 at 9 49 03 am

I believe the issue in this case is that the trainee had completed the homework requirement twice, however in other cases, there are trainees who have done multiple discussion sessions, or done the steps out of order, or done SWC Homework but DC Demo, or vice versa. The matching criteria need to be updated to fit all of these situations. 

The matching criteria need to be:
IN ANY ORDER
(Discussion)(1 or more) AND
((SWC Demo)(1 or more) OR (DC Demo)(1 or more)) AND 
((SWC Homework)(1 or more) OR (DC Demo)(1 or more))

pbanaszkiewicz:
This fixes #1226 by changing "instructor eligibility" annotation query.

Additionally a small bug related to display of unescaped training progress notes in HTML was fixed.

### DIFF
--- a/workshops/filters.py
+++ b/workshops/filters.py
@@ -294,8 +294,13 @@ def filter_trainees_by_instructor_status(queryset, name, choice):
     elif choice == 'dc':
         return queryset.filter(is_dc_instructor=True)
     elif choice == 'eligible':
-        return queryset.filter(Q(swc_eligible=True, is_swc_instructor=False) |
-                               Q(dc_eligible=True, is_dc_instructor=False))
+        # Instructor eligible but without any badge.
+        # This code is kept in Q()-expressions to allow for fast condition
+        # change.
+        return queryset.filter(
+            Q(instructor_eligible=True) &
+            (Q(is_swc_instructor=False) & Q(is_dc_instructor=False))
+        )
     else:  # choice == 'no'
         return queryset.filter(is_swc_instructor=False, is_dc_instructor=False)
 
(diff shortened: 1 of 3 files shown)


pbanaszkiewicz:
@ErinBecker I fixed the "instructor eligibility" query as you wanted in #1226:

1. Needs to have "training"
2. Needs to have "discussion"
3. Needs to have either "SWC homework" or "DC homework"
4. Needs to have either "SWC demo" or "DC demo"

Any positive number of specific elements will suffice.

There's one question I need your answer to: how to decide if person must be shown in the list of intsructor-badge eligible trainees? Right now the conditions are:
1. eligible for instructor badge (ie. all points above)
2. no SWC instructor badge and no DC instructor badge

This means as soon as somebody gets their instructor badge, they won't show up in the list of eligible trainees. This also means that we don't distinguish trainees eligible for SWC or DC badges separately.
@ErinBecker if you want to change this behavior, please let me know.

ErinBecker:
Thanks @pbanaszkiewicz. Let's leave this behavior as-is for now. Very few Instructors get both badges, so I'm not worried about that edge case right now.

Issue #1216, opened 2017-09-25
Bulk upload persons - do not refresh page after "use existing person"


maneesha:
After doing a bulk upload persons,  often times there are many people who are already in the database.  The option to use existing person is useful for this.  However, it's tedious when you have to refresh the page after each update.  It would be better check of each person that has an existing record, and update them all at once.

pbanaszkiewicz:
According to [this comment](https://github.com/swcarpentry/amy/pull/1301#issuecomment-407059034), if the provided in #1301 fix is insufficient, we'll revisit this issue.

Issue #1215, opened 2017-09-25
Bulk upload persons - change link to *not* use existing person 


maneesha:
With bulk upload of persons, if a possible match exists, we can use that person.  This is useful.

Once we have matched the new person to an existing person the display changes to

![image](https://user-images.githubusercontent.com/829690/30820611-6bba228c-a1f0-11e7-8453-c161b5485a20.png)

It's noted that an existing person will be used, but the link to use the existing person still displays.  This link should be to *not* use the existing person.

PR #1301, opened 2018-07-21
Bulk upload - a couple of fixes

pbanaszkiewicz:
This PR provides:
1. a couple of style, text or function documentation changes
2. a sort-of fix for #1216 (more on that below)
3. a fix for #1215 

Changes in bulk-upload:
1. when first uploaded, entries from CSV are matched against people in the database; as long as the entries are correct, there's no need to go manually through the table and click "match with selected person" for each entry - this is easier and faster to implement than what was proposed in #1216, but I hope it's good enough
2. when someone is matched, they can be later unmatched (#1215)

### DIFF
--- a/workshops/templates/workshops/person_bulk_add_results.html
+++ b/workshops/templates/workshops/person_bulk_add_results.html
@@ -72,14 +72,18 @@
             {% if entry.similar_persons %}
                 <ul>
                 {% for person in entry.similar_persons %}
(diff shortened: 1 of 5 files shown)


pbanaszkiewicz:
@maneesha please decide if the solution for #1216 is good enough - it differs from what you proposed in #1216 but what I did was easier and faster for me to implement.

maneesha:
I think this will work - it's hard for me to tell without seeing it in action.

pbanaszkiewicz:
OK, so please @maneesha keep it in mind when you use bulk-upload after next release, and if you think it's still refreshing too much, then we'll revisit this issue.

PR #1340, opened 2018-09-09
Instructor training seats

### LINKED ISSUE #1055: Connect instructor training sessions to partnership agreements
We need a way to say "this event satisfies or partially satisfies our obligations for that partnership agreement".  (Right now, I'm storing that as free text.)  A many-to-many between events and partnership agreements (_not_ partners) would do this.

pbanaszkiewicz:
This PR fixes #1055 by implementing the functionality from [the draft](https://github.com/swcarpentry/amy/issues/1055#issuecomment-416948374).

This feature will be presented on the next AMY meeting, and will be extended/approved based on the feedback from the admins.


pbanaszkiewicz:
Changes requested as part of this PR review:
* allow to freely assign member sites with specific learners (ie. drop Event `member_sites`)
* display a warning when after assigning learners with memberships, they exceed memberships seat limits
* add "additional seats" field to membership (in case admin wants to extend number of seats available)
* extend training requests form with assigning memberships to the selected requests, additionally display warnings if either membership isn't active and/or number of seats allowed is exceeded
* add a simple dashboard for showing stats of membership organizations
* add "seats remaining" column to the all memberships page, allow for sorting by that column

PR #1358, opened 2018-10-03
Add localhost host to debug mode

marwahaha:
I got the below error in the latest Google Chrome when trying to run AMY. The fix in this PR made it all work ok!

image

### DIFF
--- a/amy/settings.py
+++ b/amy/settings.py
@@ -102,6 +102,7 @@ ALLOWED_HOSTS = [
 ]
 if DEBUG:
     ALLOWED_HOSTS.append('127.0.0.1')
+    ALLOWED_HOSTS.append('localhost')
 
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
@marwahaha thank you for the contribution. What command do you use to start AMY developer server?

marwahaha:
I followed the instructions here and ran `make serve`, but `localhost:8000` will not work unless this is enabled. However, `127.0.0.1:8000` will work fine.

https://github.com/swcarpentry/amy#getting-started

pbanaszkiewicz:
@marwahaha well, now I can see that our README needs updates. If you'd like to tackle this, I've just added an [issue](https://github.com/swcarpentry/amy/issues/1365).

PR #1363, opened 2018-10-06
Events API: Fix 500-error when filtering

### LINKED ISSUE #1351: 500 Server Error when filtering on Events API
https://amy.software-carpentry.org/api/v1/events/published/ API throws 500 Server error when using parameters to filter the call. 

For example: https://amy.software-carpentry.org/api/v1/events/?completed=&tag=SWC&start=&start_before=&start_after=&end=&end_before=&end_after=&order_by=start

I tried with several parameter combinations and all threw 500 Server error.

pbanaszkiewicz:
This fixes #1351. The error was cryptic and well-hidden, but I narrowed
it to ill-made filter for tags.

Probably the idea was to allow people put in exact tag names, but now
it's easy to do with Django Filter options.

### DIFF
--- a/api/filters.py
+++ b/api/filters.py
@@ -8,22 +8,16 @@ from workshops.filters import (
 from workshops.models import Event, Task, Tag, Person, Badge
 
 
-def filter_tag_by_name(queryset, name, values):
-    # tags = Tag.objects.filter(name__in=values)
-    # for tag in tags:
-    #     queryset = queryset.filter(tags=tag)
-    # return queryset
-    return Tag.objects.all()
-
-
 class EventFilter(filters.FilterSet):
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
@anenadic @fmichonneau @symulation @maneesha hello everyone, I'm tagging you here because you seem to be using AMY API and I'm going to change this API endpoint:
https://amy.software-carpentry.org/api/v1/events/
(**this is not https://amy.software-carpentry.org/api/v1/events/published/**)

Instead of allowing `?tag=xxx&tag=yyy` in URL, we're switching to `?tags=xxx&tags=yyy`. Please check if your scripts are using this endpoint and update if necessary. The change is in filter syntax - now it's `tags` instead of `tag` - and in behavior (tags are conjoined, so events have to have all supplied tags to appear in the results).

PR #1374, opened 2018-10-13
Bulk upload training request manual score

pbanaszkiewicz:
@karenword @ErinBecker here's snapshot of the new "Download for manual scoring" button group:
screenshot_2018-10-13 amy training requests

The same selected requests that you can see in the screenshot generate this CSV (zipped because GitHub):
[two_results.zip](https://github.com/swcarpentry/amy/files/2475843/two_results.zip)

Uploading this CSV file without any changes on the Bulk Upload page gives this result:
screenshot_2018-10-13 amy confirm uploaded training requests manual score data

I didn't implement inline editing, like in bulk add persons, therefore in case of errors the user would have to fix the file and re-upload.

Please tell me what you think - e.g. if there's something wrong with the CSV.

### DIFF
--- a/api/renderers.py
+++ b/api/renderers.py
@@ -2,13 +2,12 @@ from rest_framework_csv.renderers import CSVRenderer
 
 from .serializers import (
     TrainingRequestWithPersonSerializer,
+    TrainingRequestForManualScoringSerializer,
 )
 
 
-class TrainingRequestCSVRenderer(CSVRenderer):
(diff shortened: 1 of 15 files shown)


ErinBecker:
This looks great to me! Thanks @pbanaszkiewicz for putting this together so quickly! 

(I had no idea you were a wizard. 😉 )

karenword:
+1 for sleep teaching!

Issue #1407, opened 2018-11-14
Grouping of questions in workshop request form


maneesha:
In the new workshop request form, there are some questions that belong together because they are asking about similar information or one question is dependent on another.  Is there a way to make this more apparent visually?  Even horizontal lines between sections may work.

pbanaszkiewicz:
@maneesha can you point the questions you want separated / joined visually together?

maneesha:
This is addressed in #1406

Issue #1406, opened 2018-11-14
Workshop request form - help text on top


maneesha:
For each section, the help text should be at the top of the list of options.

For example, in this section, the text "If you know the academic level....." should appear above the list of options.

![image](https://user-images.githubusercontent.com/829690/48514015-5dd0c180-e82b-11e8-8c0c-84298de6a0af.png)

pbanaszkiewicz:
By "section" do you mean each "question" (that's what on your screenshot), or a group of questions?

Do you want this change to only affect checkbox-style questions (like the one you posted), radio-style questions or all questions?

maneesha:
Here is what I am proposing.  @pbanaszkiewicz  Let me know if you have questions about all this.


##### Horizontal lines to group sections
There are a few horizontal lines I want to include to help break up and organize the sections.  Red color is for emphasis here; actual color should match the color scheme of the form.

##### Institutional affiliation section
* Remove the side text "Name of institution you're affiliated with" as it is only to be filled in if institution is not on list, so this should not look like a separate question

##### Conference section
* "Is this workshop part of a conference or larger event" question should go on side bar.  Help text should say "If yes, please provide conference details (name, description")

##### Number of attendees
* Move help text ("This number doesn't need to be precise....") to top of option choices

##### Domains or topic of interest
* Move help text  to top of option choices
* Move "Don't know yet" to bottom of list of choices

##### Attendees' academic level
* Move help text  to top of option choices

##### Attendees' level of computing experience
* Move help text  to top of option choices

##### Which Carpentry workshop are you requesting? 
* Move help text  to top of option choices
* See other issue about including other LC choices
* Move DC Other and SWC Other to bottom of the list of options for DC and SWC respectively

##### How will you manage travel expenses?
* Question should be moved to be at level with "Regardless of the fee due...."

![image](https://user-images.githubusercontent.com/829690/49674330-a3099d00-fa3f-11e8-9894-b23e641fa24e.png)

pbanaszkiewicz:
Hello @maneesha,

thanks for the screenshot, it's very convenient!

Below you'll find my questions.

---

> ##### Conference section
> 
> * "Is this workshop part of a conference or larger event" question should go on side bar.  Help text should say "If yes, please provide conference details (name, description")

Essentially you want to remove the checkbox and take it's label ("Is this workshop part of conference or larger event") and use it for text field label (so instead of "Conference details" text) + changes to help texts?

---

> ##### How will you manage travel expenses?
[...]

PR #1412, opened 2018-11-15
Workshop request UI changes

pbanaszkiewicz:
This PR fixes a number of issues related to workshop-request submission form.

### DIFF
--- a/amy/templates/forms/workshoprequest_confirm.html
+++ b/amy/templates/forms/workshoprequest_confirm.html
@@ -12,4 +12,5 @@ you shortly. For more information please visit:</p>
     <li>or contact <a href="mailto:team@carpentries.org">team@carpentries.org</a>.</li>
 </ul>
 
+<p><a href="{% url 'workshop_request' %}">Click here</a> to submit another workshop request.</p>
 {% endblock %}


pbanaszkiewicz:
@maneesha almost all issues (reported so far) regarding workshop requests form are resolved in this PR. I'm still waiting for your feedback on #1406 and #1407.

pbanaszkiewicz:
@maneesha: please confirm this layout:

screenshot_2018-12-09 amy request a carpentries workshop 1

pbanaszkiewicz:
@maneesha ping

maneesha:
Thanks for reminding me.  

These questions should be required:
* Institutional affililiation (require either selection from drop down or fill in text)
* Domain or topic of interest.... 
* How will you manage travel expenses....


Otherwise looks good. Thanks @pbanaszkiewicz !

pbanaszkiewicz:
Thanks @maneesha.

Unfortunately I cannot get marked "Insitutional Affiliation" as required, because it would make it impossible to not select anything in drop down. Instead we're relying on server-side verification which enforces that one of the following is true: 1) user selected a valid entry in drop down; 2) user entered some text in "If your institution isn't on the list, enter its name in the field below." field.

The same applies to "Domains or topic of interests" (first field is called `domains`, second field is called `domains_other` and is for "Other" text entry), and "How will you manage travel expenses" (first field is called `travel_expences_management` and second is called `travel_expences_management_other`).

PR #1445, opened 2019-01-14
Application group code

maneesha:


maneesha:
An additional change submitted to edit some text on the training application form.  Didn't mean to include this in the same PR.  @pbanaszkiewicz   let me know if you want this separated out.

pbanaszkiewicz:
@maneesha some code was already fixed by me, and I want `fake_database` to produce as much fake data as possible, so I'll change this PR into "Application group code", and work on `fake_database` in another PR.

Issue #1448, opened 2019-01-25
Update email addresses on 'New training request' page


tracykteal:
We'd like to update the email addresses on the 'New Training Request' page

- change training@carpentries.org to team@carpentries.org
- change partnerships@carpentries.org to memberships@carpentries.org

pbanaszkiewicz:
Fixed via #1449.

PR #1449, opened 2019-01-25
updated email addresses

elizabethwilliams8:
Addresses issue number #1448


maneesha:
@elizabethwilliams8   Can you include the django migration for this change?  If not I can do it.

tracykteal:
Thanks @maneesha, I don't know how to do that and I'm not sure @elizabethwilliams8 does either. Could you provide guidance?

maneesha:
I ran `makemigrations` and it looks like this is good as is.  I'll leave it to @pbanaszkiewicz  to verify and merge.

pbanaszkiewicz:
Approved the changes for v2.5. Thanks @elizabethwilliams8!

This fixes #1448.

@maneesha: Travis-CI should check migrations as part of testing process, but I've just now discovered it's not working correctly.

PR #1457, opened 2019-01-30
set cursor focus to search box

maneesha:
This sets the keyboard cursor to the search box when the app opens.  I didn't see any migrations needed.

### DIFF
--- a/amy/templates/navigation.html
+++ b/amy/templates/navigation.html
@@ -74,7 +74,7 @@
           <input type="hidden" name="in_persons" value="on" />
           <input type="hidden" name="in_airports" value="on" />
           <input type="hidden" name="in_training_requests" value="on" />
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
@maneesha thanks for your contribution. I wasn't aware of the `autofocus` property.

I looked for UX accessibility / usability considerations regarding `autofocus` and found [this](https://ux.stackexchange.com/questions/60026/when-is-it-appropriate-to-add-the-html5-autofocus-attribute-to-a-page). It looks like people in general agree that `autofocus` should be used on pages where the autofocused field is main point of interest. In AMY's example that would be https://amy.software-carpentry.org/workshops/search/ and "Term" field.

@maneesha if it will help you use AMY, then I'll merge, but I'm afraid it would cause issues to other users or in future.

maneesha:
Closing this as @pbanaszkiewicz  has brought up accessibility issues this may cause.

PR #1471, opened 2019-02-17
Continents

### LINKED ISSUE #1444: Search workshops, instructors, etc. by continent
We currently have the feature to search by country.  It would also be useful to search by continent or other larger global region.  For example - Europe, Asia, Africa, South America, Central America, Middle East, etc.  Are there existing libraries that define these geographic regions? Can we have user-defined regions?

Can we use a library like what's used to define Africa in this issue (https://github.com/swcarpentry/amy/issues/1368)?

pbanaszkiewicz:
This fixes #1444.

@maneesha this PR adds Continent model and allows you, as a superuser, to modify / extend / create additional continents. Currently there's just "standard" continents like Europe, Asia, Australia/Oceania, North America, etc.

This system allows you to manually create and define, for example, Middle East of Central America.

The filter bar for Events and Find Workshop Staff was extended with Continent selection.


maneesha:
This sounds great @pbanaszkiewicz !
Looking forward to seeing it in the next version.

PR #1521, opened 2019-08-06
fix email address for sysadmin

fmichonneau:
looking at the logs of Mailgun I noticed that notifications to this address bounce because it doesn't exist. `sysadmin@carpentries.org` does exist though but I'm not 100% sure we want to go there because I'm not sure the kind of emails that end up being sent to this address.


pbanaszkiewicz:
Great! I was wondering for like a year why I wasn't getting any error emails.

PR #1524, opened 2019-08-27
Upgrade Django to v2.2.4

pbanaszkiewicz:
This PR contains changes required for upgrading to Django v2.2.4 as well as some fixes.


pbanaszkiewicz:
This is still WIP - I need to investigate errors in tests + tests manually if everything is working.

PR #1530, opened 2019-09-12
fix typo

fmichonneau:
fix small typo in user profiles


pbanaszkiewicz:
Good catch, thanks @fmichonneau!

PR #1531, opened 2019-09-12
change GH link swcarpentry > carpentries

maneesha:
There may be other places that we also need to update the GH  name.


pbanaszkiewicz:
Thanks @maneesha!

PR #1545, opened 2019-09-20
Enable cache

### LINKED ISSUE #1538: "Results could not be loaded" when trying to assign admin to event or request
When trying to assign an admin to an event, the drop down does not load, and instead it says "The results could not be loaded"

This happens inconsistently.  Trying again will sometimes make the drop down list load.

image

pbanaszkiewicz:
This potentially fixes #1538 (I'm still waiting for user test results).

The potential cause for this error was no cache table (no cache at all
was set up), so the Django-select2 AutoResponseView was unable to find
a correct widget by `field_id` GET parameter.

https://github.com/applegrew/django-select2/blob/master/django_select2/views.py#L82

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -247,6 +247,20 @@ AUTH_PASSWORD_VALIDATORS = [
     },
 ]
 
+# CACHE
+# -----------------------------------------------------------------------------
+# https://docs.djangoproject.com/en/2.2/topics/cache/#database-caching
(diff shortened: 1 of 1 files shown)


pbanaszkiewicz:
According to [this comment](https://github.com/carpentries/amy/issues/1538#issuecomment-534128721), the issue was fixed.

PR #1555, opened 2019-10-02
EventForm: build expected tags in clean_curricula differently

### LINKED ISSUE #1554: Event form: improper validation of curricula and selected tags
`EventForm.clean_curricula` does try to validate selected curricula (corresponding tags must exist), but doesn't work with Mix & Match curriculum.

pbanaszkiewicz:
Previously we were using name of the curriculum (split by "-"). Now we
can use curriculum.carpentry value, which corresponds to Tag name,
unless it's mix&match -> then tag name is "Circuits". And this is what
the method now does.

A new test case was added to ensure this mechanism is working correctly
with Mix&Match curriculum.

This fixes #1554.

### DIFF
--- a/amy/workshops/forms.py
+++ b/amy/workshops/forms.py
@@ -557,10 +557,12 @@ class EventForm(forms.ModelForm):
         tags = self.cleaned_data['tags']
 
         try:
-            expected_tags = [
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
It was reported to me that this fix resolved the issue.

PR #1602, opened 2020-01-06
Update WorkshopInquiryRequest's assignment view

pbanaszkiewicz:
Permissions for that view were wrong- used the old name
(without "request" part).


pbanaszkiewicz:
This PR fails because it wasn't rebased against `develop`. I added it from GH web interface. It's safe to merge.

PR #1611, opened 2020-01-23
add link to submit self org wkshp

maneesha:
Based on suggestion from @Talishask  , this PR adds a link to the trainee dashboard view so they can submit a self organised workshop directly from that view if they see something missing.  Here's a snapshot of what it will look like:

image


Talishask:
Hey @maneesha & @pbanaszkiewicz - Is it too late to update the verbiage on this update?
If not, could we update it to:

If you'd like to let us know about a workshop you are planning or have run a self-organised workshop that is missing from this list, please fill out our [self organised workshop form](https://amy.carpentries.org/forms/self-organised/).

PR #1626, opened 2020-03-10
Fix Github OAuth2 warning

pbanaszkiewicz:
Github [deprecated](https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters)
API login through query args (now it should use HTTP basic auth).

This commit provides a simple solution that overrides headers sent to
GH servers in a GH OAuth2 social-login backend.

Unfortunately upstream remains unresponsive to PRs that fix this
urging matter, so I decided to follow custom-backend approach.

### DIFF
--- a/amy/workshops/github_auth.py
+++ b/amy/workshops/github_auth.py
@@ -1,3 +1,5 @@
+from urllib.parse import urljoin
+
 from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
I think we can drop this one since upstream package included the required changes: https://github.com/python-social-auth/social-core/blob/master/CHANGELOG.md#330---2020-03-17

PR #1631, opened 2020-03-29
Update from URL when `slug` is missing

### LINKED ISSUE #1572: Amy Event URL Validation
I believe we recently updated the validation rules for the workshop website (URL) in AMY to accept any website format. However, when trying to 'Update from URL' I am getting this error message (see thread).
Here is the AMY event: https://amy.carpentries.org/workshops/event/2019-11-26-Potsdam-Berlin/
new github page:
http://swc-bb.gitext.gfz-potsdam.de/pages/2019-11-26-Potsdam-Berlin/
New Repo: https://gitext.gfz-potsdam.de/swc-bb
old github page:
https://swc-bb.github.io/2019-11-26-Potsdam-Berlin-R/ (edited) 

Here is our slack conversation:
**Maneesha**: AMY will accept any URL in the event's record.  However to use "Update from URL" AMY needs to know the url of the github repo it is from.  If the website url is in the typical format (`username.github.io/YYYY-MM-DD-sitename`) then AMY can figure out the github repo.  If the URL is in a different format, then AMY can't figure out the github repo and therefore can't update from URL.  Let me know if that's clear.
**François**:I thought (and a quick inspection of AMY's code suggests this) that the reading of the meta data about the workshops was done by looking at the meta tags in the HTML version of the website and should therefore be independent of the whether the repository is hosted on GitHub or somewhere else.  This validation might be unnecessary.
**Maneesha**: Thanks for that @francois  We should then be able to make this work if they are still using our template, even if the URL isn't in the standard format.  If they are not using our template then it may not have the meta tags and therefore can't update from URL (example https://www.sib.swiss/training/course/2019-07-elixir-sib-dc)
**François**: yes, I agree. The checks about whether the metadata exist or not shouldn't depend on the scheme of the URL (because people could create a workshop website that doesn't use our template using GitHub). Could you please open an issue about this so we can address it in the future?

### LINKED ISSUE #1610: Venue, address and latlong info not picked up from workshop URL
Just discovered some workshops without location info in AMY and when checked in their workshop websites - location (venue, address and latlong) info was there. Tried updating workshop records in AMY by doing 'update from URL' but it didn't pick up the location data.

### LINKED ISSUE #1625: Update form fill/validation when creating AMY Events from submissions
In this issue (https://github.com/carpentries/amy/issues/1572) we identified some issues that block AMY from pulling workshop information into the event. However, when this conflict exists with the workshop webpage, all of the fields on the 'new event' side are blank. Even the information that is pulled from the workshop request form. 

Can we update the validation/autofill here so that when these hosting or webpage conflicts happen, the information that is pulled from the workshop request form still auto-fills as appropriate on the 'new event' panel? This way we only have to manually add the fields that were pulled from the workshop website?

pbanaszkiewicz:
This PR responds to numerous issues raised over functionality of "import from URL" / "update from URL":

- #1572 
- #1610
- #1625 

The solution is to basically ignore the fact that `slug` is missing from websites (mostly because they're hosted outside GitHub). Instead, only when not a single tag is to be found in HTML, then the algorithm switches to YAML version, and if again nothing is found - an error is generated.

This slight change allow to import from URL from websites such as:
* https://pages.nist.gov/2020-05-29-nist/ (used in tests below)
* https://www.hpc.tntech.edu/workshop-2020-03-16/

You can access source code (CTRL+U in Firefox for Windows) and see that these website have a bunch of `<meta>` tags, but the one named `slug` is missing content.

Using first page as an example, I was able to import from URL when creating a new event (1) and when accepting a self-organised submission (2):

**1:**
Screenshot_2020-03-29 AMY New event

**2:**
Screenshot_2020-03-29 AMY

Please note how in both cases the slug field, when empty, was highlighted red.

Closes #1572. Closes #1610. Closes #1625.

### DIFF
--- a/amy/extrequests/tests/test_selforganised_submissions.py
+++ b/amy/extrequests/tests/test_selforganised_submissions.py
@@ -2,6 +2,7 @@ from datetime import date, timedelta
 
 from django.conf import settings
 from django.urls import reverse
+from requests_mock import Mocker
 
 from autoemails.models import Trigger, EmailTemplate, RQJob
 from autoemails.tests.base import FakeRedisTestCaseMixin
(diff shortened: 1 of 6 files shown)


Talishask:
Thanks @pbanaszkiewicz this looks good. The last thing I'd add is just making sure that when creating an event from the workshop submissions and if there is an error with reading the information from the URL, the New Event panel on the right is at least populated with the information that is pulled from the workshop submission like administrator=self organized, tags, curricula, language, etc.

pbanaszkiewicz:
@Talishask I extended the way Self-Organised Submission "accept and create a new event" works, so that it will always load values from S.O.S. into Event Form (url, curricula, tag Circuits in case of mix&match, administrator, and host if provided). Other values will be filled if the website loads correctly and AMY is capable of reading meta tags from it.

PR #1637, opened 2020-04-13
PostWorkshopAction: link to survey results

pbanaszkiewicz:
This PR adds a survey result link (`reports_link`) to `PostWorkshopAction` context.

### DIFF
--- a/amy/autoemails/tests/test_postworkshopaction.py
+++ b/amy/autoemails/tests/test_postworkshopaction.py
@@ -197,7 +197,7 @@ class TestPostWorkshopAction(TestCase):
                 helpers=[p1, p3],
                 all_emails=['hg@magic.uk', 'hp@magic.uk'],
                 assignee='Regional Coordinator',
-                reports_link='https://workshop-reports.carpentries.org/?key=e18dd84d093be5cd6c6ccaf63d38a8477ca126f4slug=test-event',
+                reports_link='https://workshop-reports.carpentries.org/?key=e18dd84d093be5cd6c6ccaf63d38a8477ca126f4&slug=test-event',
             ),
         )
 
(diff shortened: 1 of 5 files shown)


amy/autoemails/tests/test_postworkshopaction.py
```diff
@@ -197,6 +197,7 @@ def testContext(self):
                 helpers=[p1, p3],
                 all_emails=['hg@magic.uk', 'hp@magic.uk'],
                 assignee='Regional Coordinator',
+                reports_link='https://workshop-reports.carpentries.org/?key=e18dd84d093be5cd6c6ccaf63d38a8477ca126f4slug=test-event',
```
fmichonneau, line 200: missing `&` between the 2 URL parameters

config/settings.py
```diff
@@ -611,3 +611,19 @@
...
+REPORTS_SALT_BACK = env.str('AMY_REPORTS_SALT_BACK', default='')
+if not DEBUG and not (REPORTS_SALT_FRONT and REPORTS_SALT_BACK):
+    raise ImproperlyConfigured(
+        "Report salts are required. See REPORT_SALT_FRONT and REPORT_SALT_BACK"
+        " in settings."
+    )
+
+REPORTS_LINK = env.str(
+    'AMY_REPORTS_LINK',
+    default='https://workshop-reports.carpentries.org/?key={hash}slug={slug}',
```
fmichonneau, line 628: the value for the environment variable is correct in https://github.com/carpentries/systems/pull/67/files but the default is missing the "&" between key and slug I think.

pbanaszkiewicz:
@fmichonneau great catch! I would scratch my head in future if it stayed that way. 😄

fmichonneau:
thanks!

PR #1642, opened 2020-04-26
Contact meta tag - split by |

### LINKED ISSUE #1568: Update Contact Field in AMY Events
Currently the 'contact:' field in an AMY event lists information pulled from the workshop repo.

The field currently lists multiple email addresses as one continuous string rather than separating them with a delimiter or comma. For example, in this workshop: https://amy.carpentries.org/workshops/event/2019-10-24-gitbash/

With the update of the WRF to request additional contacts, we would like to update the contact field in the AMY event to:
- List the emails from the 'Additional contacts:' field of the workshop request form [in place of the contacts from the workshop website]
- Separate the email addresses with a comma

While this does not cause a work stoppage, it may be an important update to integrate into the typeform survey workflow.

pbanaszkiewicz:
This PR fixes #1568 by supporting multiple values in Event contact field.

For more details please see commit messages.

### DIFF
--- a/amy/workshops/tests/test_util.py
+++ b/amy/workshops/tests/test_util.py
@@ -58,7 +58,7 @@ class TestHandlingEventMetadata(TestBase):
 <meta name="invalid" content="invalid" />
 <meta name="instructor" content="Hermione Granger|Ron Weasley" />
 <meta name="helper" content="Peter Parker|Tony Stark|Natasha Romanova" />
-<meta name="contact" content="hermione@granger.co.uk, rweasley@ministry.gov" />
+<meta name="contact" content="hermione@granger.co.uk|rweasley@ministry.gov" />
 <meta name="eventbrite" content="10000000" />
 <meta name="charset" content="utf-8" />
 </head>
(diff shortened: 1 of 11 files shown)


amy/workshops/tests/test_util.py
```diff
@@ -149,9 +149,9 @@ def test_finding_metadata_on_index(self):
             'address': 'Highway to Heaven 42, Academipolis',
             'latlng': '36.998977, -109.045173',
             'language': 'us',
-            'instructor': 'Hermione Granger, Ron Weasley',
```
fmichonneau, line 152: does this mean that names should also be split with `|` or is this is a test?

pbanaszkiewicz: This is a test for part of AMY that supports reading metatags from YAML - it's a fallback if we can't read them from HTML, and I don't think it's used very often. However, once we read the metatags from YAML, the expected output is that list values (instructors, helpers, or contact) should be joined with "|" as single strings. Originally supported format in YAML is in `self.yaml_content` (lines ca. 80 in this file):

```
instructor: ["Hermione Granger", "Ron Weasley",]
helper: ["Peter Parker", "Tony Stark", "Natasha Romanova",]
contact: ["hermione@granger.co.uk", "rweasley@ministry.gov"]
```

PR #1662, opened 2020-05-18
include redis/docker instructions in local setup

maneesha:


pbanaszkiewicz:
@maneesha if you want to additionally include instructions on how to start developing with email automation, please let me know and I'll prepare something.

maneesha:
@pbanaszkiewicz  I don't have immediate needs for that, but yes, at some point that would be great!

PR #1663, opened 2020-05-18
make Administrator field required

maneesha:
This makes the Administrator field required.

The model still has `null=True, blank=True,`.
If we have historic data where we don't have Administrator I'm ok leaving those but moving ahead I want all future data to be one of the options available in the drop down.  @pbanaszkiewicz  Is there anything else that needs to be done for this?


maneesha:
The CI check failed - it says it needs a migration. When I ran `python manage.py makemigrations` it says `No changes detected`.

pbanaszkiewicz:
It seems like this changes a lot in tests (~37 tests to fix). Migration is indeed not needed.

pbanaszkiewicz:
I'll work on fixing the issues with tests.

PR #1706, opened 2020-08-18
Membership codes

### LINKED ISSUE #1705: Add instructor training registration code to membership
For each membership term for a given organization, we give them a code that is used to register on Eventbrite. This code is required to unlock registration on Eventbrite.  Currently we store this in a separate spreadsheet, but ideally we'd want it stored here so we are not looking in another place for this.  Each organization's membership term will have a unique code that is valid for that term only.  They get a new code when their membership renews.

For now, this can be a simple text field so that we are not keeping data in two separate places.  Ideally moving ahead we'd like to use it to integrate with Eventbrite's API allowing us to match up Eventbrite registration with our applications (related to #1614).

maneesha:
Fixes #1705


maneesha:
Adds a field for registration code to the membership model. Right now the code does not do anything. It is meant to replace an external spreadsheet that tracks organization name, membership start date, membership end date, and registration code.

pbanaszkiewicz (approved):

PR #1708, opened 2020-08-22
Membership workshops summary fixes

### LINKED ISSUE #1664: Fixes to workshop summary in membership view
In the Membership summary view, it includes canceled workshops in the completed workshops count.  This should show only workshops that are in the past and have not been canceled/stalled.  "Completed" does not need to checked for it to be counted here.  We'd also like a column for "Planned" for future workshops.  

Also, because there are no limits to self organized workshops, there should only be two columns for completed and planned. 

Here's a sample of what this should look like.
image

pbanaszkiewicz:
This fixes #1664 by incorporating changes requested by @maneesha.

### DIFF
--- a/amy/fiscal/tests/test_membership.py
+++ b/amy/fiscal/tests/test_membership.py
@@ -1,5 +1,4 @@
 from datetime import timedelta, date
-import itertools
 
 from django.urls import reverse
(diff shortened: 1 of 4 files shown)


maneesha:
I just tested this out locally; looks good to me.

PR #1733, opened 2020-10-29
Make accordion work in Django form

### LINKED ISSUE #1723: Pre-approved registration box does not expand
In the Training Request form, when someone selects "Pre-approved Registration" the box should immediately expand letting them enter in a group name / code.  This does not happen on click, and the form still looks like this:

image

It's not until the user tries to submit the form that they are asked to go back and fill it in - at which point they see this:

image

marwahaha:
fixes #1723 

The accordion worked fine when I copied the HTML into a JSFiddle. I noticed that triggering the "collapse" function in JavaScript allowed the accordion to work again.

It seems that the field was hidden intentionally for the pre-filled group name. I removed that in light of #1723. 

Hopefully I did this correctly!

Unfortunately, I'm not sure the best way to test for this behavior.

### DIFF
--- a/amy/extforms/forms.py
+++ b/amy/extforms/forms.py
@@ -92,9 +92,6 @@ class TrainingRequestForm(forms.ModelForm):
             initial['group_name'] = initial_group_name
             initial['review_process'] = 'preapproved'
         super().__init__(*args, initial=initial, **kwargs)
-        if initial_group_name is not None:
-            field = self.fields['group_name']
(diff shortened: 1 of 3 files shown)


pbanaszkiewicz (changes requested):
Hi @marwahaha, thank you for the pull request. I looked around and turns out the reason accordion doesn't work is old version of Bootstrap. I think upgrading to v4.5 branch should fix the issue.

PR #1740, opened 2020-11-01
Update to Boostrap 4.5

### LINKED ISSUE #1723: Pre-approved registration box does not expand

marwahaha:
supersedes https://github.com/carpentries/amy/pull/1733

fixes #1723 

@pbanaszkiewicz  : you are correct, updating to Bootstrap 4.5.3 (and fixing the unmet peer dependency) makes the accordion work again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "@fortawesome/fontawesome-free": "^5.1.1",
     "@github/time-elements": "^3.1.1",
     "@ttskch/select2-bootstrap4-theme": "^1.0.5",
-    "bootstrap": "^4.3.1",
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz (approved):

pbanaszkiewicz:
Thank you for the contribution, @marwahaha!

PR #1748, opened 2020-11-11
Extend AskForWebsite with supporting instructor role

pbanaszkiewicz:
Applied both for condition check and for recipients of the email.

### DIFF
--- a/amy/autoemails/actions.py
+++ b/amy/autoemails/actions.py
@@ -801,6 +801,8 @@ class AskForWebsiteAction(BaseAction):
     ...     job = scheduler.enqueue_in(launch_at, action)
     """
 
+    role_names = ["instructor", "supporting-instructor"]
+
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
@fmichonneau this PR makes two corrections in how AskForWebsite worked:

1. supporting instructors are now taken into account when checking if AskForWebsite can work (at least one instructor or supporting instructor must be assigned, to have someone to send the email to)
2. supporting instructors are now added to the context (under `instructors` like normal instructors). With this approach, you will not be able to distinguish normal instructors and supporting instructors in the email body.

Please confirm you agree with both points, or provide corrections otherwise.

fmichonneau:
Thanks Piotr, this is good! 
I think it's not a problem to not be able to distinguish between instructors and supporting instructors for this particular email.

fmichonneau (approved):

PR #1771, opened 2021-01-27
Membership duration warning

pbanaszkiewicz:
Improve agreement start/end date handling and UX.

1. There's no more default value of "today" for both fields
2. "End" field gets updated to +1yr when selecting start date
3. Warning on the form shows and hides when the duration is not 1 year long.
4. Warning shows on membership detail page after adding a new membership with invalid agreement duration
5. "End" will not accept dates in past.

TODO: I asked a question about membership agreement start/end values being empty. I think we shouldn't allow for them being empty, but I want to receive confirmation from @maneesha.


maneesha:
Correct - start and end date should always be required.

PR #1806, opened 2021-02-19
feature: Filter workshop staff by knowledge domain.

### LINKED ISSUE #1788: Filter workshop staff by knowledge domain
I was wondering if it would be possible to add an option to the [Find Workshop Staff](https://amy.carpentries.org/workshops/workshop_staff/) page in Amy so that search results can be filtered by knowledge domain (which is a field each person has in the database)?

That would help when looking for instructors and helpers for research domain specific workshops.

lauryndbrown:
Screen Shot 2021-02-19 at 1 38 35 AM

Screen Shot 2021-02-19 at 1 39 07 AM

Fixes #1788

### DIFF
--- a/amy/workshops/lookups.py
+++ b/amy/workshops/lookups.py
@@ -253,7 +253,7 @@ class LanguageLookupView(LoginNotRequiredMixin, AutoResponseView):
         return results
 
 
-class KnowledgeDomainLookupView(LoginNotRequiredMixin, AutoResponseView):
+class KnowledgeDomainLookupView(OnlyForAdminsNoRedirectMixin, AutoResponseView):
     def dispatch(self, request, *args, **kwargs):
         self.subtag = "subtag" in request.GET.keys()
         return super().dispatch(request, *args, **kwargs)
--- a/amy/workshops/lookups.py
+++ b/amy/workshops/lookups.py
@@ -253,6 +253,31 @@ class LanguageLookupView(LoginNotRequiredMixin, AutoResponseView):
         return results
 
 
+
+class KnowledgeDomainLookupView(LoginNotRequiredMixin, AutoResponseView):
+    def dispatch(self, request, *args, **kwargs):
+        self.subtag = "subtag" in request.GET.keys()
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_queryset(self):
+        results = models.KnowledgeDomain.objects.all()
+
+        if self.term:
+            results = results.filter(
+                Q(name__icontains=self.term) | Q(subtag__icontains=self.term)
+            )
+
+            if self.subtag:
+                return results.filter(subtag__iexact=self.term)
+
+        results = results.annotate(person_count=Count("person")).order_by(
+            "-person_count"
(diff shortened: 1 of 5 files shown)


amy/workshops/lookups.py
```diff
@@ -222,6 +222,31 @@ def get_queryset(self):
         return results
 
 
+
```
pbanaszkiewicz, line 225: There should be two empty lines before class definition, three is PEP8 violation. Similar situation at the end of this hunk.

amy/workshops/lookups.py
```diff
@@ -222,6 +222,31 @@ def get_queryset(self):
...
+class KnowledgeDomainLookupView(LoginNotRequiredMixin, AutoResponseView):
```
pbanaszkiewicz, line 226: Since this lookup is used in admin-restricted area of AMY, I'd like you to switch to admin required mixin. This will affect tests that you had to adjust for `LoginNotRequiredMixin`.

amy/workshops/tests/test_workshop_staff_searching.py
```diff
@@ -229,6 +229,52 @@ def test_match_on_many_language(self):
...
         self.assertNotIn(self.harry, response.context['persons'])
+    
+    def test_match_on_one_domain(self):
+        """Ensure people with one particular knowledge domain preference
+        are returned by search."""
+        # prepare knowledge domains
+        self._setUpDomains()
+        # Ron has a humanities knowledge domain
+        self.ron.domains.add(self.humanities)
+        self.ron.save()
```
pbanaszkiewicz, line 240: No need to save Ron (Harry would never say that).

The domains exist in M2M relationship, so this call to `.add` here actually is an `INSERT` clause in the intermediary table between `Person` and `KnowledgeDomain` tables. `ron` is not being changed.

pbanaszkiewicz (changes requested):
Minor style changes and removal of some `.save()` calls needed. Otherwise looks good.

amy/workshops/lookups.py
lauryndbrown: @pbanaszkiewicz So I notice black changes are cluttering the PR a bit. Might be nice to have a single larger pr to handle flake8/black formatting. But no preference as long as it's looks readable to you.

amy/workshops/tests/test_workshop_staff_searching.py
lauryndbrown: Oh thank you for pointing that out!

amy/workshops/lookups.py
pbanaszkiewicz: @lauryndbrown you're right. https://github.com/carpentries/amy/issues/1812

lauryndbrown:
@pbanaszkiewicz this one should be ready for another look

pbanaszkiewicz (changes requested):
Something is causing conflicts. Other than that, I don't have any change requests.

pbanaszkiewicz (approved):

PR #1814, opened 2021-02-25
tech: Apply Black to the entire codebase.

lauryndbrown:
Updates the README and runs Black on everything with the commit skipped by the git blame. As of the creation of this PR. This is not true for the GitHub blame UI:

> The one caveat is that GitHub and GitLab do not yet support ignoring revisions using their native UI of blame. So blame information will be cluttered with a reformatting commit on those platforms. (If you'd like this feature, there's an open issue for GitLab and please let GitHub know!)

See for more info on the process: https://github.com/psf/black/#migrating-your-code-style-without-ruining-git-blame

Fixes https://github.com/carpentries/amy/issues/1812.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,1 @@
+# Migrate code style to Black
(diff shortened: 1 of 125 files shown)


pbanaszkiewicz (approved):
Looks good.

PR #1857, opened 2021-03-21
[#1780][#1853][#1849][#1852][#1851] Membership calculations, UI improvements, fixes from feedback

pbanaszkiewicz:
This fixes #1780.
This fixes #1853.
This fixes #1849.
This fixes #1852.
This fixes #1851.

Organisation details table with memberships:
obraz

---

Membership details table:
obraz

---

New membership consortium help text:
obraz

---

Membership members add another help text:
obraz

---

Membership public_status field defaults to "Private":
obraz

---

Search looks up memberships by name:
obraz

PR #1858, opened 2021-03-21
[#1838][#1848][#1850] Improvements to Membership formsets

### LINKED ISSUE #1838: Membership: clicking on discard button doesn't do anything
when editing persons or members in consortium view:
1. click on "Add another"
2. click on "Discard" 

what happens:
- the background becomes grey

what I would expect:
- the 2 entries for the form to disappear from the screen

Note that clicking "Save" after clicking "Discard" works as expected (even if the 2 form entries are filled, the person/member is not added to the membership).

### LINKED ISSUE #1848: Memberships: make formset with pre-existing fields uneditable until marked with checkbox
In the example below:
obraz

The first two forms (_Library Carpentry_ and _Maddox and Sons_) should be read-only until a checkbox is clicked. The checkbox should be positioned close to the relevant read-only form.

### LINKED ISSUE #1850: Memberships: prevent changing consortium to non-consortium if there are >1 members assigned
Current implementation allows for membership to be changed from consortium (multiple members) to non-consortium (only one member), but a validation should be implemented to prevent that change when more than one member exists.

@maneesha may provide additional logic for the change, for example _the remaining member should receive a different membership role_.

pbanaszkiewicz:
This fixes #1838.
This fixes #1848.
This fixes #1850.

Discard button closes small form:
discard_button

Existing members are only editable if corresponding checkbox is selected:
obraz

Changing from consortium to non-consortium when there are 2 members linked:
obraz


pbanaszkiewicz:
This build up on #1857, so once that PR is merged this will be much cleaner to read.

pbanaszkiewicz:
@lauryndbrown this has been rebased against develop, and should now be much easier for you to review. Thanks in advance!

amy/fiscal/tests/test_membership.py
```diff
@@ -930,6 +991,65 @@ def test_mix_adding_removing_members_from_consortium(self):
...
+            reverse("membership_members", args=[self.membership.pk]),
+            data=data,
+            follow=True,
+        )
+
+        self.assertEqual(response.status_code, 200)  # form failed
+        self.assertIn("__all__", response.context["formset"].errors[0])
+        self.assertEqual(list(self.membership.organizations.all()), [self.org_alpha])
+
+    def test_not_editing_noneditable_members_fails(self):
```
lauryndbrown, line 1023: Is the name here correct? Should the test fail? It looks like it was a success instead. :)

amy/fiscal/tests/test_membership.py
```diff
@@ -930,6 +991,65 @@ def test_mix_adding_removing_members_from_consortium(self):
...
+            "form-0-id": m1.pk,
+        }
+        response = self.client.post(
+            reverse("membership_members", args=[self.membership.pk]),
+            data=data,
+            follow=True,
+        )
+
+        self.assertEqual(response.status_code, 200)  # form failed
+        self.assertIn("__all__", response.context["formset"].errors[0])
```
lauryndbrown, line 1020: You may want to assert the specific error about form values not changing is in the __all__ list.

amy/fiscal/tests/test_membership.py
pbanaszkiewicz: Good catch - the name should be `test_not_editing_noneditable_member_works`. I'll change it.

amy/fiscal/forms.py
```diff
@@ -257,16 +280,19 @@ def __init__(self, *args, **kwargs):
...
         self.helper.layout.append(Field("id"))
         self.helper.layout.append(Field("DELETE"))  # visible; formset adds it
+        self.helper_empty_form.layout = self.helper.build_default_layout(self)
         self.helper_empty_form.layout.append(Field("id"))
         self.helper_empty_form.layout.append(
             Div(Field("DELETE"), css_class="d-none")  # hidden
         )
+        # remove EDITABLE checkbox from empty helper form
+        pos_index = self.helper_empty_form.layout.fields.index("EDITABLE")
+        self.helper_empty_form.layout.pop(pos_index)
```
lauryndbrown, line 292: Maybe one test (or part of one test) that asserts self.helper_empty_form does not include the EDITABLE field.

lauryndbrown (approved):
Looks good. Added some suggestions for the tests nothing major.

amy/fiscal/tests/test_membership.py
pbanaszkiewicz: Resolved by checking for `response.context["formset"].errors[0]` complete value.

amy/fiscal/forms.py
pbanaszkiewicz: I moved Member-related tests to a separate file and added layout tests for MemberForm and MembershipTaskForm.

PR #1860, opened 2021-03-24
feature(Consents): Users prompted to sign new required terms on login.

lauryndbrown:
Dependent on https://github.com/carpentries/amy/pull/1841
Fixes https://github.com/carpentries/amy/issues/1829

- Adds a different action required page. The middleware is currently behind a flag determined in settings.
- Changes the view of the terms from radio buttons to dropdown. Since unsetting a yes-only term required a refresh with radio buttons and looked like a broken state to me.


lauryndbrown:
When env variable is used:
Screen Shot 2021-03-24 at 12 46 22 AM

lauryndbrown:
Ok only a select few are added now. Still needs to be styled
Screen Shot 2021-03-24 at 1 23 34 AM

lauryndbrown:
Screen Shot 2021-03-24 at 12 47 41 PM

lauryndbrown:
This is currently hidden behind a flag in settings.py

amy/consents/views.py
```diff
@@ -1,7 +1,14 @@
-from consents.forms import ActiveTermConsentsForm
+from consents.forms import ActiveTermConsentsForm, RequiredConsentsForm
 from consents.models import Consent
+from consents.util import person_has_consented_to_required_terms
+from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin
+from django.forms import HiddenInput
+from django.http.response import Http404
+from django.shortcuts import redirect, render
+from rest_framework.reverse import reverse
```
pbanaszkiewicz, line 9: This `isort` doesn't do great job here and in other places. It should have understood that `django` and `rest_framework` are our third party.

amy/consents/tests/test_action_required_view.py
```diff
@@ -0,0 +1,298 @@
...
+            )
+
+            # now the dashboard is easily reachable
+            rv = self.client.get(url)
+            self.assertEqual(rv.status_code, 200)
+
+    def test_allowed_urls(self):
+        urls = [
+            reverse("logout"),
+        ]
```
pbanaszkiewicz, line 232: Is this the only allowed URL when person didn't consent to anything?

amy/consents/tests/test_action_required_view.py
```diff
@@ -0,0 +1,298 @@
...
+        # ensure we're logged in
+        self.client.force_login(self.neville)
+        self.assertEqual(person_has_consented_to_required_terms(self.neville), False)
+        with self.terms_middleware():
+            for url in urls:
```
pbanaszkiewicz, line 237: You probably want to force_login in the loop, in case there are more URLs to cover.

amy/consents/tests/test_action_required_view.py
```diff
@@ -0,0 +1,298 @@
...
+                rv = self.client.get(url)
+                # doesn't redirect to the form
+                self.assertIn(rv.status_code, [200, 302])
+                if "Location" in rv:
+                    self.assertNotEqual(rv["Location"], self.form_url)
```
pbanaszkiewicz, line 242: If "doesn't redirect to the form" then why checking 302 + Location header?

amy/consents/tests/test_action_required_view.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def test_next_param(self):
+        """Ensure a non-dispatch URL is reachable through `?next` query
+        string."""
+
+        url = reverse("autoupdate_profile")
+        self.form_url += "?{}".format(urlencode({"next": url}))
```
pbanaszkiewicz, line 249: I'd use local variable for this new form url.

amy/consents/tests/test_action_required_view.py
```diff
@@ -0,0 +1,298 @@
...
+                required_type=Term.PROFILE_REQUIRE_TYPE
+            ).prefetch_active_options()
+            data = {"person": self.neville.pk}
+            for term in terms:
+                data[term.slug] = term.options[0].pk
+            rv = self.client.post(self.form_url, data=data)
+            self.assertEqual(rv.status_code, 302)
+            self.assertEqual(rv["Location"], url)
+
+    def test_old_terms_do_not_affect_terms_middleware(self) -> None:
```
pbanaszkiewicz, line 267: I think you can remove this returned type hint.

amy/consents/views.py
lauryndbrown: @pbanaszkiewicz what do you use for ordering imports? How would you like it to run instead?

amy/consents/tests/test_forms.py
```diff
@@ -1,11 +1,12 @@
...
+from consents.forms import ActiveTermConsentsForm, RequiredConsentsForm
 from consents.models import Consent, Term, TermOption
+from django.db.models.fields import BLANK_CHOICE_DASH
 from django.test import TestCase
 from django.utils import timezone
 from workshops.models import Person
 
 
-class TestTermForm(TestCase):
+class TestActiveTermConsentsForm(TestCase):
```
pbanaszkiewicz, line 9: You introduced a new method to these forms, are you testing it?

amy/consents/tests/test_models.py
```diff
@@ -128,3 +128,44 @@ def test_term_and_term_option_should_match(self):
...
+            content="option1",
+        )
+        term2_option2 = TermOption.objects.create(
+            term=term2,
+            option_type="agree",
+            content="option2",
+        )
+        # Term options method is cached and requires only one
+        # query even during subsequent calls
+        with self.assertNumQueries(2):
```
pbanaszkiewicz, line 157: Great use of the assertion method, I'm glad I can finally see it in the wild 👍

pbanaszkiewicz (approved):
Minor comments.

amy/consents/views.py
pbanaszkiewicz: I don't use anything automatic yet. I want to use isort, I'm just surprised that it worked that way.

https://www.python.org/dev/peps/pep-0008/#imports
> Imports should be grouped in the following order:
>
>   1. Standard library imports.
>   2. Related third party imports.
>   3. Local application/library specific imports.

Anyway, don't worry about it. I'll check out isort soon and see if it has some configuration problem or something.

lauryndbrown: Oh I see. I think there's a configuration setting that can be modified. Before going down this route @pbanaszkiewicz can you tell me how you typically have handled imports?

pbanaszkiewicz:
@lauryndbrown let me know if you want to `CONSENTS` envvar set to true for deployment following merge of this PR.

amy/consents/views.py
lauryndbrown: Opps looks like I asked my question as soon as you answered. no worries. Thanks for explaining

amy/consents/tests/test_forms.py
lauryndbrown: @pbanaszkiewicz what new method are you referring to? The only additional method is the `get_terms` method that is called in the init function. Is that what you mean? That is covered by initializing the form

pbanaszkiewicz: Yes, I meant `get_terms`. It's okay if there are tests for it.

amy/consents/tests/test_action_required_view.py
lauryndbrown: Yes. I don't think anything else is reachable.

amy/consents/tests/test_action_required_view.py
lauryndbrown: Oh sorry this was pulled from the preexisting tests for the actions required view. Let me correct that.

amy/consents/views.py
```diff
@@ -22,3 +29,41 @@ def get_form_kwargs(self):
...
+            messages.success(request, "Agreement successfully saved.")
+
+            if "next" in request.GET:
+                return redirect(request.GET["next"])
+            else:
+                return redirect(reverse("dispatch"))
+        else:
+            messages.error(request, "Fix errors below.")
+    if request.method == "GET":
+        form = RequiredConsentsForm(**kwargs)
```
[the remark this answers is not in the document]
lauryndbrown: Should be handled now.

PR #1863, opened 2021-03-25
Preapproved registration

maneesha:
This changes the text in the instructor training application, with corresponding changes to the confirmation emails.

Note I tried to run migrations but get this error:
`FATAL:  password authentication failed for user "amy"`


pbanaszkiewicz:
@maneesha I'll generate the migration for this PR, but first let's get other PRs with migrations merged.

lauryndbrown:
Hmm I don't recall running into a user password error when getting setup. @maneesha Off the top of my head: 
-  Are you running with docker? `docker-compose -f docker/docker-compose.yml -p amy up -d  database redis`
- Does `make dev_database` work for you? That does run the migrations and create fake data

amy/workshops/migrations/0237_auto_20210331_1851.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.18 on 2021-03-31 18:51
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('workshops', '0236_organization_coordinates'),
```
pbanaszkiewicz, line 9: The migration should depend on '0239' and be the next number - 0240

pbanaszkiewicz (changes requested):

amy/workshops/migrations/0237_auto_20210331_1851.py
maneesha: Thanks. Is it ok to edit the dependencies and the file name manually?

pbanaszkiewicz: I'm not sure the tests would accept that if the `0239` migration was not found.

But you can do this:

```
git checkout develop
git pull
git checkout -
git merge develop
```

After this you can rename the migration as needed (or even drop `amy/workshops/migrations/0237_auto_20210331_1851.py` and recreate with `python manage.py makemigrations`).

lauryndbrown: @maneesha I'm not sure you're familiarity so forgive me if this is too basic. I just wanted to provide some color in case it was needed.

Your pull request is outdated relative to the migration code, so when you used `python manage.py makemigrations` the migration was dependent on the latest migration you had available. But that's not the most current migration in the codebase now, so you code is out of date. It's nice to have the latest code when you're creating migrations so you can ensure all the changes up to this point work well together.

So in this case, I'd recommend doing the following:

Like Piotr mentioned you'll want to pull the most recent version of develop and combine your changes with git. 

After that you will want to do a couple of things:
1. Roll back your migrations `python manage.py migrate workshops zero`
2. Delete the `amy/workshops/migrations/0237_auto_20210331_1851.py` file
3. Rerun `python manage.py makemigrations`. This will create a new file with your changes.
4. Run `make database` and `make serve` so you can see that everything runs ok locally and your changes are there.
5. Make another commit with your new changes and push.

pbanaszkiewicz (approved):
Looks good!

PR #1866, opened 2021-03-26
fix typo in documentation

zkamvar:
We do not have a `workshops_persons` table, but we do have a `workshops_person` table.


pbanaszkiewicz:
Thanks @zkamvar.

PR #1867, opened 2021-03-26
[#1835] Switch Organization.domain to accept domain + path

### LINKED ISSUE #1835: Organisation: changes to domain field
Requirements:

- ~Do not use domain as unique identifier.~
- ~Do not require domain.~
- Allow subdomains and subdirectories (i.e. www.example.com and www.example.com/departmentX  and data.example.com  are all different organizations)

EDIT: removed first 2 requirements as per feedback session 03/17

pbanaszkiewicz:
Previously `Organization.domain` was just that - only a domain. Now it can contain almost entire URL, without scheme.

Generic URL format: `scheme://netloc/path;parameters?query#fragment`

Out of it we're stripping scheme, `//` before the netloc, parameters and fragment.

Breaking change: URLs in API don't contain trailing slashes (otherwise they would not match new domain format correctly).

This fixes #1835.

This PR is built on top of PR #1858 and contains it's commits.

### DIFF
--- a/amy/fiscal/base_views.py
+++ b/amy/fiscal/base_views.py
@@ -1,3 +1,5 @@
+from urllib.parse import unquote
+
 from django.shortcuts import get_object_or_404
 from django.views.generic import FormView
 
@@ -51,3 +53,11 @@ class MembershipFormsetView(GetMembershipMixin, FormView):
 
     def get_success_url(self):
         return self.membership.get_absolute_url()
+
+
+class UnquoteSlugMixin:
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        slug_url = self.kwargs.get(self.slug_url_kwarg)
+        if slug_url is not None:
+            self.kwargs[self.slug_url_kwarg] = unquote(slug_url)
--- a/amy/fiscal/tests/test_organization.py
+++ b/amy/fiscal/tests/test_organization.py
@@ -1,7 +1,7 @@
 from django.urls import reverse
 from django_comments.models import Comment
 
-from fiscal.forms import OrganizationCreateForm
+from fiscal.forms import OrganizationForm, OrganizationCreateForm
 from workshops.models import Organization, Event
 from workshops.tests.base import TestBase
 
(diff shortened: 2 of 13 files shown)


lauryndbrown:
I'll take a look at this one after some of the changes it depends on lands.

pbanaszkiewicz:
@lauryndbrown up for a review, thanks in advance

amy/fiscal/tests/test_organization.py
```diff
@@ -48,24 +48,34 @@ def test_organization_delete(self):
...
+            (
+                "//user:password@example.com:80/path?query",
+                "user:password@example.com:80/path?query",
+            ),
+        ]
+        for domain, expected in test_data:
+            with self.subTest(domain=domain):
+                form = OrganizationForm({"domain": domain})
+                form.full_clean()
+                self.assertIn("domain", form.cleaned_data, form.errors)
```
lauryndbrown, line 77: I think the form.errors is an accidental add.

lauryndbrown (approved):
See my comments on questions. Otherwise looks good.

amy/fiscal/base_views.py
```diff
@@ -51,3 +53,11 @@ def form_valid(self, form):
 
     def get_success_url(self):
         return self.membership.get_absolute_url()
+
+
+class UnquoteSlugMixin:
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
```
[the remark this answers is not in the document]
pbanaszkiewicz: This is a very interesting case. On it's own, `UnquoteSlugMixin` won't do anything useful, but the `super()` will work when you use the mixin in a subclass.

Here's an example:

```python
class SomeView(UnquoteSlugMixin, DetailView):
    pass
```

Due to [MRO](http://www.srikanthtechnologies.com/blog/python/mro.aspx), when `SomeView` is being set up, it's going to call `UnquoteSlugMixin.setup()`, which in turns calls next ancestor through `super()` - `DetailView.setup()`.

amy/fiscal/tests/test_organization.py
pbanaszkiewicz: I sometimes debug the tests and forget to erase the last argument of `self.assert*` calls. It's for the message displayed when the tests fail; in this case I wanted to display form errors to diagnose form input data. All in all, this is pretty harmless, but I'll amend the commit to fix it.

PR #1869, opened 2021-03-27
[#1836] Affiliated organisations

### LINKED ISSUE #1836: Organisation: link affiliated organisations
M2M link, no additional information required. This is not the same as membership consortium.

pbanaszkiewicz:
This fixes #1836.

Builds on top of #1867 and therefore contains some additional commits.


pbanaszkiewicz:
@lauryndbrown this is up for a review, thanks in advance!

docs/amy_database_structure.md
```diff
@@ -143,6 +143,7 @@ The primary tables used in AMY (that will likely appear in every query) are thos
 * `fullname` Human friendly name of the organization
 * `country` Stored as the [two digit country code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)
 * `latitude` and `longitude` Stored as floating point (decimal) numbers
+* `affiliated_organizations` Many-to-many relationship between organizations
```
lauryndbrown, line 146: So https://github.com/carpentries/amy/issues/1836 says it's not the same as a consortium. Could you add a few lines on how this should be used?

lauryndbrown (approved):
It looks good to me. I'm lacking context here so it wasn't really clear to me why this many to many field is now needed. I think adding to the documentation will truly help.

docs/amy_database_structure.md
pbanaszkiewicz: Sure, good idea.

PR #1871, opened 2021-03-28
[#1854][#1855][#1856] Adjustments to rolling-over mechanism, display person membership roles

### LINKED ISSUE #1854: Memberships: when rolling-over, enable editing number of rolled-over
All seats/workshops should not automatically rollover.  User should be able to edit how many roll over.

### LINKED ISSUE #1855: Memberships: when rolling-over, user should be able to allow keeping members (if consortium) and persons
When membership is rolled over, user should see an option to keep members from current membership - but only if the membership is consortium.
In case of non-consortium the only member should be automatically copied to the new membership.

When membership is rolled over, user should see an option to keep persons from current membership.

### LINKED ISSUE #1856: Memberships: display person membership roles if they exist
1. In Person view, member affiliation should show in a “Member roles” view just like views for teaching history, badges, etc. If there are no member roles for that Person, for the admin view, this can display as “No member roles.”
2. For the users own view, this should display only if the Person has a member affiliation role.  Otherwise this should not display at all.

pbanaszkiewicz:
This fixes #1854.
This fixes #1855.
This fixes #1856.

Depends on #1869.

1. Display person membership roles in person details view as well as in the trainee dashboard (but only if person has any membership roles) - #1856
2. Allow editing values for roll-over fields when creating membership through roll-over mechanism - #1854 
3. Allow for copying over member organisations and membership person tasks when rolling over; mechanism is controlled over checkboxes - #1855

### DIFF
--- a/amy/dashboard/views.py
+++ b/amy/dashboard/views.py
@@ -11,18 +11,20 @@ from django.db.models import (
     Prefetch,
     Q,
 )
-from django.shortcuts import render, redirect
+from django.shortcuts import render, redirect, get_object_or_404
 from django.utils.html import format_html
 from django.urls import reverse
 from django.views.decorators.http import require_GET
 from django_comments.models import Comment
 
+from fiscal.models import MembershipTask
 from workshops.models import (
     Airport,
     Badge,
     Event,
     Qualification,
     Person,
+    Task,
     Organization,
     Membership,
     Tag,
@@ -107,12 +109,25 @@ def admin_dashboard(request):
 
 @login_required
 def trainee_dashboard(request):
-    # Workshops person taught at
-    workshops = request.user.task_set.select_related("role", "event")
+    qs = Person.objects.select_related("airport").prefetch_related(
+        "badges",
+        "lessons",
+        "domains",
+        "languages",
+        Prefetch(
+            "task_set",
+            queryset=Task.objects.select_related(),
+        ),
+        Prefetch(
+            "membershiptask_set",
+            queryset=MembershipTask.objects.select_related(),
+        ),
+    )
+    user = get_object_or_404(qs, id=request.user.id)
 
     context = {
         "title": "Your profile",
-        "workshops": workshops,
+        "user": user,
     }
     return render(request, "dashboard/trainee_dashboard.html", context)
 
--- a/amy/dashboard/views.py
+++ b/amy/dashboard/views.py
@@ -116,11 +116,11 @@ def trainee_dashboard(request):
         "languages",
         Prefetch(
             "task_set",
-            queryset=Task.objects.select_related(),
+            queryset=Task.objects.select_related("event", "role"),
         ),
         Prefetch(
             "membershiptask_set",
-            queryset=MembershipTask.objects.select_related(),
+            queryset=MembershipTask.objects.select_related("membership", "role"),
         ),
     )
     user = get_object_or_404(qs, id=request.user.id)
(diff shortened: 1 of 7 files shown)


amy/workshops/views.py
```diff
@@ -250,17 +251,30 @@ class PersonDetails(OnlyForAdminsMixin, AMYDetailView):
...
+                "task_set",
+                queryset=Task.objects.select_related(),
+            ),
             Prefetch(
                 "task_set",
                 to_attr="training_tasks",
                 queryset=Task.objects.filter(
                     role__name="learner", event__tags__name="TTT"
-                ),
+                ).select_related(),
```
lauryndbrown, line 271: Should this have a select related?

amy/dashboard/views.py
```diff
@@ -107,12 +109,25 @@ def admin_dashboard(request):
...
-    # Workshops person taught at
-    workshops = request.user.task_set.select_related("role", "event")
+    qs = Person.objects.select_related("airport").prefetch_related(
+        "badges",
+        "lessons",
+        "domains",
+        "languages",
+        Prefetch(
+            "task_set",
+            queryset=Task.objects.select_related(),
```
lauryndbrown, line 119: So I do not think it's generally a good idea to call select related without arguments. From the docs:

>There may be some situations where you wish to call select_related() with a lot of related objects, or where you don’t know all of the relations. In these cases it is possible to call select_related() with no arguments. This will follow all non-null foreign keys it can find - nullable foreign keys must be specified. This is not recommended in most cases as it is likely to make the underlying query more complex, and return more data, than is actually needed.

pbanaszkiewicz: Resolved

amy/fiscal/tests/test_membership.py
```diff
@@ -894,95 +909,272 @@ def setUpMembership(self):
...
-            # fields ignored
-            "workshops_without_admin_fee_rolled_from_previous": "invalid value",
-            "public_instructor_training_seats_rolled_from_previous": "invalid value",
-            "inhouse_instructor_training_seats_rolled_from_previous": "invalid value",
-        }
-
-        form = MembershipRollOverForm(data)
+    def test_form_validation_rolled_fields(self):
+        test_data = [
+            # PASS: centrally org. workshops maxed out, others set to 0
```
lauryndbrown, line 924: I'm not sure I know what `centrally org` means

amy/fiscal/tests/test_membership.py
```diff
@@ -894,95 +909,272 @@ def setUpMembership(self):
...
+                False,
+                [
+                    "workshops_without_admin_fee_rolled_from_previous",
+                    "public_instructor_training_seats_rolled_from_previous",
+                    "inhouse_instructor_training_seats_rolled_from_previous",
+                ],
+            ),
+        ]
+        for data, max_values, is_valid, failed_fields in test_data:
+            with self.subTest(max_values=max_values):
```
lauryndbrown, line 1072: Oh haven't seen subtest used before. Nice.

lauryndbrown (approved):
Everything looks ok. I appreciated the tests. Though I had a question about the comment you wrote.

amy/fiscal/tests/test_membership.py
pbanaszkiewicz: _Centrally-organised workshops_ ie. workshops that are handled by The Carpentries staff. Our regional coordinators for example help with finding instructors and helpers.
_Self-organised workshops_ are administered entirely by instructors or other staff.

pbanaszkiewicz:
The code scanning results error from CodeQL that appears here will be fixed as part of #1876.

PR #1872, opened 2021-03-29
feature(Consents): Email user when there is a new required consent to approve.

lauryndbrown:
Fixes https://github.com/carpentries/amy/issues/1828


amy/autoemails/models.py
```diff
@@ -289,6 +289,10 @@ class Trigger(ActiveMixin, CreatedUpdatedMixin, models.Model):
         ("workshop-request-response1", "Response to Workshop Request 1"),
         ("workshop-request-response2", "Response to Workshop Request 2"),
         ("workshop-request-response3", "Response to Workshop Request 3"),
+        (
+            "consent-required",
```
lauryndbrown, line 293: @pbanaszkiewicz could you tell me what a trigger is used for vs an action? Looks like the action classes construct the email. What do the triggers represent?

amy/consents/admin.py
```diff
@@ -1,5 +1,34 @@
...
+from autoemails.base_views import ActionManageMixin
+from autoemails.actions import NewConsentRequiredAction
+
+logger = logging.getLogger("amy.signals")  # TODO... why signals logger?
+scheduler = django_rq.get_scheduler("default")
+redis_connection = django_rq.get_connection("default")
+
+
+def send_consent_email(request, term: Term) -> None:
+    jobs, rqjobs = ActionManageMixin.add(
```
lauryndbrown, line 17: Saw ActionManageMixin used like this in a different view. Not certain if this is the correct way however

lauryndbrown:
TODO: this pr needs to be rebased on https://github.com/carpentries/amy/pull/1880 to work properly

amy/autoemails/models.py
pbanaszkiewicz: Trigger represents a cause for action.
Action is not a model, it's a Python object, so we needed a way to mark what users can do that would result in email sent.

amy/autoemails/actions.py
```diff
@@ -1133,3 +1134,66 @@ def get_additional_context(self, objects, *args, **kwargs):
...
+
+class NewConsentRequiredAction(BaseAction):
+    """
+    Action for asking users to consent to newly required terms. This email
+    should be sent when a new required Term is created.
+
+    How to use it:
+
+    >>> triggers = Trigger.objects.filter(active=True,
+                                          action='consent-required-new')
```
pbanaszkiewicz, line 1147: action is actually `consent-required`, without `-new`

amy/autoemails/actions.py
```diff
@@ -1133,3 +1134,66 @@ def get_additional_context(self, objects, *args, **kwargs):
...
+        """Assuming self.context is ready, overwrite email's recipients
+        with selected ones."""
+        try:
+            return self.context["all_emails"]
+        except (AttributeError, KeyError):
+            return None
+
+    def all_recipients(self) -> str:
+        """If available, return string of all recipients."""
+        # TODO what is this versus recipients?
```
pbanaszkiewicz, line 1172: One is used in sending the action, the other is used for displaying in Django Admin panel. I don't remember which one is which.

amy/autoemails/actions.py
```diff
@@ -1133,3 +1134,66 @@ def get_additional_context(self, objects, *args, **kwargs):
...
+        try:
+            terms = self.context_objects["terms"]
+            # TODO: Dependent on unset consents being stored
+            person_emails = list(
+                Consent.objects.filter(
```
pbanaszkiewicz, line 1177: So this sends one email to multiple recipients. Can you check if that would result in one email or in one email per recipient? We'd rather not reveal people's email addresses.

amy/consents/admin.py
pbanaszkiewicz: This is okay.

pbanaszkiewicz (approved):
This is looking okay.

amy/autoemails/actions.py
lauryndbrown: I messaged @pbanaszkiewicz  individually, but I'm going to summarize what we spoke about here:
- This appears to send 1 emaill with a very long to message to all the users included
- The way to avoid that is: https://anymail.readthedocs.io/en/stable/sending/templates/#batch-send
-  This appears to still send one email until it reaches mailgun. Mailgun handles separating them into multiple emails, so there's no way to double check that everything is ok without a mailgun sandbox. 
- I also hit a column length issue because my recipients list is so long. 

For now I'm going to:
- preserve the work I did to try to do bulk emails with Mailgun in another branch
- Convert this branch to manually creating an individual email per user so we can be sure it works correctly

What would be helpful in the future:
- Getting a sandbox mailgun account to test out the functionality and/or putting the mailgun account in AMY to sandbox mode so that we can test this once.

lauryndbrown: @pbanaszkiewicz saving the pr here: https://github.com/lauryndbrown/amy/compare/1828/email-user-when-new-required-consent-to-approve...lauryndbrown:lb/trying-to-mass-email-correctly?expand=1

lauryndbrown:
Bulk send message
Screen Shot 2021-04-17 at 12 04 08 PM

lauryndbrown:
Screen Shot 2021-04-17 at 12 14 42 PM

lauryndbrown:
Merging with the caveats above.

PR #1862, opened 2021-03-24
ref(consent): Add helptext to Terms.

lauryndbrown:
Adds a help text to the terms, so that not everything is written in the label.

Fixes: https://github.com/carpentries/amy/issues/1861

PR #1880, opened 2021-04-03
feature(Consents): Store unset consents when term or person is created

lauryndbrown:
Changes the way unset consents work. Stores the unset consents after a conversation with François wanting to see them in the admin view. 

Fixes https://github.com/carpentries/amy/issues/1877


lauryndbrown:
@pbanaszkiewicz once this is merged plus https://github.com/carpentries/amy/pull/1862 It should be fine to deploy the consents features so that folks can see it.

lauryndbrown:
@pbanaszkiewicz since we are now storing unset consents, I'm realizing we will have quite a few consents since each time consents change they are stored in the database. I'm assuming we're fine on space here, but if not, I did just realize that Django Admin has a default change history feature that I can explore. Let me know if space is an issue I should be worry about and I can change how the change history is stored.

amy/consents/tests/test_action_required_view.py
```diff
@@ -1,8 +1,9 @@
+from amy.consents.tests.helpers import reconsent
```
pbanaszkiewicz, line 1: We don't import `from amy`, if you need you can import relatively `from .helpers` or from package `from consents.tests.helpers`.

amy/consents/tests/test_signals.py
```diff
@@ -0,0 +1,137 @@
+from consents.models import Consent, Term
```
pbanaszkiewicz, line 1: Nice, I really like that you added tests for signals.

pbanaszkiewicz (approved):
Looks good, I left a note on imports - please check if the different style of imports works.

pbanaszkiewicz:
> 
> 
> @pbanaszkiewicz once this is merged plus #1862 It should be fine to deploy the consents features so that folks can see it.

So I should use `CONSENTS=1` envvar?



> 
> 
> @pbanaszkiewicz since we are now storing unset consents, I'm realizing we will have quite a few consents since each time consents change they are stored in the database. I'm assuming we're fine on space here, but if not, I did just realize that Django Admin has a default change history feature that I can explore. Let me know if space is an issue there.

Space is not an issue. :)

lauryndbrown:
@pbanaszkiewicz `CONSENTS=1` should work great. Thanks!

amy/consents/tests/test_action_required_view.py
lauryndbrown: That should be resolved now

PR #1883, opened 2021-04-05
fix(MakeFile): Fix command createsuperuser in the MakeFile.

lauryndbrown:
The command has a typo and would fail out before creating the superuser.
See https://docs.djangoproject.com/en/2.2/intro/tutorial02/#creating-an-admin-user

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ dev_database :
 	${MANAGE} loaddata amy/autoemails/fixtures/templates_triggers.json
 	${MANAGE} fake_database
 	${MANAGE} createinitialrevisions
-	${MANAGE} create_superuser
+	${MANAGE} createsuperuser
 
 ## node_modules : install front-end dependencies using Yarn
 node_modules : package.json


Makefile
```diff
@@ -30,7 +30,7 @@ dev_database :
 	${MANAGE} loaddata amy/autoemails/fixtures/templates_triggers.json
 	${MANAGE} fake_database
 	${MANAGE} createinitialrevisions
-	${MANAGE} create_superuser
+	${MANAGE} createsuperuser
```
pbanaszkiewicz, line 33: This wasn't broken. We have our own command `create_superuser` in `amy/workshops/management/commands/create_superuser.py`.

pbanaszkiewicz (changes requested):
I don't know what bug this is referring to. Makefile works for me, I use `make dev_database` very often.

Makefile
lauryndbrown: Oh I didn't realize. I thought it was a typo. Thanks for letting me know.

PR #1885, opened 2021-04-05
feature(Consents): Archive terms and term options from Django Admin view

lauryndbrown:
Fixes https://github.com/carpentries/amy/issues/1882
Dependent on https://github.com/carpentries/amy/pull/1880

### DIFF
--- a/amy/consents/admin.py
+++ b/amy/consents/admin.py
@@ -8,6 +8,11 @@ from django.contrib import admin
 
 
 class ArchiveActionMixin:
+    """
+    Used for AdminModels that have the CreatedUpdatedArchivedMixin
+    and need to archive rather than delete the model.
+    """
+
     change_form_template = "consents/admin_change_form_term.html"
     actions = ["archive"]
     readonly_fields = ("archived_at",)
--- a/amy/consents/admin.py
+++ b/amy/consents/admin.py
@@ -6,6 +6,10 @@ from django.contrib import admin
 class TermOptionInline(admin.TabularInline):
     model = TermOption
     extra = 0
+    readonly_fields = ("archived_at",)
+
+    def has_delete_permission(self, *args, **kwargs):
+        return False
 
 
 class TermAdmin(admin.ModelAdmin):
@@ -13,6 +17,10 @@ class TermAdmin(admin.ModelAdmin):
     inlines = [
         TermOptionInline,
     ]
+    readonly_fields = ("archived_at",)
+
+    def has_delete_permission(self, *args, **kwargs):
+        return False
 
 
 class ConsentAdmin(admin.ModelAdmin):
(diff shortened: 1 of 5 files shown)


pbanaszkiewicz:
@lauryndbrown looks like this is dependant on another PR, can you ping me once this one is cleared up?

lauryndbrown:
@pbanaszkiewicz could you take another look?

amy/consents/admin.py
```diff
@@ -1,15 +1,71 @@
...
+            self.message_user(
+                request,
+                f"Error: Cannot archive. {obj} is already archived.",
+                level=messages.ERROR,
+            )
+        else:
+            obj.archive()
+
+    def has_delete_permission(self, *args, **kwargs):
+        return False
```
pbanaszkiewicz, line 47: What's the purpose of this method? Can you document in a docstring? Edit: or comment

amy/consents/admin.py
```diff
@@ -31,9 +87,13 @@ class ConsentAdmin(admin.ModelAdmin):
     def get_term_option_type(self, obj):
         return obj.term_option.option_type if obj.term_option else None
 
+    def has_delete_permission(self, *args, **kwargs):
```
pbanaszkiewicz, line 90: Not sure if you wanted this written explicitly or is this a mistake since the parent class defines the same method.

amy/consents/models.py
```diff
@@ -85,6 +99,19 @@ class TermOption(CreatedUpdatedArchivedMixin, models.Model):
     def __str__(self):
         return f"{self.content} ({self.option_type})"
 
+    def archive(self) -> None:
+        """
+        Archive self and archive all Consent objects that have the
+        term option as their answer.
+        If the Term this term option is attached to is still active,
```
pbanaszkiewicz, line 106: ~~Something wrong with this sentence.~~
Edit: scratch that, something wrong was with me not the sentence 😅

pbanaszkiewicz (approved):

amy/consents/admin.py
lauryndbrown: Ths class does not inherit from the `ArchiveActionMixin` I'll add a comment here to make that more clear. I do not think admins should be able to archive individual user's consents.

amy/consents/admin.py
lauryndbrown: I'll add a comment with a link to the documentation. This is a feature of django admin, that lets django know that this admin view cannot delete models. Admin views can typically delete models by default. I instead want them archived from the admin view and related models archived.

lauryndbrown:
@pbanaszkiewicz I totally forgot to add a screen shot here. I altered the admin view template:

Screen Shot 2021-04-14 at 10 20 21 AM

Typically the default django delete happens on the list view. I felt that was a little bit dangerous to place and I wanted a warning message to the user to provide context about what they are doing:
Screen Shot 2021-04-14 at 11 37 41 AM

Terms:
Screen Shot 2021-04-14 at 11 41 31 AM

lauryndbrown:
Going to merge since I think it's probably ok, but let me know if there's something about the UI you dislike.

PR #1896, opened 2021-04-13
feature(Consents): Add Existing terms to the consents models in a migration. 

lauryndbrown:
Fixes https://github.com/carpentries/amy/issues/1870

- Revised tests to inherit from a central base class. 
- Fixed tests to account for the terms in the db
- changed fake_database to assume that there are terms in the database rather than create them itself.

Tested migration:
- `python manage.py migrate`
- `python manage.py migrate consents 0003_term_help_text`


lauryndbrown:
The first is yes-only. The others are yes and no.

Screen Shot 2021-04-14 at 10 06 46 AM

pbanaszkiewicz (approved):
Looking good!

PR #1903, opened 2021-04-17
[#1898] Fix django messages tags

### LINKED ISSUE #1898: Admin: Missing Bootstrap CSS
Django admin is missing bootstrap css. In the settings.py we have changed the default message tags `MESSAGE_TAGS`. So the default css that shows up in django admin does not work.

pbanaszkiewicz:
This fixes #1898 by adding default tag classes to changed message tags.

Thanks to this change, messages shown in Django Admin display with
correct styles, and current styling of messages in other parts of AMY
is maintained.

A warning message in Django Admin:
obraz

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -443,10 +443,11 @@ MANAGERS = ADMINS
 from django.contrib.messages import constants as message_constants  # noqa
 
 MESSAGE_TAGS = {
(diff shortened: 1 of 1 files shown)


lauryndbrown (approved):
Oh nice. I appreciate the simplicity of this fix.

PR #1906, opened 2021-04-18
isort - imports sorting utility

pbanaszkiewicz:
This PR adds isort to the project, enables it as a pre-commit hook and updates all existing imports. Finally, the commit revision is added to git blame exceptions to not disrupt `git blame` command.


lauryndbrown (approved):
Perfect! Thanks for adding this.

PR #1908, opened 2021-04-20
feat(Profile Archival): All consents are unset when a person is archived.

lauryndbrown:
Fixes https://github.com/carpentries/amy/issues/1888

Dependent upon https://github.com/carpentries/amy/pull/1905
I needed to move signals in Workshops app so that they could be registered by the receivers in Consents. The way it was structured before the Signal init function ran after the Consents app had registered the signal (wiping out the receiver). Putting it in a module __init__ prevents that from happening.

### DIFF
--- a/amy/consents/receivers.py
+++ b/amy/consents/receivers.py
@@ -1,9 +1,9 @@
 from django.db.models.signals import post_save
-from workshops.signals import person_archived_signal
 from django.dispatch import receiver
 
 from consents.models import Consent, Term
 from workshops.models import Person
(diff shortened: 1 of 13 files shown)


amy/consents/receivers.py
```diff
@@ -1,7 +1,9 @@
 from django.db.models.signals import post_save
+from workshops.signals import person_archived_signal
```
pbanaszkiewicz, line 2: isort would move this import to line 6

pbanaszkiewicz (approved):

PR #1910, opened 2021-04-21
[#1711] Reactivate instructor survey links

### LINKED ISSUE #1711: Re-activate fields for pre/post surveys
We used to have the fields `instructors_pre` and  `instructors_post` in the `workshops_event` table. The fields still exist but are hidden from view.  We would like to put these fields back in the `Event` view.  These fields will store the html link to view the pre/post survey results for that event.  

Requesting review from @fmichonneau  before working on this task to be sure there are no privacy concerns with storing the survey results links in AMY.

pbanaszkiewicz:
This fixes #1711 by re-introducing instructors pre/post survey links into event details page and event create/edit form.


lauryndbrown (approved):
Looks good. Do these affect any tests? Should they?

pbanaszkiewicz:
> Do these affect any tests? Should they?

No tests affected - probably because this was one of the earliest parts of AMY and it was tested in a different way (HTML scraping) than I'd do things today (testing the form directly, then integration testing).

The fields aren't required so that can also be a reason why no tests were triggered.

PR #1922, opened 2021-05-02
fix(Profile Archival): Remove more  personal information when the user is  archived.

lauryndbrown:
I missed some personal information. Removing those as well.
Fixes https://github.com/carpentries/amy/issues/1918

Before Archival:
Screen Shot 2021-05-02 at 6 02 37 PM

After Archival:
Screen Shot 2021-05-02 at 6 03 03 PM

### DIFF
--- a/amy/workshops/models.py
+++ b/amy/workshops/models.py
@@ -46,7 +46,6 @@ from workshops.mixins import (
 )
 from workshops.signals import person_archived_signal
 
-
 STR_SHORT = 10  # length of short strings
 STR_MED = 40  # length of medium strings
 STR_LONG = 100  # length of long strings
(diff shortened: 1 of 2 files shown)


amy/workshops/models.py
```diff
@@ -988,6 +987,13 @@ def archive(self) -> None:
         self.is_active = False
         self.occupation = ""
         self.orcid = ""
+        self.secondary_email = ""
+        self.gender = GenderMixin.UNDISCLOSED
+        self.gender_other = ""
+        # TODO(lb): Remove references to agreements when Consents project is launched
```
pbanaszkiewicz, line 993: Is there a story / ticket for this so that we don't forget?

pbanaszkiewicz (approved):

amy/workshops/models.py
lauryndbrown: So I think it's covered in a broad one here: https://github.com/carpentries/amy/issues/1881 Let me know if you prefer something more detailed. But I think this naturally gets removed when this ticket is done.

PR #1923, opened 2021-05-03
feature(Profile Archival): Added social auth removal to Person archive.

lauryndbrown:
Fixes https://github.com/carpentries/amy/issues/1920

Before Archival:
Screen Shot 2021-05-02 at 6 48 03 PM

After Archival:
Screen Shot 2021-05-02 at 6 52 55 PM

### DIFF
--- a/amy/workshops/tests/test_person.py
+++ b/amy/workshops/tests/test_person.py
@@ -1035,14 +1035,16 @@ def github_username_to_uid_mock(username):
         "username": "1",
         "changed": "2",
         "changedagain": "3",
+        "user-github": "4",
+        "admin-github": "5",
     }
     return username2uid[username]
 
 
+@patch("workshops.github_auth.github_username_to_uid", github_username_to_uid_mock)
 class TestPersonAndUserSocialAuth(TestBase):
     """ Test Person.synchronize_usersocialauth and Person.save."""
 
-    @patch("workshops.github_auth.github_username_to_uid", github_username_to_uid_mock)
     def test_basic(self):
         user = Person.objects.create_user(
             username="user",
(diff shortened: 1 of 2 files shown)


amy/workshops/tests/test_person.py
```diff
@@ -1384,7 +1384,18 @@ def setUp(self):
             password="pass",
         )
         self.user.data_privacy_agreement = True
+        self.user.github = "user-github"
+        self.user.is_active = True
         self.user.save()
+        self.user.synchronize_usersocialauth()
```
pbanaszkiewicz, line 1390: Looks like this would call the GitHub API for [this user](https://github.com/user-github). Can you mock the underlying `Github` class (see `github_auth.py` for details)?

amy/workshops/tests/test_person.py
```diff
@@ -1384,7 +1384,18 @@ def setUp(self):
...
+        assert self.user.social_auth.all()
+
+        # Add social auth to admin
+        self.admin.github = "admin-github"
+        self.admin.is_active = True
+        self.admin.synchronize_usersocialauth()
```
pbanaszkiewicz, line 1396: Does admin account in tests require sync with GitHub?

pbanaszkiewicz (changes requested):
I think there's some mocking required. Other than that this looks good.

amy/workshops/tests/test_person.py
lauryndbrown: Gotcha added it.

amy/workshops/tests/test_person.py
lauryndbrown: Not necessarily. I mostly have just done it in case there was any difference between the way the admin works in comparison to the normal user.

pbanaszkiewicz (approved):

PR #1924, opened 2021-05-03
[#1879] Add "Withdrawn" state to TrainingRequest

### LINKED ISSUE #1879: Add "Withdrawn" state to Instructor Training Application
We currently have the states Pending, Accepted, and Discarded.
We should also have the state Withdrawn. This will be used when a trainee informs us they are no longer interested in the Instructor Training program.

@klbarnes20 @karenword  what do you think?

pbanaszkiewicz:
This fixes #1879.

New state: 
obraz
obraz


amy/workshops/mixins.py
```diff
@@ -149,6 +149,27 @@ def active(self):
         return self.state == "p"
 
 
+class StateExtendedMixin(models.Model):
+    """State field with representation of 'Withdrawn' state, for now only used in
+    TrainingRequest.
+
+    This was rewritten instead of inherited from `StateMixin` - there were some
+    issues with `get_state_display` method for "withdrawn" state."""
```
lauryndbrown, line 157: @pbanaszkiewicz could you say more about this issue? I don't really understand why inheriting from this class causes issues?

lauryndbrown (approved):
I don't see anything blocking here. But I'm not clear on why the class was recreated.

amy/workshops/mixins.py
pbanaszkiewicz: @lauryndbrown I didn't get to the root of the problem. I only discovered that `get_state_display` didn't display correctly the new state - "Withdrawn". Instead there only was `w`, like the field didn't use the new choice values.

If I remember correctly, I rewrote pretty much the whole `StateMixin` except for the `active` method - only it was inherited.

PR #1948, opened 2021-05-08
[#1939] Make affiliated organisations symmetrical

### LINKED ISSUE #1939: Organisation: display affiliated organisations linking to current org.
If organisation A links to organisation B through affiliated organisations, we can see the link in organisation A details page, but it's not visible in org. B details page.

Display the organisations linking to current organisation on it's details page (in this example: display link to organisation A on organisation B's detail page).

pbanaszkiewicz:
Symmetrical option makes affiliated organisation automatically
"point back" to the organisation that initiated relationship.

For example, previously we could relate organistations: A->B.
To relate B with A we'd have to add another relation entry B->A.

After the change, once we set A->B then B->A is automatically set
(we end up with symmetrical relation `A<->B`).

This fixes #1939.


lauryndbrown (approved):
Looks good. I haven't seen this used very often. You probably don't need a test since this is literally built in django functionality, but you might still consider one if there's some additional logic that relies on this being symmetrical.

pbanaszkiewicz:
@lauryndbrown thanks for the suggestion, I agree that this is worth checking as it is directly a feature requested by PO. I added the test and will merge as soon as checks pass.

PR #1949, opened 2021-05-08
[#1945] Change Membership public status def. value

### LINKED ISSUE #1945: Memberships: make private by default (amend migration)
The migration `amy/workshops/migrations/0237_auto_20210321_1056.py` sets default value to `private`, but migration `amy/workshops/migrations/0228_auto_20210204_0658.py` introduced the value as public first.

The migrations will have to be changed in order to make production data private by default.

pbanaszkiewicz:
The migrations order would:
1. introduce "public_status" field with "public" default value
2. change default value to "private".

This was wrong because all the memberships would become public once
the first migration is run.

Changes introduced in this commit/PR make "private" default for the first
migration; the second migration is removed as it's now a no-op.

This fixes #1945.

### DIFF
--- a/amy/workshops/migrations/0228_auto_20210204_0658.py
+++ b/amy/workshops/migrations/0228_auto_20210204_0658.py
@@ -23,7 +23,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='membership',
             name='public_status',
-            field=models.CharField(choices=[('public', 'Public'), ('private', 'Private')], default='public', max_length=20,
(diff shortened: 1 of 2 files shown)


lauryndbrown (approved):
So I'm sure this works fine. When you're changing defaults, it's often a bit cleaner to insert the value you want then change it. So
- Change all public memberships to private 
- Set the default

If you could please add a comment at the top of the second migration mentioning that it is a no op and why (basically the PR description).

pbanaszkiewicz:
@lauryndbrown I fixed the migration 0237 with some extensive comment, thank you!

Issue #1937, opened 2021-05-05
Membership: document multiple extensions


pbanaszkiewicz:
Currently there's only one extension available and it gets overwritten if membership is extended multiple times.

1. Change the DB structure to hold multiple extensions (e.g. `[30, 15, 15]`)
2. Display multiple extensions and their cumulative value in membership
3. Add a comment “<User name> extended membership by 15 days on 2021-05-05” - similar to the story #1929.

No additional metadata needs to be stored with each extension.

PR #1950, opened 2021-05-09
[#1944] Make comments work w/o user's email

### LINKED ISSUE #1944: Comments: adding a comment from account without email address triggers form error
When user without email address tries to add a comment, then form validation prevents them.

Expected behaviour: a comment can be added even if user doesn't have email address on file.

pbanaszkiewicz:
This fixes #1944 by disabling required email field - now it's not
required.

This means that some users, who don't have email on file, will be able
to post comments.

### DIFF
--- a/amy/extcomments/tests.py
+++ b/amy/extcomments/tests.py
@@ -1,1 +1,39 @@
-# Create your tests here.
+from django.test import TestCase
+import django_comments
+
+from workshops.models import Organization, Person
+
+
+class TestEmailFieldRequiredness(TestCase):
+    def test_email_field_requiredness(self):
+        """Regression test for #1944.
+
+        Previously a user without email address would not be able to add a comment."""
+        # Arrange
+        person = Person.objects.create(
+            personal="Ron",
+            family="Weasley",
+            username="rw",
+            is_active=True,
+            email="",
+        )
+        person.set_password("testrwpassword")
+        self.client.login(username="rw", password="testrwpassword")
+
+        organization = Organization.objects.create(
+            domain="example.org", fullname="Example Organisation"
+        )
+        CommentForm = django_comments.get_form()
+
+        data = {
+            "honeypot": "",
+            "comment": "Content",
+            "name": "Ron",
+            **CommentForm(organization).generate_security_data(),
+        }
+
+        # Act
+        form = CommentForm(organization, data)
+
+        # Assert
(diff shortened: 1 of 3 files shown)


amy/extcomments/tests.py
```diff
@@ -1 +1,39 @@
...
+        # Arrange
+        person = Person.objects.create(
+            personal="Ron",
+            family="Weasley",
+            username="rw",
+            is_active=True,
+            email="",
+        )
+        person.set_password("testrwpassword")
+        self.client.login(username="rw", password="testrwpassword")
```
lauryndbrown, line 21: @pbanaszkiewicz can you assert that self.client.login returns True? 

I have a feeling it might not, and it's essential for this test.

lauryndbrown (changes requested):

amy/extcomments/tests.py
```diff
@@ -1 +1,39 @@
...
+        """Regression test for #1944.
+
+        Previously a user without email address would not be able to add a comment."""
+        # Arrange
+        person = Person.objects.create(
+            personal="Ron",
+            family="Weasley",
+            username="rw",
+            is_active=True,
+            email="",
```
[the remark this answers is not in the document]
pbanaszkiewicz: This is an interesting topic. `Person.email` is marked with `unique=True`, and additionally we know it's a string field (`CharField`).

The uniqueness is only enforced on the string values of the field. This means that if we have a case in AMY where we want to enter users with not known email, we should enter `email=None` because otherwise we'd receive Integrity Error (unique constraint failed). Additionally, this also means that there can be at most one user with `email=""`.

The test should work with both `email=""` and `email=None`, as the crucial part is not having to provide any email in the form's data.

amy/extcomments/tests.py
pbanaszkiewicz: @lauryndbrown you were right, it didn't log the user in. No wonder we're using `force_login` so much in tests.

I'm reworking the tests, and I'll resubmit this for your review afterwards.

pbanaszkiewicz:
@lauryndbrown I improved the tests according to your suggestions (thanks!), please re-review.

pbanaszkiewicz:
This PR has been extended by util function `add_comment_for_object`. #1937 (PR #1954) is dependent, and should work once this code is merged.

lauryndbrown (approved):

Issue #1875, opened 2021-03-31
Developer tools and automation


pbanaszkiewicz:
Some developer tools and improvements to complete:

- [x] docker-compose with database and redis
- [x] [pre-commit](https://pre-commit.com/) with [black](https://github.com/psf/black)
- [x] pre-commit with [flake8](https://flake8.pycqa.org/en/latest/)
- [ ] pre-commit with [isort](https://pycqa.github.io/isort/)
- [x] GitHub CI workflow for building docs
- [x] GitHub CI workflow for running tests
- [ ] [poetry](https://python-poetry.org/)
- [ ] WIP

Some items may need fine-tuning.

PR #1951, opened 2021-05-09
Improve performance with advanced statics caching

pbanaszkiewicz:
This replaces django-compressor with whitenoise - a WSGI middleware
which handles caching, compression, etc. of static files.


pbanaszkiewicz:
There's a couple of issues with this PR.

1. whitenoise setup/teardown during WSGI application setup/weardown is painfully slow, and hence it slows down testing by a lot.
2. `node_modules` contains sources for JS/CSS packages we're dependent on, but the whole directory is set as static source (`STATICFILES_DIRS` setting). This slows down the setup and `./manage.py collectstatic`, and has high impact on testing as well.
3. there are some unused dependencies (to be fixed with next commit).

All in all, here's related issue: https://github.com/evansd/whitenoise/issues/94
And here's related documentation: http://whitenoise.evans.io/en/stable/django.html#whitenoise-makes-my-tests-run-slow

It also looks like we should switch to JS development with webpack to shear off files ending up in `staticfiles` directory.

pbanaszkiewicz:
Ref #1875

lauryndbrown:
@pbanaszkiewicz are you wanting my review on this one?

pbanaszkiewicz:
@lauryndbrown no, I don't want you to review this; it's not ready yet and frankly quite some work needs to be done.

pbanaszkiewicz:
I think I have a better solution.

PR #1952, opened 2021-05-16
[#1930] Distinct membership members, person tasks

### LINKED ISSUE #1930: Membership: only allow distinct organisations+roles pairs as members
Enforce uniqueness on organisation+role pairs in membership members.

@maneesha small warning: I will have to remove non-unique entries from test-amy when this gets deployed. Production shouldn't be affected as long as this tickets gets into soonest release.

pbanaszkiewicz:
This PR fixes #1930 by enforcing uniqueness constraint on DB level, and by adding the uniqueness validation in the membership formsets.

The solution isn't perfect, but it's the only one I got working. The biggest issue was the uniqueness validation happening before I was able to write "my own" membership to the formsets.


amy/workshops/migrations/0244_auto_20210509_2051.py
```diff
@@ -0,0 +1,39 @@
+# Generated by Django 2.2.18 on 2021-05-09 20:51
+
+from django.db import migrations, models
+
+
+def find_duplicate_members(apps, schema_editor):
+    """The next migration will enforce a unique constraint in DB, which may result in
```
lauryndbrown, line 7: Since this only alerts on duplicates, is it somehow handled manually? How are the duplicates addressed?

amy/fiscal/views.py
```diff
@@ -248,6 +248,9 @@ class MembershipMembers(
 ):
     permission_required = "workshops.change_membership"
 
+    def get_formset(self, *args, **kwargs):
+        return modelformset_factory(Member, MemberForm, *args, **kwargs)
```
lauryndbrown, line 252: nit: This appears to just have been moved no changes. probably isn't necessary

amy/fiscal/base_views.py
```diff
@@ -31,25 +31,16 @@ def get_form_class(self):
...
-            instance.save()
-
-        # remove deleted objects
-        for instance in form.deleted_objects:
-            instance.delete()
-
-        return super().form_valid(form)
+    def form_valid(self, formset):
+        formset.save()  # handles adding, updating and deleting instances
+        return super().form_valid(formset)
```
lauryndbrown, line 43: Since we are saving before checking if it is valid, how does that affect things?  If so is there test coverage for this?

lauryndbrown (approved):
Formsets are definitely tricky. Relying on hidden fields can potentially result in security concerns because as you mentioned it's not much work for a user to change them. If there's a way of grabbing the membership through the database/url instead of passing it along with the form that's probably preferred since the user shouldn't change/select it.

That said as long as there's test coverage and you've appropriately noted the behavior, I think we're ok.

amy/workshops/migrations/0244_auto_20210509_2051.py
pbanaszkiewicz: Migration step will fail and the code upgrade will not work. This will probably happen on test server, but will not happen in production server.

It requires human intervention - someone who can assess and remove the duplicates.

amy/fiscal/views.py
pbanaszkiewicz: Good catch! This was moved to reflect the order of methods in a "sibling" view - `MembershipTasks`.

amy/fiscal/base_views.py
pbanaszkiewicz: Actually at this point we know the form is valid: https://ccbv.co.uk/projects/Django/3.1/django.views.generic.edit/FormView/#post
First `post()` checks `form.is_valid()`, and then `form_valid()` is executed.

As for the tests, `TestMembershipMembers` is a pretty big class containing tests for various cases of added/removed/mixed members.

In general, this change simplifies things, as the operations replaced (e.g. `for instance in form.deleted_objects: instance.delete()`) are covered by `formset.save(commit=True)` (commit is True by default).

PR #1954, opened 2021-05-16
[#1937] Multiple membership extensions

### LINKED ISSUE #1937: Membership: document multiple extensions

pbanaszkiewicz:
This fixes #1937.

The existing field `extended` was migrated into `extensions` which
is a list (array) of single extensions.

For example, `[30, 60]` means that membership was extended by
30 and then by 60 days, for a total of 90 days extension.

The total (cumulative) value is also displayed in the membership details
page.


pbanaszkiewicz:
One feature missing: extension log (see #1937 for details).

amy/fiscal/tests/test_membership.py
```diff
@@ -23,6 +23,8 @@
 )
 from workshops.tests.base import TestBase
 
+CommentModel = django_comments.get_model()
```
lauryndbrown, line 26: What is the significance of this change? Do they not return the same model?

amy/fiscal/tests/test_membership.py
```diff
@@ -865,9 +866,41 @@ def test_membership_extended(self):
             response, reverse("membership_details", args=[membership.pk])
         )
         membership.refresh_from_db()
-        self.assertEqual(membership.extended, 30)
+        self.assertEqual(membership.extensions, [30])
```
lauryndbrown, line 869: Could you add a test where the membership has been extended more than once?

lauryndbrown (approved):

amy/fiscal/tests/test_membership.py
pbanaszkiewicz: In our case it's the same model, but there's option to replace the default model with a custom one (much like `User` from `django.contrib.auth`). `.get_model` will return the actually used model.

I decided to switch since I was using mixed approach at some point and wanted to unify.

amy/workshops/models.py
```diff
@@ -140,11 +141,12 @@ class Membership(models.Model):
         help_text="If an extension is being granted, do not manually edit the end date."
         ' Use the "Extend" button on membership details page instead.'
     )
-    extended = models.PositiveIntegerField(
-        null=True,
-        blank=True,
-        help_text="Number of days the agreement was extended with. If this value was"
-        " provided, then the agreement end has been moved.",
+    extensions = ArrayField(
```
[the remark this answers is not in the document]
pbanaszkiewicz: It was discussed with Maneesha and François, and they were okay with not storing datetime metadata related to individual extension.

amy/workshops/models.py
```diff
@@ -140,11 +141,12 @@ class Membership(models.Model):
...
+        models.PositiveIntegerField(),
+        help_text="Number of days the agreement was extended with. The field stores "
```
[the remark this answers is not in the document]
pbanaszkiewicz: Thanks! I'll fix it.

amy/fiscal/tests/test_membership.py
pbanaszkiewicz: Good idea. I'll add it.

PR #1956, opened 2021-05-18
[#1932][#1933] Link rolled-over memberships

### LINKED ISSUE #1932: Membership: indicate that membership was created by rollover
When going to a membership page that was created by rolling over an existing one, add a note at the top that says “Rolled over from <MEMBERSHIP LINK>.”  Likewise, the source membership should then say “Rolled over to <MEMBERSHIP LINK>.”

### LINKED ISSUE #1933: Membership: prevent from rolling over multiple times
User should be prevented from rolling a membership over it said membership has been rolled over in past.

pbanaszkiewicz:
This fixes #1932.
This fixes #1933.

A new one-to-one relationship was introduced for membership. It indicates that was rolled over to another membership.
The new membership also contains reversed relation indicating that it was rolled to from first membership.

The difficult part of one-to-one relationships is the "reverse" side - some tricks had to be done in order to check this value properly in some places.


amy/fiscal/tests/test_membership.py
```diff
@@ -1296,3 +1306,33 @@ def test_membership_rollovers(self):
...
+            additional_public_instructor_training_seats=0,
+            inhouse_instructor_training_seats=9,
+            additional_inhouse_instructor_training_seats=0,
+        )
+        self.membership.rolled_to_membership = second_membership
+        self.membership.save()
+
+        # Act
+        response = self.client.get(
+            reverse("membership_create_roll_over", args=[self.membership.pk])
```
lauryndbrown, line 1334: On an unrelated note: Why is create membership rollover a get rather than a post?

pbanaszkiewicz: It's a form, it supports both - GET and POST.

lauryndbrown (approved):

Issue #1941, opened 2021-05-05
Membership: link events with memberships directly


pbanaszkiewicz:
Since there can be concurrent memberships active for an organisation, there needs to be a field for linking particular event with specific membership.

Rules for selecting membership in the data migration:
1. use sponsor's current membership
2. if there's many memberships, use the one which expires sooner

Calculation of memberships workshops/seats/etc should be changed to look for events pointing directly to that membership.

Neither host or sponsor fields should take part in membership calculations.

pbanaszkiewicz:
@maneesha some events don't have start or end dates. In these cases we may not be able to link the event with particular membership active in specific period of time.

Example: a university ABC wants to run a workshop in December 2021. Regional coordinator creates event `2021-12-xx-ABC-online`, but doesn't fill out start / end dates since they're not known at this point.

A human can understand that this workshop will count towards ABC's 2021 membership, but for a machine we'd have to interpret the event's slug.

pbanaszkiewicz:
Closed through #1959.

PR #1959, opened 2021-05-23
[#1941] Link events with memberships

pbanaszkiewicz:
Special migration was introduced, which tries to find membership
for selected event sponsor, according to rules laid out in #1941.

Membership event calculations were switched to use new `membership`
field instead of `sponsor`.


lauryndbrown (approved):
As long as there are good tests around `_self_organized_workshops_queryset` This looks good to me.

PR #1960, opened 2021-05-24
ref(Consents): Filtered out already answered required consents in actions required.

lauryndbrown:
Fixes https://github.com/carpentries/amy/issues/1927

### DIFF
--- a/amy/consents/forms.py
+++ b/amy/consents/forms.py
@@ -31,10 +31,10 @@ class BaseTermConsentsForm(WidgetOverrideMixin, forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         form_tag = kwargs.pop("form_tag", True)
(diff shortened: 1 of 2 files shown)


amy/consents/forms.py
```diff
@@ -105,6 +105,19 @@ class RequiredConsentsForm(BaseTermConsentsForm):
     Builds form shown on login when there are missing required consents.
     """
 
+    def _build_form(self, person: Person) -> None:
```
pbanaszkiewicz, line 108: Is there a reason why `self.terms` cannot be passed to this method as an argument similarly to `person`?

pbanaszkiewicz (approved):

amy/consents/forms.py
lauryndbrown: There's not. I'll change that.

PR #1961, opened 2021-05-24
feat(Consents): Added new-style consents to the user profile page.

lauryndbrown:
Fixes https://github.com/carpentries/amy/issues/1925
Screen Shot 2021-05-24 at 10 10 30 AM


amy/workshops/views.py
```diff
@@ -286,7 +286,22 @@ def get_context_data(self, **kwargs):
...
         context["is_usersocialauth_in_sync"] = is_usersocialauth_in_sync
+        consents = (
+            Consent.objects.filter(person=self.object)
+            .active()
+            .select_related("term", "term_option")
+        )
 
+        context["consents"] = {
+            "May contact": consents.filter(term__slug="may-contact")[0],
+            "Consent to publish profile": consents.filter(term__slug="public-profile")[
```
lauryndbrown, line 297: Solution seems a little strange. Happy to put this in the database as an extra field instead.

pbanaszkiewicz: Do we have a way to figure out the term question, e.g. "May contact"?

On a related note, I think using a mapping between `term.slug` and a label ("May contact") should be used to reduce the number of queries.

lauryndbrown: Yes. I'm just not storing a label with the same wording is all. for instance the privacy policy is: *I have read and agree to the data privacy policy of The Carpentries.* I'll revise this. Maybe it's not as big of an issue if they don't map perfectly.

amy/workshops/views.py
```diff
@@ -286,7 +286,22 @@ def get_context_data(self, **kwargs):
...
+                0
+            ],
+            "Consent to include name when publishing lessons": consents.filter(
+                term__slug="may-publish-name"
+            )[0],
+            "Privacy policy agreement": consents.filter(term__slug="privacy-policy")[0],
```
pbanaszkiewicz, line 303: This PR looks good, except for the additional queries the `context["consents"]` introduces - can you perhaps improve to no more than 1 query?

lauryndbrown:
@pbanaszkiewicz that should be taken care of now.

lauryndbrown:
Also noticed that it was a little confusing to click into person edit and see the old consents. I removed them from there. They are viewable on the consents tab on this page:
Screen Shot 2021-05-27 at 11 34 12 AM

pbanaszkiewicz (approved):

PR #1965, opened 2021-05-29
[#1875] Pipenv instead of requirements.txt

pbanaszkiewicz:
Rationale for using [Pipenv](https://pipenv.pypa.io/en/latest/) is to secure a software chain by using pinned & hashed dependencies. Additionally it simplifies development by clearly distinguishing main/dev packages from all installable dependencies.


.github/workflows/python-test.yml
```diff
@@ -17,8 +17,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.7', '3.8', '3.9' ]
-        
+        python-version: [ '3.9' ]
```
lauryndbrown, line 20: What prompted this change?

pbanaszkiewicz: Pipenv only supports a single Python version, so I decided to move testing to the same version that's supported by pipenv.

I can move back testing to 3.7+ but long term I'd like to switch to v3.9 completely.

Pipfile
```diff
@@ -0,0 +1,55 @@
...
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+django = "~=2.2"
+argon2-cffi = "*"
+django-anymail = "*"
+django-compressor = "*"
+django-contrib-comments = "*"
```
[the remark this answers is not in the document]
pbanaszkiewicz: Dependencies are pinned in the `Pipfile.lock` (also security hashes are there). The `Pipfile` only contains upgrade instructions, so in our case we only want Django v2.2.x (LTS), and the rest can be upgraded to any version.

Upgrading will change versions in lock file, so we'll see if anyone updates them.

lauryndbrown (approved):

pbanaszkiewicz:
I'm moving this ticket to v3.2 since I didn't have time to work on pipenv changes in Ansible scripts used for deploying AMY.

pbanaszkiewicz:
I think I got the ansible scripts right, but the test deployment will tell.

PR #1978, opened 2021-06-09
[#1976] Linters configuration paths (GitHub CI)

### LINKED ISSUE #1976: bug: Precommit ordering does not match Github Action's linter
The isort is run with no flags in Github actions linter and so sorts everything in alphabetical order. Which doesn't match what we do in pre-commit.

Example:
https://github.com/carpentries/amy/pull/1975/checks?check_run_id=2784851480

Potential solutions could be:
- Github super linter (https://github.com/github/super-linter) has a PYTHON_ISORT_CONFIG_FILE env variable. Where you could add a config file to it, but that probably requires duplicating what you have in the pre-commit file.
- Alternatively the isort github action could be used https://pycqa.github.io/isort/docs/configuration/github_action/

pbanaszkiewicz:
Fixes #1976.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -23,9 +23,10 @@ jobs:
         env:
           DEFAULT_BRANCH: develop
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


lauryndbrown (approved):
Tested the commit here: https://github.com/carpentries/amy/pull/1979

PR #1985, opened 2021-06-14
feat(Consents): Remove old actions required form.

lauryndbrown:
Part of the work that fixes https://github.com/carpentries/amy/issues/1881.

Removes the unused actions required form in favor of now only using the new consents based form.


amy/api/tests/test_api_structure.py
```diff
@@ -1,6 +1,7 @@
 from django.urls import reverse
 from rest_framework.test import APITestCase
 
+from amy.workshops.tests.base import consent_to_all_required_consents
```
pbanaszkiewicz, line 4: Please fix to `workshops.tests.base` (without `amy.`).

amy/workshops/urls.py
```diff
@@ -4,86 +4,171 @@
 
 urlpatterns = [
     # utility views
-    path('log/', views.changes_log, name='changes_log'),
-    path('version/<int:version_id>/', views.object_changes, name='object_changes'),
-    path('workshop_staff/', views.workshop_staff, name='workshop_staff'),
-    path('workshop_staff/csv/', views.workshop_staff_csv, name='workshop_staff_csv'),
-
+    path("log/", views.changes_log, name="changes_log"),
```
pbanaszkiewicz, line 7: I guess the GH linter workflow didn't allow you to leave the file styling untouched? Unfortunately I didn't fix it in time to ignore `urls.py`.

But on the other hand, it's just easier to keep no exceptions from Black, and the formatted file is still quite readable.

pbanaszkiewicz (approved):
Looks good to me (except the imports).

amy/workshops/urls.py
lauryndbrown: @pbanaszkiewicz that's right. I couldn't pass the linter it without making an exception for this file.

amy/api/tests/test_api_structure.py
lauryndbrown: Oh hmm.. thought I'd fixed that in my editor. Good catch.

Issue #1987, opened 2021-06-24
Bug: exclude migrations and urls from GitHub linting


pbanaszkiewicz:
See this sample run: https://github.com/carpentries/amy/runs/2878956531

```
2021-06-21 20:25:23 [INFO]   
2021-06-21 20:25:23 [INFO]   ----------------------------------------------
2021-06-21 20:25:23 [INFO]   ----------------------------------------------
2021-06-21 20:25:23 [INFO]   Linting [PYTHON_ISORT] files...
2021-06-21 20:25:23 [INFO]   ----------------------------------------------
2021-06-21 20:25:23 [INFO]   ----------------------------------------------
2021-06-21 20:25:23 [INFO]   ---------------------------
2021-06-21 20:25:23 [INFO]   File:[/github/workspace/amy/consents/migrations/0006_auto_20210614_1234.py]
2021-06-21 20:25:23 [INFO]    - File:[0006_auto_20210614_1234.py] was linted with [isort] successfully
2021-06-21 20:25:23 [INFO]   ----------------------------------------------
2021-06-21 20:25:23 [INFO]   ----------------------------------------------
2021-06-21 20:25:23 [INFO]   The script has completed
2021-06-21 20:25:23 [INFO]   ----------------------------------------------
2021-06-21 20:25:23 [INFO]   ----------------------------------------------
2021-06-21 20:25:23 [ERROR]   ERRORS FOUND in PYTHON_BLACK:[1]
2021-06-21 20:25:24 [INFO]   ERROR! Failed to call GitHub Status API!
2021-06-21 20:25:24 [INFO]   ERROR:[]
2021-06-21 20:25:24 [ERROR]   ERRORS FOUND in PYTHON_FLAKE8:[1]
2021-06-21 20:25:24 [INFO]   ERROR! Failed to call GitHub Status API!
2021-06-21 20:25:24 [INFO]   ERROR:[]
2021-06-21 20:25:24 [INFO]   ERROR! Failed to call GitHub Status API!
2021-06-21 20:25:24 [INFO]   ERROR:[]
2021-06-21 20:25:24 [FATAL]   Exiting with errors found!
```

This can probably be fixed through setting file exclusion: https://github.com/github/super-linter#filter-linted-files

Similarly precommit configuration excludes migrations and `urls.py`.

pbanaszkiewicz:
Fixed in https://github.com/carpentries/amy/commit/67be688225ea3d0a2bf32fe0d65f36a035a2be24

PR #1994, opened 2021-07-09
Alacarte membership

maneesha:
Adds a new variant to the memberships.

This will be used when institutions purchase "a la carte" workshops rather than buying them as a membership package.


pbanaszkiewicz:
Linting errors are result of invalid configuration for the linters. There's a pending issue: #1987

pbanaszkiewicz (approved):

Issue #1996, opened 2021-07-11
Bug: W3, EU countries not recognized by django-countries


pbanaszkiewicz:
Recent update of django-countries package stopped recognizing these options:

```python
COUNTRIES_OVERRIDE = {
    "EU": _("European Union"),
    "US": _("United States"),
    "W3": _("Online"),
}
```

pbanaszkiewicz:
Fixed in #1997.

PR #1997, opened 2021-07-11
[#1996] Fix *CountriesFilter override countries

pbanaszkiewicz:
AllCountriesFilter and AllCountriesMultipleFilter use COUNTRIES_OVERRIDE
to substitute custom countries used to narrow the total list of
countries.

For example: `['W3']`
will be replaced with: `[('W3', 'Online')]`
since this value is set in the `settings.py`.


pbanaszkiewicz:
Merging since I want to release this today.

Issue #1946, opened 2021-05-06
Error in calculating remaining instructor training seats


maneesha:
**_Note: using test server here so I can show examples without messing with or exposing real data._**  

Go here 

https://test-amy.carpentries.org/fiscal/membership/3/ (Conner Ltd Bronze membership Mar 26, 2018 - Aug 23, 2019)
See they have used 7 of 8 public instructor training seats, 1 remaining.

Go to a TTT event like https://test-amy.carpentries.org/workshops/event/2020-11-03-warnermouth-ttt/ 
Assign a learner to the membership above (Conner Ltd Bronze membership Mar 26, 2018 - Aug 23, 2019)

Works as expected, https://test-amy.carpentries.org/fiscal/membership/3/ now shows 8 seats used, 0 remaining.

Delete the learner task that was just assigned.  See it goes back to showing they have used 7 of 8 public instructor training seats, 1 remaining.  This much works as expected.

Now that we have 1 remaining, go to Training Requests. Select an accepted application, and assign it to the same membership  (Conner Ltd Bronze membership Mar 26, 2018 - Aug 23, 2019). We now see the warning _Warning: Membership "Conner Ltd Bronze membership Mar 26, 2018 - Aug 23, 2019 (consortium)" is using more training seats than it's been allowed._

Going back to https://test-amy.carpentries.org/fiscal/membership/3/, it looks like this seat was allocated to the In house seats, rather than public seats as we still see 7 of 8 public seats used, and 1 of 0 in house seats used.

pbanaszkiewicz:
It seems that adding, editing tasks directly correctly affects numbers in membership linked with specific task.

There's a bug that the field doesn't show for public/in-house when assigning training request with a membership:
![obraz](https://user-images.githubusercontent.com/72821/117549989-df40ed80-b03d-11eb-88af-f3c0c4516647.png)

Another bug is that we're unable to match training request entry with a person, it should not try to reach GitHub on the test server.

I can confirm that due to the form field missing, the training request is automatically allocated as in-house.

maneesha:
@pbanaszkiewicz  Re-opening this issue as we are continuing to see this bug.  I'm  sharing some examples in slack (as they contain real data).

pbanaszkiewicz:
Fixed through #1998.

PR #1998, opened 2021-07-11
[#1946] Fix: warning msg for exceeding remaining seats

pbanaszkiewicz:
Calculation in some specific cases would result in a warning message
even though the remaining number of seats was >=1.


pbanaszkiewicz:
Merging since I want to release this today.

PR #2002, opened 2021-07-19
fix(Consents): Trainee dashboard user edit page updates consents correctly.

lauryndbrown:
The user edit page was updating the older version of consents.
Fixes https://github.com/carpentries/amy/issues/1991


pbanaszkiewicz (approved):
Looks good to me.

PR #2013, opened 2021-07-28
feat(Consents): Updated Merge person to handle new consents.

lauryndbrown:
Fixes https://github.com/carpentries/amy/issues/1881
Screen Shot 2021-07-28 at 3 29 27 PM

### DIFF
--- a/amy/workshops/util.py
+++ b/amy/workshops/util.py
@@ -23,6 +23,7 @@ from django.db import IntegrityError, models, transaction
 from django.db.models import Q
 from django.http import Http404
 from django.shortcuts import redirect, render
+from django.utils import timezone
 from django.utils.http import is_safe_url
 from django_comments.models import Comment
 import django_rq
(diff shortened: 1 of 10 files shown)


amy/workshops/util.py
```diff
@@ -994,6 +996,51 @@ def assign(request, obj, person_id):
...
+            consents_to_archive.append(consent_a)
+        elif consent_b.created_at < consent_a.created_at:
+            consents_to_archive.append(consent_b)
+        else:
+            # If they were created at the same time rather than being
+            # nondeterministic archive both and when the user logs in again
+            # they can consent to the term once more.
+            consents_to_archive.append(consent_a)
+            consents_to_archive.append(consent_b)
+            consents_to_recreate.append(
```
pbanaszkiewicz, line 1031: If both consents are archived, why only one is being recreated?

If only one object is being created, then `consents_to_recreate` doesn't have to be a list, and you don't need to use `bulk_create` down below.

pbanaszkiewicz: Okay, I think I understand why only one consent is being recreated. It will appear for the `base_obj` user, who is the result of a merge.

pbanaszkiewicz (approved):
Minor suggestion, otherwise looks good.

amy/workshops/util.py
lauryndbrown: @pbanaszkiewicz that's correct. I can add a better comment there. Basically, if I couldn't determine which consent to pick I archived both and required the new merged user to reconsent. I can also just go with the base_obj's consent instead though. Did you have feelings about that?

PR #2017, opened 2021-08-01
[#2014] Fix: redirect if only 1 group of results is singular

### LINKED ISSUE #2014: Search feature not working properly
The search feature is not working properly.  For example, searching for "florida" should return all organisations with "florida" in the name, all persons with "florida" in their email address or profile, etc.  Instead it goes to one single person's record.

pbanaszkiewicz:
Fix the bug: improper redirect to a single result if from many group
results only one group is exactly 1 element long.

This fixes #2014.


pbanaszkiewicz:
@maneesha if the workaround I proposed on Slack is not suitable for you, I can release this bugfix as v3.2.2.

lauryndbrown (approved):

PR #2020, opened 2021-08-06
[#2015] Edit membership extensions

### LINKED ISSUE #2015: Enable editing of membership date extensions.
Currently, once  a membership end date has been extended, this can not be edited. We would like to able to edit this as things can change due to changes in member contracts or user error.

pbanaszkiewicz:
Fixes #2015.

This PR introduces a way to edit extensions in Membership Edit page.

Screenshot:
obraz

### DIFF
--- a/amy/fiscal/forms.py
+++ b/amy/fiscal/forms.py
@@ -8,6 +8,7 @@ from django.dispatch import receiver
 from django.urls import reverse
 from markdownx.fields import MarkdownxFormField
 
+from fiscal.fields import FlexibleSplitArrayField
 from fiscal.models import MembershipTask
 
 # this is used instead of Django Autocomplete Light widgets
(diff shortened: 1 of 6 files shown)


amy/fiscal/forms.py
```diff
@@ -92,6 +93,14 @@ def save(self, *args, **kwargs):
 
 class MembershipForm(forms.ModelForm):
     helper = BootstrapHelper(add_cancel_button=False)
+    extensions = FlexibleSplitArrayField(
+        forms.IntegerField(),
```
lauryndbrown, line 97: I'm not as familiar with this. Should the base field be specified here?

lauryndbrown (approved):

amy/fiscal/forms.py
pbanaszkiewicz: Yes, this is composite field. It's postgres-specific, you can read more here: https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/forms/#splitarrayfield

PR #2032, opened 2021-08-09
feat: Bulk email using Mailgun.

lauryndbrown:
Fixes https://github.com/carpentries/amy/issues/1904
Send an email with Anymail's `get_merge_data` rather than having AMY send out the indivdual emails. 

Documentation: 
- https://anymail.readthedocs.io/en/stable/sending/templates/#anymail.message.AnymailMessage.merge_data
- Note that the limit on Mailgun's side is 1000 emails: https://documentation.mailgun.com/en/latest/user_manual.html#batch-sending

Scheduled job:
Screen Shot 2021-08-08 at 11 25 54 PM

Emails (note that they are marked as spam since I am using a test email client):
Screen Shot 2021-08-08 at 11 24 48 PM

Screen Shot 2021-08-08 at 11 25 09 PM


amy/autoemails/actions.py
```diff
@@ -1144,7 +1150,7 @@ class NewConsentRequiredAction(BaseAction):
     ...     job = scheduler.enqueue_in(launch_at, action)
     """
 
-    launch_at = timedelta(hours=1)
+    launch_at = timedelta(seconds=1)  # TODO CHANGE BACK
```
lauryndbrown, line 1153: I need to do this

amy/autoemails/tests/test_action.py
```diff
@@ -345,3 +345,63 @@ def testOverrideSettings(self):
         self.assertEqual(email.bcc, [])
         self.assertEqual(email.reply_to, ["reply-to@example.org"])
         self.assertEqual(email.from_email, "sender@example.org")
+
```
lauryndbrown, line 348: TODO remove test not testing anything

amy/autoemails/actions.py
```diff
@@ -127,6 +127,11 @@ def all_recipients(self) -> str:
         """If available, return string of all recipients."""
         return ""
 
+    def get_merge_data(self) -> Optional[Dict]:
+        """If available return a dict containing per user customizations
+        See https://anymail.readthedocs.io/en/stable/sending/templates/#anymail.message.AnymailMessage.merge_data"""  # noqa
```
pbanaszkiewicz, line 132: Recently I found out Flake8 didn't complain about my long line inside `"""` string. You can check if it complains here without `# noqa` comment.

amy/autoemails/actions.py
```diff
@@ -1175,3 +1181,9 @@ def check(term: Term):
 
     def get_additional_context(self, objects, *args, **kwargs):
         return dict()
+
+    def get_merge_data(self) -> Optional[Dict]:
+        # Returning an empty dictionary is necessary
+        # if you want to send an individual email to each recipient
+        # but have no per-user configuration
```
pbanaszkiewicz, line 1188: Can we have this comment available in the ancestor's method too?

amy/autoemails/actions.py
```diff
@@ -1175,3 +1181,9 @@ def check(term: Term):
...
+        return {}
```
pbanaszkiewicz, line 1189: We may need to add user details here. Did you ask @fmichonneau and others how the email message should look like? Maybe they'd like to add user name to the email body?

amy/autoemails/models.py
```diff
@@ -222,6 +223,12 @@ def build_email(
             body=body.text,
         )
         msg.attach_alternative(body.html, "text/html")
+        # https://anymail.readthedocs.io/en/stable/sending/templates/#batch-sending-with-merge-data
+        # https://anymail.readthedocs.io/en/stable/sending/templates/#anymail.message.AnymailMessage.merge_data
+        # When set the recipients in the two column each get an individual message
```
pbanaszkiewicz, line 228: I don't understand. What's `two column` in this case?

amy/autoemails/models.py
```diff
@@ -222,6 +223,12 @@ def build_email(
...
+        msg.merge_data = merge_data
+        if merge_data is not None:
+            assert len(msg.to) <= settings.BULK_EMAIL_LIMIT
```
pbanaszkiewicz, line 231: We definitely have more users than 1000. Will there be a mechanism to invoke numerous bulk emails when a new consent comes in?

amy/consents/admin.py
```diff
@@ -34,18 +35,24 @@ def send_consent_email(request, term: Term) -> None:
...
+                logger=logger,
+                scheduler=scheduler,
+                triggers=triggers,
+                context_objects={
+                    "term": term,
+                    "person_email": emails_to_send,
+                },
+                object_=term,
+            )
+            emails_to_send = []
```
pbanaszkiewicz, line 55: I think we can calculate `emails_to_send` upfront:

```python
N = 1000
emails_to_send = [emails[i:i+N] for i in range(0, len(emails), N]
for email in emails_to_send:
    jobs, rqjobs = ActionManageMixin.add(...)
```

pbanaszkiewicz: Another thing: are there tests for this function? Should they be updated?

pbanaszkiewicz (approved):
Minor comments left.

amy/autoemails/actions.py
lauryndbrown: I don't think I ever received any copy. @fmichonneau do you have a sense of what you would like the email to look like for consents? I just have a bare-bones message.

amy/autoemails/models.py
lauryndbrown: Oh sorry I meant every recipient gets an individual message.

amy/consents/admin.py
lauryndbrown: @pbanaszkiewicz no I don't think there are. Good call out. I'll add some this weekend.

amy/autoemails/actions.py
fmichonneau: Let's talk about it when we meet. I'll draft something after that. Thanks!

amy/autoemails/actions.py
lauryndbrown: @pbanaszkiewicz looks like it does complain without the noqa

PR #2036, opened 2021-08-10
[#2005] Membership Extend: use new end date instead of number of days

### LINKED ISSUE #2005: Issue membership extensions by date, rather than number of days
Right now we can issue membership extensions by X number of days.  This is useful for rough guidelines but can cause confusion when it comes to exact dates.

For example, Membership X was supposed to end 2021-07-23.  We grant a three month extension, so the new end date should be 2020-10-23.   AMY requires extensions to be entered by number of days.  Entering a 90 day extension would make the end date by 2021-10-21.  This small difference can cause confusion when we communicate with member sites.

We would like to change this interface so the user enters in a new end date from a calendar, leaving AMY to calculate the datediff.

@Talishask  Please let me know if I've captured this issue accurately and completely.

pbanaszkiewicz:
This fixes #2005.

membership_extend


lauryndbrown (approved):
Looks good

PR #2040, opened 2021-08-21
[#1657] Null event latitude and longitude

### LINKED ISSUE #1657: Fixes needed to how lat/lon is treated
Latitude and longitude are appropriately stored as floats, for example in the [Airport class](https://github.com/carpentries/amy/blob/develop/amy/workshops/models.py#L397) and the [Event class](https://github.com/carpentries/amy/blob/develop/amy/workshops/models.py#L1015).  

However, because (0, 0) is a valid set of lat/long coordinates,  testing for something like `if not lat` or `if not long` returns false negatives.  See the for example the [admin dashboard](https://github.com/carpentries/amy/blob/develop/amy/templates/dashboard/admin_dashboard.html#L78) and the [event details table](https://github.com/carpentries/amy/blob/develop/amy/templates/includes/event_details_table.html#L55).  

Once this is resolved we should also fix the [default coordinates](https://github.com/carpentries/amy/blob/develop/amy/workshops/models.py#L1322) for online events to be (0,0).

pbanaszkiewicz:
This fixes #1657 by changing logic for handling event latitude and longitude.

Apart from small fixes to how these values are displayed, there's some fixes to various Javascript scripts used with them.

### DIFF
--- a/amy/static/online_country.js
+++ b/amy/static/online_country.js
@@ -1,52 +1,35 @@
-var prev_venue,
-    prev_address,
-    prev_latitude,
-    prev_longitude,
-    online_selected_before = false;
-
-var id_country = '#id_country, #id_event-country',
-    id_venue = '#id_venue, #id_event-venue',
-    id_address = '#id_address, #id_event-address',
-    id_latitude = '#id_latitude, #id_event-latitude',
-    id_longitude = '#id_longitude, #id_event-longitude';
-
-if ($(id_country).val() == 'W3') {
-    online_selected_before = true;
-    $(id_venue).prop('disabled', true);
-    $(id_address).prop('disabled', true);
-    $(id_latitude).prop('disabled', true);
-    $(id_longitude).prop('disabled', true);
-}
-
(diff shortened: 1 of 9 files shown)


amy/static/online_country.js
```diff
@@ -1,52 +1,35 @@
...
+  const id_longitude = "#id_longitude, #id_event-longitude";
 
-if ($(id_country).val() == 'W3') {
-    online_selected_before = true;
-    $(id_venue).prop('disabled', true);
-    $(id_address).prop('disabled', true);
-    $(id_latitude).prop('disabled', true);
-    $(id_longitude).prop('disabled', true);
-}
+  if ($(id_country).val() == "W3") {
```
lauryndbrown, line 9: I assume in this case w3 is online country? would it be possible to put this in a constant you can use.

lauryndbrown (approved):
Just a small suggestion. Looks good.

amy/static/online_country.js
pbanaszkiewicz: Improved, thanks!

PR #2058, opened 2021-09-21
[#2051] Membership counting queryset adjustment

### LINKED ISSUE #2051: Counting workshops with memberships
I have seen some discrepancies in how workshops are being counted on the membership page. For example: 

On a [member view page like this](https://amy.carpentries.org/fiscal/membership/253/) we see that  they have completed 4 of 6 allowed workshops plus 1 self organized workshop.

Yet when I do `SELECT * FROM workshops_event WHERE membership_id = 253`  I get 8 workshops total (1 self organized and 7 centrally organized).  

Where is this discrepancy coming from?

pbanaszkiewicz:
This fixes #2051 by removing constraint dates on calculation - now the membership start and end dates are not considered. The link to membership is enough.

### DIFF
--- a/amy/fiscal/tests/test_membership.py
+++ b/amy/fiscal/tests/test_membership.py
@@ -172,8 +172,8 @@ class TestMembership(TestBase):
         self.setUpTasks()
         self.assertEqual(self.current.workshops_without_admin_fee_per_agreement, 10)
         self.assertEqual(self.current.workshops_without_admin_fee_total_allowed, 10)
-        self.assertEqual(self.current.workshops_without_admin_fee_completed, 6)
-        self.assertEqual(self.current.workshops_without_admin_fee_planned, 4)
+        self.assertEqual(self.current.workshops_without_admin_fee_completed, 7)
+        self.assertEqual(self.current.workshops_without_admin_fee_planned, 3)
         self.assertEqual(self.current.workshops_without_admin_fee_remaining, 0)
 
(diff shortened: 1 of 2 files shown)


amy/fiscal/tests/test_membership.py
```diff
@@ -172,8 +172,8 @@ def test_workshops_without_admin_fee(self):
         self.setUpTasks()
         self.assertEqual(self.current.workshops_without_admin_fee_per_agreement, 10)
         self.assertEqual(self.current.workshops_without_admin_fee_total_allowed, 10)
-        self.assertEqual(self.current.workshops_without_admin_fee_completed, 6)
-        self.assertEqual(self.current.workshops_without_admin_fee_planned, 4)
+        self.assertEqual(self.current.workshops_without_admin_fee_completed, 7)
+        self.assertEqual(self.current.workshops_without_admin_fee_planned, 3)
```
pbanaszkiewicz, line 176: Changes in the values are from the loosened constraints in calculation queryset.
6->7 was increased because event from outside membership date range is now counted in.
4->3 change was due to `completed` value increase (planned = available - completed).

lauryndbrown (approved):
This looks good to me. It looks like the query for `_self_organized_workshops_queryset` and `_workshops_without_admin_fee_queryset` have a lot in common and there could be a helper they both call, but that's just a nit. This looks good to me.

PR #2071, opened 2021-09-29
[#2056] Fix for negative remaining values when rolling over

### LINKED ISSUE #2056: Allow zero values for rolled over benefits
In creating a new membership by rolling over an existing one, when you try to hit submit, it takes you back to the input for "Public instructor training seats rolled from previous." 
I'm guessing this is because zero values are not allowed here.  They should be - we will often have memberships where no instructor training seats or workshops are rolled over.  Roll over values will likely be the exception.

pbanaszkiewicz:
This fixes #2056.

Previous behavior: when rolling over, max value for public/inhouse
training seats was set to the "remaining value" in respective seat
field. Even if that remaining value was negative.

This is now changed so that the max value is no lesser than 0.

### DIFF
--- a/amy/fiscal/tests/test_membership.py
+++ b/amy/fiscal/tests/test_membership.py
@@ -1498,9 +1498,8 @@ class TestMembershipCreateRollOver(TestBase):
 
     def test_membership_rollover_negative_remaining_values(self):
         """If current membership has used more seats/workshops than allowed, and
-        therefore its remaining values are negative, the roll-over form should
-        have max values for rolled-over fields set to 0, instead of these negative
-        values.
+        its remaining values are negative, the roll-over form should have max values
+        for rolled-over fields set to 0, instead of these negative values.
 
         This is a regression test for https://github.com/carpentries/amy/issues/2056.
         """
@@ -1510,7 +1509,7 @@ class TestMembershipCreateRollOver(TestBase):
             public_instructor_training_seats=0,
             inhouse_instructor_training_seats=0,
         )
-        # add two instructor training seats, so as the remainings will be -1
+        # add two instructor training seats, so the remaining seats will be -1
         event = Event.objects.create(
             slug="event-centrally-organised",
             host=self.org_beta,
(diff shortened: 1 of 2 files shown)


amy/fiscal/tests/test_membership.py
```diff
@@ -1484,6 +1496,93 @@ def test_membership_rollovers(self):
...
+
+        This is a regression test for https://github.com/carpentries/amy/issues/2056.
+        """
+        # Arrange
+
+        self.setUpMembership(
+            public_instructor_training_seats=0,
+            inhouse_instructor_training_seats=0,
+        )
+        # add two instructor training seats, so as the remainings will be -1
```
lauryndbrown, line 1513: Nit:  Change this to 
```
add two instructor training seats, so the remaining seats will be -1
```

amy/fiscal/tests/test_membership.py
```diff
@@ -1484,6 +1496,93 @@ def test_membership_rollovers(self):
...
+                    "name": "Test Membership",
+                    # doesn't matter
+                    "workshops_without_admin_fee_rolled_from_previous": 3,
+                    "public_instructor_training_seats_rolled_from_previous": value,
+                    "inhouse_instructor_training_seats_rolled_from_previous": value,
+                    **self.data,
+                }
+
+                # Act
+                response = self.client.post(
```
lauryndbrown, line 1564: Can you check the response code here?

amy/fiscal/views.py
```diff
@@ -529,12 +529,20 @@ def get_initial(self) -> Dict[str, Any]:
         }
 
     def get_form_kwargs(self) -> Dict[str, Any]:
+        # if any of the values is negative, use 0
+        max_values = {
+            "workshops_without_admin_fee_rolled_from_previous": max(
+                self.membership.workshops_without_admin_fee_remaining, 0
+            ),
+            "public_instructor_training_seats_rolled_from_previous": max(
```
lauryndbrown, line 537: Is anything checking max values in the response?

lauryndbrown (approved):

pbanaszkiewicz:
@lauryndbrown thanks! I corrected comments and added some assertions, as you suggested.

Issue #1807, opened 2021-02-19
Re-name columns named "end" 


maneesha:
Two tables in AMY use `end` as a column name. Since this is also a [PostgreSQL reserved word](https://www.postgresql.org/docs/current/sql-keywords-appendix.html) it causes problems when running SQL queries.  SQLite was more forgiving, but as PostgreSQL is stricter, this should be fixed to call `start` and `end` to `start_date` and `end_date`.  

Tables with `end` column:
- extrequests_selforganisedsubmission
- workshops_event

pbanaszkiewicz:
A migration and possibly quite some changes in templates required.

maneesha:
We had a community member create a PR long ago that has been sitting idle.  I am not sure if the PR does everything it needs to, and would be reluctant to merge it right away as it would break many of our external queries.  

It makes writing SQL queries a bit annoying as we have to do something to escape the reserved words.  I'm not sure if this could potentially cause other problems as well, so I would defer to @elichad  and @pbanaszkiewicz  on what to do next.

pbanaszkiewicz:
Note: we now have other models with these fields too, for example `CommunityRole`.

PR #2075, opened 2021-10-08
PR #1807 renamed field start and end

za:
https://github.com/carpentries/amy/issues/1807


maneesha:
Thank you for offering this @za 
Because there are many potential dependencies in renaming these fields (see notes in #1807) I would recommending closing this for now, and considering everything at once.

Issue #1527, opened 2019-09-04
Google Analytics tracking code


fmichonneau:
@pbanaszkiewicz here is the tracking code you can use to enable google analytics in amy.

```html
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-118122261-3"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'UA-118122261-3');
</script>
```

pbanaszkiewicz:
At first this seemed easy fix to me, but then I started thinking about GDPR and all these cookie-acceptance banners we see now everywhere.

I assume that as long as we [don't send Personally Identifiable Information (PII)](https://support.google.com/analytics/answer/6366371?hl=en&ref_topic=2919631) to Google Analytics, we're safe with GDPR. [IP anonymization](https://support.google.com/analytics/answer/2763052?hl=en&ref_topic=2919631) may be useful to turn on too.

We never had any cookie-accept banner, though. @fmichonneau is there a chance a legal advisor could tell us what we should do?

fmichonneau:
let's wait on this. We're in the process of having more clarity around GDPR but it might be a while before we have answers to this type of questions.

fmichonneau:
@pbanaszkiewicz you can use the following tracking code. It's GDPR compliant and no cookies are set up (it means that we can't identify repeat visitors but I'd rather not have cookie banners).

```html
<script>
  var _paq = window._paq = window._paq || [];
  _paq.push(["setDoNotTrack", true]);
  _paq.push(["disableCookies"]);
  _paq.push(['trackPageView']);
  _paq.push(['enableLinkTracking']);
  (function() {
    var u="https://carpentries.matomo.cloud/";
    _paq.push(['setTrackerUrl', u+'matomo.php']);
    _paq.push(['setSiteId', '5']);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.async=true; g.src='//cdn.matomo.cloud/carpentries.matomo.cloud/matomo.js'; s.parentNode.insertBefore(g,s);
  })();
</script>
<noscript><p><img src="https://carpentries.matomo.cloud/matomo.php?idsite=5&amp;rec=1" style="border:0;" alt="" /></p></noscript>
```

fmichonneau:
[...]

PR #2077, opened 2021-10-16
#1527 Cookie tracking policy

pbanaszkiewicz:
Footer with new paragraph with a link to new cookie policy:
obraz

Matomo Tracking code was added to base HTML template. This tracking is compliant both with GDPR and Cookie Law.


fmichonneau:
thanks, it looks good!

PR #2088, opened 2021-10-28
update AMY contributor list

maneesha:


pbanaszkiewicz (approved):
We have some duplicates here (Chris Medrela and Greg Wilson), but I guess it's okay.

maneesha:
The existing contributor list shows me  - and some other people - twice.  This is because of how people identify themselves in their profile settings.  It's a rather hidden setting.  I had to change that myself a few months ago to make myself not look like I had two identities.

PR #2122, opened 2022-01-25
Remove covid-19 and add workshops webpage

sheraaronhurt:
now that online workshops are permanent we do not need to mention covid-19.


pbanaszkiewicz:
Hey @sheraaronhurt! Similar changes should be made in file `amy/templates/mailing/workshopinquiry.html`, do you want to add them in this PR?

sheraaronhurt:
@pbanaszkiewicz  yes I would :) Do I need to put in another PR?

Issue #2109, opened 2021-12-10
Typo -- "instructors" 


maneesha:
The word "instructors" is spelled as "instrutors" in several places:

https://github.com/carpentries/amy/search?q=instrutors

PR #2134, opened 2022-02-18
Typos

maneesha:
Fixes typos as noted in #2109 (created one migration).

Note it does not edit the typos in previous migrations because I don't know what happens if you edit an old migration.


pbanaszkiewicz (approved):

pbanaszkiewicz:
> Note it does not edit the typos in previous migrations because I don't know what happens if you edit an old migration.

This is fine.

PR #2150, opened 2022-04-01
Fix docker-compose command

maneesha:


pbanaszkiewicz:
@maneesha do you have latest docker installed locally? `docker-compose` was changed to `docker compose` a few releases ago: https://docs.docker.com/compose/cli-command/

pbanaszkiewicz:
Closing as the change is not needed.

Issue #2149, opened 2022-04-01
Training Request Search


fmichonneau:
If you search for someone's name that only has a training request and not an AMY profile, the regular search interface will not return any result in the "training request" tab. It's only possible to find this request if you search through it in the Training request view (https://amy.carpentries.org/requests/training_requests/)

reported by @sheraaronhurt

maneesha:
Note this has to do with how the search is operating on full names for training requests.  Expected search results are returned if you enter in only the first or only the last name.

maneesha:
Fixed with #2151

PR #2151, opened 2022-04-01
Training search

maneesha:
Adds functionality to search trainees by personal and family name together.  This does add some slightly redundant code, but only twice.  We can consider refactoring if needed.

Addresses #2149

### DIFF
--- a/amy/dashboard/views.py
+++ b/amy/dashboard/views.py
@@ -558,17 +558,33 @@ def search(request):
             )
             results_combined += airports
 
-            training_requests = list(
-                TrainingRequest.objects.filter(
+            if len(tokens) == 2:
+                name1, name2 = tokens
+                complex_q = (
                     Q(group_name__icontains=term)
-                    | Q(family__icontains=term)
+                    | (Q(personal__icontains=name1) & Q(family__icontains=name2))
+                    | (Q(personal__icontains=name2) & Q(family__icontains=name1))
                     | Q(email__icontains=term)
+                    | Q(secondary_email__icontains=term)
                     | Q(github__icontains=term)
                     | Q(affiliation__icontains=term)
                     | Q(location__icontains=term)
                     | Q(user_notes__icontains=term)
                 )
-            )
+                training_requests = list(TrainingRequest.objects.filter(complex_q))
+
+            else:
+                training_requests = list(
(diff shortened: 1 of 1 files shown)


maneesha:
As I think about this more, I'm not entirely pleased with this approach as if someone has a space in their personal or family names, this will not work. For example, searching for *Mary Ann* will not return the person named *Mary Ann Smith*.

pbanaszkiewicz:
@maneesha I'm okay with your changes.

@maneesha @fmichonneau I also conducted some testing of PostgreSQL's Full Text Search.
Here's how the code would look like:
```python
            persons = list(
                Person.objects.annotate(
                    name_search=SearchVector('personal', 'middle', 'family')
                ).filter(
                    Q(name_search=term)
                    | Q(email__icontains=term)
                    | Q(secondary_email__icontains=term)
                    | Q(github__icontains=term)
                ).order_by("family")
            )
```
Diff:
obraz
In my opinion, it's much much cleaner.

This code also solves for the case you mentioned (`Mary Ann Smith`). I created two accounts for that name (1: personal="Mary Ann", family="Smith", 2: personal="Mary", middle="Ann", family="Smith") and they both appear in the results:
obraz

There's a caveat though. In some cases the results would not contain the same set of people as previously. Here's search results for term `peter`:
obraz
Here's database persons with highlighted term `peter`:
obraz
Clearly `Shawn Peterson` wasn't returned by the Full Text Search.

@maneesha @fmichonneau Please let me know if you would like to switch to Full Text Search.

pbanaszkiewicz (approved):

pbanaszkiewicz:
@maneesha I don't know why `Shawn Peterson` is not returned. My guess is that PostgreSQL engine doesn't "think" `Peter` is close enough to `Peterson`.

maneesha:
It seems illogical to me that  `Shawn Peterson` is not returned  and `epeterson@gmail.com`  is returned.  Looking at the code, the search is including the `github` field.  Was `peter` in the github name for `Shawn Peterson`?

If that's not the case, then I don't like this approach as the logic doesn't make sense.

pbanaszkiewicz:
On a call today we agreed with @maneesha to not use full-text search for this kind of searching.

Issue #2147, opened 2022-03-24
Default dates for roll over memberships


maneesha:
When a new membership is created by rolling over an existing membership, by default the new membership should start the day after the previous one and last for one calendar year.

For example, membership start 2021-03-01 ends on 2022-02-28.  New membership should start 2022-03-01 and end on 2023-02-28.  
This avoids overlapping dates for start/end of consecutive memberships.

maneesha:
@pbanaszkiewicz  @elichad  is there any other information needed to implement this feature?

elichad:
This has been fixed by #2152 .

PR #2152, opened 2022-04-01
make new membership start day after previous ends

maneesha:
Addresses #2147

### DIFF
--- a/amy/fiscal/forms.py
+++ b/amy/fiscal/forms.py
@@ -236,9 +236,6 @@ class MembershipCreateForm(MembershipForm):
             "membership."
         ).format(reverse("organization_add"))
 
-        # set up a layout object for the helper
-        self.helper.layout = self.helper.build_default_layout(self)
-
         # add warning alert for dates falling within next 2-3 months
         STANDARD_PACKAGES_PREPOPULATION_WARNING = (
             "<b>Workshops without admin fee per agreement</b> and "
--- a/amy/fiscal/views.py
+++ b/amy/fiscal/views.py
@@ -509,7 +509,7 @@ class MembershipCreateRollOver(
             "consortium": self.membership.consortium,
             "public_status": self.membership.public_status,
             "variant": self.membership.variant,
-            "agreement_start": self.membership.agreement_end,
+            "agreement_start": self.membership.agreement_end + timedelta(days=1),
             "agreement_end": date(
                 self.membership.agreement_end.year + 1,
                 self.membership.agreement_end.month,
--- a/amy/static/amy_utils.js
+++ b/amy/static/amy_utils.js
@@ -315,10 +315,16 @@ $(document).ready(function() {
   const duration_warning = function(start_element, end_element, warning_element) {
     const next_year = start_element.datepicker("getDate");
     const end_date = end_element.datepicker("getDate");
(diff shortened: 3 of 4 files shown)


amy/fiscal/views.py
```diff
@@ -509,7 +509,7 @@ def get_initial(self) -> Dict[str, Any]:
             "consortium": self.membership.consortium,
             "public_status": self.membership.public_status,
             "variant": self.membership.variant,
-            "agreement_start": self.membership.agreement_end,
+            "agreement_start": self.membership.agreement_end + timedelta(days=1),
```
pbanaszkiewicz, line 512: Wouldn't it be better if the agreement ended one day before full year from the start?

For example, agreement starts on April 1st 2022, and ends on March 31th, 2023. The new agreement will start also on April 1st 2023.

pbanaszkiewicz: @pbanaszkiewicz to confirm the end date for membership

amy/fiscal/forms.py
```diff
@@ -236,9 +236,6 @@ def __init__(self, *args, **kwargs):
             "membership."
         ).format(reverse("organization_add"))
 
-        # set up a layout object for the helper
-        self.helper.layout = self.helper.build_default_layout(self)
-
```
pbanaszkiewicz, line 241: @maneesha this one fixes an issue with agreement dates warning not showing up when creating new membership.

amy/static/calendar_popup.js
```diff
@@ -39,6 +39,7 @@ $(document).ready(function() {
         if (e.date && !$('#id_agreement_end').val()) {
             d = e.date;
             d.setFullYear(d.getFullYear() + 1);
+            d.setDate(d.getDate() - 1);
```
pbanaszkiewicz, line 42: @maneesha this one changes date when new membership is created.

When user selects e.g. `2022-05-15` for the agreement start date, then `2023-05-14` will be selected for agreement end date. (Previously it would have been `2023-05-15`.)

amy/static/amy_utils.js
```diff
@@ -315,10 +315,16 @@ $(document).ready(function() {
...
-      warning_element.removeClass("d-none");
+    if (next_year && end_date) {
+      next_year.setFullYear(next_year.getFullYear() + 1);
+      next_year.setDate(next_year.getDate() - 1);
+
+      if (next_year.getTime() != end_date.getTime()) {
+        warning_element.removeClass("d-none");
+      } else {
+        warning_element.addClass("d-none");
+      }
```
pbanaszkiewicz, line 327: Couple of small fixes to accommodate new agreement end date calculation and some corner case of missing value in a date field.

pbanaszkiewicz:
@maneesha I added small change for when user is creating a new membership. Now the agreement end date will be 1 day lower than previously calculated.

For example, when user selects 2022-05-15 as agreement start date, then new agreement end date will be 2023-05-14 (+1 year, -1 day). Previously it would have been +1 year (2023-05-15).

Apart from that I fixed some small issues, and one pretty big one: agreement duration warning was not being shown.

Please take a look and decide if you like the changes. I'm okay with your changes, btw.

pbanaszkiewicz (approved):

PR #2330, opened 2023-02-06
Seeding scripts

### LINKED ISSUE #2239: Create script to apply only new data from fixtures
When deploying it's not really possible to use `loaddata` and only add new entries.

Create a script that will add new entries from selected fixtures; it should accept a table column by which existence of an entry should be determined.

pbanaszkiewicz:
This fixes #2239 by providing an alternative to fixtures from Django.

Seeding scripts give more control to the developer. This way it was possible to implement a check before given entity gets created.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203911669015361


amy/workshops/utils/seeding.py
```diff
@@ -0,0 +1,59 @@
...
+    def _info(msg: str) -> None:
+        if logger:
+            logger.info(msg)
+
+    class_name = model_class.__name__
+
+    _info(f"Start of {class_name} deprecation.")
+
+    for i, model_id in enumerate(model_id_list):
+        if not model_class._default_manager.filter({lookup_field: model_id}).exists():
```
elichad, line 52:
```suggestion
        if not model_class._default_manager.filter(**{lookup_field: model_id}).exists():
```

amy/workshops/utils/seeding.py
```diff
@@ -0,0 +1,59 @@
...
+            _info(f"{i} {class_name} <{model_id}> doesn't exist, skipping.")
+            continue
+
+        _info(f"{i} {class_name} <{model_id}> exists, removing.")
+        model_class._default_manager.filter({lookup_field: model_id}).delete()
```
elichad, line 57:
```suggestion
        model_class._default_manager.filter(**{lookup_field: model_id}).delete()
```

elichad (changes requested):
The `deprecate_models` method doesn't work for me if I add names to `DEPRECATED_X` variables - suggestions should fix it. Looks good otherwise.

Does the seeding get run automatically during the deployment process?

amy/workshops/utils/seeding.py
pbanaszkiewicz: Good catch! Thanks, fixing now.

elichad (approved):

PR #2333, opened 2023-02-07
Run `create_superuser` to create super user on application's start

### LINKED ISSUE #2310: Prepare AMY staging instance for actual use
New test AMY server is running, but it lacks some features from the other server.

- [x] Run fixtures (should be accompanied by #2239)
- [x] Scaffold non-admin users for AMY database
- [ ] Add default admin user

pbanaszkiewicz:
This fixes #2310.

### DIFF
--- a/amy/workshops/management/commands/create_superuser.py
+++ b/amy/workshops/management/commands/create_superuser.py
@@ -1,5 +1,4 @@
 from django.core.management.base import BaseCommand, CommandError
-from django.utils.crypto import get_random_string
 
 from communityroles.models import CommunityRole, CommunityRoleConfig
 from workshops.models import Person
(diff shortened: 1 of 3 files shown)


amy/workshops/management/commands/create_superuser.py
```diff
@@ -8,16 +9,35 @@ class Command(BaseCommand):
...
+                username=username,
                 personal="admin",
                 family="admin",
-                email="admin@example.org",
-                password="admin",
+                email=email,
+                password=password,
             )
-            print("Created admin user")
+            print(f"Created admin user with password {password}")
```
pbanaszkiewicz, line 40: I'm worried this may get censored in the AWS CloudWatch.

pbanaszkiewicz:
The idea here is to run `create_superuser` command when the container starts, so that admin user is created and we get to know their password.

I can see the CodeQL checks are complaining, so maybe I'll drop the password printing and accept the fact that new installation will have `admin:admin` superuser.

pbanaszkiewicz:
Added tests to make sure the command behaves like I want.

It will create the superuser account only if one doesn't exist. The existing superuser account will not be changed.

When moving this code to production, we can easily create account for user `admin` and disable it; this way the script doesn't do any harm.

pbanaszkiewicz:
Conflict resolved. @elichad Can you review this PR?

elichad:
> When moving this code to production, we can easily create account for user admin and disable it; this way the script doesn't do any harm.

Is it worth creating that user now so we don't forget about it later?

elichad:
Also, I haven't looked through the new CI/CD setup yet - is the database recreated from scratch every time the test server is re-deployed, or does it persist?

pbanaszkiewicz:
@elichad 

> Is it worth creating that user now so we don't forget about it later?

Good idea. I have just created this account on production server and disabled its permissions and access.
CC @maneesha @froggleston 

> Also, I haven't looked through the new CI/CD setup yet - is the database recreated from scratch every time the test server is re-deployed, or does it persist?

No, database persists. It is treated as a completely separate service from the application.

Issue #2328, opened 2023-02-01
Review of AMY docs and corresponding UI operations


elichad:
Dumping all my notes from a thorough read of the AMY docs and corresponding operations on the test server. Many of them are small things and I'll make a PR to fix as many of those as possible. Some probably deserve their own issues.

## Questions:
- [ ] what criteria make an instructor or other Person 'active' vs 'inactive?' Am I right in thinking that people are made inactive manually?
- [ ] under what criteria would you archive a Person?
- [ ] Training requests: Is there any particular reason why we allow matching trainees to multiple TTT events? I expected it'd throw an error but it didn't.

## Outdated things in docs:
- [ ] Organization: Affiliated organizations not used - hide/remove?
- [ ] Membership: Person roles (Trainer missing from docs)
- [ ] Event: URL validation error suggests pointing to a github repo rather than a github.io site
![url_validation_out_of_date](https://user-images.githubusercontent.com/20683271/216090680-22a65e33-0e3e-4c3a-9621-055210b96d7f.png)
- [ ] Add Person to Event: Title and URL are not used - hide/remove?
- [ ] Docs: various references to separate DC/LC/SWC badges/demos/instructors etc
- [ ] Find workshop staff: Badge filter has been changed to a couple of checkboxes

## Miscellaneous issues:
- [ ] Docs: anchor links don't have a proper icon, just a hard-to-see box 
![image](https://user-images.githubusercontent.com/20683271/216082069-1204657f-e159-4e16-9da8-b368e96765bf.png)
- [ ] dropdowns start to 404 if the page is kept open for a long time (5-10 min), displaying "The results could not be loaded." I assume this is a form timeout and possibly deliberate? But it's not that clear that that's what's happening.
- [ ] Documentation missing for some of the Person tabs - Consents and Community Roles
- [ ] New Event: certain tags must be added to match the selected curriculum, but the warning if you don't doesn't tell you which tags are needed. (Also, I wonder if we need those tags to be user-selectable at all or if we can add them automatically?)
- [ ] Docs: "**URL** Enter in the workshop's url (to the GitHub page, not the repo)" could be clearer as often "GitHub page"=repo
- [ ] New Event: Sometimes an extra blank entry appears when entering the first email into 'additional people to contact.' I assume this isn't causing problems though.
![image](https://user-images.githubusercontent.com/20683271/216087338-302ead36-9e03-4a89-8400-dfe9c512e626.png)
[...]

PR #2337, opened 2023-02-09
Improvements to AMY docs and UI

### LINKED ISSUE #2331: Different URLs on /workshops/events vs /workshops/event/<slug>
Minor, but a bit odd:
The list of all events at `/workshops/events` displays the repository URL, while the individual event page at `/workshops/event/<slug>` displays the only website URL.
Most obvious in real AMY where both URLs are usually set.

elichad:
Fixes some items in #2328, fixes #2331


mkdocs.yml
```diff
@@ -44,7 +44,7 @@ nav:
 
 
 markdown_extensions:
-    - toc:
-        permalink: 
+  - toc:
+      permalink: True
```
pbanaszkiewicz, line 48: This was set to a particular character, ¶ probably (or similar indicator of a permalink). How does the permalink look now?

amy/workshops/forms.py
```diff
@@ -564,22 +564,24 @@ def clean_curricula(self):
...
-            expected_tags = []
+            expected_tags = set()
 
+        missing_tags = set()
         for tag in expected_tags:
             if not tags.filter(name=tag):
-                raise forms.ValidationError(
-                    "You must add tags corresponding to these curricula."
-                )
+                missing_tags.add(tag)
```
pbanaszkiewicz, line 580: Alternatively:
```
missing_tags = expected_tags - set(tags.values_list('name', flat=True))
```

amy/workshops/forms.py
```diff
@@ -564,22 +564,24 @@ def clean_curricula(self):
...
+        if missing_tags:
+            raise forms.ValidationError(
+                f"You must add tags corresponding to these curricula. Missing tags: {', '.join(missing_tags)}"
+            )
```
pbanaszkiewicz, line 585: So this method `clean_curricula` validates tags based on selected curricula. I think it should be `clean_tags`, so that at least the error shows up in proper place.

Also it could be beneficial for user experience if the fields in the form switched places.

amy/workshops/forms.py
```diff
@@ -580,7 +580,8 @@ def clean_curricula(self):
                 missing_tags.add(tag)
         if missing_tags:
             raise forms.ValidationError(
-                f"You must add tags corresponding to these curricula. Missing tags: {', '.join(missing_tags)}"
+                f"""You must add tags corresponding to these curricula. """
+                f"""Missing tags: {', '.join(missing_tags)}"""
```
pbanaszkiewicz, line 584: In lines 583 and 584 you don't need to use `"""`. You can use `"` just fine.

Additionally, in L583 you defined an f-string, which is not used and will highlight in some linters. Drop `f` and you should be fine.

pbanaszkiewicz (approved):
Very minor styling could be improved, but overall it looks good.

mkdocs.yml
pbanaszkiewicz: obraz
Current permalink.

Looks good.

elichad: It was meant to be this icon https://fontawesome.com/icons/link?s=solid&f=classic, I believe, but the Material theme doesn't support that icon in the same way as the old theme presumably did (I had no luck trying to force it). `True` just sets it to the theme default.

amy/workshops/forms.py
elichad: I agree - though updating this was less straightforward than I thought. In short, `curricula` wasn't in `cleaned_data` at the time `clean_tags` was run during validation, so I've changed this method to be called during `clean()` - following these Django docs https://docs.djangoproject.com/en/4.1/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other. Can you do a quick re-review?

pbanaszkiewicz: @elichad Looks okay except for the way validation error is handled.

We have this pattern:

```python
errors = {}

if validation_failed_for_a_field:
    errors["field"] = ValidationError("Text")

# ...

if errors:
    raise ValidationError(errors)
```

For example: `CommunityRoleForm.clean`, `amy/communityroles/forms.py`.

elichad: Ah, thanks for the reference, done

amy/workshops/forms.py
```diff
@@ -602,6 +599,26 @@ def save(self, *args, **kwargs):
 
         return res
 
+    def clean(self):
+        cleaned_data = super().clean()
+        if not cleaned_data:
+            return cleaned_data
+
+        errors: defaultdict[str, list[ValidationError]] = defaultdict(list)
```
pbanaszkiewicz, line 607: Nice typing!

pbanaszkiewicz (approved):
Looks good to me.

PR #2339, opened 2023-02-10
hotfix: check for null values when loading custom keys

### LINKED ISSUE #2336: Server Error when editing community roles
Reported by @ErinBecker at `/workshops/person/<id>/edit/#communityroles` and `/communityroles/role/<another id>/edit/` in production.

Reproduced on test AMY:
```
TypeError at /communityroles/role/25/edit/
the JSON object must be str, bytes or bytearray, not list
Error during template rendering
In template /webapps/test-amy.carpentries.org/repo/amy/templates/bootstrap4/field.html, error at line 39

{% crispy_field field %}
```

Also encountered locally when creating a community role and immediately editing it:
```
Exception Value: | 'NoneType' object is not iterable
/home/eli/amy/amy/communityroles/fields.py, line 20, in get_context

```

elichad:
Fixes #2336 .

I wrote some tests, but do we actually use custom keys at all?

### DIFF
--- a/amy/communityroles/fields.py
+++ b/amy/communityroles/fields.py
@@ -17,7 +17,10 @@ class CustomKeysWidget(forms.TextInput):
 
     def get_context(self, name: str, value: str, attrs: dict):
         value_deserialized = json.loads(value)
-        value_deserialized_dict = dict(value_deserialized)
+        if value_deserialized is not None:
+            value_deserialized_dict = dict(value_deserialized)
+        else:
+            value_deserialized_dict = {}
         default_values = dict([(label, "") for label in self.labels])
         context_value = default_values | value_deserialized_dict
 
--- a/amy/communityroles/fields.py
+++ b/amy/communityroles/fields.py
@@ -1,8 +1,11 @@
 import json
+import logging
 
 from django import forms
 from django.http import QueryDict
 from django.utils.datastructures import MultiValueDict
+
+logger = logging.getLogger("amy")
 
 
 class CustomKeysWidget(forms.TextInput):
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
Custom keys have been introduced quite recently. I think the original idea was to customize some community roles with them, e.g. when a role was related to a specific lesson we could add "lesson url" custom key and fill it out when assigning the role.

amy/communityroles/fields.py
```diff
@@ -17,7 +17,10 @@ def apply_labels(self, labels: list[str]) -> None:
 
     def get_context(self, name: str, value: str, attrs: dict):
         value_deserialized = json.loads(value)
-        value_deserialized_dict = dict(value_deserialized)
+        if value_deserialized is not None:
+            value_deserialized_dict = dict(value_deserialized)
+        else:
+            value_deserialized_dict = {}
```
pbanaszkiewicz, line 23: `json.loads` can potentially return other types that will fail to dict transformation, for example:

```python
>>> S = '["asd", "dsa", "zxc"]'
>>> json.loads(S)
['asd', 'dsa', 'zxc']
>>> dict(json.loads(S))
ValueError: dictionary update sequence element #0 has length 3; 2 is required
```

I'm wondering if `try-except` wouldn't be better here. You can add logging of the failed value, too, so in any case we could diagnose faster.

amy/communityroles/tests/test_fields.py
```diff
@@ -0,0 +1,37 @@
...
+        self.assertListEqual(subwidget_labels, custom_keys)
+        self.assertListEqual(subwidget_values, list(values.values()))
+
+    def test_custom_keys_none(self):
+        self.widget.apply_labels([])
+        context = self.widget.get_context(
+            name="communityroles-custom_keys", value="null", attrs={}
+        )
+
+        self.assertEqual(context["widget"]["subwidgets"], [])
```
pbanaszkiewicz, line 37: Can you follow Arrange-Act-Assert pattern? https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/

I usually separate each section within a test with comments like so:

```python
def test():
  # Arrange
  widget = Widget()

  # Act
  widget.click()

  # Assert
  assert widget.focused is True
```

pbanaszkiewicz (changes requested):

elichad:
@pbanaszkiewicz this one is the hotfix that I mentioned yesterday

pbanaszkiewicz (approved):

pbanaszkiewicz:
Looks good to me!

PR #2346, opened 2023-02-17
Fix internal errors from filters

### LINKED ISSUE #2327: Bug: Internal Server Error: /reports/membership_trainings_stats/
Production issue. Same type as #2313 but on the Membership training statistics page.

When searching for an organization:
```
FieldError at /reports/membership_trainings_stats/
Cannot resolve keyword 'organization' into field. Choices are: additional_inhouse_instructor_training_seats, additional_public_instructor_training_seats, agreement_end, agreement_link, agreement_start, communityrole, consortium, contribution_type, emergency_contact, event, extensions, id, inhouse_instructor_training_seats, inhouse_instructor_training_seats_rolled_from_previous, inhouse_instructor_training_seats_rolled_over, instructor_training_seats_inhouse_remaining, instructor_training_seats_inhouse_total, instructor_training_seats_inhouse_utilized, instructor_training_seats_public_remaining, instructor_training_seats_public_total, instructor_training_seats_public_utilized, member, membershiptask, name, organizations, persons, public_instructor_training_seats, public_instructor_training_seats_rolled_from_previous, public_instructor_training_seats_rolled_over, public_status, registration_code, rolled_from_membership, rolled_to_membership, rolled_to_membership_id, task, variant, workshops_without_admin_fee_per_agreement, workshops_without_admin_fee_rolled_from_previous, workshops_without_admin_fee_rolled_over
```
Other issues appear when ordering or filtering with the checkboxes:
```
Cannot resolve keyword 'instructor_training_seats_total' into field
Cannot resolve keyword 'instructor_training_seats_remaining' into field
Cannot resolve keyword 'instructor_training_seats_utilized' into field.
```

elichad:
Fixes #2327 and introduces tests for membership filters.
Includes combining some columns in the membership training stats view as requested in that issue.

### DIFF
--- a/amy/fiscal/filters.py
+++ b/amy/fiscal/filters.py
@@ -48,7 +48,7 @@ def filter_training_seats_only(queryset, name, seats):
         return queryset
 
 
-def filter_nonpositive_remaining_seats(queryset, name, seats):
+def filter_negative_remaining_seats(queryset, name, seats):
     """Limit Memberships to only entries with negative remaining seats."""
     if seats:
         return queryset.filter(instructor_training_seats_remaining__lt=0)
@@ -81,9 +81,9 @@ class MembershipFilter(AMYFilterSet):
         widget=widgets.CheckboxInput,
     )
 
-    nonpositive_remaining_seats_only = django_filters.BooleanFilter(
+    negative_remaining_seats_only = django_filters.BooleanFilter(
         label="Only show memberships with less than zero remaining seats",
-        method=filter_nonpositive_remaining_seats,
+        method=filter_negative_remaining_seats,
         widget=widgets.CheckboxInput,
     )
 
@@ -125,9 +125,9 @@ class MembershipTrainingsFilter(AMYFilterSet):
         widget=widgets.CheckboxInput,
     )
 
-    nonpositive_remaining_seats_only = django_filters.BooleanFilter(
+    negative_remaining_seats_only = django_filters.BooleanFilter(
         label="Only show memberships with less than zero remaining seats",
-        method=filter_nonpositive_remaining_seats,
+        method=filter_negative_remaining_seats,
         widget=widgets.CheckboxInput,
     )
 
--- a/amy/fiscal/filters.py
+++ b/amy/fiscal/filters.py
@@ -65,7 +65,7 @@ def filter_nonpositive_remaining_seats(queryset, name, seats):
 
 class MembershipFilter(AMYFilterSet):
     organization_name = django_filters.CharFilter(
-        label="Organisation name",
+        label="Organization name",
         field_name="organizations__fullname",
         lookup_expr="icontains",
     )
--- a/amy/fiscal/filters.py
+++ b/amy/fiscal/filters.py
@@ -1,7 +1,6 @@
 from datetime import date
 
 from django.forms import widgets
-from django.db.models import Q
 import django_filters
 
 from workshops.fields import Select2MultipleWidget, Select2Widget
@@ -121,13 +120,13 @@ class MembershipTrainingsFilter(AMYFilterSet):
     )
 
     training_seats_only = django_filters.BooleanFilter(
-        label="Only show memberships with more than zero allowed training seats (public or in-house)",
+        label="Only show memberships with more than zero allowed training seats",
         method=filter_training_seats_only,
         widget=widgets.CheckboxInput,
     )
 
     nonpositive_remaining_seats_only = django_filters.BooleanFilter(
-        label="Only show memberships with less than zero remaining seats (public or in-house)",
+        label="Only show memberships with less than zero remaining seats",
         method=filter_nonpositive_remaining_seats,
         widget=widgets.CheckboxInput,
     )
@@ -137,12 +136,9 @@ class MembershipTrainingsFilter(AMYFilterSet):
             "name",
             "agreement_start",
             "agreement_end",
-            "instructor_training_seats_public_total",
-            "instructor_training_seats_public_utilized",
-            "instructor_training_seats_public_remaining",
-            "instructor_training_seats_inhouse_total",
-            "instructor_training_seats_inhouse_utilized",
-            "instructor_training_seats_inhouse_remaining",
+            "instructor_training_seats_total",
+            "instructor_training_seats_utilized",
+            "instructor_training_seats_remaining",
         ),
     )
 
--- a/amy/fiscal/filters.py
+++ b/amy/fiscal/filters.py
@@ -1,6 +1,7 @@
 from datetime import date
 
 from django.forms import widgets
+from django.db.models import Q
 import django_filters
 
 from workshops.fields import Select2MultipleWidget, Select2Widget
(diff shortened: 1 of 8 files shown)


amy/workshops/filters.py
```diff
@@ -139,13 +139,13 @@ def filter(self, qs, value):
...
             return ordering.order_by("family", "middle", "personal")
-        elif any(v in ["-lastname"] for v in value):
+        elif any(v in ["-family"] for v in value):
             return ordering.order_by("-family", "-middle", "-personal")
-        elif any(v in ["firstname"] for v in value):
+        elif any(v in ["personal"] for v in value):
             return ordering.order_by("personal", "middle", "family")
-        elif any(v in ["-firstname"] for v in value):
+        elif any(v in ["-personal"] for v in value):
             return ordering.order_by("-personal", "-middle", "-family")
```
pbanaszkiewicz, line 149: I also fixed this filter in here: https://github.com/carpentries/amy/pull/2335/files#diff-76c8f195f620772cf38ce87a84d610ada62278260c666e317c8387d8f274adf7

I would prefer my approach unless you find something wrong with it. I think the extra choices were intentionally set to `"lastname"` and `"firstname"`, as we don't have these fields in any model, and `NamesOrderingFilter` allows to filter by additional fields if provided.

amy/fiscal/tests/test_filters.py
```diff
@@ -0,0 +1,454 @@
...
+from django.test import TestCase
+
+from fiscal.filters import MembershipFilter, MembershipTrainingsFilter
+from fiscal.models import Membership
+from workshops.models import Event, Member, MemberRole, Organization, Person, Role, Task
+
+
+class TestMembershipFilter(TestCase):
+    @classmethod
+    def setUpClass(cls) -> None:
```
pbanaszkiewicz, line 14: Is there a particular reason why you chose `setUpClass` and `tearDownClass` class methods?

[The docs](https://docs.python.org/3/library/unittest.html#class-and-module-fixtures) say it should be used with caution:

> Note that shared fixtures do not play well with [potential] features like test parallelization and they break test isolation. They should be used with care.

In all other cases we used `setUp()` and `tearDown` methods.

pbanaszkiewicz: I am a little afraid about parallelization of these tests (I use it often with `make fast_test` or `python manage.py test --parallel` commands).

amy/fiscal/tests/test_filters.py
```diff
@@ -0,0 +1,454 @@
...
+            variant="Bronze",
+            agreement_start=date.today(),
+            agreement_end=date.today(),
+            contribution_type="Financial",
+            public_instructor_training_seats=1,
+        )
+        cls.member = Member.objects.create(
+            membership=cls.membership, organization=cls.organization, role=member_role
+        )
+        # create a test membership that sphould be filtered out
```
pbanaszkiewicz, line 33: Small typo: `sphould` -> `should`.

amy/fiscal/tests/test_filters.py
```diff
@@ -0,0 +1,454 @@
...
+        # Arrange
+        name = "active_only"
+        value = True
+
+        # Act
+        result = self.filterset.filters[name].filter(self.qs, value)
+
+        # Assert
+        self.assertTrue(self.membership in result)
+        self.assertFalse(self.membership2 in result)
```
pbanaszkiewicz, line 157: There's also `assertIn` and `assertNotIn`, they may provide better insight if the condition fails.

amy/fiscal/tests/test_filters.py
```diff
@@ -0,0 +1,454 @@
...
+                + F("inhouse_instructor_training_seats")
+                + F("additional_inhouse_instructor_training_seats")
+                + Coalesce("inhouse_instructor_training_seats_rolled_from_previous", 0)
+                - Count(
+                    "task",
+                    filter=Q(task__role__name="learner", task__seat_public=False),
+                )
+                - Coalesce("inhouse_instructor_training_seats_rolled_over", 0)
+            ),
+        )
```
pbanaszkiewicz, line 80: Since it looks like we have this same query in multiple places (twice in this file, one in `fiscal.views.AllMemberships.queryset` and one in `reports.views.membership_trainings_stats`) then I propose to extract it out, for example into a utility function, and use it in all 4 places instead. Hopefully this would lead to less changes in future and less errors in the calculations since they are quite difficult.

amy/fiscal/tests/test_filters.py
```diff
@@ -0,0 +1,454 @@
...
+
+        # Act
+        for field in fields.keys():
+            results[field] = self.filterset.filters[filter_name].filter(
+                self.qs, [field]
+            )
+
+        # Assert
+        # we don't have any unexpected fields
+        self.assertListEqual(list(fields.keys()), list(expected_results.keys()))
```
pbanaszkiewicz, line 238: Does `assertEqual(fields.keys(), expected_results.keys())` also work? `.keys()` returns a set-like object, so we should be able to compare them.

amy/fiscal/tests/test_filters.py
```diff
@@ -0,0 +1,454 @@
...
+        # each field was filtered correctly
+        for field in fields.keys():
+            self.assertQuerysetEqual(results[field], expected_results[field])
+
+
+class TestMembershipTrainingsFilter(TestCase):
+    """
+    Duplicate of TestMembershipFilter with some fields removed.
```
pbanaszkiewicz, line 246: What fields were removed?

pbanaszkiewicz (changes requested):
In general this looks good and definitely as a new feature rather than a bugfix.

There are some questions left in the review:
1. What spelling we should use, @maneesha?
2. Which fix for the name ordering filter we should choose?
3. What is the impact of `setUpClass` / `tearDownClass` on parallelization of test runs?

Because of these questions I'm going to leave the review as "Request changes" and not push it into milestone https://github.com/carpentries/amy/milestone/95.

amy/fiscal/filters.py
```diff
@@ -58,7 +58,7 @@ def filter_nonpositive_remaining_seats(queryset, name, seats):
 
 class MembershipFilter(AMYFilterSet):
     organization_name = django_filters.CharFilter(
-        label="Organisation name",
+        label="Organization name",
```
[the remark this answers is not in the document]
elichad: We use 'Organization' everywhere else in the UI. I don't know if British English spelling was agreed on or not, but in this case the 'z' is acceptable in British English anyway.
(I've just been down the rabbit hole and it turns out [the Oxford English Dictionary even prefers the -ize/-ization endings...](https://en.wikipedia.org/wiki/Oxford_spelling#Language_tag_comparison))

amy/workshops/filters.py
elichad: Your approach looks good, I'll revert these changes

amy/fiscal/tests/test_filters.py
elichad: Habit from a past project, and a desire to keep the number of database calls down. Happy to change to `setUp/tearDown`. (and actually learned about the helper setup methods in TestBase since I wrote this, so will look at using those too)

amy/fiscal/tests/test_filters.py
elichad: It does, I think I iterated this a few times and then didn't think about simplifying 😄

amy/fiscal/tests/test_filters.py
elichad: Good idea. I moved this to a new Manager for the Membership object - could do with another review

pbanaszkiewicz (approved):

PR #2357, opened 2023-03-04
Colors for different builds

### LINKED ISSUE #2322: Different colour scheme for test-amy / local builds
It'd be useful to have a visual indicator of which site you're on - it's currently quite easy to mix up production AMY and staging/local AMY pages if you have both open.

I think the nav background colour would be the easiest and most visually obvious thing to change.

pbanaszkiewicz:
This fixes #2322.

There will be 2 banners available: local and testing.

Local looks like this:
obraz

Testing looks like this:
obraz

Every site will have this banner. It will stay on top, but will hide beneath the navigation bar:
banner_hides_beneath

### DIFF
--- a/amy/static/css/amy.css
+++ b/amy/static/css/amy.css
@@ -22,6 +22,27 @@ footer {
     font-size: 0.75rem;
 }
 
+.banner {
+    display: block;
+    text-align: center;
+    font-size: 1rem;
+    padding: 0.5em;
+}
+
+.banner-local {
+    color: #222;
+    background-color: #000000;
+    opacity: 0.8;
+    background: repeating-linear-gradient(-45deg, hsl(56, 100%, 50%), hsl(56, 100%, 50%) 30px, hsl(56, 80%, 50%) 30px, hsl(56, 80%, 50%) 60px);
+}
+
+.banner-testing {
+    color: #EEE;
+    background-color: #000000;
+    opacity: 0.9;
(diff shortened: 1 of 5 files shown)


elichad:
Could we also have a smaller indicator in the nav bar so that there's always something visible even when scrolled down? Example below.

Top of page:
Screenshot with local banner and yellow 'Local' indicator next to the AMY home link in the nav bar

Scrolled down:
Screenshot with banner hidden but nav bar indicator still present

maneesha:
An indication that you are on a test or local build should always be visible.  I'm ok with having the bar always be visible but if others feel it takes up too much space, I also like Eli's suggestions of changing the menu colors.  We can also consider a different body background color for local and test builds.  Anything that would indicate to the user that they are in a different space.

pbanaszkiewicz:
@elichad @maneesha I think we could make it sticky to the bottom of the page and it would not hide. What do you think?

pbanaszkiewicz:
I was following this design from Django docs for development version: https://docs.djangoproject.com/en/dev/

elichad:
I'd favour sticky at the top so it's harder to miss. It doesn't take up too much space and I'll get used to it!

pbanaszkiewicz:
obraz

Changes:
1. banner stays on top and is never covered by the navigation menu
2. banner background no longer contains diagonal stripes
3. sharper color contrast between font color and background color for increased readability

pbanaszkiewicz:
I didn't want to change navigation bar or page background colors because it would impact readability or require a lot of further work to improve CSS of elements displayed in the navigation bar or on the page.

elichad:
I'm happy with this version.

maneesha:
Looks good to me!

PR #2359, opened 2023-03-10
Add survey links to instructor profile

### LINKED ISSUE #2016: Include survey links in instructor profile view
When a person logs in to AMY, they can see their own profile with a list of workshops they have taught.  Next to each workshop we would like to see a link to the survey results for that workshop (coming from the links in that event's page).

elichad:
Fixes #2016 .

I used the `instructor_pre` field, but we could also use the `reports_link` function to generate the URL on the fly. I chose the former as some events don't seem to have survey links at all (especially older ones) and some point to surveymonkey links. Any thoughts on this?

Side question though: how/when is `instructor_pre` actually set? I assumed it's automatic if we can generate the links automatically, but I can't find it in the code.

### DIFF
--- a/amy/templates/dashboard/instructor_dashboard.html
+++ b/amy/templates/dashboard/instructor_dashboard.html
@@ -103,6 +103,7 @@ Your email address and GitHub id can not be updated here, as they are tied to yo
             <th>Dates</th>
             <th>Venue</th>
             <th>Your role</th>
+            <th>Survey results</th>
           </tr>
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
@elichad According to documentation, `instructors_pre` is not used, but I can see it appears on `EventForm`, probably editable by the user. @maneesha can you confirm that?

pbanaszkiewicz (approved):

PR #2361, opened 2023-03-15
Reduce hardcoding of consent labels in UI

### LINKED ISSUE #2302: Consents: add admin field text representation of a consent term
When displaying all consents for a given person, we're missing a (possibly) short form text describing what the consent stands for.

For example, `may-contact` should have a corresponding textual representation of `May contact`. Note that it's not always the same as changing from hyphenated to title case.

elichad:
Fixes #2302 .

Also propose updating the `content` field of the `public-profile` consent to "Do you consent to publish your profile
on The Carpentries website?" in the AMY Django admin interface.

### DIFF
--- a/amy/consents/migrations/0007_term_short_description.py
+++ b/amy/consents/migrations/0007_term_short_description.py
@@ -26,20 +26,15 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        # add field and make it temporarily nullable
+        # add field
         migrations.AddField(
             model_name="term",
             name="short_description",
-            field=models.CharField(null=True, max_length=100),
+            field=models.CharField(default="", max_length=100),
+            preserve_default=False,
         ),
         # set descriptions for existing terms
         migrations.RunPython(
             set_term_descriptions, reverse_code=migrations.RunPython.noop
-        ),
-        # make field non-nullable
-        migrations.AlterField(
-            model_name="term",
-            name="short_description",
-            field=models.CharField(max_length=100),
         ),
     ]
--- a/amy/consents/migrations/0007_term_short_description.py
+++ b/amy/consents/migrations/0007_term_short_description.py
@@ -1,0 +1,45 @@
+# Written by @elichad 2023-03-15
+
+from django.db import migrations, models
+
+term_descriptions = {
+    "may-contact": "May contact",
+    "public-profile": "Consent to making profile public",
+    "may-publish-name": "Consent to include name when publishing lessons",
+    "privacy-policy": "Privacy policy agreement",
+}
+
+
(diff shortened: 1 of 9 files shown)


amy/consents/migrations/0007_term_short_description.py
```diff
@@ -0,0 +1,45 @@
...
+    dependencies = [
+        ("consents", "0006_auto_20210614_1234"),
+    ]
+
+    operations = [
+        # add field and make it temporarily nullable
+        migrations.AddField(
+            model_name="term",
+            name="short_description",
+            field=models.CharField(null=True, max_length=100),
```
pbanaszkiewicz, line 33: You should be able to create the field with `""` as default value, therefore there should be no need switch `null=True` and then `null=False` in a migration.

amy/templates/dashboard/instructor_dashboard.html
```diff
@@ -37,15 +37,15 @@
       </th><td>{{ user.secondary_email|default:"—" }}</td></tr>
     <tr><th>Gender:</th><td>{{ user.get_gender_display }}{{ user.gender_other }}</td></tr>
     <tr>
-      <th>Can we contact you?</th>
+      <th>{{ consents_content.may_contact }}</th>
```
pbanaszkiewicz, line 40: Is this the same as `consents.public_profile.term.content`? If yes, then I'm not sure we need `consents_content` object.

pbanaszkiewicz (approved):
I left some comments where I'm not sure the solution is needed, but hey I may be wrong so let's have a discussion :) The rest looks good.

amy/templates/dashboard/instructor_dashboard.html
elichad: The content is the same, but if the user hasn't set that consent, `consents.public_profile` won't exist.

amy/consents/migrations/0007_term_short_description.py
elichad: Wouldn't you then have to do the same thing with `blank=True` and `blank=False` instead? I don't think we want this field to be permitted to be blank as it may appear in the UI through lines like this https://github.com/carpentries/amy/pull/2361/files#diff-a97c97acc5cd76b1c48999bacdc2b399b62f678806ab48b559dcc0944c366536R278

pbanaszkiewicz: I *think* `blank` is only applied when creating model instances. You definitely should be fine if you provide one-off value for the migration (default value of `""`). `makemigrations` may even ask you for it.

elichad:
Merging now. I've also made that change to the content of `public-profile` as suggested in the initial comment.

PR #2375, opened 2023-04-05
Allow different term content on instructor training application

### LINKED ISSUE #2367: Add "consent to publish profile" to instructor training application
Based off the conversation in #2162, we should have a consent in the instructor training application that reads:

"The Carpentries will not make public any of your personal data when going
through the application process. However, once certified, do you consent to
your name and any linked social media accounts being published on The
Carpentries website?"

This should be migrated to the publish profile consent once the person profile is created from the application.

elichad:
Requires #2363.

Fixes #2367 by adding a new field `training_request_content` to the Term model. This can optionally be set to display different content for the term on the instructor training request form. If unset, the form will fall back on the existing `content`.

Phrasing for the `may-contact` training request content needs to be properly agreed on - @maneesha do you have thoughts on this?


pbanaszkiewicz:
I merged my code changes so you should do a rebase against `develop` now.

amy/consents/migrations/0009_term_training_request_content.py
```diff
@@ -0,0 +1,53 @@
+# Generated by Django 3.2.18 on 2023-04-05 10:56
+
+from django.db import migrations, models
+
+term_training_request_content = {
+    "may-contact": (
+        "The Carpentries will contact you about the progress of your application."
+        " Once certified, may we also contact you for volunteering opportunities,"
+        " surveys, feedback, etc.?"
```
elichad, line 9: @maneesha @froggleston seeking your feedback about this phrasing. This will be displayed for the may-contact consent on the instructor training request form (unchanged elsewhere)

amy/extforms/forms.py
```diff
@@ -177,13 +177,14 @@ def set_consent_fields(self, terms: Iterable[Term]) -> None:
 
     def create_consent_field(self, term: Term) -> forms.ChoiceField:
         options = [(opt.pk, option_display_value(opt)) for opt in term.options]
+        label = term.training_request_content or term.content
```
pbanaszkiewicz, line 180: I think there is a similar function in another form (ConsentForm?). Should we put this line over there, too?

pbanaszkiewicz: Additionally you marked the field `training_request_content` as nullable. This means it can have 2 "empty" values: `None` and `""`.

For `CharFields` I always prefer to have 1 empty value (`""`) and only accept nullable if there's a good reason for it.

I think here it would be better to keep the field non-nullable with a default value of `""`, so that we are handling it correctly in forms and there are no hidden surprises.

amy/consents/migrations/0009_term_training_request_content.py
froggleston: The only change I might make would be `other volunteering opportunities`

pbanaszkiewicz (approved):
Looks good to me after changes

Issue #2314, opened 2023-01-19
404 on Training requests view


maneesha:
This page is often returning a 404: https://amy.carpentries.org/requests/training_requests/

maneesha:
@elichad  @pbanaszkiewicz Can we close this issue because it will be resolved when #2315 is implemented?

elichad:
@maneesha the 404 should be resolved by addressing #2315, but the underlying difficulty of displaying a large set of results remains. I just realised this view isn't paginated, which might be the real root of the problem (considering the events page is much faster to load on my machine even with ~4500 results).

elichad:
I've done some poking around on this today.
I think the key to this would be to use `serverSide: true` in the DataTable that renders the training requests. This would prevent the client from loading all 7000 items before creating the pagination. But we'd have to write some new code to handle the [subsequent requests](https://datatables.net/manual/server-side) from the DataTable to the Django view. There are a few Python packages out there that purport to handle this (which can be found by searching "django datatable serverside"), but it's not clear if they're maintained or reliable.
Using Django's inbuilt pagination would avoid that, but the DataTable is used in the specific case of training requests because of old issue #1375.

maneesha:
Per @elichad we can close this as it is partly addressed by loading a filtered queryset and addressed in issue about datatables (#1512)

PR #2381, opened 2023-04-13
Set default filtering for training requests

### LINKED ISSUE #2315: Add  filter for default training requests view
The [training requests view](https://amy.carpentries.org/requests/training_requests/) returns all training requests (currently over 7000 results).  This makes the page load very slowly (possibly related to #2314).

Default search results should be:
* State:  Pending or accepted
* Is Matched: Unmatched 

We should then be able to adjust search as we currently do.

Edit to note in effect, the link to Training requests should go [here](https://amy.carpentries.org/requests/training_requests/?search=&group_name=&state=no_d&matched=u&affiliation=&location=&order_by=) instead of [here](https://amy.carpentries.org/requests/training_requests/).

elichad:
Fixes #2315.
Stopgap for #2314 - the filtered view still performs a bit slowly on the full AMY database ([link](https://amy.carpentries.org/requests/training_requests/?search=&group_name=&state=no_d&matched=u&affiliation=&location=&order_by=)), but it shouldn't be catastrophic any more.

This implementation only sets the defaults if the user has chosen not to filter at all - i.e. it is still possible to see all training requests by explicitly selecting the blank options for these fields in the filter UI (or setting the URL query string to `?state=&matched=`)
Note that setting defaults for requests that do include user-selected filter choices is [not recommended](https://django-filter.readthedocs.io/en/main/guide/tips.html?highlight=default#using-initial-values-as-defaults).

### DIFF
--- a/amy/extrequests/tests/test_training_request.py
+++ b/amy/extrequests/tests/test_training_request.py
@@ -290,7 +290,32 @@ class TestTrainingRequestsListView(TestBase):
         self.second_training.tags.add(self.ttt)
 
     def test_view_loads(self):
+        """
+        View should default to settings:
+            state=no_d (Pending or accepted)
+            matched=u (Unmatched)
+        """
+        # Act
         rv = self.client.get(reverse("all_trainingrequests"))
+
+        # Assert
+        self.assertEqual(rv.status_code, 200)
+        self.assertEqual(
+            set(rv.context["requests"]),
+            {self.second_req},
+        )
+
+    def test_view_loads_all_on_request(self):
+        """
+        Explicitly setting state and matched to null should return all requests.
+        """
+        # Arrange
+        query_string = "state=&matched="
+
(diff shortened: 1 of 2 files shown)


amy/extrequests/tests/test_training_request.py
```diff
@@ -290,7 +290,32 @@ def setUp(self):
...
+
+    def test_view_loads_all_on_request(self):
+        """
+        Explicitly setting state and matched to null should return all requests.
+        """
+        # Arrange
+        query_string = "state=&matched="
+
+        # Act
+        rv = self.client.get(reverse("all_trainingrequests"), QUERY_STRING=query_string)
```
pbanaszkiewicz, line 316: Interesting, I didn't know about `QUERY_STRING=`.

pbanaszkiewicz (approved):
Looks good to me

PR #2382, opened 2023-04-14
Add Titanium membership variant, remove prepopulation of membership benefits

### LINKED ISSUE #2100: Remove auto-fill of membership benefits
Right now, membership benefits (number of workshops and number of instructor training seats) gets automatically filled in based on membership level.  Because these benefits will change and may get customized, we would like to remove this feature and have these fields manually entered in.

### LINKED ISSUE #2368: Add "Titanium" membership variant
Add a new membership variant: Titanium.
No benefits are associated with this membership.
This would require updating the [membership choices](https://github.com/carpentries/amy/blob/develop/amy/workshops/models.py#L180) and running the migration.

Is there anything else we would need to do?  @pbanaszkiewicz  @elichad

elichad:
Fixes #2368, fixes #2100.

Pins black to 22.x (matching our precommit/lint jobs) to avoid some reformatting elsewhere in `workshops/views.py` that happens under black 23.x.


Pipfile
```diff
@@ -48,7 +48,7 @@ requests-mock = "~=1.9.3"
 tblib = "~=1.7.0"
 flake8 = "*"
 mypy = "*"
-black = "*"
+black = "~=22.12.0"
```
pbanaszkiewicz, line 51: I think I'd rather have black unpinned and allow it to reformat some files now.

Plus I don't think you should be seeing reformatting applied to `workshops/views.py` since this file is not touched by your changes and would not trigger pre-commit hook.

pbanaszkiewicz (changes requested):
Please bring unpinned black. Otherwise the changes look good.

Pipfile
elichad: I wrote the wrong name 🤦 it was reformatting `workshops/models.py`. As I've reverted it and not made further changes to that file, the formatting won't change now anyway.

Perhaps we should bump our CI linting, pre-commit file etc. to use black 23.x and reformat the whole repo accordingly once we have v4.1 ready to go?

pbanaszkiewicz: @elichad 
> Perhaps we should bump our CI linting, pre-commit file etc. to use black 23.x and reformat the whole repo accordingly once we have v4.1 ready to go?

That's a good idea.

pbanaszkiewicz (approved):

PR #2416, opened 2023-05-10
Remove 'evaluated_by' and 'discarded' fields from TrainingProgress

### LINKED ISSUE #2407: Remove the "Evaluated by" field
Each trainee progress step has a field for "Evaluated by"
I think this just becomes the person who did the data entry.  If it is not used in any way, we should remove it.  
Assiging to https://app.asana.com/0/412899495476717/list to decide if she wants to see this data.

### LINKED ISSUE #2409: Remove the "Discarded" field in training progress
I don't know what this is or why we might ever use it.

elichad:
Fixes #2409
Fixes #2407 

Discussed with instructor training team and agreed that the very few existing TrainingProgress records which are currently marked as discarded will be otherwise unchanged (so some will become valid passes). Manually confirmed in AMY that this will not make anyone eligible for an instructor badge who wasn't eligible before.


amy/trainings/views.py
```diff
@@ -134,31 +124,16 @@ def all_trainees(request):
     if request.method == "POST" and "discard" in request.POST:
         # Bulk discard progress of selected trainees
         form = BulkAddTrainingProgressForm()
```
pbanaszkiewicz, line 126: Should these 3 lines remain? We could probably leave the "elif" from L149 as "if" and skip the check for "discard" in POST since that form was removed.

pbanaszkiewicz (approved):
Looks good to me, I left only 1 small comment.

amy/trainings/views.py
elichad: Good point. I'll make this change.

pbanaszkiewicz:
Hey @elichad I was looking over git log history and I noticed that this PR was merged into a feature branch. Do you want to merge these changes into development branch, too?

elichad:
@pbanaszkiewicz my plan is to merge the checkout work into this `feature/instructor-checkout-changes` branch, and then merge that branch into `develop` when it's complete enough for user testing.

PR #2420, opened 2023-05-16
Rename Discussion to Welcome Session, remove deprecated TrainingRequirements

### LINKED ISSUE #2410: Rename Community Discussions -> Welcome Sessions
We will re-purpose the existing "discussion" field to be for attending a welcome session.
Doing this will mean that we will not be able to easily differentiate whether a trainee did a community discussion under the old requirement or a Welcome session under the new requirement.  
We can assume that people before [DATE] likely did the old style discussion and people after [DATE] likely did the new welcome session but there will likely be a few months overlap where those distinctions weil not be clear.

### LINKED ISSUE #2418: Clean up unused options/fields in training requirements
Last year when we moved to a single Carpentries  badge, it created unused options in the database model for LP-specific steps towards lesson contribution and teaching demo.  We will now have another unused option for lesson contribution.  We should remove these options.  If removing them from the django admin panel is enough, then we can do that at any time.  

Eli note: amy/trainings/forms.py and amy/scripts/seed_training_requirements.py both reference LP-specific TrainingRequirements - these should be removed & there could be others too.

elichad:
Noticed that these PRs aren't checked by CI as they're not against develop - will think about the best way to approach this

Fixes #2418
Fixes #2410


.github/workflows/cicd_main.yml
```diff
@@ -6,7 +6,7 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
```
pbanaszkiewicz, line 9: File `cicd_develop.yml` should be invoked for all PRs, I think. I'm not sure what this change does compared to other pipelines.

.github/workflows/test.yml
```diff
@@ -5,11 +5,12 @@ on:
 
 jobs:
   test:
-    name: Test on Python ${{ matrix.python-version }}
+    name: ${{ matrix.test-type }} tests on Python ${{ matrix.python-version }}
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: [ '3.11' ]
+        test-type: ["Unit", "Migration"]
```
pbanaszkiewicz, line 13: Is this required? What if we skipped `test-type` matrix entry and just had two sequential steps, one `Unit tests` and the other `Migration tests`?

amy/scripts/seed_training_requirements.py
```diff
@@ -9,7 +9,14 @@
...
 
-DEPRECATED_TRAINING_REQUIREMENTS: list[str] = []
+DEPRECATED_TRAINING_REQUIREMENTS: list[str] = [
+    "DC Homework",
+    "SWC Homework",
+    "LC Homework",
+    "DC Demo",
+    "SWC Demo",
+    "LC Demo",
+]
```
pbanaszkiewicz, line 19: Just want to make sure: were these migrated in production? I'm not sure what would happen if these were removed for existing users.

pbanaszkiewicz: Okay, I saw a migration removing these entries. We should not keep 2 systems for managing "static" (seeded) objects. I would suggest sticking to the seeding script, unless a data migration is actually required.

pbanaszkiewicz (commented):
I added some comments and I think we should discuss the approach regarding CI/CD pipeline and data migration vs seeding.

.github/workflows/cicd_main.yml
elichad: This PR is against `feature/instructor-checkout-changes` rather than `develop`, so needs explicit inclusion in a pipeline somewhere. I'm not sure if `cicd_main.yml` is the best place to do this, it was just the most convenient

.github/workflows/test.yml
elichad: It's not required, but the migration tests take some time to run (~3 minutes for just the ones that were added in this PR). Separating them into the parallel job keeps the whole suite a bit shorter. But I wouldn't mind reverting this for now and re-assessing later if it becomes a problem.

amy/scripts/seed_training_requirements.py
elichad: Okay, I can remove that step from the migration. This is one thing I am uncertain of - are the seeding scripts run when deploying new versions to production?

amy/scripts/seed_training_requirements.py
```diff
@@ -22,13 +29,7 @@
 
 TRAINING_REQUIREMENTS: list[TrainingRequirementDef] = [
     {"name": "Training", "url_required": False, "event_required": True},
-    {"name": "DC Homework", "url_required": True, "event_required": False},
-    {"name": "SWC Homework", "url_required": True, "event_required": False},
-    {"name": "Discussion", "url_required": False, "event_required": False},
```
pbanaszkiewicz, line 27: "Discussion" is not being deprecated as it's being switched to "Welcome Session". The migration `0259_...` should probably only migrate the name "Discussion" <-> "Welcome Session", without changes the seeding manages.

amy/workshops/migrations/0259_remove_deprecated_training_requirements.py
```diff
@@ -0,0 +1,87 @@
...
+    except TrainingRequirement.DoesNotExist:
+        pass
+
+
+def delete_outdated_requirements(apps, schema_editor) -> None:
+    TrainingRequirement = apps.get_model("workshops", "TrainingRequirement")
+    TrainingProgress = apps.get_model("workshops", "TrainingProgress")
+
+    # migrate SWC/DC/LC specific progress to generic demo/lesson contribution
+    # this was already done in production AMY but not in local databases
```
pbanaszkiewicz, line 34: What do you mean by "local databases"? Do you mean development databases, or test-amy2 database?

This migration looks reasonable (except for lines 52-54 where Training Requirements are removed), but I'm trying to understand if it's truly needed.

elichad: Development databases. When fake training progresses are created, they choose from all available TrainingRequirements, so my local dev database was populated with lots of SWC Demos, LC Homeworks, etc.
We don't have auto-generated fake data on test-amy2 so this migration isn't required there (though it would have been required if we had generated some).

elichad: I've removed the creation/deletion of TrainingRequirements from this step of the migration since that's covered by the seeding, but this section still remains for discussion.

.github/workflows/cicd_feature_branch.yml
```diff
@@ -0,0 +1,15 @@
+# CI for feature branches - contains only test runs
+
+name: CI (feature)
```
pbanaszkiewicz, line 3: 👍🏼

amy/workshops/models.py
```diff
@@ -656,27 +656,21 @@ def passed_either(*reqs):
...
-            "SWC Homework",
-            "DC Homework",
-            "LC Homework",
-        ]
-        DEMO_TRAININGPROGRESS_NAMES = ["Demo", "SWC Demo", "DC Demo", "LC Demo"]
+        LESSON_CONTRIBUTION_NAMES = ["Lesson Contribution"]
 
         return self.annotate(
             passed_training=passed("Training"),
             passed_lesson_contribution=passed_either(*LESSON_CONTRIBUTION_NAMES),
```
pbanaszkiewicz, line 663: Can we use `passed` instead of `passed_either` with literal string `"Lesson Contribution"` as an argument?

pbanaszkiewicz (approved):
Looks good. I had only 1 comment. But we have to remember about seeding in production before deploying v4.2.

amy/workshops/models.py
elichad: Yes - this is coming in a later PR which updates the checkout requirements after the change from "Lesson Contribution" -> "Get Involved"

PR #2430, opened 2023-05-30
Allow longer auth tokens for password reset

### LINKED ISSUE #2429: Reverse for 'password_reset_confirm' with keyword arguments '{'uidb64': '*******', 'token': '******-********************************'}' not found.
UID/Token replaced with asterisks for security.
```
Reverse for 'password_reset_confirm' with keyword arguments 
'{'uidb64': '*******', 'token': '******-********************************'}' not found. 
1 pattern(s) tried: ['account/reset/(?P<uidb64>[0-9A-Za-z_\\-]+)/(?P<token>[0-9A-Za-z]{1,13}-[0-9A-Za-z]{1,20})/$']
```

We are seeing tokens where the second part is 32 characters and so the password reset fails. Not sure if this has been happening in all cases or just some, but it crops up every couple of days. Django's own [auth/urls.py](https://github.com/django/django/blob/main/django/contrib/auth/urls.py) no longer specifies a length.

elichad:
Fixes #2429 .

@pbanaszkiewicz ideally this should be released as a hotfix rather than waiting for v4.2.

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -27,112 +27,186 @@ from workshops.utils.access import login_required
 from workshops.views import logout_then_login_with_msg
 
 urlpatterns = [
-    path('', RedirectView.as_view(pattern_name='dispatch')),
+    path("", RedirectView.as_view(pattern_name="dispatch")),
     path(settings.ADMIN_URL, admin.site.urls),  # {% url 'admin:index' %}
-
-    path('autoemails/', include('autoemails.urls')),
-    path('api/v1/', include('api.urls')),  # REST API v1
-    path('dashboard/', include('dashboard.urls')),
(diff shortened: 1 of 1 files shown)


config/urls.py
```diff
@@ -27,112 +27,186 @@
...
+    ),
+    path(
+        "account/password_reset/done/",
+        auth_views.PasswordResetDoneView.as_view(
+            template_name="account/password_reset_done.html",
+        ),
+        name="password_reset_done",
+    ),
+    path(
+        "account/reset/<uidb64>/<token>/",
```
elichad, line 74: The whole file has been re-formatted with Black, but this is the only path that I actually changed.

pbanaszkiewicz (approved):

PR #2446, opened 2023-06-23
Update bulk training progress form to work with Get Involved step

### LINKED ISSUE #2445: Update bulk add training progress form
The bulk add training progress form should work for the Get Involved step.

Extra note/bug: submitting the form for a Welcome Session added the session twice to the selected trainees, but only once? Perhaps I double-clicked submit...

elichad:
In the interest of time, I'm raising this not against feature/instructor-checkout-changes but instead against the branch used in https://github.com/carpentries/amy/pull/2431 for review. Will not merge until https://github.com/carpentries/amy/pull/2431 is merged.

Fixes #2445.

This PR will:
- add fields needed for the Get Involved step to the bulk training progress form
- make the bulk form use the same JS as the individual training progress form to dynamically display fields based on options
- expand tests of the bulk form to test adding different requirements

Screenshot showing the bottom of the all trainees page. One trainee is selected and the bulk add progress form displays the fields associated with the Get Involved requirement

### DIFF
--- a/amy/trainings/forms.py
+++ b/amy/trainings/forms.py
@@ -118,6 +118,13 @@ class BulkAddTrainingProgressForm(forms.ModelForm):
         required=True,
     )
 
+    involvement_type = forms.ModelChoiceField(
+        label="Type of involvement",
+        required=False,
+        queryset=Involvement.objects.default_order().filter(archived_at__isnull=True),
+        widget=RadioSelect(),
+    )
+
     helper = BootstrapHelper(
         additional_form_class="training-progress",
         submit_label="Add",
@@ -129,8 +136,10 @@ class BulkAddTrainingProgressForm(forms.ModelForm):
         # the template where this form is used
         "requirement",
         "state",
+        "involvement_type",
         "event",
         "url",
+        "date",
         "notes",
     )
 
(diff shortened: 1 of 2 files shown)


froggleston (approved):
LGTM 👍

amy/trainings/forms.py
```diff
@@ -129,8 +136,10 @@ class BulkAddTrainingProgressForm(forms.ModelForm):
         # the template where this form is used
         "requirement",
         "state",
+        "involvement_type",
         "event",
         "url",
+        "date",
```
pbanaszkiewicz, line 142: Is there a widget for user when they click on the "date of occurrence" field?

pbanaszkiewicz: If not, then at least there should be a format specified in `help_text` for the field.

pbanaszkiewicz (approved):
Minor UX question.

amy/trainings/forms.py
elichad: Yes, there's a widget. I think this is used by default with DateFields. It's still possible to type in a date instead, so good idea to include the required format in the help text.

PR #2458, opened 2023-07-07
Add accessibility testing workflow

elichad:
Sets up a workflow that can use either [pa11y](https://pa11y.org) or [Google Lighthouse](https://developer.chrome.com/docs/lighthouse/overview/) for accessibility tests. Both have benefits, but for now I've only enabled pa11y as most of our issues fall into the [axe-core](https://dequeuniversity.com/rules/axe/html/4.7) ruleset used by both.

I've set the workflow up to be allowed to fail, as it's not realistic to fix all our a11y issues as part of this PR; instead we will plan to resolve issues over time.


config/settings.py
```diff
@@ -547,8 +547,8 @@
 
 # Debug Toolbar
 # -----------------------------------------------------------------------------
-DEBUG_TOOLBAR_PATCH_SETTINGS = False
```
elichad, line 550: This setting has been deprecated for some time: https://django-debug-toolbar.readthedocs.io/en/latest/changes.html#id31

config/settings.py
```diff
@@ -547,8 +547,8 @@
...
 INTERNAL_IPS = ["127.0.0.1", "::1"]
+DEBUG_TOOLBAR_CONFIG = {"SHOW_COLLAPSED": True}
```
elichad, line 551: The debug panel always loads collapsed for me on my dev instance, but for the a11y tests it was always expanded. Not sure why the difference, but the panel has a lot of a11y failures when expanded (increased the total count by ~25%...)

.github/workflows/accessibility.yml
```diff
@@ -0,0 +1,139 @@
...
+          - 5432:5432
+      redis:
+        image: redis
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 6379:6379
```
pbanaszkiewicz, line 51: I don't think we need redis for our test suite.
If other pipelines use it, we could disable it there too.

.pa11yci
```diff
@@ -0,0 +1,164 @@
...
+        "http://127.0.0.1:8000/workshops/events/metadata_changed/",
+        "http://127.0.0.1:8000/workshops/version/1/",
+        "http://127.0.0.1:8000/workshops/persons/bulk_upload/",
+        "http://127.0.0.1:8000/workshops/persons/add/",
+        "http://127.0.0.1:8000/workshops/persons/merge/?person_a=2&person_b=3",
+        "http://127.0.0.1:8000/workshops/person/1/",
+        "http://127.0.0.1:8000/workshops/person/1/edit/",
+        "http://127.0.0.1:8000/workshops/task/1/",
+        "http://127.0.0.1:8000/workshops/workshop_staff/",
+        "http://127.0.0.1:8000/account/logout/"
```
pbanaszkiewicz, line 141: I assume pa11y doesn't support "HOST" setting and we'll have to maintain this list of URLs?

config/settings.py
pbanaszkiewicz: I'm guessing is that it uses cookie or some other storage to remember last state (collapsed/expanded).

docs/accessibility_testing.md
```diff
@@ -0,0 +1,55 @@
+# Accessibility Testing in AMY
+
+## Automated Testing
+
+The automated tests aim to test one page corresponding to each template in `amy/templates` (including base templates and the `includes` subfolder).
+
+The following pages cannot be effectively tested automatically at the moment. Problems include:
+
+* page inaccessible to logged-in admin (`/terms/action_required``)
+* objects unavailable (e.g. no available signups for instructors to view)
```
pbanaszkiewicz, line 10: You mean that our dev_database command could create more objects? I think we would be able to implement this.

pbanaszkiewicz (approved):
LGTM, minor comments/questions/suggestions left.

pbanaszkiewicz:
@elichad Please remember to select the release version you want this in.

docs/accessibility_testing.md
elichad: Yes, that's what I mean. I created #2468 for this

.pa11yci
elichad: Yes, we have to maintain it. pa11y [can use a sitemap](https://github.com/pa11y/pa11y-ci#sitemaps), though we would still need to manually specify URLs where we use custom actions. (Lighthouse doesn't support using a sitemap)
Django does support [creating sitemaps](https://docs.djangoproject.com/en/4.2/ref/contrib/sitemaps/), but it doesn't seem particularly straightforward.

.github/workflows/accessibility.yml
elichad: Ok, I removed redis from this workflow and `test.yml` and it seems to work fine.

PR #2466, opened 2023-07-11
Correct links in changes_log view

### LINKED ISSUE #2456: "Recently changed" page links point to wrong records
On https://amy.carpentries.org/workshops/log/, the links to the edited records are wrong. Each link points to a Version, and that Version can be associated with an entirely different record, possibly on an entirely different model.
Example from test-amy:
`Pluto changed [TrainingProgress object (53)](https://test-amy2.carpentries.org/workshops/version/291/)`, but that link goes to a version of `Person Buttered Popcorn` instead.

elichad:
Fixes #2456.

Though now I've done this, I realise that displaying just the first Version created as part of each Revision loses some helpful information (e.g. if a Task is created, this updates an Event and a Person; only the Event name is shown in the changes_log view). As we barely use this view, I'm not going to propose further changes at this time.

### DIFF
--- a/amy/templates/workshops/changes_log.html
+++ b/amy/templates/workshops/changes_log.html
@@ -9,7 +9,13 @@
     {% for change in log %}
     <tr>
       <td>{{ change.date_created|date:'M j, P' }} </td>
-      <td>{% if change.user %}{{ change.user.personal }}{%else%}Unknown user{%endif%} changed <a href="{% url 'object_changes' change.pk %}">{{ change.version_set.all|first }}</a></td>
+      <td>
+        {% if change.user %}
+          {{ change.user.personal }}
+        {%else%}
+          Unknown user
+        {%endif%} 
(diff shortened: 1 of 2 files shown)


amy/templates/workshops/changes_log.html
```diff
@@ -9,7 +9,13 @@ <h3>Recently changed</h3>
...
     <tr>
       <td>{{ change.date_created|date:'M j, P' }} </td>
-      <td>{% if change.user %}{{ change.user.personal }}{%else%}Unknown user{%endif%} changed <a href="{% url 'object_changes' change.pk %}">{{ change.version_set.all|first }}</a></td>
+      <td>
+        {% if change.user %}
+          {{ change.user.personal }}
+        {%else%}
+          Unknown user
+        {%endif%} 
+        changed <a href="{% url 'object_changes' change.version_set.first.pk %}">{{ change.version_set.all|first }}</a></td>
```
pbanaszkiewicz, line 18: Can you use `change.version_set.first` inside `<a>...</a>`?

pbanaszkiewicz (approved):

PR #2471, opened 2023-07-14
Improve event validation and option filtering for training progress and instructor badge awarding

### LINKED ISSUE #2342: Recording trainee participation at TTT event
For Instructor Training, we record trainee participation twice.

First, we record them by giving them a role of `learner` at the event they attended, matching them to a membership if applicable.

Then we record that they completed the `training` step towards trainee progress.  Here, we are able to note whether they passed, failed, or were asked to repeat.  We then add in the associated training event.  In this step, we would like to see this automatically fill in the TTT event they were a learner at.  If they were a learner at more than one TTT event, then we should see a drop down that is limited to just the events they were a learner at, rather than seeing all events as options.

EDIT to add additionally -- As expected, a trainee may have multiple `training` steps completed.  However, we should not allow multiple `training` steps for the same event.

### LINKED ISSUE #2343: Recording event associated with instructor badge
When we award an Instructor badge, we have the optional field for event.
This should be mandatory field.  Similar to #2342, we would like to see this automatically fill in the most recent TTT event that they "passed" the "training" requirement for. We should not have options for multiple events because this should be a unique condition.

elichad:
Fixes #2342, fixes #2343 :

- when awarding instructor badge, auto-fill badge and event selection where possible. Where multiple passed Trainings exist for a trainee, filter event field options to only include those events (though this generally shouldn't happen as only one passed Training should exist)
- when creating a training progress, filter event field options to include only events where the selected trainee has a learner task
- when validating training progress, ensure that 
    1. trainee has a learner task for the chosen event
    2. trainee does not already have a progress for the chosen event
- when adding training progress in bulk, if there are some failures due to the above validation, create all progresses that can be created. Instead of the usual success message, show the errors and info about how many trainee progresses were created/skipped (screenshot).
- start some documentation for where we use design patterns that might be useful elsewhere (motivated by how it took me a while to realise the community role creation view had the dynamic behaviour I was looking for)

Screenshot of Trainees page, with messages at the top: "Error: Training: Trainee Jay Dean already has a training progress for event 2023-05-29-hodgefort-ttt", repeat for Wanda Higgins, and "Changed progress of 2 trainees. 2 trainees were skipped due to errors."


amy/workshops/lookups.py
```diff
@@ -79,7 +89,15 @@ def get_queryset(self):
...
-        results = models.Event.objects.filter(tags__name="TTT")
+        results = models.Event.objects.ttt()
+
+        # trainee is provided through the TrainingProgress creation views
+        # for which learner tasks are relevant
+        if trainee := self.request.GET.get("trainee"):
+            learner_tasks = models.Task.objects.filter(
+                role__name="learner", person__id=trainee
+            )
+            results = results.filter(task__in=learner_tasks)
```
pbanaszkiewicz, line 100: Is this a similar case as above? Perhaps we should have a separate class for this case, too.

amy/workshops/models.py
```diff
@@ -2671,6 +2671,34 @@ def clean_event(
         elif not requirement.event_required and self.event:
             return self.get_not_required_error(requirement)
 
+        if self.event and hasattr(self, "trainee") and self.trainee:
```
pbanaszkiewicz, line 2674: `trainee` is a FK, we should always have this attribute. It seems non-nullable, so maybe it's enough to check `if self.event:`?

amy/workshops/models.py
```diff
@@ -2671,6 +2671,34 @@ def clean_event(
...
+            # check this trainee has a learner task for this event
+            try:
+                Task.objects.get(
+                    person=self.trainee,
+                    event=self.event,
+                    role=Role.objects.get(name="learner"),
+                )
```
pbanaszkiewicz, line 2681: One concern I have about this code is that we're using a SELECT query during creation of training progress. Maybe such code could live in a form `clean_event` method?

amy/workshops/models.py
```diff
@@ -2671,6 +2671,34 @@ def clean_event(
...
+                    "This progress cannot be created without a corresponding learner "
+                    f"task. Trainee {self.trainee} does not have a learner task for "
+                    f"event {self.event}."
+                )
+                return ValidationError(msg)
+
+            # check this trainee doesn't already have a training progress for this event
+            return self.clean_event_existing_progress(self.event, self.trainee)
+
+    def clean_event_existing_progress(self, event, trainee):
```
pbanaszkiewicz, line 2693: Please add type hints.

amy/workshops/models.py
```diff
@@ -2671,6 +2671,34 @@ def clean_event(
...
+        if existing_progress := trainee.trainingprogress_set.filter(event=event):
+            if not (existing_progress.count() == 1 and self in existing_progress):
+                msg = (
+                    f"Trainee {self.trainee} already has a training progress "
+                    f"for event {self.event}."
+                )
+                return ValidationError(msg)
```
pbanaszkiewicz, line 2700: Could we have a UNIQUE constraint on `(trainee, event)` pair that would move this check to database level?

One thing that should be checked is how the DB treats pairs containing NULL values for `event`.

amy/workshops/views.py
```diff
@@ -2186,11 +2187,22 @@ def get_initial(self, **kwargs):
...
-            tasks = Person.objects.get(
-                pk=self.request.GET["person"]
-            ).get_training_tasks()
-            if tasks.count() == 1:
-                initial.update({"event": tasks[0].event})
+            initial.update(
+                {
+                    "badge": Badge.objects.get(name="instructor"),
+                }
+            )
```
pbanaszkiewicz, line 2194: Probably can be simplified:
```suggestion
            initial["badge"] = Badge.objects.get(name="instructor")
```

amy/workshops/views.py
```diff
@@ -2186,11 +2187,22 @@ def get_initial(self, **kwargs):
...
+                trainee__id=self.request.GET["person"],
+                requirement=TrainingRequirement.objects.get(name="Training"),
+                state="p",
+            )
+            if progresses.count() == 1:
+                initial.update(
+                    {
+                        "event": progresses[0].event,
+                    }
+                )
```
pbanaszkiewicz, line 2205: We could use try-except here:
```suggestion
            try:
                progress = TrainingProgress.objects.get(
                    trainee__id=self.request.GET["person"],
                    requirement=TrainingRequirement.objects.get(name="Training"),
                    state="p",
                )
                initial["event"] = progress.event
            except (TrainingProgress.DoesNotExist, TrainingProgress.MultipleObjectsReturned):
                pass
```

amy/workshops/models.py
```diff
@@ -2671,6 +2671,34 @@ def clean_event(
...
+                    f"task. Trainee {self.trainee} does not have a learner task for "
+                    f"event {self.event}."
+                )
+                return ValidationError(msg)
+
+            # check this trainee doesn't already have a training progress for this event
+            return self.clean_event_existing_progress(self.event, self.trainee)
+
+    def clean_event_existing_progress(self, event, trainee):
+        if existing_progress := trainee.trainingprogress_set.filter(event=event):
```
pbanaszkiewicz, line 2694: `existing_progress` should be in plural form

amy/trainings/views.py
```diff
@@ -125,18 +126,45 @@ def all_trainees(request):
...
+                        url=form.cleaned_data["url"],
+                        date=form.cleaned_data["date"],
+                        notes=form.cleaned_data["notes"],
+                    )
+                    progress.full_clean()
+                    progress.save()
+                except ValidationError as e:
+                    errors.append(e)
+
+            if len(errors) > 0:
```
pbanaszkiewicz, line 147: Idiomatic Python version:
```suggestion
            if errors:
```

pbanaszkiewicz (changes requested):
I left some comments. I think this is heading in a right direction, but I would like you to consider some of my proposals.

amy/workshops/lookups.py
elichad: This lookup is only used in the training views (TrainingProgressForm, BulkAddTrainingProgressForm, TraineeFilter), so I'm not sure it makes as much sense to separate it out.

elichad:
I'm bumping this to the 4.3 milestone for time reasons. There's some more work needed here to address Piotr's review comments, and it's only a convenient extra to the checkout project rather than anything essential.

elichad:
@pbanaszkiewicz I finally got around to addressing your comments on this PR. Ready for a re-review!

amy/trainings/utils.py
```diff
@@ -0,0 +1,24 @@
...
+def raise_validation_error_if_no_learner_task(trainee: Person, event: Event) -> None:
+    """Raises a ValidationError if trainee does not have a learner task for the event.
+    Does nothing if trainee or event is None."""
+    if not trainee or not event:
+        return
+    try:
+        Task.objects.get(
+            person=trainee,
+            event=event,
+            role=Role.objects.get(name="learner"),
```
pbanaszkiewicz, line 15: 1 query less if you switch to `role__name="learner"`.

amy/trainings/tests/test_training_progress.py
```diff
@@ -166,12 +172,34 @@ def test_event_must_be_blank(self):
...
+            event=event,
+        )
+        p2 = TrainingProgress(
+            requirement=self.event_required,
+            trainee=self.admin,
+            event=event,
+        )  # do not save to DB as this violates the unique constraint we want to test
+        p1.full_clean()  # should be no error if only this progress exists
+        with self.assertValidationErrors(["__all__"]):
+            p2.full_clean()
```
pbanaszkiewicz, line 201: Can you check the validation error here?

pbanaszkiewicz (approved):
Looks good to me!

PR #2490, opened 2023-07-26
Update color palette for better accessibility

elichad:
This is achieved by overriding the Sass colors set in Bootstrap 4. I followed the guidance in the [Bootstrap 4 docs](https://getbootstrap.com/docs/4.0/getting-started/theming/#sass). `custom_bootstrap.css` and `custom_bootstrap.css.map` are both generated by Sass based on `custom_bootstrap.scss`.

Most of the palette is copied from the [Bootstrap 5 palette](https://getbootstrap.com/docs/5.0/customize/color/), but some customisation was needed to meet WCAG 2.1 AA contrast throughout, especially for links in striped tables. There is no change to the meaning of any color.

This PR will resolve the majority of contrast-related a11y issues across the whole UI, reducing pa11y failures from ~4400 to ~1300.

Example screenshots which use most of the palette:
Old palette:
Screenshot of all trainees page.
New palette:
Screenshot of all trainees page with updated colors. Green, blue, red are darker. Cyan is lighter. Navbar elements are lighter gray.


amy/static/scss/custom_bootstrap.scss
```diff
@@ -12,12 +15,13 @@ $gray-300: #dee2e6;
 $gray-400: #ced4da;
 $gray-500: #adb5bd;
 $gray-600: #6c757d;
+$gray-650: #5a636a; // custom
```
pbanaszkiewicz, line 18: Could you perhaps use some calculate function to get this value, for example:
```scss
$gray-650: hsl(208, 7%, 46% - 8%);  // hsl values for $gray-600
```
This is an example, as the value you entered for `$gray-650` is `hsl(206, 8%, 38%)` - very similar to what I put above, but not quite the same.

...
After some thinking I decided to add here that using `hsl` or any other color function only makes sense if you modified another color by e.g. increasing or decreasing individual component (like lightness or saturation, or hue). If there's too many components changed then it doesn't really make any sense.

amy/static/css/custom_bootstrap.css
pbanaszkiewicz, line 1: Should this filename contain `.min`?

amy/static/scss/custom_bootstrap.scss
```diff
@@ -0,0 +1,98 @@
+// custom_bootstrap.scss
+
+// compile with Sass:
+// $ npm install -g sass
```
pbanaszkiewicz, line 4: I'd prefer to put `sass` as dev dependency in `package.json`. This way everyone following readme will have it installed locally if they need to regenerate this file.

pbanaszkiewicz (approved):
Looks good. Kudos for such a big improvement in accessibility in one go!

I just want you to confirm if there is any impact on deployment with this new (S)CSS file introduced. I guess once we regenerate it, we should run `collectstatic` again?

Do you think it's worth updating README to capture these instructions in case someone wants to extend or modify our SCSS?

amy/static/scss/custom_bootstrap.scss
elichad: I switched to hsl, as it's _mainly_ a lightness change (difference of 2 in the other components between `gray-600` and `gray-700`).

PR #2491, opened 2023-07-26
[Emails] Template context variables documentation links

### LINKED ISSUE #2482: [Emails] Display variables available in the context for every signal
As a user I want to be able to see what variables, and with what type, are available in the email template body context.

pbanaszkiewicz:
This fixes #2482 by providing links to documentation for every variable used in email template context (e.g. person, event, award, etc.).

It builds on top of #2481.

~~WIP. Needs testing and perhaps some code improvements / refactoring.~~

### DIFF
--- a/amy/templates/emails/email_template_create.html
+++ b/amy/templates/emails/email_template_create.html
@@ -6,4 +6,13 @@
 
 {% crispy form %}
 
+<div class="row">
+  <div class="col-12 col-lg-2">
+    <strong>Template variables</strong>
+  </div>
+  <div class="col-12 col-lg-10">
+    Visible after saving the template.
+  </div>
+</div>
+
 {% endblock %}
--- a/amy/templates/emails/email_template_create.html
+++ b/amy/templates/emails/email_template_create.html
@@ -8,7 +8,7 @@
 
 <div class="row">
   <div class="col-12 col-lg-2">
-    <strong>Template variables</strong>
+    <strong>Variables available in the email template body</strong>
   </div>
   <div class="col-12 col-lg-10">
     Visible after saving the template.
--- a/amy/templates/emails/email_template_edit.html
+++ b/amy/templates/emails/email_template_edit.html
@@ -9,7 +9,7 @@
 
 <div class="row">
   <div class="col-12 col-lg-2">
-    <strong>Template variables</strong>
+    <strong>Variables available in the email template body</strong>
   </div>
   <div class="col-12 col-lg-10">
     <ul>
--- a/amy/templates/emails/email_template_edit.html
+++ b/amy/templates/emails/email_template_edit.html
@@ -1,9 +1,23 @@
 {% extends "base_nav.html" %}
 
 {% load crispy_forms_tags %}
+{% load emails %}
 
 {% block content %}
 
 {% crispy form %}
+
+<div class="row">
+  <div class="col-12 col-lg-2">
+    <strong>Template variables</strong>
+  </div>
+  <div class="col-12 col-lg-10">
+    <ul>
+    {% for key, value in body_context_annotations.items %}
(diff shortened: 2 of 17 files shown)


pbanaszkiewicz:
@elichad ready for your review! Thanks.

amy/templates/emails/email_template_create.html
```diff
@@ -6,4 +6,13 @@
 
 {% crispy form %}
 
+<div class="row">
+  <div class="col-12 col-lg-2">
+    <strong>Template variables</strong>
+  </div>
+  <div class="col-12 col-lg-10">
+    Visible after saving the template.
```
elichad, line 14: Hmm, it would be nice to show the variables as soon as a signal is selected, but I know that would increase the complexity a lot (is it even possible for this case?)
I think we can manage with this and make it part of the workflow to save before entering the full body text, since we shouldn't be creating tons of templates.

amy/templates/emails/scheduled_email_edit.html
```diff
@@ -15,4 +16,17 @@ <h1>
 
 {% crispy form %}
 
+<div class="row">
+  <div class="col-12 col-lg-2">
+    <strong>Template variables</strong>
+  </div>
+  <div class="col-12 col-lg-10">
+    <ul>
+    {% for key, value in body_context_annotations.items %}
```
elichad, line 25: Can these template variables still be used when editing a scheduled email? I had assumed the templates were rendered once and weren't re-rendered before sending.
If they can't be used, this row should be removed from here and `scheduled_email_detail.html`.

elichad (commented):
This is looking good! I have just a couple of questions/suggestions before approving

amy/templates/emails/email_template_create.html
pbanaszkiewicz: I think it's not impossible, but since I have a ton of other features to implement I didn't focus on this case.

One option I can see is to load the variables and their definitions to a JSON object, then switch display in this depending on what user is choosing in the dropdown.

amy/templates/emails/email_template_edit.html
```diff
@@ -1,9 +1,23 @@
...
 
 {% block content %}
 
 {% crispy form %}
 
+<div class="row">
+  <div class="col-12 col-lg-2">
+    <strong>Template variables</strong>
+  </div>
+  <div class="col-12 col-lg-10">
```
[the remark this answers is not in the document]
pbanaszkiewicz: I'm going to update the wording so that it's clearer what they refer to.

About template validation: yes, I think I could build that in here. The previous system contained similar validation (for proper Django template format).

amy/templates/emails/scheduled_email_edit.html
pbanaszkiewicz: @elichad You're correct.

There are 2 renderings:
1. raw user input --(django template enigne)-> Markdown format
2. Markdown format --(markdown compiler)-> HTML

Currently first rendering occurs when an email is scheduled, and the second should be performed by the worker (at the moment it isn't - I will get back to this).

The advantage of this approach is simplicity, but the clear disadvantage is the final form of the email which can't be updated with newer versions of the objects included. So potentially we need to shift the rendering more towards the worker - it would have to render Django template and then Markdown.

I'll remove these lines, however they may be brought back in future.

amy/templates/emails/email_template_edit.html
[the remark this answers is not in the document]
pbanaszkiewicz: @elichad Actually the validation is not possible at this point.

We'd get the best results if we used Django Template Engine behavior for checking if all variables are accessible. However, to do that we have to have context with real data. At this point we're working with email templates without any context, so proper validation is not possible.

pbanaszkiewicz:
@elichad Please take a look again and consider my responses to your questions. Thanks!

elichad (approved):
LGTM now!

PR #2492, opened 2023-07-27
Introduce separate views for Get Involved submissions

### LINKED ISSUE #2472: Actions when trainee submits "Get involved"
After a trainee successfully submits a "Get involved" step, the trainee can not currently view, edit, or delete their submission, and the trainee is able to create another submission.

The trainee should see the following: 

* The trainee should see a summary of their submission with a message that says "Thank you. Your submission will be reviewed in 7-10 days."   
* The trainee should be able to edit or delete their submission.  
* The trainee should not be able to submit another entry.

### LINKED ISSUE #2473: Actions when admin reviews trainee's "Get involved" submission
When an admin reviews a trainee's "Get involved" submission:

* If the submission is accepted, a summary of the submission should remain in the admin and in the trainee views.  The trainee should not be able to create any more submissions.
* If the submission is not accepted (asked to repeat), a summary of submission should remain in the admin and in the trainee views.  Both views should also include the notes that show why the submission was not accepted.  The trainee should then be able to create a new submission.  This should continue until a submission is accepted.

elichad:
Removes the Get Involved submission form from the training progress view and instead provides create/edit/delete views for the submission. Fixes #2472; fixes #2473.

Allows trainees to view their most recent submission, and edit or delete their submission if it has not yet been reviewed.

These views need some permissions to be set so that people can't edit or delete other TrainingProgresses by changing the ID in the URLs! I'll work on this next - #2493.


elichad:
@pbanaszkiewicz Sorry, I forgot to request your review when I opened this PR (though I know you've been away so maybe it makes no difference!). Could you prioritise reviewing this when you have time please?

amy/templates/dashboard/training_progress.html
```diff
@@ -75,38 +75,48 @@ <h2>Your progress towards becoming a Carpentries Instructor</h2>
...
           {% include 'includes/progress_trainee_view.html' with progress=progress_get_involved %}
           {% if progress_get_involved.state == "n" %}
-          <p>Please wait while we evaluate your submission.</p>
+          <p>Your submission will be evaluated within 7-10 days.</p>
+          <p>You may edit or delete your submission before it is evaluated. If you delete it, you will be allowed to create a new submission afterwards.</p>
+          {% endif %}
+          <p>Details of your submission are below.</p>
+          <p>
+            <strong>Activity:</strong> {{ progress_get_involved.involvement_type }}<br/>
+            <strong>Date:</strong> {% firstof progress_get_involved.date|date:'F j, Y' "No date provided" %}<br/>
```
pbanaszkiewicz, line 89: Nice, I didn't know about `firstof`. How does particular tag work when `progress_get_involved.date` is empty (e.g. `None`)? Does `|date` filter complain?

amy/templates/dashboard/training_progress.html
```diff
@@ -75,38 +75,48 @@ <h2>Your progress towards becoming a Carpentries Instructor</h2>
...
+            <strong>URL:</strong> {% firstof progress_get_involved.url "No URL provided" %}<br/>
+            <strong>Notes:</strong> {% firstof progress_get_involved.trainee_notes "No notes provided" %}
+          </p>
+          {% if progress_get_involved.state == "n" %}
+              <p class="edit-object float-left">
+                <a href="{% url 'getinvolved_update' progress_get_involved.pk %}" class="btn btn-primary">Edit</a>  
+              </p>
+              <div class="delete-object float-left" style="padding-left: 2%;">
```
pbanaszkiewicz, line 97: Can you use [`.p`](https://getbootstrap.com/docs/4.6/utilities/spacing/#notation)-style class here? What I'm mostly concerned about is the percentage, as I think we're not using this unit for spacing anymore.

amy/dashboard/views.py
```diff
@@ -209,8 +216,6 @@ def autoupdate_profile(request):
 
 @login_required
 def training_progress(request):
```
pbanaszkiewicz, line 218: Given how it has been simplified, it should be turned into a generic `TemplateView`. Can be done later as long as there's a ticket for it.

amy/dashboard/views.py
```diff
@@ -229,7 +234,33 @@ def training_progress(request):
...
+
+
+class GetInvolvedCreateView(LoginRequiredMixin, RedirectSupportMixin, AMYCreateView):
+    # permission_required = TODO
+    model = TrainingProgress
+    form_class = GetInvolvedForm
+    template_name = "get_involved_form.html"
+    success_url = reverse_lazy("training-progress")
+    success_message = (
+        "Thank you. Your Get Involved submission will be reviewed within 7-10 days."
```
pbanaszkiewicz, line 254: Almost the same message ("Your submission will be evaluated within 7-10 days."). Optionally you could store the "7-10 days" part somewhere as a configurable.

amy/dashboard/views.py
```diff
@@ -229,7 +234,33 @@ def training_progress(request):
...
+    )
+
+    def get_context_data(self, **kwargs) -> dict:
+        context = super().get_context_data(**kwargs)
+        context["title"] = str("Submit your Get Involved activity")
```
pbanaszkiewicz, line 259: This string doesn't need conversion to string.

amy/dashboard/views.py
```diff
@@ -238,25 +269,33 @@ def training_progress(request):
...
+            return self.form_valid(get_involved_form)
 
-            messages.success(
-                request,
-                "Thank you. Your Get Involved submission will be reviewed within 7-10 "
-                "days.",
-            )
-            return redirect(reverse("training-progress"))
+        else:
+            return self.form_invalid(get_involved_form)
```
pbanaszkiewicz, line 277: This logic should not be handled in `post()` method. If you need to set instance, you can probably override `get_form_kwargs()`:

```python
    def get_form_kwargs(self):
        kwargs = super().get_form_kwargs()
        base_training_progress = TrainingProgress(
            trainee=self.request.user,
            state="n",  # not evaluated yet
            requirement=TrainingRequirement.objects.get(name="Get Involved"),
        )
        kwargs["instance"] = base_training_progress
        return kwargs
```

There probably is no access to `request` anymore and so it should be saved in `dispatch()` or `get()` and `post()` methods. We have a few views that use this technique.

amy/dashboard/views.py
```diff
@@ -229,7 +234,33 @@ def training_progress(request):
...
+        "Thank you. Your Get Involved submission will be reviewed within 7-10 days."
+    )
+
+    def get_context_data(self, **kwargs) -> dict:
+        context = super().get_context_data(**kwargs)
+        context["title"] = str("Submit your Get Involved activity")
+        return context
+
+    def post(self, request):
+        self.object = None
```
pbanaszkiewicz, line 263: Not sure what this line is for here, can you explain? (If it's still needed after moving to `get_form_kwargs`)

amy/dashboard/views.py
```diff
@@ -238,25 +269,33 @@ def training_progress(request):
...
+    # permission_required = TODO
+    model = TrainingProgress
+    form_class = GetInvolvedForm
+    template_name = "get_involved_form.html"
+    success_url = reverse_lazy("training-progress")
+    success_message = "Your Get Involved submission was updated successfully."
+
+    def get_context_data(self, **kwargs) -> dict:
+        context = super().get_context_data(**kwargs)
+        context["title"] = str("Update your Get Involved submission")
```
pbanaszkiewicz, line 290: String conversion not needed.

amy/dashboard/views.py
```diff
@@ -238,25 +269,33 @@ def training_progress(request):
...
-        "title": "Your training progress",
-        "get_involved_form": get_involved_form,
-        "progress_training": progress_training,
-        "progress_get_involved": progress_get_involved,
-        "progress_welcome": progress_welcome,
-        "progress_demo": progress_demo,
-    }
-    return render(request, "dashboard/training_progress.html", context)
+
+class GetInvolvedUpdateView(LoginRequiredMixin, RedirectSupportMixin, AMYUpdateView):
```
pbanaszkiewicz, line 280: I think `RedirectSupportMixin` is not needed here, as `ModelFormMixin` (part of Django's `UpdateView`) handles redirect after form submission.

amy/dashboard/views.py
```diff
@@ -229,7 +234,33 @@ def training_progress(request):
...
+        "title": "Your training progress",
+        "progress_training": progress_training,
+        "progress_get_involved": progress_get_involved,
+        "progress_welcome": progress_welcome,
+        "progress_demo": progress_demo,
+    }
+    return render(request, "dashboard/training_progress.html", context)
+
+
+class GetInvolvedCreateView(LoginRequiredMixin, RedirectSupportMixin, AMYCreateView):
```
pbanaszkiewicz, line 247: I think `RedirectSupportMixin` is not needed here, as `ModelFormMixin` (part of Django's `CreateView`) handles redirect after form submission.

amy/dashboard/views.py
```diff
@@ -238,25 +269,33 @@ def training_progress(request):
...
+    success_url = reverse_lazy("training-progress")
+    success_message = "Your Get Involved submission was updated successfully."
+
+    def get_context_data(self, **kwargs) -> dict:
+        context = super().get_context_data(**kwargs)
+        context["title"] = str("Update your Get Involved submission")
+        return context
+
+
+class GetInvolvedDeleteView(LoginRequiredMixin, RedirectSupportMixin, AMYDeleteView):
```
pbanaszkiewicz, line 294: `RedirectSupportMixin` not needed here, as `AMYDeleteView` handles redirection to `success_url` on its own.

amy/templates/get_involved_form.html
pbanaszkiewicz, line 1: Minor: you may want to configure your editor to automatically create empty line at the end of files (setting called "Insert Final Newline" in Visual Studio Code), as it's highlighted here in the HTML files that they don't have line ending characted at the end of the file.

amy/templates/get_involved_form.html
```diff
@@ -0,0 +1,24 @@
...
+<p>
+  Only submit an activity that you have already completed.
+</p>
+<p>
+  If submitting a GitHub contribution, please be sure your contribution is to a repository in one of the 
+  <a href="https://docs.carpentries.org/topic_folders/communications/tools/github_organisations.html">
+    GitHub organisations owned by The Carpentries
+  </a>.
+</p>
+{% crispy form %}
```
pbanaszkiewicz, line 23: Can this be replaced with the following?

```
{% include "includes/get_involved_form_embedded.html" with get_involved_form=form %}
```

amy/dashboard/tests/test_get_involved_views.py
```diff
@@ -0,0 +1,234 @@
...
+from django.urls import reverse
+
+from trainings.models import Involvement
+from workshops.models import TrainingProgress, TrainingRequirement
+from workshops.tests.base import TestBase
+
+
+class TestGetInvolvedCreateView(TestBase):
+    def setUp(self):
+        super()._setUpUsersAndLogin()
```
pbanaszkiewicz, line 12: Please be aware that this is using an admin user to perform the requests. Perhaps it would be better to create an instructor account and authorize it in `self.client`.

amy/dashboard/tests/test_get_involved_views.py
```diff
@@ -0,0 +1,234 @@
...
+                self.admin.pk,
+                "https://example.org",
+                self.get_involved.pk,
+                self.github_contribution.pk,
+                date(2023, 7, 27),
+            )
+        ]
+        self.assertEqual(got, expected)
+
+    def test_create_view_submission_invalid_notes(self):
```
pbanaszkiewicz, line 115: I feel like this is testing form validation and could be moved to form-specific test (which will perform faster than this).

amy/dashboard/tests/test_get_involved_views.py
```diff
@@ -0,0 +1,234 @@
...
+            url="https://example.org",
+            date=date(2023, 7, 27),
+        )
+
+    def test_delete_view_get_request_not_allowed(self):
+        # Act
+        rv = self.client.get(reverse("getinvolved_delete", args=[self.progress.pk]))
+
+        # Assert
+        self.assertEqual(rv.status_code, 405)
```
pbanaszkiewicz, line 223: You probably don't have to test this, I hope it has been tested by tests for AMYDeleteView.

pbanaszkiewicz (changes requested):
@elichad I added some comments, mostly I'm concerned about the `post()` method used in 1 or 2 views. The rest is minor things.

amy/templates/dashboard/training_progress.html
elichad: `date` does not complain in that case - I think it returns an empty string. So then `firstof` falls back on `"No date provided"`. It's tested in `test_get_involved_details_not_provided`

amy/dashboard/views.py
elichad: Good point, made #2503

amy/templates/get_involved_form.html
elichad: Thanks for catching that, I assumed I had it on but I guess it was `black` handling the Python files!

amy/templates/get_involved_form.html
elichad: It can, though the `...embedded.html` file is a holdover from earlier experiments that I had meant to delete! Do you still think it would be better to use the `include` even though this template is fairly simple?

amy/dashboard/tests/test_get_involved_views.py
elichad: I can't find any tests for AMYDeleteView in our code - though I would similarly hope this case is covered by tests for the underlying Django DeleteView.
(I copied many of these tests over from `TestCRUDViews` in `test_training_progress.py` - this is one of those)

amy/dashboard/views.py
elichad: Ah, that's much better, thank you

amy/dashboard/views.py
elichad: It's no longer needed, but it was an addition to get the call to either `form_valid` or `form_invalid` to work.

amy/dashboard/views.py
elichad: I think the slight extra detail is useful given that this message appears at the top of the page and not in the 'Get Involved' section - though I have changed 'reviewed' to 'evaluated' in one of the other PRs for consistency.

I'll leave the 7-10 days part as it is for now - I don't anticipate us changing this in the near future.

elichad:
@pbanaszkiewicz I made some updates/comments, ready for your re-review.

amy/templates/get_involved_form.html
pbanaszkiewicz: @elichad If the "embedded" file is going away then there's no need to use `include` here.

amy/dashboard/tests/test_get_involved_views.py
pbanaszkiewicz: 405 is our custom mechanism for handling `GET` requests performed on this delete view. If we don't have tests specific for AMYDeleteView then it's okay to keep this test here.

amy/dashboard/tests/test_get_involved_views.py
```diff
@@ -21,14 +20,32 @@ def setUp(self):
...
+
+        # set up and log in as a trainee
+        self.user = Person(
+            username="trainee_alice",
+            personal="Alice",
+            family="Trainee",
+            email="alice_trainee@example.com",
+            is_active=True,
+        )
+        self.user.set_password("password")
```
pbanaszkiewicz, line 40: Alternatively you can use `Person.objects.create_user()` (method of `PersonManager`) which accepts password.

pbanaszkiewicz (approved):
Looks good!

PR #2494, opened 2023-07-28
[Emails] Prevent changes to emails in specific states

### LINKED ISSUE #2459: [Emails] Prevent changes to scheduled emails in specific states
As a system administrator, I want users to not be able to cancel a successfully run scheduled email, or I don't want them to reschedule a cancelled email, or I don't want to allow any changes to a locked email.

There may be more forbidden transitions than these to.

pbanaszkiewicz:
Fixes #2459.

Based on PR #2491.

This PR introduces whitelist of statuses the scheduled email has to be in in order to edit, reschedule or cancel it.

One of the interesting techniques used here is database-level locking connected with filtering for allowed statuses. This will ensure that an email is in proper status before and during the POST request.

### DIFF
--- a/amy/emails/models.py
+++ b/amy/emails/models.py
@@ -125,8 +125,6 @@ ScheduledEmailStatusActions = {
     "edit": [
         ScheduledEmailStatus.SCHEDULED,
         ScheduledEmailStatus.FAILED,
-        # note: will not re-send the email or affect emails already sent
-        ScheduledEmailStatus.SUCCEEDED,
     ],
     "reschedule": [
         ScheduledEmailStatus.SCHEDULED,
--- a/amy/emails/models.py
+++ b/amy/emails/models.py
@@ -111,12 +111,44 @@ class EmailTemplate(ActiveMixin, CreatedUpdatedMixin, models.Model):
 
 
 class ScheduledEmailStatus(models.TextChoices):
-    SCHEDULED = "scheduled"
-    LOCKED = "locked"
-    RUNNING = "running"
-    SUCCEEDED = "succeeded"
-    FAILED = "failed"
-    CANCELLED = "cancelled"
+    SCHEDULED = "scheduled"  # editing, cancelling allowed
+    LOCKED = "locked"  # nothing allowed
+    RUNNING = "running"  # nothing allowed
+    SUCCEEDED = "succeeded"  # editing allowed with note about re-sending
+    FAILED = "failed"  # editing, cancelling allowed
+    CANCELLED = "cancelled"  # allowed to re-schedule
+
+
+# List of states when specific actions are allowed. This is used in per-object
+# permissions to block or allow specific actions.
+ScheduledEmailStatusActions = {
+    "edit": [
+        ScheduledEmailStatus.SCHEDULED,
+        ScheduledEmailStatus.FAILED,
+        # note: will not re-send the email or affect emails already sent
+        ScheduledEmailStatus.SUCCEEDED,
+    ],
+    "reschedule": [
+        ScheduledEmailStatus.SCHEDULED,
+        ScheduledEmailStatus.FAILED,
+        ScheduledEmailStatus.CANCELLED,
+    ],
+    "cancel": [
+        ScheduledEmailStatus.SCHEDULED,
+        ScheduledEmailStatus.FAILED,
+    ],
+}
+
+
+# Displayed in scheduled email details view.
+ScheduledEmailStatusExplanation = {
+    ScheduledEmailStatus.SCHEDULED: "Scheduled to be sent",
+    ScheduledEmailStatus.LOCKED: "Locked for sending; worker is processing it",
+    ScheduledEmailStatus.RUNNING: "Sending in progress",
+    ScheduledEmailStatus.SUCCEEDED: "Sent successfully",
+    ScheduledEmailStatus.FAILED: "Sending failed; worker will re-try soon",
+    ScheduledEmailStatus.CANCELLED: "Sending cancelled",
+}
 
 
 class ScheduledEmail(CreatedUpdatedMixin, models.Model):
--- a/amy/emails/tests/test_controller.py
+++ b/amy/emails/tests/test_controller.py
@@ -207,6 +207,39 @@ class TestEmailController(TestCase):
         )
         self.assertEqual(latest_log.author, person)
 
+    def test_reschedule_cancelled_email(self) -> None:
+        # Arrange
+        old_scheduled_date = datetime(2023, 7, 5, 10, 00, tzinfo=UTC)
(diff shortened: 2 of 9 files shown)


pbanaszkiewicz:
@elichad This is ready for your review.

elichad:
@pbanaszkiewicz this PR contains a lot of the commits from #2491, could you rebase it?

pbanaszkiewicz:
@elichad Whoops, I completely missed that. Rebased and ready for review. Thanks!

pbanaszkiewicz:
@elichad Can you review this PR? Thanks in advance!

amy/templates/emails/scheduled_email_reschedule.html
```diff
@@ -13,6 +13,12 @@ <h1>
 
 {% include "emails/scheduled_email_header_details.html" with scheduled_email=scheduled_email %}
 
+{% if scheduled_email.state == 'cancelled' %}
+<div class="alert alert-warning" role="alert">
+  <strong>Warning</strong>: if you reschedule a cancelled email, it will become scheduled again and may be sent immediately.
```
elichad, line 18:
```suggestion
  <strong>Warning</strong>: if you reschedule a cancelled email, it will become scheduled again and will be sent immediately if the scheduled time is in the past.
```

amy/emails/views.py
```diff
@@ -143,6 +155,15 @@ class ScheduledEmailUpdate(OnlyForAdminsMixin, FlaggedViewMixin, AMYUpdateView):
...
     object: ScheduledEmail
 
+    # Will lock this object in the database for the duration of the request.
+    # Most specifically, we want to lock it when we're saving the form. This way the DB
+    # helps us make sure the data is consistent.
+    # Additionally, we're limiting the queryset to only those objects that can be edited
+    # (see ScheduledEmailStatusActions).
+    queryset = ScheduledEmail.objects.filter(
+        state__in=ScheduledEmailStatusActions["edit"]
+    ).select_for_update()
```
elichad, line 165: I wondered if this would lock the object as soon as the user navigates to the Update page, but this appears not to be the case (I assume I misunderstood how `select_for_update()` works):
In one tab I opened the Update view for a scheduled email, then in another tab I cancelled that email. When I submitted the update, I got a 404 error because the email can no longer be edited.

Could we add a check during the form validation stage for these views that shows an error to the user if the object no longer has the right status for the operation? (can be a separate issue)

elichad (approved):
Sorry for the long delay. This looks good to me! I've added a couple of minor comments and one suggestion for a new issue

amy/emails/models.py
```diff
@@ -111,12 +111,44 @@ def get_absolute_url(self) -> str:
...
+
+
+# List of states when specific actions are allowed. This is used in per-object
+# permissions to block or allow specific actions.
+ScheduledEmailStatusActions = {
+    "edit": [
+        ScheduledEmailStatus.SCHEDULED,
+        ScheduledEmailStatus.FAILED,
+        # note: will not re-send the email or affect emails already sent
+        ScheduledEmailStatus.SUCCEEDED,
```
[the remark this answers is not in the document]
pbanaszkiewicz: I admit it's debatable. I wanted to not prevent user from amending the email just because I may not see the use case they may have.

@maneesha Once email is sent successfully, should users be able to edit it? This will not affect the email whatsoever, including re-sending it.

amy/emails/tests/test_controller.py
```diff
@@ -207,6 +207,39 @@ def test_reschedule_email(self) -> None:
...
+        signal = "test_email_template"
+        EmailTemplate.objects.create(
+            name="Test Email Template",
+            signal=signal,
+            from_header="workshops@carpentries.org",
+            cc_header=["team@carpentries.org"],
+            bcc_header=[],
+            subject="Greetings {{ name }}",
+            body="Hello, {{ name }}! Nice to meet **you**.",
+        )
```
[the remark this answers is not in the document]
pbanaszkiewicz: Fixed, thanks!

amy/emails/tests/test_views.py
```diff
@@ -419,6 +435,74 @@ def test_view(self) -> None:
...
+                    ScheduledEmailStatus.FAILED,
+                    ScheduledEmailStatus.SUCCEEDED,
+                ]
+            ]
+        )
+
+        # Act
+        results = queryset.all()
+
+        # Assert - none of the defined scheduled emails can be retrieved with this query
```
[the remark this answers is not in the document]
pbanaszkiewicz: Good catch, I'm fixing these mistakes.

amy/templates/emails/scheduled_email_reschedule.html
pbanaszkiewicz: I'll use your suggestion, but it doesn't capture exactly what I had in my mind. The email can be sent (almost) immediately if the scheduled time is in the past or very close to current time.

amy/emails/views.py
pbanaszkiewicz: `select_for_update()` is a Django way of creating a `SELECT ... FOR UPDATE` PostgreSQL query. It's one of the methods of per-row explicit locking: https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS

This locks the row for the duration of a single transaction, and I wanted to use this because there's a risk someone will be updating an entry that simultaneously was fetched by the email worker and made "locked" (as in the scheduled email status).

You will have hard time reproducing this one on your local machine, the best way is to have a command that spawns multiple clients trying to do one single edit. I've had some very bad experience with this kind of behavior which resulted in a race condition that was difficult to catch.


> Could we add a check during the form validation stage for these views that shows an error to the user if the object no longer has the right status for the operation? (can be a separate issue)

That is possible, but I don't know how (yet). I'll add a low-priority issue for this.

pbanaszkiewicz: Ticket created: https://github.com/carpentries/amy/issues/2531

pbanaszkiewicz:
@elichad updated, now I'm waiting for feedback from @maneesha before merging.

amy/templates/emails/scheduled_email_reschedule.html
elichad: Makes sense. I was thinking only about capturing the part of the behaviour that might be unexpected

amy/emails/models.py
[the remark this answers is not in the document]
pbanaszkiewicz: @maneesha decided to block editing.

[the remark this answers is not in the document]
maneesha: We should actually have enforcement that an email **can not** be edited after it has been sent, as this would create a situation where what is in AMY is different than what was actually set.

PR #2496, opened 2023-07-31
Update curriculum links

maneesha:
Updates curriculum links that broke (with trailing `/` or `/index.html`) after new lesson template transition.


elichad (approved):
LGTM, thanks @maneesha!

PR #2500, opened 2023-08-01
Set permissions for Get Involved views

### LINKED ISSUE #2493: Set permissions for Get Involved CRUD views
- [ ] trainee cannot edit/delete their submission after it's been evaluated
- [ ] trainee cannot add a second submission unless they have been asked to repeat
- [ ] trainee cannot access submissions belonging to other trainees
- [ ] trainee cannot access TrainingProgresses for steps other than Get Involved

Essential for 4.2 release.

elichad:
Fixes #2493. Requires #2492 to be merged first.


amy/dashboard/views.py
```diff
@@ -245,57 +245,135 @@ def training_progress(request):
...
+        num_existing = existing_progresses.count()
+        if num_existing == 0:
+            return True
+        else:
+            num_asked_to_repeat = existing_progresses.filter(state="a").count()
+            if num_asked_to_repeat == num_existing:
+                return True
+            else:
+                return False
+
```
pbanaszkiewicz, line 274: This should become part of the form validation, for example in the `clean()` method. Combined with no use of `post()` to handle form valid/non-valid check (see my review of #2492), this will simplify the view.

amy/dashboard/views.py
```diff
@@ -245,57 +245,135 @@ def training_progress(request):
...
+        if get_involved_form.is_valid() and user_may_create_submission:
             return self.form_valid(get_involved_form)
-
         else:
+            if not user_may_create_submission:
+                self.form_invalid_message = (
+                    "You already have an existing submission. "
+                    "You may not create another submission unless your previous "
+                    'submission has the status "asked to repeat."'
+                )
```
pbanaszkiewicz, line 306: This can also be handled by form `clean()` method:

```python
...

    def clean(self) -> None:
        if not self.user_may_create_submission():
            raise ValidationError("You already have an existing submission.")

```

amy/dashboard/views.py
```diff
@@ -245,57 +245,135 @@ def training_progress(request):
...
+            if (
+                progress.trainee == self.request.user
+                and progress.requirement == get_involved
+                and progress.state == "n"
+            ):
+                return True
+            else:
+                return False
+        except TrainingProgress.DoesNotExist:
+            return False
```
pbanaszkiewicz, line 340: There's a useful method `self.get_object()` available. It will get correct `TrainingProgress` from the database for you.
If you construct your queryset in a specific way, you'll disable users from getting others' training progresses. This is semantically more correct, as permissions involve e.g. groups and allowance to perform generic action (e.g. edit a training progress), not checking if selected training progress contains specific training requirement.

So I propose to move some of what you have in `has_permission` to queryset. The changes would look like so:

```python
class GetInvolvedUpdateView(
    LoginRequiredMixin, PermissionRequiredMixin, AMYUpdateView
):
    # we could add this anyway, so that in future it's easier to manage permissions
    # - but if users currently don't have these permissions then I'd comment them out
    # or remove `PermissionsRequiredMixin`
    permission_required = ["workshops.view_trainingrequirement", "workshops.change_trainingprogress"]

    # model = TrainingProgress  # not used with queryset

    form_class = GetInvolvedForm
    template_name = "get_involved_form.html"
    success_url = reverse_lazy("training-progress")
    success_message = "Your Get Involved submission was updated successfully."

    def get_queryset(self) -> QuerySet[TrainingProgress]:
        return TrainingProgress.objects.filter(
            trainee=self.request.user,  # you may need to save `self.request = request` in dispatch() or get/post
            requirement__name="Get Involved",
            state="n",
        )

    def get_context_data(self, **kwargs) -> dict:
        context = super().get_context_data(**kwargs)
        context["title"] = str("Update your Get Involved submission")
        return context
```

amy/dashboard/views.py
```diff
@@ -245,57 +245,135 @@ def training_progress(request):
...
```
pbanaszkiewicz, line 376: The same thing as with `GetInvolvedUpdateView`.

pbanaszkiewicz (changes requested):
Is the order of commits correct?

"[add permission tests](https://github.com/carpentries/amy/pull/2500/commits/11c4bb629753193424eb93317d3a4a3b27fb923f)" comes before "[Set up permissions and extra tests for views [](https://github.com/carpentries/amy/pull/2500/commits/8aaaf4c43c25cf9a5554c699be213db24bb0c133)https://github.com/carpentries/amy/issues/2493[]](https://github.com/carpentries/amy/pull/2500/commits/8aaaf4c43c25cf9a5554c699be213db24bb0c133)"

As for the permissions, I suggest a different approach by using a queryset. This will keep permissions semantically separate from checking object's field values.

I didn't look at the tests, but I suspect they may require some adjustments after `get_queryset` is introduced. Perhaps this method can be then tested in a unit-test fashion, e.g.:

```
def test_get_queryset():
    # Arrange
    request = RequestFactory.get("/")
    view = GetInvolvedUpdateView(request)

    # Act
    queryset = view.get_queryset()

    # Assert
    ...
```

We have a few tests like this - they're used to test outside of `self.client` and even don't require extensive data preparation like tests inheriting from `TestBase`.

amy/dashboard/views.py
```diff
@@ -245,57 +245,135 @@ def training_progress(request):
...
-            data=request.POST, instance=base_training_progress
-        )
+        # select data specifically to prevent user from submitting extra args
+        data = {
+            "involvement_type": request.POST.get("involvement_type"),
+            "date": request.POST.get("date"),
+            "url": request.POST.get("url"),
+            "trainee_notes": request.POST.get("trainee_notes"),
+        }
+        get_involved_form = GetInvolvedForm(data=data, instance=base_training_progress)
```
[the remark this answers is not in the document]
elichad: I was wondering if a malicious actor could post extra args and update the state of their TrainingProgress or something, but thinking about it, I am sure that this must be covered in Django base behaviour (and tests I put in confirm this)

elichad:
@pbanaszkiewicz I implemented your suggestions, thank you! Ready for a re-review.

Note that 83dc220 is a refactoring commit that combines the `Test...View` and `Test...ViewPermissions` classes for each view to clean things up, so you might prefer to look at the other commits separately to that one to understand all the changes.

amy/dashboard/tests/test_get_involved_views.py
```diff
@@ -125,45 +125,43 @@ def test_create_view_works(self):
...
-    def test_update_view_loads(self):
         # Act
-        rv = self.client.get(reverse("getinvolved_update", args=[self.progress.pk]))
+        rv_get = self.client.get(reverse("getinvolved_add"))
+        rv_post = self.client.post(reverse("getinvolved_add"), data)
 
         # Assert
-        self.assertEqual(rv.status_code, 200)
+        self.assertEqual(rv_get.status_code, 302)
+        self.assertEqual(rv_post.status_code, 302)
```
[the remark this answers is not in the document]
pbanaszkiewicz: I found other similar tests here that don't check redirection URL for unauthorized users.

pbanaszkiewicz (approved):
Looks good. I left minor comment about a few tests.

PR #2504, opened 2023-08-04
Add cancel button and fix validation bug on Get Involved form

elichad:
Form was raising a server error if submitted blank, because the validation on `trainee_notes` was expecting `involvement_type` to be set.
Note that `involvement_type` is always required in `TrainingProgress` creation, but it has `required=False` set on the form. This is because at least one _other_ field will always be required, but this is only determined at validation time, so showing an asterisk on only one field and not the rest doesn't quite represent how the form works.


pbanaszkiewicz (approved):
LGTM

PR #2505, opened 2023-08-04
[#2426] Hide Anymail errors

### LINKED ISSUE #2426: Errors in forms on test server
When filling out https://test-amy2.carpentries.org/forms/request_training/ 
we get the error:

```
AnymailRequestsAPIError at /forms/request_training/
Mailgun API response 401 (Unauthorized): 'Forbidden'
```

While the email should not go out, we should still be able to create forms in the test environment.

pbanaszkiewicz:
This fixes #2426 by hiding Anymail errors.
The original bug report is caused by test-amy2 attempting to send email without authorization. There's a couple of ways to fix this:
1. disable sending emails in test-amy2 (through AMY_LIVE_EMAIL env variable)
2. catch the exceptions where they are caused (this is the way I chose)
3. set up test-amy2 in a way that it can send real emails
4. switch to sandbox/test-only mode in test-amy2 + mailgun (probably solution that would take the most time).

I wanted to fix this issue quickly, and I didn't want to disable sending emails altogether since we may want to do this in future. So I decided on solution no. 2.

### DIFF
--- a/amy/workshops/base_views.py
+++ b/amy/workshops/base_views.py
@@ -1,6 +1,7 @@
 from smtplib import SMTPException
 from typing import Optional
 
+from anymail.exceptions import AnymailRequestsAPIError
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.mixins import PermissionRequiredMixin
@@ -342,7 +343,7 @@ class AutoresponderMixin:
 
         try:
             email.send()
-        except SMTPException as e:
+        except (SMTPException, AnymailRequestsAPIError) as e:
             if not fail_silently:
                 raise e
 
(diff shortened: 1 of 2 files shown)


amy/workshops/base_views.py
```diff
@@ -342,7 +343,7 @@ def autoresponder(self, form, fail_silently=True):
 
         try:
             email.send()
-        except SMTPException as e:
+        except (SMTPException, AnymailRequestsAPIError) as e:
```
pbanaszkiewicz, line 346: I wasn't able to reproduce this bug in tests, because Django was switching to a `locmem` email backend even if I explicitly enabled Anymail through environmental variables.

Thus I didn't include any regression tests for this bug.

elichad (approved):
This looks fine to me for now. 
I assume we will come back to the topic of sending emails from test-amy when the new email system is ready.

elichad:
Actually, are there any cases where this exception might be raised in production?

pbanaszkiewicz:
> Actually, are there any cases where this exception might be raised in production?

@elichad Possibly if our Mailgun account expires, or they have an outage or something.

PR #2510, opened 2023-08-09
Allow admins to submit/update training progress with non GitHub URL

### LINKED ISSUE #2507: Error when "Ask to Repeat" due to non-Carpentries repo
See this trainee as an example: https://test-amy2.carpentries.org/workshops/person/74/
They have one "Get Involved" step that is to a non-Carpentries repo.  When the admin tries to mark that as "Ask to Repeat" they get an error that it is a non-Carpentries repo. 
We want the admin to be able to mark this as "Ask to repeat" so that the trainee can go back and make the appropriate changes.

elichad:
Fixes #2507 by moving the relevant validation out of the TrainingProgress model and into the GetInvolvedForm.

Required due to submissions made before the rollout potentially not having GitHub URLs where they would now be required - admins may still need to update these to change the state or add notes.

Required for v4.2.

### DIFF
--- a/amy/api/tests/test_export.py
+++ b/amy/api/tests/test_export.py
@@ -557,7 +557,7 @@ class TestExportingPersonData(BaseExportingTest):
                 },
                 "involvement_type": {
                     "name": "GitHub Contribution",
-                    "display_name": "Submitted a contribution to a Carpentries repository",  # noqa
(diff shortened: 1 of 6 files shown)


pbanaszkiewicz (approved):
LGTM!

PR #2512, opened 2023-08-09
Update conditions for progress messages at top of page

### LINKED ISSUE #2508: Your Training Progress Top Text Validation
The text at the top of the 'Your training progress' page does not seem to be validating correctly. I have tried a few different combinations of progress status to see which messages would show up.

In this case, the message should not show up because training is failed:
image

If I remove the welcome session progress the message returns to the appropriate language:
image

Another case where this is showing incorrectly when the training is asked to repeat:
image

If I remove the demo progress the message returns to the appropriate language:
image

elichad:
Fixes #2508 .

More explanation in https://github.com/carpentries/amy/issues/2508#issuecomment-1671458935.

Required for v4.2 so we aren't displaying misleading information about checkout deadlines.

### DIFF
--- a/amy/dashboard/tests/test_instructor_dashboard.py
+++ b/amy/dashboard/tests/test_instructor_dashboard.py
@@ -116,6 +116,64 @@ class TestInstructorStatus(TestBase):
             "certification",
         )
 
+    def test_deadline_shown_when_training_passed(self):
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz (approved):
LGTM

PR #2515, opened 2023-08-10
Remove use of text-white to override text colour

elichad:
Fixes a contrast bug introduced by #2490 . The override is no longer needed as the text colour choice is handled better by Bootstrap under the new palette.

Current `develop` branch:
Button with cyan background and white text
This branch:
Button with cyan background and dark grey text


pbanaszkiewicz (approved):

pbanaszkiewicz:
@elichad I'm going to merge this as it has passed the code review, and because we have the deployment tomorrow. This is a small change and we can let it in at the last minute.

PR #2516, opened 2023-08-13
[Emails] Use django-flags for feature flag management

### LINKED ISSUE #2485: [Emails] Try another approach to feature flags
...Other than setting an environment value.

pbanaszkiewicz:
This fixes #2485 by replacing envvar-managed feature flag `EMAIL_MODULE_ENABLED` with [`django-flags`](https://cfpb.github.io/django-flags/).

To enable or disable `EMAIL_MODULE` feature flag `/admin/feature_flags/`.

~~WIP. Todo:~~
- [x] 1. Mention enabled feature flags in the top site banner
- [x] 2. Add view listing enabled feature flags
- [x] 3. Add option to enable or disable feature flags by the user
- [x] 4. add tests


pbanaszkiewicz:
@elichad Ready for your testing. At this moment I don't know if we want to display enabled feature flags or not.

pbanaszkiewicz:
@elichad I simplified enabling and disabling the feature flag - you can do it from UI now. Please review when you can.

config/settings.py
```diff
@@ -632,3 +650,37 @@
...
+# ------------
+FLAGS = {
+    # Enables instructor recruitment views; this should be enabled in all environments.
+    # The flag itself was not migrated from envvar to django-flags mechanism.
+    # "INSTRUCTOR_RECRUITMENT": [{"condition": "boolean", "value": True}],
+    # ------------
+    # Enables the new email module. It's only for logged-in users and it has to be
+    # enabled through `?enable_email_module` GET parameter or the same parameter stored
+    # in request session.
+    "EMAIL_MODULE": [
```
elichad, line 681: How would we enable this flag globally in production? 

I looked at the [django-flags usage guide](https://cfpb.github.io/django-flags/usage/) and saw that "Additional conditions can be added in the Django admin for any defined flag." I think that would be a good approach, so I tried setting a boolean True condition in the Django admin on my local instance, but it didn't seem to have any effect.

amy/workshops/tests/test_template_tags.py
```diff
@@ -0,0 +1,47 @@
...
+        condition1 = Condition(condition="session", value="test")
+        condition2 = Condition(condition="user", value="test")
+        condition3 = Condition(condition="boolean", value="test")
+        flag = Flag(name="TEST_FLAG", conditions=[condition1, condition2, condition3])
+        # Act
+        result = first_parameter_condition(flag)
+        # Assert
+        self.assertIsNone(result)
+
+    def test_url_strip_param_value(self):
```
elichad, line 33: Is there a case where a full url will be passed to this function? If so, should it also handle the case where multiple params are passed?
If it's just for stripping values from parameter conditions, the name could be changed to make that clearer

config/settings.py
```diff
@@ -632,3 +650,37 @@
...
         "SITE_BANNER_STYLE accepts only one of 'local', 'testing', 'production'."
     )
+
+# Feature Flags
+# -----------------------------------------------------------------------------
+# These flags are used to enable/disable features in AMY.
+# See https://cfpb.github.io/django-flags/ for more details.
+
+# ------------
+# The system for enabling or disabling feature flags by users themselves should only be
```
elichad, line 660: Could you flesh out this documentation into an actual page in the developer docs? It would be helpful to explain the different ways a flag can be set through the UI/URL using these conditions, and how the conditions should be updated during the course of development.

elichad (changes requested):
This is cool! Thanks for doing the work to set up django-flags, this definitely looks like an improvement on our current system.

Comments are mostly about clarity and documentation

amy/emails/utils.py
```diff
@@ -15,21 +16,22 @@
...
-    is disabled."""
-    return settings.EMAIL_MODULE_ENABLED is True
+@conditions.register("session")  # type: ignore
+def session_condition(value, request: HttpRequest, **kwargs):
+    """Additional condition for django-flags. It reads a specific value from
+    request session."""
+    return request.session.get(value, False)
 
 
 def feature_flag_enabled(func):
```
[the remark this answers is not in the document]
pbanaszkiewicz: Good idea. I'll rewrite it to become email module agnostic, and I'll move it to `workshops` app.

amy/templates/navigation.html
```diff
@@ -93,6 +94,7 @@
               <div class="dropdown-divider"></div>
               {% if user.is_superuser %}
               <a class="dropdown-item" href="{% url 'admin:index' %}">Django Admin</a>
+              <a class="dropdown-item" href="{% url 'feature_flags' %}">Feature flags</a>
```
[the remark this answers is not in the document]
pbanaszkiewicz: I'll change visibility for non-admins, but I'd argue that we should be able to tell easily what features are enabled in production. For example, we could partially deploy some feature "in the dark" and after deployment we would make sure that feature is still disabled.

config/settings.py
pbanaszkiewicz: It works for me with this setting:
image

I'm even unable to disable the flag using my own feature flags page 😄

amy/workshops/tests/test_template_tags.py
pbanaszkiewicz: Would it sound better as `url_strip_query_string`?

amy/templates/base.html
```diff
@@ -33,9 +33,19 @@
   <body>
 
     {% if SITE_BANNER_STYLE == "local" %}
-    <section class="banner banner-local">This is a local version of AMY.</section>
+    <section class="banner banner-local">
+      This is a local version of AMY.
+      {% if user.is_admin %}
+      You have {{ FEATURE_FLAGS_ENABLED|length }} feature flag{{ FEATURE_FLAGS_ENABLED|length|pluralize }} <a href="{% url 'feature_flags' %}">enabled</a>.
```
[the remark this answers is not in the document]
pbanaszkiewicz: @elichad I'm going to show this line in the banner unconditionally for the users. The link will require logging in, so in future if you want to allow unauthorized users to participate in testing, there may be additional work required to enable that view for them, or another approach to enabling the feature flag will have to be taken.

amy/workshops/tests/test_template_tags.py
pbanaszkiewicz: Realistically this function should not be used to strip URLs out of the parameter value. I'll change the name to `parameter_strip_value` - the `parameter` sounds generic but in the context of specific [django flag condition](https://cfpb.github.io/django-flags/conditions/#parameter) it should be clearer.

pbanaszkiewicz:
@elichad Thank you, I reworked this PR after your feedback. Ready for round 2!

config/settings.py
elichad: Ah, so it does. I had set "Required" to be true, and that prevented the flag from being enabled. I'm not sure why.

elichad (approved):

amy/templates/navigation.html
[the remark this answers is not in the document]
pbanaszkiewicz: Just hiding the "Enable" and "Disable" links may not be enough, as the query parameters enabling the feature flags may still be active.

In my opinion, we should have a process for permanently disabling or enabling a feature flag after a production release. We can use django-admin to add boolean condition to a flag like here:
https://github.com/carpentries/amy/pull/2516#discussion_r1302105008
If we add boolean "false", the flag will not be able to activate.

I'll create an issue to continue discussion there.

amy/templates/base.html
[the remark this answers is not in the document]
pbanaszkiewicz: > For membership codes, I was thinking to just omit the anonymous condition from the flag, because that should then allow the flag to be set with a URL parameter even without access to the view (right?).

Yes, it would work like that.

config/settings.py
pbanaszkiewicz: Honestly I don't know that either. I think all conditions marked as required must be true, or only one non-required must be true for the flag to be enabled. https://cfpb.github.io/django-flags/usage/#defining-flags

PR #2517, opened 2023-08-14
Make explicit where Instructors can include accommodations they need when signing up for workshop

### LINKED ISSUE #2436: Allow Instructors to indicate accommodations they need to teach effectively
Suggested by @bkmgit in #2435.

An Instructor can currently provide information about accommodations in the notes when expressing interest in teaching a workshop, but we don't explicitly ask for this anywhere. We could make this part of the instructor's AMY profile to avoid them having to provide the same information repeatedly.

elichad:
Addresses but does not close #2436.

### DIFF
--- a/amy/dashboard/forms.py
+++ b/amy/dashboard/forms.py
@@ -302,7 +302,9 @@ class SignupForRecruitmentForm(forms.ModelForm):
         required=False,
         widget=forms.Textarea,
         label="Your notes",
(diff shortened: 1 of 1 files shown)


elichad:
@maneesha are you happy with this wording?

froggleston (approved):
LGTM

maneesha (approved):

PR #2518, opened 2023-08-15
Make GitHub organization check case insensitive

elichad:
Fixes a bug in v4.2 reported by @karenword in Slack:

> a trainee is reporting that this link is throwing an error claiming it is not associated with a Carpentries repo: `https://github.com/LibraryCarpentry/lc-r/issues/<id>` .

This arises because we have the organization in lowercase (`librarycarpentry`) in our code. Github organization URLs are case insensitive, so we should accept both.


pbanaszkiewicz (approved):
LGTM although I prefer `.lower()` over `.casefold()` (mostly because I'm unfamiliar with the latter).

PR #2519, opened 2023-08-15
Update release procedure documentation

elichad:
I tried to capture all the extra undocumented steps we took.

I've also created a new file where we can add manual deployment steps that should be completed for a particular release.

### DIFF
--- a/docs/procedures.md
+++ b/docs/procedures.md
@@ -36,6 +36,8 @@ of new features in case they would have to be released to production
 5. Write to <amy@lists.carpentries.org> mailing list.
    The suggested subject of the new thread is "[AMY] New release v2.X.Y".
 
+6. Post release announcement on Slack with user-appropriate notes on the important changes.
+
 ## Release Procedure
 
 We assume that you want to release AMY v2.X.Y.
@@ -120,4 +122,12 @@ What are the changes:
 
 ## Deployment procedure using Ansible
 
-Moved to relevant repository `README.md`.
+1. Back up the database through the AWS console (RDS > Databases > prod-amy > Actions (top right) > Take snapshot).
+
+2. Check for pending maintenance through the AWS console (RDS > Databases > prod-amy > Maintenance and Backups (below Summary section)) and complete it if needed.
+
+3. Complete any [Manual Deployment Steps](./manual_deployment_steps.md) noted for before deployment of this release.
+
+4. Run the Ansible deployment as described in the relevant repository `README.md`.
+
+5. Complete any [Manual Deployment Steps](./manual_deployment_steps.md) noted for after deployment of this release.
--- a/docs/procedures.md
+++ b/docs/procedures.md
@@ -61,34 +61,38 @@ Execute the following commands on your local machine, not production.
         $ git checkout main
         $ git pull origin main
 
-4.  Merge `develop` into `main` branch (be careful, as there are sometimes conflicts that need to be manually resolved):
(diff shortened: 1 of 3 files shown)


docs/procedures.md
```diff
@@ -129,4 +119,12 @@ What are the changes:
 
 ## Deployment procedure using Ansible
 
-Moved to relevant repository `README.md`.
+1. Back up the database through the AWS console (RDS > Databases > prod-amy > Actions (top right) > Take snapshot).
```
pbanaszkiewicz, line 122: I'm not sure how I feel about publishing our production cluster name. Maybe a more generic instruction would be better, e.g. `RDS -> Databases -> AMY database cluster -> Actions (top right) -> Take snapshot`?
This is also more futureproof (cluster name can change).

pbanaszkiewicz: We should have a naming scheme, too. As you saw, naming the snapshot is not easly.

What about "v4.2.0_YYYYMMDD" scheme? Do you remember if it allowed dots and underscores in the snapshot name?

pbanaszkiewicz (approved):
Provided minor comments, otherwise looks good to me.

docs/procedures.md
froggleston: From the AWS snapshot field hint: `Snapshot identifier is case insensitive, but stored as all lower-case, as in "mysnapshot". Cannot be null, empty, or blank. Must contain from 1 to 255 alphanumeric characters or hyphens. First character must be a letter. Cannot end with a hyphen or contain two consecutive hyphens.`

docs/procedures.md
```diff
@@ -46,28 +45,21 @@ Execute the following commands on your local machine, not production.
...
 
         $ git remote -v
-        origin	git@github.com:chrismedrela/amy.git (fetch)
-        origin	git@github.com:chrismedrela/amy.git (push)
-        upstream	git@github.com:carpentries/amy.git (fetch)
-        upstream	git@github.com:carpentries/amy.git (push)
+        origin	git@github.com:carpentries/amy.git (fetch)
+        origin	git@github.com:carpentries/amy.git (push)
 
 3.  Make sure your local `develop` and `main` branches are up to date:
```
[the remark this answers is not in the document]
froggleston: Good spot - we should have some guidance here about what release branch naming is/should look like. From our discussions, it looks like `develop` will also end up being used in our DEV testing workflow, rather than UAT, so it makes sense to update the docs to reflect this now?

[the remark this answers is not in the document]
elichad: I've updated this to use a release branch instead

docs/procedures.md
```diff
@@ -61,34 +61,38 @@ Execute the following commands on your local machine, not production.
         $ git checkout main
         $ git pull origin main
 
-4.  Merge `develop` into `main` branch (be careful, as there are sometimes conflicts that need to be manually resolved):
+4. Create a release branch `release/vX.Y.Z` from `develop`. Major/minor release branches should be based on the `HEAD` of `develop`, but bugfix releases may be based on older commits to avoid including features intended for the next major/minor release.
```
pbanaszkiewicz, line 64: > on older commits...

"... such as previous release branch.

For more details on managing branches, see https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow"

Some ideas on extending this sentence.

pbanaszkiewicz (approved):
Looks good to me.

PR #2521, opened 2023-08-17
[Emails] Disable some notifications for non-admin users

### LINKED ISSUE #2498: [Emails] Disable notification toast for user-triggered emails
As a training/instructor user, I don't want to see automation email toast notification.

This applies currently only to `instructor_signs_up_for_workshop`, but in future there may be other instances that should not show the notification.

pbanaszkiewicz:
This fixes #2498.

Two notifications generated by email actions (`messages_missing_template` and `messages_action_scheduled`) now add extra tag `"only-for-admins"`.

New condition in `base.html` checks for this tag when displaying messages. If the tag is present on the message, then that message is shown only for administrator users. If the tag is not present, then the message is shown as usual.

Writing unit tests for this was a little bit tricky, as I wasn't able to inject adding messages to a view. So instead I moved the logic to the template tag and tested it instead of testing the view output.

### DIFF
--- a/amy/dashboard/templatetags/notifications.py
+++ b/amy/dashboard/templatetags/notifications.py
@@ -1,6 +1,7 @@
 from typing import cast
 
 from django import template
+from django.conf import settings
 from django.contrib.messages.storage.base import Message
 from django.http import HttpRequest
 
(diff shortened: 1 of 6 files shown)


pbanaszkiewicz:
@elichad Ready for your review.

PR #2523, opened 2023-08-21
Remove question about learner count from WRF

### LINKED ISSUE #2308: Remove question about learner count from workshop request form
Following conversation with @danielle06 we would like to completely remove the question on the workshop request and inquiry forms about "Anticipated number of attendees."

elichad:
Fixes #2308 by completely removing this field from the DB (issue comments say that this data is not used)


pbanaszkiewicz (approved):
LGTM. Interesting case with the alter field before it is removed through migration.

danielle06:
@elichad , saw the original pr was closed. Wasn't sure if that meant that the count field was able to be removed - double checking because it still shows on the form.

elichad:
@danielle06 Sorry for the confusion! The PR is closed as the development is complete, but this change hasn't been released yet. The question should disappear from the form after the next release (v4.3, planned for late October). Is that okay?

danielle06:
@elichad , yes! No worries and thank you for the explanation. I wanted to make sure I was understanding correctly :D

Issue #2380, opened 2023-04-12
Use member registration code on all workshop request forms


maneesha:
~This may need to be broken out into several issues.~  *DONE*

On the workshop request forms, we would like to add the following question:

> Are you affiliated with a Carpentries member organisation?

- If no, proceed as normal.
- If yes, the user should be asked for a registration code.  This should be verified to be a code associated with a current membership.  
- If it is not a currently valid code,  give the user a warning (example text below). Allow them to proceed. Include the warning text in the confirmation email that the user receives.
- Add a filter to the workshop request view to filter for requests with invalid codes. (See new issue #2398)
- When that workshop request form is accepted and turned into an event, the membership associated with that code should automatically fill in the "membership" field. (See new issue #2399)

maneesha:
Suggested warning text:

> It seems you have entered an invalid member code.  Please make sure you have the current year's code and do not have any typos.  If the code looks correct, please proceed to submit your form. A member of our team will follow up with you if there are any problems that require your attention.

maneesha:
For workshop requests, the code should be valid as of 60 days **before** the membership start date, and end as of the actual membership end date.

elichad:
Discussed in depth with WAT on 18 October. Summary below.

1. We can remove the "are you affiliated with a member org" question - as the member code isn't a required field, we can just have that by itself. That said, we could dynamically show/hide some questions based on the affiliation question. I'd rather avoid this if we can because dynamic things add more room for error.
2. would be convenient to auto-fill the institution after a code is entered, but this is tricky to navigate when members may share /donate workshops with other institutions
3. if a user enters an institution which has a membership, but doesn't enter the member code, create a way of flagging this to admins. This might happen where someone isn't in contact with membership contacts at their institution, or just has forgotten their code.
4. agreed that code validation should be hard, not soft - users should know who to speak to to check their code; we should make sure our error messaging is clear enough for member contacts to troubleshoot.
5. member code question should not be asked on SOW form or inquiry form, as we don't use that info and there's more scope for confusion (users submitting these forms may be completely unaware of the membership program)
6. extra task: Send email to member contact when a workshop request is submitted using their code. This reduces the chance of someone taking a donated/shared code and requesting more workshops than they were meant to without the knowledge of the lead contact.

PR #2532, opened 2023-09-18
Add membership code & validation to workshop requests

elichad:
Progress on #2380 but doesn't close it.

I deliberately haven't implemented the validation for workshop inquiries or self-organised submissions yet as I'd like @pbanaszkiewicz's thoughts on the current approach before I copy the implementation to the other forms (I have been thinking about if/how to unify the form validation for all three with some sort of CommonRequestBaseForm class)


amy/extrequests/forms.py
```diff
@@ -440,6 +453,31 @@ def institution_label_from_instance(obj):
         essentially works just like `Model.__str__`."""
         return "{}".format(obj.fullname)
 
+    @feature_flag_enabled("ENFORCE_MEMBER_CODES")
+    def clean_membership_info(self, request: HttpRequest) -> dict:
```
pbanaszkiewicz, line 457: If form method is called `clean_xyz`, then I assume it is run for validation of a single field `xyz`, which is not a case here. I'd rename to `clean_member_code` if we have access to `self.cleaned_data` in per-field validation, otherwise rename to `validate_member_code`.

If we do have access to `self.cleaned_data` in per-field validation then calling this method in `clean()` will not be necessary.

amy/extrequests/forms.py
```diff
@@ -440,6 +453,31 @@ def institution_label_from_instance(obj):
...
+        code = self.cleaned_data.get("member_code", "")
+        if affiliation in ["yes", "unsure"] and code:
+            # ensure that code belongs to a membership
+            try:
+                Membership.objects.get(registration_code=code)
+            except Membership.DoesNotExist:
+                errors["member_code"] = ValidationError(
+                    "This code is invalid. "
+                    "Please contact your Member Affiliate to verify your code."
+                )
```
pbanaszkiewicz, line 469: This is a different message than suggested here: https://github.com/carpentries/amy/issues/2380#issuecomment-1505763664
> It seems you have entered an invalid member code. Please make sure you have the current year's code and do not have any typos. If the code looks correct, please proceed to submit your form. A member of our team will follow up with you if there are any problems that require your attention.

amy/workshops/models.py
```diff
@@ -2876,6 +2876,32 @@ class CommonRequest(SecondaryEmailMixin, models.Model):
         verbose_name="Department/School/Library affiliation (if applicable)",
     )
 
+    MEMBER_AFFILIATION_CHOICES = (
+        ("yes", "Yes"),
+        ("no", "No"),
+        ("unsure", "I'm not sure"),
+    )
+    member_affiliation = models.CharField(
```
pbanaszkiewicz, line 2884: Why is this field needed at all? Will it be used for something else than making `member_code` required or not?

Because if we want to have a soft validation, we don't really need this field.

amy/workshops/models.py
```diff
@@ -2876,6 +2876,32 @@ class CommonRequest(SecondaryEmailMixin, models.Model):
...
+        max_length=STR_MED,
+        null=False,
+        default="no",
+        blank=True,
+        choices=MEMBER_AFFILIATION_CHOICES,
+        verbose_name="Are you affiliated with a Carpentries member organization?",
+        help_text="This may be the same as your institution above, "
+        "or another institution.",
+    )
+    member_code = models.CharField(
```
pbanaszkiewicz, line 2894: What are the values to be used here? Can you provide some samples?

What is the "secrecy level" of this code, can this leak to third parties and cause harm?

amy/extrequests/forms.py
```diff
@@ -440,6 +453,31 @@ def institution_label_from_instance(obj):
...
 
+    @feature_flag_enabled("ENFORCE_MEMBER_CODES")
+    def clean_membership_info(self, request: HttpRequest) -> dict:
+        errors = dict()
+        affiliation = self.cleaned_data.get("member_affiliation")  # yes/no/unsure
+        code = self.cleaned_data.get("member_code", "")
+        if affiliation in ["yes", "unsure"] and code:
+            # ensure that code belongs to a membership
+            try:
+                Membership.objects.get(registration_code=code)
```
pbanaszkiewicz, line 464: I assume here you want to add other validation in future, for example like "is membership active"?

amy/extrequests/forms.py
```diff
@@ -440,6 +453,31 @@ def institution_label_from_instance(obj):
...
+        elif affiliation == "yes" and not code:
+            errors["member_code"] = ValidationError(
+                "This field is required if you selected 'Yes' above."
+            )
+        elif affiliation == "no" and code:
+            errors["member_code"] = ValidationError(
+                "This field must be empty if you selected 'No' above."
+            )
+
+        return errors
```
pbanaszkiewicz, line 479: I don't think we have a place in AMY where we do "soft validation". Here are my three ideas how to implement it. It's definitely not easy.

**Idea 1: code validation as background process in the browser**

Once the code is filled out by the user, we'd make a request to the API to see if there's a membership matching criteria.

With JS we'd dynamically show warnings in the form for the user, based on the responses from the server.

This potentially exposes the API to third parties as users don't have to be logged in in these forms, so some throttling should be implemented.


**Idea 2: dynamic field requesting user to confirm the message popup**

After user provides the code and there's something wrong with it, we amend the form by adding a new required checkbox field next to that member code.

This is clever because it uses Django default behavior to "break" form validation, but at the same time it's easy for the user to correct the form without changing member code - thus soft validation.

The downside is that altering a form like that is not an easy task.

**Idea 3: remembering form state between requests**

This idea uses Django session storage to remember if user say the warning message.

When first attempt on validating the form throws the error like in your current implementation, the server should also store some variable in the session.

This variable, when present in the session, would then change the member code validation and prevent throwing the same error twice. Finally the variable should be removed from the server session.

pbanaszkiewicz (commented):
@elichad I provided some questions and suggestions.

amy/extrequests/forms.py
elichad: Yes, this will come later #2533

amy/workshops/models.py
elichad: These values are manually created by using a short name for the institution followed by two digits, e.g. `alpha55` for Alpha Organization, or `seu78` for Some Example University (SEU). The number of characters in the name part of the code can vary - the longest code I've seen is 9 characters total, but the corresponding DB field on Membership model permits up to 40.
For membership renewals, the trailing digits are changed, with no pattern. 

This code is only shared by us with designated member contacts, but as they will pass it on to those booking workshops/training there is a risk of it leaking that way. If a malicious actor got hold of a code they would theoretically be able to book workshops/training seats with it without the member contact knowing (unless the member contact was carefully validating the numbers in their quarterly emails).
But note that sometimes members will donate or share workshops/seats to/with other institutions, so just checking the code matches the institution in this form won't work

amy/workshops/models.py
elichad: The intent is for the `member_code` to eventually be displayed dynamically if the user marks Yes to the affiliation question. But depending on how we decide to approach validation, it's true that we may not need this question in the end.

amy/extrequests/forms.py
elichad: Per-field validators do have access to `self.cleaned_data`, but the `request` argument is needed here for `@feature_flag_enabled` to work, so I think this has to be called manually from `clean()`.

This raises a side question: do we plan to remove feature flags once they're enabled and stable in production? If so I guess I should plan for some refactoring/cleanup at the end of this project.

amy/extrequests/forms.py
elichad: Thanks for all these - I thought of idea 2 myself as well. I think idea 2 would be best as then it makes it easy to flag to the admins when the code needs to be double-checked.

amy/extrequests/forms.py
elichad: The current error is what I was given by the workshops team more recently in Slack along with the request for hard validation of the code.
I think this message will need updating when extra validation is added anyway, to include possible reasons for the error, so I'll include it as part of my conversations with them.

amy/extrequests/forms.py
elichad: Created #2535 from this comment

elichad:
@pbanaszkiewicz Thanks for your comments. I will meet with the workshops team next week to discuss some of these points (the team lead is away this week).
In the meantime, I would like to start work on some other tasks which will depend on the model updates in this PR. Would you be okay approving this as a basic implementation, knowing that the details of the validation/error messages will change later? If not, I'll split out the model updates into their own PR and come back to the validation parts when we've nailed them down better.

pbanaszkiewicz:
@elichad Yeah, I'm okay with iterative approach to this new feature.

pbanaszkiewicz (approved):

amy/extrequests/forms.py
pbanaszkiewicz: We could remove them in the long-term, but for now we can use the database to set the feature flag dynamically to always be enabled (we discussed this a few weeks back with one of my PRs).

PR #2534, opened 2023-09-18
Improve sending emails for EmailSendMixin

### LINKED ISSUE #2426: Errors in forms on test server

pbanaszkiewicz:
Silence the email errors in views using the EmailSendMixin.

This fixes #2426.

### DIFF
--- a/amy/extforms/views.py
+++ b/amy/extforms/views.py
@@ -79,7 +79,7 @@ class WorkshopRequestCreate(
     page_title = "Request a Carpentries Workshop"
     template_name = "forms/workshoprequest.html"
     success_url = reverse_lazy("workshop_request_confirm")
-    email_fail_silently = False
+    email_fail_silently = True
 
     autoresponder_subject = "Workshop request confirmation"
     autoresponder_body_template_txt = "mailing/workshoprequest.txt"
@@ -167,7 +167,7 @@ class WorkshopInquiryRequestCreate(
     page_title = "Inquiry about Carpentries Workshop"
     template_name = "forms/workshopinquiry.html"
     success_url = reverse_lazy("workshop_inquiry_confirm")
-    email_fail_silently = False
+    email_fail_silently = True
 
     autoresponder_subject = "Workshop inquiry confirmation"
     autoresponder_body_template_txt = "mailing/workshopinquiry.txt"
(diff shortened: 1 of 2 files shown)


pbanaszkiewicz:
@elichad ready for your review.

amy/extforms/views.py
```diff
@@ -79,7 +79,7 @@ class WorkshopRequestCreate(
     page_title = "Request a Carpentries Workshop"
     template_name = "forms/workshoprequest.html"
     success_url = reverse_lazy("workshop_request_confirm")
-    email_fail_silently = False
```
elichad, line 82: Is there a way to do this only in non-production environments?

pbanaszkiewicz: We can make condition based on the environment, just like in the case of top banner in test-amy or local env, but I'm not sure this is a good idea.

If the fail is silenced in test-amy then the forms should work (Maneesha's issue is fixed). In production we never had any issues with Mailgun, so this was never a problem.

elichad:
Pinging @maneesha / @froggleston - this PR fixes a problem where we can't submit workshop request forms on test-amy, but it does so by ignoring errors from sending these emails in general.
We haven't encountered these errors in production, but they could appear under rare circumstances (e.g. if the Mailgun server goes down).

What are your thoughts on taking this approach until the new email system is ready?

maneesha:
I am not comfortable with silent failures so I would like to see this ignored in test and local builds only.  Errors should be raised in production.  @pbanaszkiewicz  can you share why it is not a good idea to make it conditional based on environment?

pbanaszkiewicz:
@elichad @maneesha @froggleston I'm going to decline this PR. I changed a single deployment environment variable (https://github.com/carpentries/systems/commit/12539f7647e81d8a73e3159f05e4a0a69c18fbc7) that controls the system used for sending emails in test-amy2. The reason is that:

1. I don't want to keep conditions in code that change the way it works based on the environment (I want to keep staging and production parity)
2. We're not sending real emails in that environment.

PR #2541, opened 2023-10-09
[Emails] Refactor receivers into class-based actions

### LINKED ISSUE #2530: [Emails] Refactor signal receivers to simplify overhead
There's some overhead on every receiver. It looks like they could use a base receiver class with some overridden methods.

pbanaszkiewicz:
This fixes #2530 by moving all existing signals/receivers into class-based actions.

Three class based actions currently exist:

1. A base one which just adds new scheduled emails
2. An update action: it fetches exactly one scheduled email and updates its fields
3. A cancel action: it fetches multiple scheduled emails and cancels all of them.

### DIFF
--- a/amy/emails/actions/admin_signs_instructor_up_for_workshop.py
+++ b/amy/emails/actions/admin_signs_instructor_up_for_workshop.py
@@ -1,58 +1,50 @@
+from datetime import datetime
 from typing import Any
 
-from django.dispatch import receiver
 from typing_extensions import Unpack
 
-from emails.controller import EmailController, EmailControllerMissingRecipientsException
-from emails.models import EmailTemplate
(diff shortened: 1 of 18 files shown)


pbanaszkiewicz:
@elichad Ready for your review!

elichad (approved):
I really like this refactoring! Everything feels a lot more readable now and it's clearer to see the differences between the different actions.

pbanaszkiewicz:
> I really like this refactoring! Everything feels a lot more readable now and it's clearer to see the differences between the different actions.

@elichad Thank you! That's really nice to hear. I think I found and marked some more places in the upcoming PR https://github.com/carpentries/amy/pull/2551.

Issue #2535, opened 2023-09-19
[Member codes] Implement soft validation for member codes


elichad:
_Originally posted by @pbanaszkiewicz in https://github.com/carpentries/amy/pull/2532#discussion_r1329269737_

I don't think we have a place in AMY where we do "soft validation". Here are my three ideas how to implement it. It's definitely not easy.

**Idea 1: code validation as background process in the browser**

Once the code is filled out by the user, we'd make a request to the API to see if there's a membership matching criteria.

With JS we'd dynamically show warnings in the form for the user, based on the responses from the server.

This potentially exposes the API to third parties as users don't have to be logged in in these forms, so some throttling should be implemented.

**Idea 2: dynamic field requesting user to confirm the message popup**

After user provides the code and there's something wrong with it, we amend the form by adding a new required checkbox field next to that member code.

This is clever because it uses Django default behavior to "break" form validation, but at the same time it's easy for the user to correct the form without changing member code - thus soft validation.

The downside is that altering a form like that is not an easy task.

**Idea 3: remembering form state between requests**

This idea uses Django session storage to remember if user say the warning message.

When first attempt on validating the form throws the error like in your current implementation, the server should also store some variable in the session.

This variable, when present in the session, would then change the member code validation and prevent throwing the same error twice. Finally the variable should be removed from the server session.

_Originally posted by @pbanaszkiewicz in https://github.com/carpentries/amy/pull/2532#discussion_r1329269737_
            
Note from @elichad :
Final decisions about the member codes validation are yet to be made, so we could also decide not to implement soft validation at all.

PR #2544, opened 2023-10-13
Enforce membership code on instructor training application

### LINKED ISSUE #2379: Enforce membership code on instructor training application
In the [instructor training application](https://amy.carpentries.org/forms/request_training/), the user has the option to choose "Profile Creation for Pre-approved Trainees" and is then asked to enter a code.  Currently there is no enforcement about what this code should be.

We would like to see the following.  ~This may need to be broken down into several issues.~ *DONE* 

* Check to see if the code is associated with a currently active membership. If so, proceed as normal.
* If not, give the user a warning (example text below).  Allow them to proceed.  Include the warning text in the confirmation email that the user receives.
* Add a filter to the instructor training request view to filter for requests with invalid codes. The admin user is already able to edit applications, so they can edit the code if need be. (EDIT: This bullet point has its own issue #2396)
* When assigning a person a learner task at TTT event, it should automatically assign the learner to the membership referenced by their code. (EDIT: This bullet point has its own issue #2397)

elichad:
Fixes #2379.

Adds code enforcement to the instructor training application in a similar way to workshop requests.

I changed the `group_name` field to `member_code` for consistency between `TrainingRequest` and `CommonRequest`, and because I think it's much clearer given the terminology we're using currently.


elichad:
@karenword a couple of questions for you about member code validation on the instructor training application. You don't need to look at the code changes in this PR.

### Validation

Initially we discussed having a "soft" validation where the user is prompted to check their code is correct, but will still be allowed to submit an invalid code. However, this turns out to be difficult to implement in the Django code framework that we use, and I'd prefer not to do it if we can avoid it. 

Here's how I've currently implemented the validation for different reasons for the invalid code:
- Code doesn't match a membership - raises an error, can't submit
- Request submission date is outside of relevant membership dates (+ 90 days either side) - raises an error, can't submit
- Membership has no seats remaining - no error, can submit, the request should be flagged on the admin side instead (my thought being that it's harder for a member contact to determine/debug if they've overused seats, and a trainee might not understand how this works)

What are your thoughts on this? Is this too restrictive?

I noticed that the [application form](https://amy.carpentries.org/forms/request_training/) currently says at the top "If you do not have a registration code, please enter the name of the institution you are joining through." Is this still accurate and expected, or should preapproved trainees always now have a code? If this statement is still true, we can't do the validation as I've described and will have to figure out the softer way.

### Error message

Here's the error message I've written: 

>This code is invalid. This may be due to a typo, an expired code, or a code that has not yet been activated. Please confirm that you have copied the code correctly, or your site contact to verify your code.

I used the phrase "site contact" because it's at the top of the form too, but I haven't seen it elsewhere - if we have a more commonly used term for member-contact-from-trainee's-perspective, let me know.

This is a bit different to [Maneesha's original suggestion](https://github.com/carpentries/amy/issues/2379#issuecomment-1505754083) as the validation style is changed. Let me know any feedback you have.

karenword:
thanks for the heads up @elichad -- this is definitely a significant change and will take some thought. I will ping @Talishask and @klbarnes20 for feedback as well as putting this on our team meeting agenda for discussion next Tuesday.

klbarnes20:
My first impression is that this will lead to a lot of people being unable to submit (and thus a lot of confused emails). I regularly get messages that codes are expired or have already used all their seats when I'm connecting training requests to trainings. But definitely interested in what @Talishask has to say as well.

amy/extforms/forms.py
```diff
@@ -192,11 +195,32 @@ def create_consent_field(self, term: Term) -> forms.ChoiceField:
         )
         return field
 
+    @feature_flag_enabled("ENFORCE_MEMBER_CODES")
+    def validate_member_code(self, request: HttpRequest) -> dict:
```
pbanaszkiewicz, line 200: This return type is not precise. This dictionary will contain ValidationError if returned, so `dict[str, ValidationError]`. Additionally I can see in L204 that `None` is being returned, so this would become: `None | dict[str, ValidationError]` (newer Python) or `Optional[dict[str, ValidationError]]` (more common format).

amy/extforms/forms.py
```diff
@@ -192,11 +195,32 @@ def create_consent_field(self, term: Term) -> forms.ChoiceField:
...
+        errors = dict()
+        member_code = self.cleaned_data.get("member_code", "")
+
+        if not member_code:
+            return
```
pbanaszkiewicz, line 212: Alternatively this could return empty dict.

amy/extforms/forms.py
```diff
@@ -88,6 +90,7 @@ class Meta:
         }
 
     def __init__(self, *args, **kwargs):
+        self.request_http = kwargs.pop("request", None)
```
pbanaszkiewicz, line 94: This name is a little bit off. Usually it was named `self.request`.

amy/extforms/forms.py
pbanaszkiewicz: Actually, if the feature flag is disabled, then this function would return `None`, so I'd still keep `None` here and just correct the return type annotation in L199.

amy/extforms/tests/test_training_request_form.py
```diff
@@ -83,7 +101,7 @@ def test_request_added(self):
         # with open('email.eml', 'wb') as f:
         #     f.write(msg.message().as_bytes())
 
-    def test_review_process_validation(self):
+    def test_review_process_validation__preapproved_code_empty(self):
         # 1: shouldn't pass when review_process requires member_code
```
pbanaszkiewicz: This comment can now become a docstring for this test. Same in L117.

amy/extforms/forms.py
```diff
@@ -199,20 +200,29 @@ def create_consent_field(self, term: Term) -> forms.ChoiceField:
...
-        #       and update text accordingly
         try:
-            Membership.objects.get(registration_code=member_code)
+            # find relevant membership - may raise Membership.DoesNotExist
+            membership = Membership.objects.get(registration_code=member_code)
+            # confirm that membership is currently active
+            if (
+                not membership.agreement_start
+                <= date.today()
+                <= membership.agreement_end
```
pbanaszkiewicz: You could create a `Membership.active_on_date(d: date) -> bool` method for checking if membership is active on a given date. This would allow for some unit testing and fine-tuning in future if we wanted to extend the range the membership is considered active by a few days.

amy/extforms/forms.py
```diff
@@ -215,11 +215,10 @@ def validate_member_code(self, request: HttpRequest) -> dict:
...
             # confirm that membership is currently active
-            if (
-                not membership.agreement_start
-                <= date.today()
-                <= membership.agreement_end
-            ):
+            # grace period: 90 days before and after
+            valid_start = membership.agreement_start - timedelta(days=90)
+            valid_end = membership.agreement_end + timedelta(days=90)
+            if not valid_start <= date.today() <= valid_end:
```
pbanaszkiewicz, line 221: Okay, I was commenting on this line in the previous commit which looked like:
```python
            if (
                not membership.agreement_start
                <= date.today()
                <= membership.agreement_end
            ):
```
Looks like you added a grace period - something I mentioned as possible opportunity if that code was turned into a method.

I still think this could be turned into a method.

pbanaszkiewicz (approved):
The changes look good to me.

amy/extforms/forms.py
elichad: I did this to avoid ambiguity between the http request and the training request that this form describes. I'm not entirely happy with the break from convention but I think it might prevent future confusion

elichad:
Summary of conversation with @karenword and @Talishask:
- Soft validation is strongly preferred (#2535)
- If soft validation is not possible
    - do not apply hard validation
    - Keep internal matching to support attendance workflow, where possible
    - Still would like filter to identify invalid codes on the training requests page

Sentence at the top of the application form: If you do not have a registration code, please enter the name of the institution you are joining through.
- We can remove this from the first paragraph.

Error message phrasing options:
- This code is invalid. This may be due to a typo, an expired code, or a code that has not yet been activated. Please confirm that you have copied the code correctly, or confirm the code with the Membership Contact for your group 
- This code is invalid. This may be due to a typo, an expired code, or a code that has not yet been activated. Please confirm that you have copied the code correctly, or ask the Membership Contact to verify your code for your group

I am going to merge this PR after addressing @pbanaszkiewicz's recommendations, as there is code here that lays useful groundwork for further development. However, things _will_ be changed on the validation side in future PRs according to the summary above.

PR #2551, opened 2023-10-22
[Emails] Completed instructor training not yet badged

### LINKED ISSUE #2540: [Emails] Implement "Completed instructor training but not yet badged"
TBC

pbanaszkiewicz:
This fixes #2540.
Add "Instructor Training Completed but not yet badged" email.

This email will be triggered when trainee is given a pass mark on training requirement.
It will update with other training requirements being awarded, and once all requirements have been passed it will cancel.

The email is scheduled to run 2 months after the training has completed (trainee's training requirement event's end date).

From developer's standpoint there's a few small differences between this action and other emails:
1. strategy run function, `run_instructor_training_completed_not_badged_strategy`, is also responsible for getting trainee's training completion date (it may not be available when marking as "passed" other requirements than the training requirement)
2. receiver classes inheriting from `BaseAction`, `BaseActionUpdate` and `BaseActionCancel` now share the same functions (called `get_scheduled_at`, `get_context`, `get_generic_relation_object`, `get_recipients`)
3. I eliminated some repeated tests for receiver classes, since they are well-tested by tests for the base classes; I kept integration tests.

### DIFF
--- a/amy/emails/actions/instructor_training_completed_not_badged.py
+++ b/amy/emails/actions/instructor_training_completed_not_badged.py
@@ -1,11 +1,14 @@
-from datetime import datetime
+from datetime import date, datetime
 import logging
 
+from django.contrib.contenttypes.models import ContentType
 from django.http import HttpRequest
 from typing_extensions import Unpack
 
 from emails.actions.base_action import BaseAction, BaseActionCancel, BaseActionUpdate
+from emails.models import ScheduledEmail
 from emails.signals import (
+    INSTRUCTOR_TRAINING_COMPLETED_NOT_BADGED_SIGNAL_NAME,
     Signal,
     instructor_training_completed_not_badged_remove_signal,
     instructor_training_completed_not_badged_signal,
@@ -17,16 +20,14 @@ from emails.types import (
     StrategyEnum,
 )
 from emails.utils import two_months_after
-from workshops.models import Award, Person, TrainingProgress
+from workshops.models import Person, TrainingProgress
 
 from .instructor_training_approaching import EmailStrategyException
 
 logger = logging.getLogger("amy")
 
 
-def instructor_training_completed_not_badged_strategy(
-    person: Person,
-) -> StrategyEnum:
+def instructor_training_completed_not_badged_strategy(person: Person) -> StrategyEnum:
     logger.info(f"Running InstructorTrainingCompletedNotBadged strategy for {person}")
 
     person_annotated = (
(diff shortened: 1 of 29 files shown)


pbanaszkiewicz:
@elichad ready for your review!

amy/emails/actions/instructor_training_completed_not_badged.py
```diff
@@ -0,0 +1,286 @@
...
+        and bool(person_annotated.passed_demo)
+    )
+
+    ct = ContentType.objects.get_for_model(person)  # type: ignore
+    has_email_scheduled = ScheduledEmail.objects.filter(
+        generic_relation_content_type=ct,
+        generic_relation_pk=person.pk,
+        template__signal=INSTRUCTOR_TRAINING_COMPLETED_NOT_BADGED_SIGNAL_NAME,
+        state="scheduled",
+    ).exists()
```
elichad, line 77: We may be adding progress after this email has been sent, since the email is intended to act as a reminder. Could you also check for running/succeeded state and prevent a new email from being created in that case?

elichad (changes requested):
Nice work on this. I have a few suggestions here and in comments:
- email should be cancelled if Training progress has been updated to a non-passed state (currently it just updates)
- make sure the new EmailStrategyExceptions are handled in a way that passes the error back to the user through the UI. I got the Django debug page by creating a Training progress using an event with no end date

amy/emails/actions/instructor_training_completed_not_badged.py
pbanaszkiewicz: I see your point.

Potentially, in the worst case scenario, as soon as first email reminder was sent, there's a new training progress passed causing another email to schedule. It would run immediately if the event's end date hadn't change and it's >= 2 months in the past.

But this can also be aligned with what business (@maneesha) wants. I mean, another one or more reminders may not cause harm.

@maneesha Consider the scenario above and let me know if only one reminder should be sent at most.

maneesha: Right now, we only send this email once, at 60 days after training, letting them know their 90 day deadline is approaching.  In our current workflows, no more emails are sent.

This may also be one that is better suited to run via a cronjob rather than the triggers and conditions other emails use.

pbanaszkiewicz:
@elichad I updated code in most of the places you mentioned in the comments above:
1. fixed invalid `context_type` passed to signals
2. switched to `person.instructor_eligible` field
3. extended the context with training requirements that the person has no grade on
4. wrapped strategy runs in try-except to catch EmailStrategyException and inform user about it

I also asked @maneesha about two scenarios you mentioned. Will wait for her response until making further changes.

amy/emails/actions/instructor_training_completed_not_badged.py
pbanaszkiewicz: Okay, I'll change the implementation to run only once, and then we'll see after tests how good this works - and optionally change it to cronjob type.

amy/trainings/views.py
```diff
@@ -74,16 +79,73 @@ def get_context_data(self, **kwargs):
...
+            run_instructor_training_completed_not_badged_strategy(
+                instructor_training_completed_not_badged_strategy(person),
+                request=self.request,
+                person=person,
+                training_completed_date=event.end if event else None,
+            )
+        except EmailStrategyException as exc:
+            messages.error(
+                self.request,
+                f"Error when running instructor training completed strategy. {exc}",
```
elichad, line 119: I'm not sure this message will make much sense to the user, so I suggest something more straightforward.
```suggestion
                f"Error when creating or updating scheduled email. {exc}",
```

pbanaszkiewicz: Good suggestion, thank you @elichad!

pbanaszkiewicz:
@elichad @maneesha Updated. Now if there exists email in locked, running or succeeded state, the strategy quits. This means that there should be at most one reminder sent to the trainee.

@elichad Additionally I fixed this issue you spotted:
> email should be cancelled if Training progress has been updated to a non-passed state (currently it just updates)

Turns out strategy was evaluated just before applying updates to a training progress in the `TrainingProgressUpdate` view.

elichad (changes requested):
I found a new bug when creating the email template for this signal. The template was created, but couldn't be displayed:

```
ValueError at /emails/template/52e92ade-00e7-4590-a50f-9c5cb6f770fb/
no signature found for builtin type <class 'datetime.date'>
```
I think this is to do with trying to render a date as part of the `body_context_annotations`, as I experimented:
```
>>> import datetime
>>> a = datetime.date(2023,5,11)
>>> a.__annotations__
AttributeError: 'datetime.date' object has no attribute '__annotations__'
```

Other than that, the new changes look good.

pbanaszkiewicz:
@elichad Thanks for letting me know about this issue. Trying to fix it was painful, and eventually I wasn't able to retain the feature causing the issue to occur.

The problem was with our system to help users investigate what template variables they have available for a given email template. It was done through annotations on context type objects (classes with suffix "Context" defined in `amy/emails/types.py` as inheriting from `TypedDict`). When Django template was attempting to reference a value written in such Context as `date` (type referencing to Python builtin `datetime.date`), the template system was attempting to call and then inspect the type. Finally it was giving up with a nasty error.

Since I didn't want to use other types than date, and also I noticed that we also have `list[something]` in the same context class, and there was no hacky way to get rid of the error, I felt compelled to eventually remove `model_documentation_link` and display a `repr()` version of a class/annotation used in the context class.

This is the result. While not pretty, at least it works:
image

elichad (approved):
I'm happy for this to be merged now, though I think we should come back to the `body_context_annotations` again as we approach user testing, and figure out how to make them as useful as possible (e.g. with some supporting documentation to guide people to the right model docs).

PR #2552, opened 2023-10-23
Remove member affiliation question; remove code question from SO and WI forms

elichad:
This comes from a discussion with the workshop admin team, as mentioned in https://github.com/carpentries/amy/issues/2380#issuecomment-1768639578 .

PR #2553, opened 2023-10-23
Add further checks for invalid code to workshop request forms

### LINKED ISSUE #2393: Update workshop request form: Warning to admin that workshop may not be allowed
We have a separate issue to validate member codes in the workshop request form  (#2380).

Once this feature is implemented, we should be able to add in another level of validation for the admin who may be accepting the workshop request.  If the workshop requested exceeds the workshops allowed (online or inperson), the admin should get a warning before being able to accept that workshop request and create a new event.

### LINKED ISSUE #2533: [Member codes] Add further checks for invalid code to all workshop request forms
Consider different reasons for the invalid code:
    Wrong code entirely
    Outside of membership dates
    Number of seats already used
    Number of workshops already used 

Error text guidance included in https://app.asana.com/0/0/1204384657108800 and https://app.asana.com/0/0/1204122479379163

elichad:
Depends on #2552 . Closes #2533. Updated 08 Nov.

This PR:
- checks that membership is active at time of form submission
- ~checks that membership has workshops remaining~
- adds info about related membership to WR details page (edit: this fixes #2393!)


elichad:
@sheraaronhurt @Talishask I want to check this with you: 

What is your usual workflow when a member has no workshops remaining but wants to request another workshop? How do you update AMY to represent this? Do you treat it as a paid COW separate from the membership?
I'm trying to understand if any nuance is necessary when validating the code and discovering there are no workshops left, or if it's okay to just block submission in that case.

sheraaronhurt:
@elichad thanks for checking. While this is a rare occurrence, yes they will contact us and we give them additional workshops at a discounted rate. In AMY it would be tracked by the discounted workshops section. @Talishask feel free to chime in :)

elichad:
So as I understand it, we agree on extra workshops with the member, but there's nothing in AMY that indicates 'this member is allowed X more workshops than the original contract (and should be allowed to submit X further requests).' We just remember that they have extras when we're processing the requests that come through with their code. Looking at the code, the 'discounted workshops' field is actually just counting all the Events connected with the membership, and subtracting the original allowance.

If that's how things work, then those members won't actually be able to submit requests if they've used their original allowance, because the form will see '0 workshops remaining' in AMY and block the submission.

To solve that, I think we need to add an 'additional workshops' number field to the membership, in the same manner as the 'additional instructor training seats' that we have at the moment. That way, when a member buys an additional 2 workshops, we can edit their membership to record that, and then they will be able to submit 2 more workshop requests after that.

Does that make sense?

Talishask:
Great questions.

**Member Workshops**
When members need additional workshops they typically submit a workshop request form before contacting us and then we discuss this with them. My current process is to update the number of allowed workshops once the member is invoiced. In most cases Danielle is already working with them.  The downside of updating the allowed workshops field is that we lose data on added workshops, so this is not optimal. We used to have a 'Members included in membership" and an 'Additional workshops' field but we lost the second field with some updates last year. I agree and would be excited to get that field back.

Given that members typically submit requests before they are invoiced for additional workshops, I would recommend we have a warning internally but do not block any submissions due to '0 workshops remaining.

**Discounted Workshops Field**
The current discounted workshops field assumes that any COWs above the allowed workshops are discounted. That is not correct. Typically these are workshops that are either above the limit or not yet paid. It would be more helpful for workshop allowances to be structured just the same as IT seats with "allowed", "used", "remaining".

**Alacarte Workshops**
Alacarte workshops are a little bit different. Our current workflow is to create the membership (so that we can tie the event to it because planning can start while we are working on paperwork), but wait to add the allowed workshop until the invoicing happens. Essentially we are waiting to confirm payment before adding the allowed workshop. We changed this order for a la carte workshops because they are easier to fall through the cracks and not get invoiced. I have a report that lets me see workshops where the allowed is zero then I check to see if we received payment or not and update the allowed workshops once we invoice it. It might be a good idea to have a conversation with the TT about some of these things to help me find better solutions and standardize more with our regular workflows.

elichad:
@Talishask Thanks for this information, I'll schedule a short call with you to talk through a couple of follow-up questions I have

elichad:
@pbanaszkiewicz this is now ready for review!

amy/extrequests/templatetags/request_membership.py
```diff
@@ -0,0 +1,34 @@
...
+from django.utils.safestring import mark_safe
+
+from workshops.models import Membership
+
+register = template.Library()
+
+
+@register.simple_tag
+def membership_description(membership: Membership):
+    if not isinstance(membership, Membership):
```
pbanaszkiewicz, line 13: Is it expected for this tag to work on objects other than Membership?

If it's expected then the type hinting is not correct. Otherwise I'd remove this condition.

pbanaszkiewicz, line 13: Is it expected for this tag to work on objects other than Membership?

If it's expected then the type hinting is not correct. Otherwise I'd remove this condition.

amy/extrequests/templatetags/request_membership.py
```diff
@@ -0,0 +1,34 @@
...
+from django import template
+from django.utils.safestring import mark_safe
+
+from workshops.models import Membership
+
+register = template.Library()
+
+
+@register.simple_tag
+def membership_description(membership: Membership):
```
pbanaszkiewicz, line 12: My suggestion is to turn this tag into a Django template. There's a lot of template generation going on compared to "difficult template logic".

Here, the logic that may be difficult or "ugly" to achieve in the Django template is the calculation of alert type. I suggest to use a template tag to determine if this message is just informational or a warning.

amy/extrequests/templatetags/request_membership.py
```diff
@@ -0,0 +1,34 @@
...
+        alert_type = "warning"
+    else:
+        alert_type = "info"
+    info = (
+        f'<div class="alert alert-{alert_type}">'
+        "Related membership: "
+        f'<a href="{membership.get_absolute_url()}">{membership}</a>.'
+        "<br>"
+        f"This membership has <strong>{workshops_remaining}</strong> "
+        f'workshop{"s" if workshops_remaining !=1 else ""} remaining.'
```
pbanaszkiewicz, line 28: In the Django template you can use [`pluralize`](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#pluralize) template filter to achieve the plural form of "workshop".

amy/extrequests/templatetags/request_membership.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+    if not active:
+        info += "<br>This membership is <strong>not currently active</strong>."
+    info += "</div>"
+
+    return mark_safe(info)
```
pbanaszkiewicz, line 34: After reading https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.html.format_html I think we should review our use of `mark_safe`, as `format_html` may be more appropriate.

amy/extrequests/utils.py
```diff
@@ -72,3 +64,15 @@ def member_code_valid_training(
         raise MemberCodeValidationError("Membership has no training seats remaining.")
 
     return True
+
+
+def get_membership_from_code_if_exists(code: str) -> Membership | None:
```
pbanaszkiewicz, line 69: Naming suggestion: `get_membership_by_code_or_none`, as `or_none` automatically suggest what is returned if membership is not found.

On the other hand, my suggestion may be flawed by suggesting fetching the membership "by code or by None".

pbanaszkiewicz (approved):
Looks good, I left some suggestion but nothing too important.

amy/extrequests/templatetags/request_membership.py
elichad: Yes, I agree. Created #2574

elichad:
Thanks Piotr! I addressed all your comments so I'll go ahead and merge.

PR #2554, opened 2023-10-24
Add code seat availability check to IT application

elichad:
Adds seat availability to the validity criteria for the member code check

### DIFF
--- a/amy/extforms/tests/test_training_request_form.py
+++ b/amy/extforms/tests/test_training_request_form.py
@@ -7,7 +7,7 @@ from django.urls import reverse
 
 from consents.models import Term, TermOptionChoices
 from extforms.views import TrainingRequestCreate
-from workshops.models import Membership, Role, TrainingRequest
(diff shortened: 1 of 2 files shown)


amy/extforms/tests/test_training_request_form.py
```diff
@@ -62,6 +62,30 @@ def setUpMembership(self):
             agreement_end=date.today() + timedelta(weeks=26),
             contribution_type="financial",
             registration_code="valid123",
+            public_instructor_training_seats="1",
+            inhouse_instructor_training_seats="1",
```
pbanaszkiewicz, line 66: Minor: is this a string value?

pbanaszkiewicz (approved):
Looks good, I left a very minor question.

amy/extforms/tests/test_training_request_form.py
elichad: Nope, they are ints, not sure why I put them in as strings

PR #2558, opened 2023-10-27
Add filter to training requests view for invalid codes

### LINKED ISSUE #2396: Add filter to training requests view for invalid code
Once we implement #2379  we should have an additional filter on the [training requests view](https://amy.carpentries.org/requests/training_requests/) for applications with invalid codes.  

The admin user can then go in and edit the code as needed.

elichad:
Fixes #2396. The filter itself is simple change thanks to the `member_code_override` field added in #2549.

I also added tests for the whole filterset, and implemented a small fix to the state="Pending or accepted" filter option as a result (it wasn't excluding withdrawn requests).

### DIFF
--- a/amy/extrequests/filters.py
+++ b/amy/extrequests/filters.py
@@ -65,6 +65,12 @@ class TrainingRequestFilter(AMYFilterSet):
         widget=widgets.CheckboxInput,
     )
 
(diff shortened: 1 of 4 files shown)


pbanaszkiewicz (approved):
Looks good.

PR #2559, opened 2023-10-27
Add a filter to the workshop request view for requests where an active member did not use their code

elichad:
Depends on #2558 because this edits the same `test_filters.py` file which was introduced there.

Requested in https://github.com/carpentries/amy/issues/2380#issuecomment-1768639578.


pbanaszkiewicz (approved):
Also looks good.

PR #2560, opened 2023-10-30
Add soft validation of member codes to training request update form

### LINKED ISSUE #2550: [Member codes] Add soft validation of member codes to TrainingRequestUpdateForm
This is on the admin side of the training requests. Check with ITT before making changes.

elichad:
Fixes #2550.

The implementation is very similar to #2549, but allows the actual reason for failure to be shown (rather than the list of possible reasons).

While testing I discovered some `CharField` and `TextField` fields on `TrainingRequest` which had `blank=True, null=True`, so I changed these to have `default=""` instead in line with best practices.


amy/workshops/models.py
```diff
@@ -2255,7 +2255,7 @@ class TrainingRequest(
     nonprofit_teaching_experience = models.CharField(
         max_length=STR_LONGEST,
         blank=True,
-        null=True,
+        default="",
```
pbanaszkiewicz, line 2258: Thanks for these changes.

amy/workshops/migrations/0269_alter_trainingrequest_default_for_text_fields.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.2.20 on 2023-10-30 09:20
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    replaces = [('workshops', '0269_alter_trainingrequest_nonprofit_teaching_experience'), ('workshops', '0270_auto_20231030_0919')]
```
pbanaszkiewicz, line 8: What causes this? Can you use a rebased codebase or alter migration `0269_alter_trainingrequest_nonprofit_teaching_experience` (it's okay before we release to prod, and even safer if it wasn't released on test-amy)?

amy/extrequests/forms.py
```diff
@@ -1407,6 +1407,79 @@ class Meta:
...
+    def member_code_valid(self, code: str) -> tuple[bool, str]:
+        """Returns True if the code matches an active Membership with available seats,
+        including a grace period of 90 days before and after the Membership dates.
+        If there is no match, returns False and a detailed error.
+        """
+        try:
+            # find relevant membership - may raise Membership.DoesNotExist
+            membership = Membership.objects.get(registration_code=code)
+        except Membership.DoesNotExist:
+            return False, f'No membership found for code "{code}".'
```
pbanaszkiewicz, line 1459: You could use custom exceptions instead of returning a flag and text.
For example:

```python

class MemberCodeValidationError(ValueError):  # or inherit from ValidationError
  pass

...

    def member_code_valid(self, code: str) -> bool:
        """Returns True if the code matches an active Membership with available seats,
        including a grace period of 90 days before and after the Membership dates.
        If there is no match, throws an exception with detailed error.
        """
        try:
            # find relevant membership - may raise Membership.DoesNotExist
            membership = Membership.objects.get(registration_code=code)
        except Membership.DoesNotExist as exc:
            raise MemberCodeValidationError(f'No membership found for code "{code}".') from exc

        ...
        return True

```

amy/extrequests/forms.py
```diff
@@ -1407,6 +1407,79 @@ class Meta:
...
+                error_msg = (
+                    "This code is invalid: "
+                    f"{error_detail} "
+                    "Tick the checkbox below to ignore this message."
+                )
+                errors["member_code"] = ValidationError(error_msg)
+
+        return errors
+
+    def member_code_valid(self, code: str) -> tuple[bool, str]:
```
pbanaszkiewicz, line 1450: I'd consider making this a function receiving `code` and `timestamp` (to be used instead of `self.instance.created_at`) and moving outside the class - as a util function. I'd also add unit tests for it.

pbanaszkiewicz (approved):
Looks good, but I strongly suggest moving `member_code_valid` to exception-based approach.

amy/workshops/migrations/0269_alter_trainingrequest_default_for_text_fields.py
elichad: I squashed two uncommitted migrations to make this one, and this line resulted from that. I'll delete it.

PR #2561, opened 2023-10-30
Update autoresponses with member codes

elichad:
- adds `member_code` field to the WRF autoresponse
- adds `member_code_override` field to the TR autoresponse if it is set to `True` (i.e. if it wasn't used, don't include it as it may not have been displayed at all)
- adds a warning to the TR autoresponse if `member_code_override=True` (i.e. if the member code is invalid)


pbanaszkiewicz (approved):
LGTM!

PR #2565, opened 2023-11-06
Remove nonexistent field from knowledge domain lookup

elichad:
The `KnowledgeDomainLookupView` filters on the `subtag` field, which doesn't actually exist on the `KnowledgeDomain` model. 
(I think this might have been copied across from the equivalent view for the `Language` model.)

This lookup is only used in one view, and it doesn't get much use, so this fix can go into the next regular release.


pbanaszkiewicz (approved):
Thanks for fixing that!

PR #2570, opened 2023-11-10
Instructor Training application - add question about Eventbrite URL and associated filter

### LINKED ISSUE #2562: Add question to IT application form about Eventbrite signup
Idea that came up in a discussion with @Talishask.

Add a field to the IT application form for preapproved trainees asking for the URL of the Eventbrite event they signed up for (if applicable). Add this field to the training request filters. Consider auto-matching applications to training based on the URL they provide.

This will help with filtering applications and matching them to training. Currently the ITT has to manually search each attendee by name, email, or member code and match them to the training.

This could offer a significant timesave for the ITT, though it would need to be removed or adjusted significantly when we switch ticketing system.

elichad:
Fixes #2562.

Marking draft as need to check with the ITT if this is okay with them as I slightly jumped the gun... (pinging @karenword and I'll link to this on your agenda)

This change will allow preapproved trainees to enter an Eventbrite URL for the event they've registered for, if applicable.
On the admin side, the "All Training Requests" page will display the event ID (from Eventbrite), and this can be searched for using the filters on the left. You can enter the ID or the full event URL for the search.

Side note: You may already know that are a few different forms that Eventbrite event URLs can take (some registrant-facing, some owner-facing). These example URLs all reference the same event with an ID `711575811407`, and all three forms are handled in this implementation. If there are more forms that you're aware of, please let me know...
- "https://www.eventbrite.com/e/online-instructor-training-7-8-november-2023-tickets-711575811407"
- "https://www.eventbrite.com/e/711575811407"
- "https://www.eventbrite.com/myevent?eid=711575811407"

And here are some screenshots.

Instructor training application form:
IT form with a new question in the "Preapproved" section: "If you have already registered for an event through Eventbrite, enter the URL of that event."

All training requests view. The Eventbrite ID is pulled from the URL in the request. Personally I'm not so happy with the increased width of the table since it means more need for horizontal scrolling.
Training requests view with a new column "Eventbrite ID," which shows a 12-digit number for some requests

Filter by event ID or URL. Search term in screenshot is `https://www.eventbrite.com/myevent?eid=920035475452`, could also have used `920035475452` or even just a subset of those digits like `5452` to get the same result:
image

Note that we don't actually need to show the Eventbrite ID in a column in order to use the search filter (but it makes things clearer that the search returns what you wanted).


karenword:
Thanks @elichad this looks great! 

I played with one of our Eventbrite pages, and it looks like the page link in the address bar at the point when trainees are prompted to complete the application looks like this:

https://www.eventbrite.com/e/online-instructor-training-20-23-november-2023-tickets-719778495857?aff=oddtdtcreator

By allowing us to search attendees by event before they have been accepted and matched, this could make a valuable impact on the amount of time spent processing attendance. Thank you!!

karenword:
I think a couple of other points worth noting for clarity:
- the main object for flexibility is on the point at which trainees enter the link. Our search can be specific to the event ID, but we have to expect that they will submit whatever link they find. It is also possible that they may try to submit non-eventbrite links such as the [confirmation page](https://carpentries.org/instructor-training/confirmation/) or the [calendar page header](https://carpentries.github.io/instructor-training/training_calendar.html#november-20-23-2023) or who knows what else.
- we did not discuss the value of displaying the Eventbrite ID in the table, but I agree that expanding the width of the table is not desirable. If we are searching by eventbrite ID, I think it is reasonable to assume that the search won't malfunction, and after applications have been matched the event slug will show, which is more usable information for us anyway.

karenword:
It is possible that, by asking people to submit a specific event link, we will inadvertently imply that they need to submit a new application whenever they want to cancel & re-register for a new event. We may need new language (either in AMY or in Eventbrite) that deters this behavior.

elichad:
@karenword 

> the main object for flexibility is on the point at which trainees enter the link. Our search can be specific to the event ID, but we have to expect that they will submit whatever link they find. It is also possible that they may try to submit non-eventbrite links such as the [confirmation page](https://carpentries.org/instructor-training/confirmation/) or the [calendar page header](https://carpentries.github.io/instructor-training/training_calendar.html#november-20-23-2023) or who knows what else.

I'll add a check that the URL the trainee enters is pointing to the Eventbrite domain, and some help text to say they can find this link in their confirmation email.

Conveniently, the URL flexibility is the same for both the trainee and the admin, because internally the code tries to pull the ID out of the URL and use that as much as possible. This means that when searching, you can enter a URL like "https://www.eventbrite.com/myevent?eid=719778495857" and it will match a trainee who entered "https://www.eventbrite.com/e/online-instructor-training-20-23-november-2023-tickets-719778495857?aff=oddtdtcreator". (This also means it doesn't matter if either link ends up with extra stuff after the `?`, which is quite common).

Theoretically, any other form of URL that Eventbrite could throw at us should work, as long as it contains the 12-digit ID somewhere in it (and no other long strings of digits) - the ones I listed in my top comment are just the ones I've explicitly tested.

> we did not discuss the value of displaying the Eventbrite ID in the table, but I agree that expanding the width of the table is not desirable. If we are searching by eventbrite ID, I think it is reasonable to assume that the search won't malfunction, and after applications have been matched the event slug will show, which is more usable information for us anyway.

Do you think you would cross-reference the search results with your Eventbrite registrations anyway? If so, I think it would be okay to remove the column.

> It is possible that, by asking people to submit a specific event link, we will inadvertently imply that they need to submit a new application whenever they want to cancel & re-register for a new event. We may need new language (either in AMY or in Eventbrite) that deters this behavior.

Good point. I think we should try to add this text at the point of cancellation if possible as I wouldn't expect people to remember what they read when signing up. Do we have custom text that trainees see when they cancel?

karenword:
> Do you think you would cross-reference the search results with your Eventbrite registrations anyway? If so, I think it would be okay to remove the column.

We will be looking at our Attendance sheets when we match trainees, so I think that should be sufficient cross-referencing so that we don't need -- pinging @klbarnes20 for any further comment

> Good point. I think we should try to add this text at the point of cancellation if possible as I wouldn't expect people to remember what they read when signing up. Do we have custom text that trainees see when they cancel?

As it happens, I have an agenda item to look into this question for other reasons. AFAIK we do not currently have a customized cancellation confirmation email. However, I should note that the point at which trainees are prompted by us to complete the application occurs during registration, when they have to check a box in Eventbrite saying they have completed it. So I think that is probably the time point when they should be told not to do it again if they already did it for a previous event. It couldn't hurt to also tell them when they cancel, but they will have to remember it when they are prompted upon re-registration.

elichad:
> However, I should note that the point at which trainees are prompted by us to complete the application occurs during registration, when they have to check a box in Eventbrite saying they have completed it.

Ah, I didn't realise this! I'll take care to avoid creating a loop with the instructions in AMY then. (This should actually work nicely if we ask people to paste the URL of the registration page)

> So I think that is probably the time point when they should be told not to do it again if they already did it for a previous event. 

I agree.

amy/extforms/forms.py
```diff
@@ -248,6 +252,12 @@ def validate_member_code(
 
         return errors
 
+    def clean_eventbrite_url(self):
+        """Check that entered URL includes 'eventbrite' in the domain."""
+        eventbrite_url = self.cleaned_data.get("eventbrite_url", "")
+        if eventbrite_url and "eventbrite" not in urlparse(eventbrite_url).netloc:
```
pbanaszkiewicz, line 258: If user provides `https://eventbrite.example.org/123123123` this fill validate. Do you think this is serious enough that we should whitelist eventbrite domain?

Also, you can use `urlparse(eventbrite_url).hostname`, as in some cases netloc contains more info than what we need.

amy/templates/requests/all_trainingrequests.html
```diff
@@ -62,6 +64,13 @@
             &lt;{{ req.email|urlize }}&gt;
           </td>
           <td>{{ req.member_code|default:"—" }}</td>
+          <td>
+            {% if req.eventbrite_url %}
+              {% eventbrite_id_from_url req.eventbrite_url %}
+            {% else %}
+              —
```
pbanaszkiewicz, line 71: Can you use `&mdash;` entity for a default value here?

Edit: For context, I think I introduced the default value of `&mdash;`, but then changed my mind and started using `—` as I was able to type it from keyboard. Now I'm not able to type it easily, so I prefer the original approach of `&mdash;` (it's also less confusing).

amy/templates/includes/trainingrequest_details.html
```diff
@@ -30,6 +30,8 @@
     <tr><th>Continue with registration code marked as invalid:</th>
         <td>{{ object.member_code_override|yesno }}</td></tr>
 {% endif %}
+    <tr><th>Eventbrite URL:</th>
+        <td>{{ object.eventbrite_url|default:"&mdash;" }}</td></tr>
```
pbanaszkiewicz, line 34: You can make it clickable, but then I'd for sure use domain whitelisting to make sure someone doesn't use spoof the original domain (e.g. by using homoglyphs).

pbanaszkiewicz (approved):
Looks good, some suggestions left.

amy/extforms/forms.py
elichad: I found that Eventbrite have a lot of local domains as well as `eventbrite.com`, e.g. `eventbrite.co.uk`, `eventbrite.fr`, so it's harder to whitelist them.
As the link isn't clickable anywhere, I think this will be okay without further validation. But maybe for safety, I'll add something on the admin side that does a stricter check and shows a warning if the link doesn't look legitimate.

amy/templates/includes/trainingrequest_details.html
elichad: Yeah, I spoke about this with Rob last week and we thought that doing this safely would be more effort than it's worth. I don't anticipate these links being used directly very often, their main purpose is to be searchable.

PR #2571, opened 2023-11-10
Never hide training requests with invalid code when filtering

elichad:
Currently the `member_code_override` `BooleanFilter`, when unchecked, will only display training requests where the override hasn't been used.
It would make more sense that when the filter is unchecked, _all_ requests should be shown, regardless of override usage.


pbanaszkiewicz (approved):
Looks good!

PR #2572, opened 2023-11-14
Automatically assign learner TTT task membership based on code in training application

elichad:
Depends on #2553 so marking as draft until that is merged.

This PR adds two features:

1. on the 'Add task' page, helps the user find the active membership when assigning a learner task (which uses a new lookup)
2. on the 'All trainees' page, adds a checkbox to auto-assign membership seats based on member codes used. This will raise errors for requests that did not provide a code or that provided an invalid code. 

Screenshots:
New Task form with person, role, and event fields filled, and only one displayed option for the membership
Match form showing new checkbox "Automatically match seats to memberships" with help text "Assigned users will take instructor seats based on the registration code they entered."
All trainees page showing warnings, errors, and info message after bulk matching trainees with the auto-assign membership feature. The warnings indicate that a membership is inactive at the time of the training, the errors indicate that some requests did not include valid codes, and the info message says "Accepted and matched 3 people to training, which raised 6 warning(s). 3 request(s) were skipped due to errors."
Note that I get lots of extra warnings raised on my dev instance because the membership dates are randomised, but they're the same warnings as you may see currently when you match trainees through this page. It's only the errors that are new.

### DIFF
--- a/amy/extrequests/utils.py
+++ b/amy/extrequests/utils.py
@@ -83,8 +83,6 @@ def get_membership_warnings_after_match(
 ) -> list[str]:
     warnings = []
 
-    today = date.today()
-
     remaining = (
         membership.public_instructor_training_seats_remaining
         if seat_public
@@ -97,7 +95,7 @@ def get_membership_warnings_after_match(
         )
 
     # check if membership is active
-    if not (membership.agreement_start <= today <= membership.agreement_end):
+    if not (membership.agreement_start <= date.today() <= membership.agreement_end):
         warnings.append(
             f'Membership "{membership}" is not active.',
         )
--- a/amy/extrequests/utils.py
+++ b/amy/extrequests/utils.py
@@ -2,7 +2,11 @@ from datetime import date
 
 from django.core.exceptions import ValidationError
 
-from workshops.models import Event, Membership
+from workshops.models import Event, Membership, Role, Task, TrainingRequest
+
+# ----------------------------------------
+# Utilities for validating member codes
+# ----------------------------------------
 
 
 class MemberCodeValidationError(ValidationError):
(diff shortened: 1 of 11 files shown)


elichad:
@pbanaszkiewicz this and #2570 are both ready for review now. I'd prefer if you reviewed this one first.

amy/extrequests/views.py
```diff
@@ -571,53 +601,49 @@ def all_trainingrequests(request):
...
-
-            messages.success(
-                request,
-                "Successfully accepted and matched selected people to training.",
-            )
+            for msg in warnings:
+                messages.warning(request, msg)
+            if errors:
+                for msg in errors:
+                    messages.error(request, msg)
```
pbanaszkiewicz, line 633: Nit: Can be moved outside the if-statement, ie. above L631. This way we keep similar code close to each other, and logic related to displaying the final message is separated out.

amy/extrequests/views.py
```diff
@@ -571,53 +601,49 @@ def all_trainingrequests(request):
...
+                    f'{"person" if changed_count==1 else "people"} to training, '
+                    f"which raised {len(warnings)} warning(s). "
+                    f"{len(errors)} request(s) were skipped due to errors."
+                )
+                messages.info(request, info_msg)
+            else:
+                messages.success(
+                    request,
+                    "Successfully accepted and matched selected people to training.",
+                )
```
pbanaszkiewicz, line 646: I don't like how how view is expanding, but to be honest I spent like 30 minutes trying to restructure this code and the results were not satisfactory.

If you have some ideas how we could improve this logic, let me know.

Things I like:
1. separate function to determine warnings, `get_membership_warnings_after_match`

Things concerning me:
1. use of `if membership_auto_assign` twice, then use of `if membership` (which is basically the same as `if not membership_auto_assign`) - this seems like two different logic flows intertwined
2. errors not calculated in separate function (just like warnings)
3. not a lot of code moved outside to separate functions (makes testing difficult)

amy/static/task_form.js
```diff
@@ -0,0 +1,31 @@
...
+            },
+        },
+    });
+    // update variables when a selection is made
+    $("#id_task-person").on("select2:select", (e) => {
+        const data = e.params.data;
+        person = data.id;
+    });
+    $("#id_task-role").on("change", (e) => {
+        console.log(e.target.value);
```
pbanaszkiewicz, line 24: Leftover?

amy/extrequests/utils.py
```diff
@@ -76,3 +76,40 @@ def get_membership_or_none_from_code(code: str) -> Membership | None:
         return Membership.objects.get(registration_code=code)
     except Membership.DoesNotExist:
         return None
+
+
+def get_membership_warnings_after_match(
```
pbanaszkiewicz, line 81: Can you add tests for this function?

pbanaszkiewicz (approved):
Left some comments / suggestions. In general it looks good, but if you have some ideas how that long view can be improved, I'm all ears.

elichad:
@pbanaszkiewicz I addressed your comments, and I did some refactoring to fully separate the `membership_auto_assign` flow from the `not membership_auto_assign` flow and move a few more things to utils. I think there's probably still room for improvement, but it's better than it was. Could you re-review?

pbanaszkiewicz (approved):
Looks good!

PR #2573, opened 2023-11-14
add CLDT tag

maneesha:


elichad:
Thanks for adding the tag. I'll note here that there are a few places in the AMY code (and Redash queries) that specifically exclude/include the "TTT" tag when handling events, so we might run into a bit of weird behaviour around CLDT events showing up where they shouldn't.

maneesha:
Thanks @elichad, and I am making related changes such as https://github.com/carpentries/carpentries.org/pull/1737

PR #2579, opened 2023-11-23
Make agreement_link required field in Membership model

### LINKED ISSUE #2307: Make "Agreement document" required field in membership model
At the Membership team meeting today, we discussed that the field for "Agreement Document" should be a required field.  Keeping this issue open to note other changes to fields in the membership model we should make.  

Tagging @elletjies  FYI, please feel free to add anything else here

elichad:
Fixes #2307.


elichad:
Pinging @Talishask @elletjies to give you a heads up that as part of the next release, the "Link to member agreement" field will become required when creating or editing a membership.

pbanaszkiewicz (approved):
Looks good.

Talishask:
Apologies for being late is it possible to undo this?

There are a few types of memberships that do not have a membership agreement document so I am unsure how this will affect those? can we discuss what this will look like?

elletjies:
@Talishask in these events can we share the Asana task link or die google drive link?

Talishask:
If we do not have consistent information to put in the field then the probably should not be required. If we end up putting random information into the database that will be messy. This will be to you all to decide, my going away recommendation would be to have a reason to do this extra work and be consistent with whatever is decided to go in the field.

PR #2580, opened 2023-11-26
[Emails] New membership onboarding email

### LINKED ISSUE #2577: [Emails] Implement "New / renewing membership starting (member onboarding)" email

pbanaszkiewicz:
This fixes #2577.

### DIFF
--- a/amy/emails/actions/__init__.py
+++ b/amy/emails/actions/__init__.py
@@ -22,4 +22,9 @@ from emails.actions.instructor_training_completed_not_badged import (
     instructor_training_completed_not_badged_remove_receiver,
     instructor_training_completed_not_badged_update_receiver,
 )
+from emails.actions.new_membership_onboarding import (
+    new_membership_onboarding_receiver,
+    new_membership_onboarding_remove_receiver,
+    new_membership_onboarding_update_receiver,
+)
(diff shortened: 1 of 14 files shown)


pbanaszkiewicz:
@elichad This is quite close to completion, you may take a look.

I'm still not quite happy about some of the solutions here.

1. When I removed all contacts from the membership, I expected the scheduled email to be cancelled. It wasn't and I will be looking into this case.
2. I'm using a "hack" to determine if membership can or can't be deleted. If it can be removed, then I'm running the strategy to remove related emails. This was required because if the membership had been removed (before strategy is run), then running the strategy would fail.

pbanaszkiewicz:
No longer WIP.

Issues from previous comment were addressed:
> 1. When I removed all contacts from the membership, I expected the scheduled email to be cancelled. It wasn't and I will be looking into this case.
> 2. I'm using a "hack" to determine if membership can or can't be deleted. If it can be removed, then I'm running the strategy to remove related emails. This was required because if the membership had been removed (before strategy is run), then running the strategy would fail. I'll look into other approach here.

Ad 1: membership tasks "billing contact" or "programmatic contact" are now required for the email to be created. Once there are no such contacts, the email will be cancelled.

Ad 2: The "hack" is still in place. Unfortunately I don't see any better solution there.

PR #2606, opened 2024-01-23
[Chore] Unconditionally enable "ENFORCE_MEMBER_CODES" feature flag

pbanaszkiewicz:
This enables the feature flag "ENFORCE_MEMBER_CODES", which is required for testing by anonymous users.


maneesha (approved):
Approved (assuming this means that the member code feature flag will always be on)

PR #2689, opened 2024-08-19
Cldt seats

froggleston:
This PR is a WIP to implement CLDT seat allocation and tracking

### DIFF
--- a/amy/templates/fiscal/all_memberships.html
+++ b/amy/templates/fiscal/all_memberships.html
@@ -17,7 +17,8 @@
         <th>Variant</th>
         <th>Dates</th>
         <th>Contribution</th>
-        <th>Remaining instructor training seats</th>
+        <th>IT seats</th>
+        <th>CLDT seats</th>
         <th class="additional-links"></th>
       </tr>
       {% for membership in all_memberships %}
@@ -41,7 +42,10 @@
           <td>{{ membership.agreement_start|date:"Y-m-d" }} &mdash; {{ membership.agreement_end|date:"Y-m-d" }}</td>
           <td>{{ membership.get_contribution_type_display }}</td>
           <td {% if membership.instructor_training_seats_remaining <= 0 and membership.instructor_training_seats_total > 0 or membership.instructor_training_seats_remaining < 0 and membership.instructor_training_seats_total == 0 %}class="table-danger"{% endif %}>
-            {{ membership.instructor_training_seats_remaining }}
+            {{ membership.instructor_training_seats_remaining }} / {{ membership.instructor_training_seats_total }}
+          </td>
+          <td {% if membership.cldt_seats_remaining <= 0 and membership.cldt_seats_total > 0 or membership.cldt_seats_remaining < 0 and membership.cldt_seats_total == 0 %}class="table-danger"{% endif %}>
+            {{ membership.cldt_seats_remaining }} / {{ membership.cldt_seats_total }}
           </td>
           <td>
             <a href="{% url 'membership_details' membership.pk %}" title="View membership"><i class="fas fa-info-circle"></i></a>
--- a/amy/workshops/models.py
+++ b/amy/workshops/models.py
@@ -139,6 +139,9 @@ class Member(models.Model):
 
 class MembershipManager(models.Manager):
     def annotate_with_seat_usage(self):
+        cldt_tag = Tag.objects.get(name="CLDT")
+        no_cldt_tags = Tag.objects.filter(name__in=["SWC", "DC", "LC", "TTT", "ITT", "WiSE"])
+
         return self.get_queryset().annotate(
             instructor_training_seats_total=(
                 # Public
@@ -152,7 +155,10 @@ class MembershipManager(models.Manager):
                 + Coalesce("inhouse_instructor_training_seats_rolled_from_previous", 0)
             ),
             instructor_training_seats_utilized=(
-                Count("task", filter=Q(task__role__name="learner"))
+                Count("task", filter=Q(
+                    task__role__name="learner",
+                    task__event__tags=[no_cldt_tags]
+                ))
             ),
             instructor_training_seats_remaining=(
                 # Public
@@ -161,7 +167,11 @@ class MembershipManager(models.Manager):
                 # Coalesce returns first non-NULL value
                 + Coalesce("public_instructor_training_seats_rolled_from_previous", 0)
                 - Count(
-                    "task", filter=Q(task__role__name="learner", task__seat_public=True)
+                    "task", filter=Q(
+                        task__role__name="learner",
+                        task__seat_public=True,
+                        task__event__tags=[no_cldt_tags]
+                    )
                 )
                 - Coalesce("public_instructor_training_seats_rolled_over", 0)
                 # Inhouse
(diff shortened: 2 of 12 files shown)


amy/workshops/models.py
```diff
@@ -140,6 +140,10 @@ class Meta:
 
 class MembershipManager(models.Manager):
     def annotate_with_seat_usage(self):
+        cldt_tag = Tag.objects.get(name="CLDT")
+        # TTT/ITT included
+        ttt_tags = Tag.objects.exclude(name__in=["SWC", "DC", "LC", "WiSE", "CLDT"])
```
pbanaszkiewicz, line 145: In here (L143, L145) use tag names:

```python
cldt_tag_name = "CLDT"
ttt_tag_names = ["TTT", "ITT"]
```

Then the lookups using `ttt_tags` or `cldt_tag` have to be changed:
1. from `task__event__tags=[ttt_tags]` to `task__event__tags__name__in=ttt_tag_names`
2. from `task__event__tags=[cldt_tag]` to `task__event__tags__name__in=[cldt_tag_name]`

amy/workshops/models.py
```diff
@@ -560,6 +634,31 @@ def inhouse_instructor_training_seats_remaining(self) -> int:
...
+        """
+        a = self.cldt_seats
+        b = self.additional_cldt_seats
+        c = self.cldt_seats_rolled_from_previous or 0
+        return a + b + c
+
+    @cached_property
+    def cldt_seats_utilized(self) -> int:
+        """Count number of learner tasks that point to this membership."""
+        return self.task_set.filter(role__name="learner", seat_public=False).count()
```
pbanaszkiewicz, line 652: Since this property should mimick the query in `MembershipManager.annotate_with_seat_usage`, there's additional filtering missing:

1. `cld_seats_utilized()` should also filter by event tag CLDT
2. in `MembershipManager.annotate_with_seat_usage`, `_cldt_seats_utilized` should also filter by `seat_public`
3. in `MembershipManager.annotate_with_seat_usage`, `_cldt_seats_remaining` should filter by `seat_public` in the same way (currently it uses `public=True`)
4. both `public_instructor_training_seats_utilized()` and `inhouse_instructor_training_seats_utilized()` should also filter by TTT tags (TTT, ITT)

Generally speaking the model properties should come down to the same equations as in annotations.

amy/templates/reports/membership_trainings_stats.html
```diff
@@ -35,6 +36,9 @@
         <td>{{ result.instructor_training_seats_total }}</td>
         <td>{{ result.instructor_training_seats_utilized }}</td>
         <td>{{ result.instructor_training_seats_remaining }}</td>
+        <td>{{ result._cldt_training_seats_total }}</td>
+        <td>{{ result._cldt_training_seats_utilized }}</td>
+        <td>{{ result._cldt_training_seats_remaining }}</td>
```
pbanaszkiewicz, line 41: Django template is complaining about these names. Apparently they can't start with `_`, so my proposition is to use these names instead:

1. `cldt_seats_total_annotation`
2. `cldt_seats_utilized_annotation`
3. `cldt_seats_remaining_annotation`

Changes should be applied here and in `MembershipManager.annotate_with_seat_usage`.

amy/templates/reports/membership_trainings_stats.html
```diff
@@ -16,6 +16,7 @@
         <th rowspan="2">Agreement</th>
         <th rowspan="2">Contribution</th>
         <th colspan="3" class="text-center">Instructor training seats (combined public and in-house)</th>
+        <th colspan="3" class="text-center">CLDT seats</th>
```
pbanaszkiewicz, line 19: This introduces 3 new columns in the row that's following (lines 23-25 below). Can you duplicate those lines? This way the header displays correctly:

image

amy/templates/fiscal/all_memberships.html
```diff
@@ -41,7 +42,10 @@
           <td>{{ membership.agreement_start|date:"Y-m-d" }} &mdash; {{ membership.agreement_end|date:"Y-m-d" }}</td>
           <td>{{ membership.get_contribution_type_display }}</td>
           <td {% if membership.instructor_training_seats_remaining <= 0 and membership.instructor_training_seats_total > 0 or membership.instructor_training_seats_remaining < 0 and membership.instructor_training_seats_total == 0 %}class="table-danger"{% endif %}>
-            {{ membership.instructor_training_seats_remaining }}
+            {{ membership.instructor_training_seats_remaining }} / {{ membership.instructor_training_seats_total }}
```
pbanaszkiewicz, line 45: This notation (`remaining count` / `total count`) is clever, but I'm wondering if users will know this is remaining / total, and not utilized / total.

amy/workshops/models.py
froggleston: Updated all calculations and set CLDT to public=False as these events are member based and not open to the public.

amy/templates/reports/membership_trainings_stats.html
froggleston: I've changed the CLDT annotation variable names but not the others. We can do this after merge/rebase?

amy/templates/reports/membership_trainings_stats.html
froggleston: Done!

amy/templates/fiscal/all_memberships.html
```diff
@@ -41,7 +42,10 @@
...
+          </td>
+          <td {% if membership._cldt_seats_remaining <= 0 and membership._cldt_seats_total > 0 or membership._cldt_seats_remaining < 0 and membership._cldt_seats_total == 0 %}class="table-danger"{% endif %}>
+            {{ membership._cldt_seats_remaining }} / {{ membership._cldt_seats_total }}
```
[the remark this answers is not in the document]
froggleston: Done!

amy/templates/fiscal/all_memberships.html
froggleston: I'll check this with the CT, but as these calculations have been in place for the other seat types, I wouldn't want to change it without consulting them.

amy/templates/reports/membership_trainings_stats.html
pbanaszkiewicz: I think only CLDT annotations started with `_`, so only they needed to be updated with that prefix. Other annotations work just fine because they don't have name clash with model properties.

amy/workshops/models.py
```diff
@@ -143,8 +145,8 @@ def annotate_with_seat_usage(self):
         # cldt_tag = Tag.objects.get(name="CLDT")
         # # TTT/ITT included
         # ttt_tags = Tag.objects.exclude(name__in=["SWC", "DC", "LC", "WiSE", "CLDT"])
-        cldt_tag_name = "CLDT"
-        ttt_tag_names = ["TTT", "ITT"]
+        # cldt_tag_name = "CLDT"
+        # ttt_tag_names = ["TTT", "ITT"]
```
pbanaszkiewicz, line 149: Can you remove this commented-out code?

amy/workshops/consts.py
```diff
@@ -5,3 +5,6 @@
 STR_LONG = 100  # length of long strings
 STR_LONGEST = 255  # length of the longest strings
 STR_REG_KEY = 20  # length of Eventbrite registration key
+
+CLDT_TAG_NAME = ["CLDT"]
```
pbanaszkiewicz, line 9: I'd keep this plural (`nameS`) since this is a list.

maneesha:
@froggleston I think this is now obsolete with the new benefits model, correct?

froggleston:
Hey @maneesha - yes, I believe we'll have another go at this once we have a follow-up about the general approach to tags vs proper lesson program entities/enums.

PR #2725, opened 2024-11-27
[Recruitment] Disallow multiple signups for the same person

### LINKED ISSUE #2495: Admin can add multiple signups for the same Instructor at the same workshop
An Instructor can only sign up to help at a workshop once. However, an admin can add more signups for that same Instructor. This has potential to create confusion if the states are changed for both signups.

Trying to confirm both causes a database error (#2449), but accepting one and declining the other means different emails will be sent and different status will be shown to the Instructor depending on the order of those actions. 

The correct action would be to delete the duplicate, but we should make it harder to run into this problem in the first place.

Screenshot of a fake event recruitment, with two signups for the same Instructor, Marcus Aguirre.

pbanaszkiewicz:
This fixes #2495 by disallowing multiple signups (both from admin and from the instructor) in open recruitments. Indirectly this also resolves an issue of multiple confirmation emails sent out in the new email system.

### DIFF
--- a/amy/dashboard/forms.py
+++ b/amy/dashboard/forms.py
@@ -3,6 +3,7 @@ from django.core.exceptions import ValidationError
 from django.db.models import Q
 from django_countries.fields import CountryField
 
+from communityroles.models import CommunityRole
 from recruitment.models import InstructorRecruitment, InstructorRecruitmentSignup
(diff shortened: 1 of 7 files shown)


pbanaszkiewicz:
WIP - tests pending.

Issue #2755, opened 2025-02-08
Remove `autoemails` application and code


pbanaszkiewicz:
This has to be done after the migrations were applied to remove models from database.

See for details: https://docs.djangoproject.com/en/5.1/howto/delete-app/

PR #2796, opened 2025-04-23
fix(#2795): Make autoemails removal migration depending on workshops.0277 migration

### LINKED ISSUE #2795: Retro-fix order of migrations
This was broken in production environment. `autoemails/migrations/0020_remove_rqjob_trigger_remove_trigger_template_and_more.py` should be depending on `workshops/migrations/0277_merge_20250206_0536.py`.

pbanaszkiewicz:
This enforces order: 0277 is removing relationships to RQJobs, then 0020 is removing RQJobs.

Fixes #2795.


pbanaszkiewicz:
This is no longer valid because in #2755 the application is removed.

Issue #2764, opened 2025-02-12
Add field for Mastodon


maneesha:
We should add a field for Mastodon to the Person model.  This should be a `URLField`.

pbanaszkiewicz:
Closed through #2809 .

PR #2797, opened 2025-04-25
Add field for Mastodon in Person model

maneesha:
Addresses #2764


maneesha:
I think this was made to the wrong branch.  See #2798 instead

PR #2801, opened 2025-05-06
[Emails] Don't cancel emails for 7days post workshop during that time after the workshop

### LINKED ISSUE #2792: Post-Workshop 7 day email gets canceled when event is updated
If a workshop gets edited in the week following the workshop end date before the post-workshop email gets sent, the email gets canceled.  

Example: If workshop start date is 2025-04-01 and end date is 2025-04-02, the post workshop email gets scheduled for 2025-04-09 (as expected).  If the workshop is edited between 2025-04-02 and 2025-04-09, the email gets canceled. 

This is likely because one of the conditions for the email is that the workshop must be in the future.  

We are able to work around this by re-scheduling the canceled email.  However, the email should remain scheduled for the original date.

pbanaszkiewicz:
This fixes #2792.

### DIFF
--- a/amy/emails/actions/post_workshop_7days.py
+++ b/amy/emails/actions/post_workshop_7days.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timedelta
+from datetime import UTC, datetime, time, timedelta
 import logging
 from typing import Any, Unpack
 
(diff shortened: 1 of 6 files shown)


maneesha:
As noted in slack, this looks good to me.

PR #2962, opened 2026-05-04
Add airport to test superuser

maneesha:
Adds the now-required airport field when creating superuser `admin` in dev database.

### DIFF
--- a/src/workshops/management/commands/create_superuser.py
+++ b/src/workshops/management/commands/create_superuser.py
@@ -27,6 +27,7 @@ class Command(BaseCommand):
                 family="admin",
                 email="admin@example.org",
                 password="admin",
+                airport_iata="AAA",
             )
             logger.info("Created admin user")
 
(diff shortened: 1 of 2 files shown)


maneesha:
@pbanaszkiewicz What is the preferred styling when lines are too long?

pbanaszkiewicz:
@maneesha if you run `uv run ruff format` it will automatically correct the line for you. Depending on the length, it will either use this format:

```python
def func(
    arg1, arg2, arg3, arg4, arg5
):
    ...
```

or this style:


```python
def func(
    arg1,
    arg2,
    arg3,
    arg4,
    arg5,
):
    ...
```

maneesha:
@pbanaszkiewicz Now tests are failing because of the added field. Would you recommend I manually update each test? Or is there another way about this?

pbanaszkiewicz:
@maneesha Both `mypy` and tests are failing. You should follow `mypy` (see the 'lint' pipeline step) output - there are 13 errors, should be easy to fix. The tests are most likely failing because they all depend on `create_superuser`.

PR #2985, opened 2026-06-22
Testing out embedding Pretix in AMY

maneesha:
@pbanaszkiewicz As discussed, this PR begins to try to embed Pretix in AMY.  I will share credentials with you tomorrow.  

For now, you can run AMY locally with this PR and log in as an active user and go to `/dashboard`. You will see the embedded widget at the top of the page with a list of dates. Select any date and use the voucher `abcde`. This now takes you to an external page to continue registration. The email and attendee name are automatically filled in from the AMY profile. This is the step we'd like to keep within AMY. This will help integrate the profile creation form questions with event registration questions and integrate event registration with AMY event tasks.  

You can see some cors headers settings I tried changing with @froggleston.  I commented those back out because they weren't working but I wanted you to see them.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -166,6 +166,8 @@ THIRD_PARTY_APPS = [
     "djangoformsetjs",
     "django_better_admin_arrayfield",
     "flags",
+    "corsheaders",
+    "sslserver",
 ]
 LOCAL_APPS = [
     "src.workshops.apps.WorkshopsConfig",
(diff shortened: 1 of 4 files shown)


config/settings.py
```diff
@@ -615,6 +615,8 @@
     raise ImproperlyConfigured("SITE_BANNER_STYLE accepts only one of 'local', 'testing', 'production'.")
 
 PROD_ENVIRONMENT = bool(SITE_BANNER_STYLE == "production")
+LOCAL_ENVIRONMENT = bool(SITE_BANNER_STYLE == "local")
+print(LOCAL_ENVIRONMENT)
```
pbanaszkiewicz, line 619: needs to be removed - I left it unintentionally!

Issue #3001, opened 2026-07-29
Add option about online event to WRF travel expenses question


maneesha:
On the workshop [request](https://amy.carpentries.org/forms/request_workshop/) and [inquiry](https://amy.carpentries.org/forms/inquiry/) forms we have the question "How will you manage travel expenses for Carpentries Instructors?"

This should have one more option to say "Not applicable; Online workshop."

PR #3019, opened 2026-08-11
Travel expenses for online workshop

maneesha:
Addresses #3001


maneesha:
@pbanaszkiewicz In this case, the error about line too long refers to the generated migration file.  Do you fix these migration files or ignore the error?

maneesha:
I tried using `ruff` to format the lines but am still getting the error (to different lines now since some formatting was applied).

src/extrequests/migrations/0041_alter_workshopinquiryrequest_travel_expences_management.py
```diff
@@ -0,0 +1,32 @@
...
+        migrations.AlterField(
+            model_name="workshopinquiryrequest",
+            name="travel_expences_management",
+            field=models.CharField(
+                blank=True,
+                choices=[
+                    ("", "Not sure yet."),
+                    (
+                        "booked",
+                        "Hotel and airfare will be booked by site; ground travel and meals/incidentals will be reimbursed within 60 days.",
```
pbanaszkiewicz, line 21: @maneesha You can break a long string like this:

```python
                        "Hotel and airfare will be booked by site; ground travel and"
                        " meals/incidentals will be reimbursed within 60 days."
```

The important thing is that there is no comma between the strings and it only works in brackets/parens.

